PR #252, opened 2013-05-10
More shared_ptr ambiguity fixes

Shadowrom:
MSVC 2012 is strange ;) 
- moved another using std::tr1::shared_ptr into the namespace.
- encasulated an include statement into a ifndef
- expanded an #if statement by (_MSC_VER >= 1700) beacause VC 2012
  defines __cplusplus = 199711L but delivers C++11 libs.

Now it builds fine here too.

### DIFF
--- a/src/include/osl_pvt.h
+++ b/src/include/osl_pvt.h
@@ -33,12 +33,14 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "OpenImageIO/dassert.h"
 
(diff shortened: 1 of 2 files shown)


MrMormon:
Should this be closed?

lgritz:
Good question. The osl_pvt.h change is irrelevant now, but I wonder if the oslconfig.h change is still helpful to us. I don't have a Windows machine to know if _MSC_VER >= 1700 (which I think corresponds to MSVS 2012?) is tantamount to assuming it's C++11 enough for our purposes.  Do you know?

lgritz:
We've long since switched to C++11 as minimum. Closing.

PR #282, opened 2013-08-29
Demonstrate simple material layering in OSL

amartinez-cg:
Inspired by a conversation at the OSL BOF at Siggraph 2013, this pull request adds material layering functionality to ubersurface.osl, as well as a demonstrative test.  It is intended to be an example of one possible way to implement material layering within OSL, but by no means do I claim it is the best method.


lgritz:
Closing this PR -- it's obviously going to be conceptually replaced by what you're working on now.

PR #314, opened 2013-12-16
Fix build on Ubuntu 13.10/GCC.

Anteru:
Had to do some minor tweaks to get OSL build on a stock Ubuntu 13.10 using the built-in libraries.


lgritz:
I think this was either fixed long ago, or is no longer relevant because the minimum required versions of OIIO include the ASSERT_MSG macro. Closing.

PR #355, opened 2014-04-22
Guess a Unix standard value for OPENIMAGEIOHOME when undefined

keryell:
This way the package openimageio-tools can work out-of-the-box on Debian &
Ubuntu.
Of course, if they are compiled with the right options for openShadingLanguage...


keryell:
This makes sense.

PR #403, opened 2014-08-19
Optimization: Be less conservative in implementation of unread_after().

lgritz:
It was really implemented as if it were "unread AT OR after". It's used to determine whether it's safe to elide assignments to variables that aren't subsequently use. But sometimes the assignment is a function where the variable assigned to is also one of the arguments. For example,

    x = func(x,y)

x is read on that op, but not AFTER. But because of how the unread_after also discounted elision if used on that op, it was less aggressive than it could have been in eliminating statements like this.

I'm seeing around a 1/2 - 1% speedup on production frames because of this, with a similar decrease in post-optimized symbols and instructions. Not dramatic, but every little bit counts.

### DIFF
--- a/src/liboslexec/runtimeoptimize.cpp
+++ b/src/liboslexec/runtimeoptimize.cpp
@@ -1161,7 +1161,7 @@ RuntimeOptimizer::unread_after (const Symbol *A, int opnum)
 
     // For all else, check if it's either never read at all in this
     // layer or it's only read earlier and we're not part of a loop
-    return !A->everread() || (A->lastread() < opnum && !m_in_loop[opnum]);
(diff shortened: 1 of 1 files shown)


fpsunflower:
It looks like this review was forgotten about (github still thinks it can merge it). Is that really the case?

fpsunflower:
It looks like it just never got committed -- the code still has the original `<`, not `<=`. Larry, do you remember if there ended up being an issue with this?

lgritz:
I'm still sitting on this one, need to return to it.

What happened is that I in my pre-commit tests, I saw an image difference that needs to be investigated. The change should not have been able to make any visual changes.

lgritz:
Sheesh, has it been that long? In my mind, I had put this off for a couple weeks. Oops. I'll get back to this.

PR #434, opened 2014-11-18
Fix load shader with closure array in/out parameter

monsieurgustav:
A shader with a closure array parameter compiles, but does not load at runtime.
This commit fixes it.

e.g.
shader Test(closure color param[4], output closure color outParam[4])
{ ... }


monsieurgustav:
I close it as the patch has been "pull requested" again from UbisoftMotionPictures.
https://github.com/imageworks/OpenShadingLanguage/pull/544

PR #445, opened 2014-12-21
Fix compilation with GCC 4.2.1 (OS X 10.7).

johnhaddon:
GCC was preferring to cast float4 to Dual2<float4> and call the locally defined versions of shuffle() etc, rather than call the correct OIIO::simd versions. Fully qualifying the calls with the namespace fixes that.

This doesn't seem ideal, as it's quite verbose, and also prone to being rebroken when new code is introduced. I did try removing the `using namespace OIIO::simd` to prevent the latter problem, but then even more verbosity would be required. Perhaps removing the `using` and replacing it with a short namespace alias would be preferred?

I have mailed a completed CLA to opensource@imageworks.com.

### DIFF
--- a/src/liboslexec/noiseimpl.h
+++ b/src/liboslexec/noiseimpl.h
@@ -427,17 +427,17 @@ template<> OIIO_FORCEINLINE Dual2<float4> negate_if (Dual2<float4> val, int4 b)
 template<int i0, int i1, int i2, int i3>
 OIIO_FORCEINLINE Dual2<float4> shuffle (Dual2<float4> a)
 {
-    return Dual2<float4> (shuffle<i0,i1,i2,i3>(a.val()),
-                          shuffle<i0,i1,i2,i3>(a.dx()),
-                          shuffle<i0,i1,i2,i3>(a.dy()));
+    return Dual2<float4> (OIIO::simd::shuffle<i0,i1,i2,i3>(a.val()),
+                          OIIO::simd::shuffle<i0,i1,i2,i3>(a.dx()),
+                          OIIO::simd::shuffle<i0,i1,i2,i3>(a.dy()));
 }
 
 template<int i>
 OIIO_FORCEINLINE Dual2<float4> shuffle (Dual2<float4> a)
 {
-    return Dual2<float4> (shuffle<i>(a.val()),
-                          shuffle<i>(a.dx()),
-                          shuffle<i>(a.dy()));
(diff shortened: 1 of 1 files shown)


lgritz:
Sorry for the delay, John. This LGTM, so I would like to merge it.

I know you sent us a personal CLA, but can you also get your employers to send us the corporate CLA?
(Unless you did it on your own time AND have a clear agreement with IE that they don't have any claims to IP you produce on your own.)

johnhaddon:
Thanks Larry. It is pretty ugly though, and prone to getting broken again (I don't think I'd remember to type the full namespace for new code). If you'd rather not merge, then it is enough for me to know that it looks OK, and I can maintain a local patch till such a time I can upgrade to a recent compiler/os. I suspect I may also be able to compile with clang on 10.7 as another option.

As for the CLA, it was on my own time, but let me get you a corporate CLA as well - I think we'll be getting stuck into OSL a bit more seriously in 2015, and getting one in place early makes sense.

lgritz:
CLA - the concern is that most employers make you sign an IP agreement when you start that can easily be interpreted as giving them claim on any IP related to their business that you produce, even on your own time. But even if you have a carve-out with them for this work, as you said, it probably doesn't hurt to clear the way for any future contributions that are officially from them.

As for the patch: Up to you. I'm happy to merge it, but if it's truly specific to using an old compiler and isn't bothering anybody else, I'm also happy to just not worry about it and have it be something you maintain on your side.

johnhaddon:
I've set the corporate wheels in motion - hopefully it won't take too long to reach the big cog upstairs. If you're happy, then a merge would be nice once that's done - it'll make life a bit simpler till I can get the compiler upgraded. Thanks!

Any word on when there might be an official 1.6 release? We're planning on releasing a Gaffer/Appleseed/OSL package at that point, which would be a nice little milestone for open source lighting and rendering.

lgritz:
OK, I'll merge it when we get the CLA. It sure can't hurt, even if you won't strictly need the namespace qualifier after you upgrade your compiler.

I think I'll aim to release 1.6 in about 2 weeks. I need to get OIIO 1.5 finalized, and then on the OSL side I think there is only one more thing I want to squeeze in that would alter APIs.

johnhaddon:
I'm told an Image Engine CLA has been sent to Sony legal. Let me know if it hasn't arrived and I'll chase it up...

lgritz:
Yes, got the CLA yesterday! I will rebase and merge this now.

lgritz:
Merged.

PR #449, opened 2015-01-05
Typo fix: exepected -> expected

nicholasbishop:


lgritz:
Thanks for the fix. Merged.

PR #450, opened 2015-01-07
oslcomp fix: add directory holding stdosl.h to the include path.

lgritz:
This is the way it had always been, but I broke it with the compile_buffer patch. That caused problems for situations where other headers in that directory were assumed to be includable without having to separately specify the path on the command line.


fpsunflower:
LGTM

PR #451, opened 2015-01-09
oslc fix: allow comma operator

lgritz:
Recently noticed that the "comma operator" does not work as expected. You can see this with a common idiom of incrementing multiple counters in a for loop, like this:

    for (int i=0; i < j; ++i, a *= 2) ...

It's not specific to 'for', of course; in general the comma operator should work in the same places it would in C. To do this properly, we need to split the grammatical 'expression' into the simple kind and a compound_expression, and also add a new AST node for the comma operator (because it's special: a,b has the type and value of b).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,8 @@ Language, standard libary, and compiler changes (for shader writers):
   directly seen by the camera will directly reproduce the closure weight
   in the final pixel, regardless of it being a surface or volume. Thus,
   you don't need to weight it by PI. #427 (1.6.2)
+* The "comma operator" now works in the way you would expect in other
+  C-like languages. #451 (1.6.2)
 
 API changes, new options, new ShadingSystem features (for renderer writers):
 * New ShadingSystem attribute int "profile", if set to 1, will include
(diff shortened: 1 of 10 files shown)


fpsunflower:
I was trying to come up with some additional tests and after some double checking, I don't think that `,` works as "return last". I think its actually that `,` has a lower precedence than `=`. So the following code:

```c
int main() {
  int a, b;
  a = 1, 2, 3;
  b = 4, 5, 6;
  printf("%d %d\n", a, b);
}
```

Prints `1 4`, not `3 6`.

Another good test case might be:
```c
int lo, hi;
if (a < b)
   lo = a, hi = b;
else
   hi = a, lo = b;
```

fpsunflower:
On the other hand:

```c
int c = (a, b);
```

Does return `b`.

So maybe your code is working as expected already? Might be good to add the cases above to make it clear.

lgritz:
Yes, I see what you mean. Um... I'll augment the test to add these cases, and then see what, if anything, is broken.

lgritz:
Yeah, some minor mistakes. Fixing.

lgritz:
OK, I think I have it. Much more thorough test case.

fpsunflower:
LGTM

PR #453, opened 2015-01-13
Bug fix in variable declaration + assignment of nested structs

lgritz:
There was a case involving this kind of setup:

    struct InnerStruct { ...fields... };
    struct OuterStruct { InnerStruct inner; }
    OuterStruct x = ...;
    OuterStruct y = x;    <----- THIS

where incorrect oso code would be generated, to wit: directly assigning x.inner to y.inner (illegal!), instead of assigning the individual fields of x.inner to the corresponding field of y.inner.

We'd solved this a while back for ordinary variable assignment, with the solution implemented in ASTassign_expression::codegen_assign_struct, which needed to operate recursively to handle structs-within-structs as far down as necessary. But the way initialization assignments like the above were handled by ASTvariable_declaration was naive and botched it for nested structs.

This patch moves ASTassign_expression::codegen_assign_struct() into the parent class (ASTnode) so that it may be used by ASTvariable_declaration to do it correctly there as well.

Oy, this has been wrong all along. I hope there aren't too many shaders that were broken but that nobody noticed, and that in now getting correct code generated, have an appearance change. But it was plainly wrong before.

Also fixed what I think is a minor latent bug in codegen_assign_struct, as well as fixed some needless clutter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -368,7 +368,7 @@ TESTSUITE ( aastep arithmetic array array-derivs array-range
             render-microfacet render-oren-nayar render-veachmis render-ward
             shortcircuit spline splineinverse string 
             struct struct-array struct-array-mixture
-            struct-err struct-layers struct-with-array 
+            struct-err struct-in-struct-init struct-layers struct-with-array 
             struct-within-struct ternary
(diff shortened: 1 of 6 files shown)


fpsunflower:
LGTM

PR #454, opened 2015-01-13
gettextureinfo constant fold fix: needs to call RS::get_texture_info

lgritz:
gettextureinfo constant fold fix: needs to call RS::get_texture_info(), but was incorrectly bypassing that straight to texturesys()->get_texture_info(). That's bad for a renderer that has a custom RS::get_texture_info and only a "mock" TextureSystem.


fpsunflower:
LGTM

PR #456, opened 2015-01-15
Only merge instances when optimize level is at least 1

lgritz:
Optimize level 0 seems like it should do very little optimization, and that counts instance merging.


aconty:
LGTM

PR #457, opened 2015-01-15
Fix two bugs related to renderer outputs

lgritz:
1. When deciding whether an output param is a renderer output, we should have checked if the renderer output list contained the layer-specified output name, as well as the unqualified (any layer) name. We did it right for the group-specific output list, but not for the renderer-wide output list.

2. Don't consider a layer unused if its outputs are renderer outputs. We were optimizing away layers, or skipping them during "lazy" evaluation, if they weren't setting any "globals" and didn't have any downstream connections. But if it sets a renderer output, it needs to be run unconditionally.

### DIFF
--- a/src/liboslexec/oslexec_pvt.h
+++ b/src/liboslexec/oslexec_pvt.h
@@ -533,6 +533,11 @@ public:
     ///
     void outgoing_connections (bool out) { m_outgoing_connections = out; }
 
+    /// Does this instance have any renderer outputs?
+    bool renderer_outputs () const { return m_renderer_outputs; }
+    /// Set whether this instance has renderer outputs
+    void renderer_outputs (bool out) { m_renderer_outputs = out; }
+
     int maincodebegin () const { return m_maincodebegin; }
     int maincodeend () const { return m_maincodeend; }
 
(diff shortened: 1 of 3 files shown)


src/liboslexec/oslexec_pvt.h
```diff
@@ -594,7 +599,8 @@ class ShaderInstance {
 
     /// Does it appear that the layer is completely unused?
     ///
-    bool unused () const { return run_lazily() && ! outgoing_connections(); }
+    bool unused () const { return run_lazily() && ! outgoing_connections()
```
aconty: Doesn't run_lazily() include the other flags? Should this be just !outgoing_connections() && !render_outputs()?

lgritz: I think you are right. I don't mind that it's explicit here, and I can imagine a future in which we separate the concepts a little more. But if you think this detracts from understanding of the code, I can remove that clause and rely on run_lazily().

aconty: I'm fine with it too

aconty:
LGTM

PR #458, opened 2015-01-15
Extend get_symbol to retrieve the symbol from a particular named layer.

lgritz:
When no layer name is specified, it produces the old behavior of finding the matching symbol in the LAST layer in which the name is found.

The point here is to help if you have a shader group (network) that has multiple layers (nodes) with identically-named output parameters, and you want to retrieve a particular one which is not necessarily the last one that executes.


aconty:
LGTM

src/liboslexec/shadingsys.cpp
```diff
@@ -265,9 +265,26 @@ ShadingSystem::execute (ShadingContext &ctx, ShaderGroup &sas,
...
+
+
+const void*
+ShadingSystem::get_symbol (ShadingContext &ctx,
+                           ustring symbolname, TypeDesc &type)
+{
+    ustring layername;
+    size_t dot = symbolname.find('.');
+    if (dot != ustring::npos) {
+        // If the name contains a dot, it's intended to be layer.symbol
```
fpsunflower: This looks like an expensive feature (because of the dynamic ustring creation) -- wouldn't it be better to guide people to the other function call if they need per layer access ?

src/liboslexec/shadingsys.cpp
```diff
@@ -265,9 +265,26 @@ ShadingSystem::execute (ShadingContext &ctx, ShaderGroup &sas,
 
 
 const void*
-ShadingSystem::get_symbol (ShadingContext &ctx, ustring name, TypeDesc &type)
+ShadingSystem::get_symbol (ShadingContext &ctx, ustring layername,
```
fpsunflower: What happens if you ask for a symbol from a layer which was not executed due to lazy layers?

lgritz: It returns NULL, can't be found.

lgritz: And that's exactly how I found the other bug I sent fixed to last night!

lgritz: Revision: it definitely returns NULL if the layer was *eliminated* during optimization. But I think it returns the valid memory location (whose contents will be uninitialized) if the layer existed but was never run (or hasn't run *yet*).

fpsunflower: Makes sense, though in combination with the other fix -- the layers who's symbols you ask for should in theory have been run already.

Up to the user to declare upfront they will be reading certain symbols if they want them to be initialized properly.

src/liboslexec/shadingsys.cpp
lgritz: This was really helpful in making testshade not get overly complicated. But I can certainly add a comment to the description of this function pointing out that get_symbol("layer.param") is much more expensive than get_symbol("layer", "param").

src/liboslexec/shadingsys.cpp
lgritz: Right. If they know what's good for them, they won't ask for symbols that they didn't say would potentially be used as renderer outputs.

src/liboslexec/shadingsys.cpp
lgritz: Updated the comment.

fpsunflower:
LGTM

PR #459, opened 2015-01-17
New option: opt_merge_instances_with_userdata

lgritz:
The existing opt_merge_instances, when on (default), will merge shader instances within a group when they are completely isomorphic, having the same code, symbols, instance parameter values, and upstream connections. When two matching instances are found, they are combined into one.

However... we have discovered an interesting edge case, wherein the two retrieve userdata (geometry-bound parameters) and even though they are both retrieving what appears to be the same-named userdata, the particular renderer (not mine!) allows a level of flexibility where the renderer can bind different userdata ("hook the 's1' interpolated variable to this instance's 's' input parameter"), and that happens in a way where OSL per se doesn't really know that it's been bound to a different name than the name of the parameter. So if they have two instances, where different alternate userdata values are bound (again, in a way not discernable to OSL itself), it would be bad to merge the instances. Of course, they can turn off opt_merge_instances to avoid any spooky bugs, but that turns off all instance merging, even in nodes where no userdata is involved, and so the problem can't crop up.

Whew, okay, now that we're all on the same page, it should be clear how it's useful to have a new opt_merge_instances_with_userdata, which if turned off will prevent instances from merging if they need userdata (though other instances not using any userdata will still merge, assuming that the old opt_merge_instances is still turned on). The default for the new option is on, so in general nobody but this one site (or others in the future who need this) will need to care about it.

To implement this, of course we need to track this in the right places. Fairly straightforward. In fact, the code changes are smaller than the explanation above of why we're doing it.

Along the way, I also spotted a bug: although we were tracking whether shaders wrote global variables (P, N, Ci, and so on), and in theory were forcing instances that write globals to run unconditionally (rather than lazily), it turned out that we never set the flag. So now that's restored. It's not a total coincidence that I found this, because the places where we need to track whether an instance reads userdata is basically the same spot as where we should have been (and now are) tracking whether any globals are written.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,7 +352,8 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             group-outputs groupstring
             hyperb ieee_fp if incdec initops intbits isconnected
             layers layers-Ciassign layers-lazy layers-repeatedoutputs
-            logic loop matrix message metadata-braces miscmath missing-shader
+            logic loop matrix message mergeinstances-nouserdata
+            metadata-braces miscmath missing-shader
             noise noise-cell
             noise-gabor noise-gabor2d-filter noise-gabor3d-filter
             noise-perlin noise-uperlin noise-simplex noise-usimplex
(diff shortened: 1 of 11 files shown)


lgritz:
They say "untested code is broken code", and indeed my first stab at this PR did not work properly. Added test. Fixed bugs. Reposted PR.

PR #461, opened 2015-01-29
Fixes for compiling, linking and installing OSL on the Windows platform

darkhorse64:
Instead of a big large commit, I have tried to split the changes into consistent chunks.

* The first commit tells OSL to find missing implementations of math functions in OIIO (only for Visual Studio 2010). It requires the latest master or RB 1.5

* The second commit is needed for oslquery.dll to link. Windows requires complete implementations of classes. For the same reason, the list of files for some targets is completed.

* The third commit should now seem familiar :-)

* The fourth commit copies some settings from OIIO for linking with Boost and OpenEXR (only for Visual Studio)

* The fifth commit requires perhaps CMake experts advice because I am not familiar with this tool. I have found that, without these changes, CMake configure does not complete when dynamic build is selected. The RUNTIME keyword usage is also needed for INSTALL to copy all needed files. I have checked that it works for static and dynamic builds on Windows. I have taken inspiration from the oiio_install_targets macro for the changes, so I believe they should be OK, but caveat emptor !


lgritz:
Math funcs - LGTM.

SIMD stuff - LGTM.

CMake stuff - LGTM, pending my trying it on Linux and OSX to double check that nothing breaks.

OSLQuery - this one is tricky. This particular OSLQuery init method takes a ShaderGroup*. The idea was that it would never be needed from liboslquery itself, which is specifically designed to be used by apps that don't need or want to drag the entire OSL runtime (and LLVM!) with it. On the other hand, an app that already has the full OSL runtime may also want to call OSLQuery, and this init() method in particular, so we put that method in shadingsys.cpp. I fear that by moving it to OSLQuery, you have just broken this all in a slightly different way, which is to make liboslquery huge and forcing it to pull 10x more stuff with it.

As an alternative, maybe we can employ some trickery: Perhaps add one more source module to liboslquery, which contains just an empty stub implementation of init(ShaderGroup*). Then the fully-elaborated init() can live in shadingsys.cpp as before, and liboslexec can pull in all the other compilation units of oslquery (but NOT the one containing the init stub!). Does that make some kind of sense?

I would perhaps suggest splitting the OSLQuery parts into a separate PR, and making this one just be for the rest of the straightforward compilation issues (all of which I am willing to merge, after some quick tests).

darkhorse64:
I am OK with this alternative. I'll let you merge the commits you approve of. Then, I'll submit a new patch for OSLquery from this point.

lgritz:
OK, I have merged the four uncontroversial fixes and will close this pull request. Let's open a new one to address the OSLQuery problem.

PR #462, opened 2015-02-01
Fixes for linking and installing OSL on the Windows platform

darkhorse64:
The first commit is a leftover from a previous commit. Sorry about that

The second commit is the promised patch. Notice that dynamic linking requires to include `oslquery.cpp` in the list of source files for `liboslexec` (because `OSLQuery::open()` is needed). I have not provided a stub for this method because I guess a dependency from `liboslexec` towards `liboslquery` is not a concern.

Another concern with this patch is that `OSLQuery` symbols are now present into two different librairies which may not please all linkers. In any case, both builds (static/dynamic) are OK for Win32.

### DIFF
--- a/src/liboslexec/CMakeLists.txt
+++ b/src/liboslexec/CMakeLists.txt
@@ -27,6 +27,13 @@ if (NOT BUILDSTATIC)
         )
 endif ()
 
+# oslquery symbols used in oslexec
+if (NOT BUILDSTATIC)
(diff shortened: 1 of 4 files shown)


lgritz:
This looks reasonable. I'll take it from you that it works on WIn32, and I'll test on OSX and Linux tomorrow. If it checks out there, I'll merge it. Thanks for the fix!

lgritz:
Merged.

PR #463, opened 2015-02-04
Clear derivs in gettextureinfo output

aconty:
This was affecting shaders using the "averagecolor" attribute and corrupting
computations later on.


fpsunflower:
Good catch. Which floating point value was the shader fetching from the texture?

lgritz:
LGTM, good catch.

aconty:
It was "averagecolor"

PR #464, opened 2015-02-06
Another round of fixes for Win32

darkhorse64:
All these changes are quite small and should be self explanatory


lgritz:
LGTM, I will merge

PR #466, opened 2015-02-12
Small spelling mistake - missing 't' in alignment

sambler:
I'm guessing the default value is preventing the compiler from detecting this name mismatch?


lgritz:
LGTM.

Name mismatch? I don't think it's considered an error for the name of a parameter in the declaration of a function to be different than the name in its separate definition. The parameter TYPES must match, but not the names. But nevertheless, yes, it's a typo. I'll merge this, thanks for the fix.

PR #467, opened 2015-02-12
Do not optimize away layers connected to globals

darkhorse64:
When a layer is connected **only** to globals, it gets optimized away.

Let's take the swirl shader from the testsuite and break it into two parts

```
shader
swirl1 (float swirl = 0,
       output float s = 0, output float t = 0)
{
    float s1 = -1 + 2*u;
    float t1 = -1 + 2*v;

    float r = hypot (s1, t1);

    s = s1*cos(swirl*r) - t1*sin(swirl*r);
    t = s1*sin(swirl*r) + t1*cos(swirl*r);
}
```

```
shader
swirl2 (string texturename = "grid.tx",
       float s = u, float t = v,
       output color Cout = 0)
{
    Cout = texture(texturename, s, t,
                           "wrap", "periodic");
}
```

If you run this command
```
testshade -g 512 512 -od uint8 --o Cout out.tif --param swirl 2.0 --layer alayer swirl1 --layer blayer swirl2 --connect alayer s blayer s --connect alayer t blayer s
```
connecting outputs from the first layer to inputs of the second layer, you get the expected result.

out

But if you run 
```
testshade -g 512 512 -od uint8 --o Cout out.tif --param swirl 2.0 --layer alayer swirl1 --layer blayer swirl2 --connect alayer s blayer u --connect alayer t blayer v
```
writing the outputs of the first layer to (u,v), you get

out wrong

OSL does not take into account connections to globals and therefore considers that the first layer does nothing.

The change in ```shadingsys.cpp``` is not related to this issue but fixes out of range access to ```m_instoverrides``` vector. I am unsure if ```std::vector``` does not perform this check anyway but it feels safer.

### DIFF
--- a/src/liboslcomp/symtab.cpp
+++ b/src/liboslcomp/symtab.cpp
@@ -193,6 +193,16 @@ Symbol::print (std::ostream &out, int maxvals) const
         if (symtype() == SymTypeParam && ! lockgeom())
             out << " lockgeom=0";
     }
+    if (symtype() == SymTypeGlobal) {
+        if (connected())
+            out << " connected";
+        if (connected_down())
+            out << " down-connected";
+        if (!connected() && !connected_down())
+            out << " unconnected";
+        if (renderer_output())
+            out << " renderer-output";
+    }
     out << "\n";
     if (symtype() == SymTypeConst) {
         out << "\tconst: ";
--- a/src/liboslexec/instance.cpp
+++ b/src/liboslexec/instance.cpp
@@ -357,6 +357,14 @@ ShaderInstance::copy_code_from_master (ShaderGroup &group)
             "should not have copied m_instsymbols yet");
     m_instsymbols = m_master->m_symbols;
 
+    // Handle connections to globals
+    for (int c = 0;  c < nconnections();  ++c) {
+        const Connection &con (connection (c));
+        Symbol *dstsym (symbol (con.dst.param));
+        if (dstsym->symtype() == SymTypeGlobal)
+			dstsym->valuesource (Symbol::ConnectedVal);
+	}
+
     // Copy the instance override data
     // Also set the renderer_output flags where needed.
     ASSERT (m_instoverrides.size() == (size_t)std::max(0,lastparam()));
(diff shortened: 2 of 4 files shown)


lgritz:
Er, actually, I don't think that it's supposed to be valid to connect anything to a global. Shader parameters may be connected to other layers, but globals are globals, no connections. 

I think the real bug is that it should have rejected your attempt to connect the first layer's output params to the second layer's "globals."

darkhorse64:
There are several reasons for my mistake:
* I am familiar with Gelato Shading Language which supports this feature and I assumed that it was the same for OSL (wrongly; when I read the OSL manual, there is nothing that says connections to globals are allowed). 
* ```ShadingSystemImpl::decode_connected_param``` allows connections to globals and the code comments say the same thing.
* Connections to globals work if there is also at least one connection to a layer input in the list of connections. So, I assumed it was an oversight. It was only a matter of a few changes to make it work.

Anyway, please close the request if you think this is not a desirable feature for OSL.

darkhorse64:
````ShadingSystemImpl::decode_connected_param``` allows connections to globals and the code comments say the same thing. Moreover, connections to globals work if there is at least one connection to a layer input. Therefore, I assumed it was an oversight.

Anyway, if you think it is not a desirable feature for OSL, please close this request.

lgritz:
I think decode_connected_param deals with globals because it's possible to connect a global to a shader layer *input* param. I never intended to allow a shader layer *output* param to be connected to a global. I think that, in some sense, is the bug -- that it doesn't catch that nonsensical situation.

I think I want to disallow this, and there are basically two reasons that it would be especially problematic:

1. We have been operating on the assumption that only shader parameters can be connected to things upstream, and (your simple example notwithstanding) I can only assume that making this change would lead to many bugs in places inside the shading system and optimizer where we say if(connected) and thereafter assume it's a parameter and if(global) and thereafter never consider that it might be upstream connected. That would be a nightmare to untangle.

2. It may have (incorrectly) accepted the command, but the reason it seemed to work to you basically boils down to the fact that at some point in execution of that second layer, it will copy the results of the first layer's outputs into u and v. But there's just one u and v for the whole shader group, so any OTHER layers of the group that run subsequently will see the new values of u,v. It would work out for the trivial two-layer setup you used as an example, but for complex shader networks it sounds like a recipe for subtle shader bugs, especially considering that layer execution ordering is nondeterministic because of lazy evaluation. Literally from point to point in the scene, depending on which variables get pulled on by different conditional branches that may execute or not, the layer execution order may vary, and therefore so will if (or when) the globals get overwritten. Ick.

I think the best thing to do is close this PR (sorry), understand that connecting a shader output to a global is disallowed, and I'll look at the code and propose the right fix that would correctly flag that as an error.

darkhorse64:
Hmm ... Looks like you made a very compelling case not to do it

PR #470, opened 2015-02-18
Add RendererServices::supports()

lgritz:
This is just a placeholder for future functionality. I wanted to get the method into the class before the next stable release branch is made (hopefully soon).

### DIFF
--- a/src/include/OSL/rendererservices.h
+++ b/src/include/OSL/rendererservices.h
@@ -60,6 +60,19 @@ public:
     RendererServices (TextureSystem *texsys=NULL);
     virtual ~RendererServices () { }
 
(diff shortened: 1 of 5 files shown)


aconty:
LGTM, what if you make it return an integer so you can ask supports("uvsets") and it answers with the limit number, or something like that. Just thinking out loud trying to anticipate the needs for a query.

lgritz:
Quite right, I will make that change.

In fact, it's funny, just a few days ago that I was wishing that OIIO's ImageOutput::supports() return int rather than bool. Maybe I'll change that in both places.

lgritz:
Updated based on your suggestion.

aconty:
LGTM

PR #472, opened 2015-02-19
Texture handle optimizations.

lgritz:
TL;DR: If the runtime optimizer can figure out which texture file is needed for a texture call, it can ask for an opaque texture handle ahead of time and avoid the filename-to-handle lookup on each individual texture call.

Details:

Additions to RendererServices:
* Calls to retrieve opaque TextureHandle* from the underlying texture system,   given a known texture name.
* New (additional) entry points for texture(), texture3d(), environment(), and get_texture_info() that take a TextureHandle* rather than a filename.

Note that renderer implementations will need to add the new virtual functions to their subclass of RendererServices. It's ok to define them trivially (i.e., just do nothing and return false) as long as the renderer doesn't turn on the opt_texture_handle (see below).

New optimization option: "opt_texture_handle". For now, the default is false. A renderer should explicitly turn it on if it supports this feature with nontrivial versions of those aforementioned RendererServices methods.

Now then, when generating JIT code for a texture call, if at that point it can figure out exactly which texture is being referenced (if the filename is a known string constant, at that point in optimization), it will generate a call to an alternate texture function that takes the handle directly, rather than the filename.

The bottom line is that this speeds up texture calls by about 10% (using OIIO's texture system, for other texture systems YMMV).


src/liboslexec/llvm_gen.cpp
```diff
@@ -2168,11 +2177,12 @@ LLVMGEN (llvm_gen_texture)
...
     args.push_back (rop.ll.void_ptr (rop.llvm_get_pointer (Result, 1)));
     args.push_back (rop.ll.void_ptr (rop.llvm_get_pointer (Result, 2)));
-    if (alpha) {
-        args.push_back (rop.ll.void_ptr (alpha));
+    if (alpha || texture_handle) {
+        args.push_back (rop.ll.void_ptr (alpha    ? alpha    : rop.ll.void_ptr_null()));
         args.push_back (rop.ll.void_ptr (dalphadx ? dalphadx : rop.ll.void_ptr_null()));
         args.push_back (rop.ll.void_ptr (dalphady ? dalphady : rop.ll.void_ptr_null()));
-        rop.ll.call_function ("osl_texture_alpha", &args[0], (int)args.size());
+        rop.ll.call_function (texture_handle ? "osl_texture_handle" : "osl_texture_alpha",
```
fpsunflower: I'm a bit confused by this part -- it looks like `osl_texture_handle` deals with (optional) alpha but the regular `osl_texture` needs a different `osl_texture_alpha` call?

Is there a reason for the asymmetry? Should `osl_texture` and `osl_texture_alpha` be merged? or do we need a `osl_texture_handle`/`osl_texture_handle_alpha` pair?

lgritz: Yes, osl_texture and osl_texture_alpha should be merged. I decided not to replicate that unnecessary split for the handle routines, and I want to come back and merge the redundancies. Should I do it as part of the same review?

fpsunflower: Makes sense. I'd say its worth doing now while its fresh on your mind. Reading through the code the condition `if (alpha || texture_handle)` jumped out as being slightly odd.

lgritz: OK, I'll do that now and amend the review.

lgritz: Updated.

lgritz:
Is this ok now that I merged those texture functions?

fpsunflower:
LGTM

PR #474, opened 2015-02-27
Have the context lazily grab texture_thread_info.

lgritz:
For short shaders that don't use texture at all (and therefore don't
need the thread-specific pointer), this was causing undue overhead.


fpsunflower:
LGTM!

PR #475, opened 2015-02-28
Restore compatibility with OIIO 1.5

lgritz:
A few #if's do the trick.


lgritz:
Thomas Dinges confirmed that this fixed his issue, so I merged.

PR #476, opened 2015-03-02
Use rotl32 from fmath.h for sufficiently new OIIO

lgritz:
This is related to this OIIO PR: https://github.com/OpenImageIO/oiio/pull/1079

This OSL-side patch simply makes OSL work whether or not the new OIIO patch is applied. Works before, works after.


aconty:
LGTM

PR #477, opened 2015-03-02
Allow get_context() to supply a TextureSystem::Perthread* to use for that context.

lgritz:
More texture per-thread tweaking: allow get_context() to supply aTextureSystem::Perthread* to use for that context.

This is helpful for renderers that want to save a bit of time by allocating such things up front.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -400,7 +400,8 @@ public:
     /// lookup automatically (and at some additional cost).  The context
     /// can be used to shade many points; a typical usage is to allocate
     /// just one context per thread and use it for the whole run.
-    ShadingContext *get_context (PerThreadInfo *threadinfo=NULL);
+    ShadingContext *get_context (PerThreadInfo *threadinfo=NULL,
+                                 TextureSystem::Perthread *texture_threadinfo=NULL);
 
     /// Return a ShadingContext to the pool.
     ///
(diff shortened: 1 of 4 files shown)


fpsunflower:
I was actually hoping we could remove the concept of per-thread texture systems from OSL altogether since its really just an implementation detail of OIIO's texture system (which an OSL based renderer may not even use).

Any advanced user that cares about the performance can override the implementation of `texture()` and related calls in `RendererServices` (the default implementations can just rely on the auto fetching that OIIO does by default).

lgritz:
I like the idea of making it so that renderers that use OIIO's texture system can generally get away with just relying on the OSL default implementation of RendererServices::texture() and not need to override it at all. Yes, a renderer can override TS::texture() (as ours does), but I don't advocate that people do that unless they have a specific need, and don't wish to make that a requirement for the best performance. We also in many places assume that any renderer that does provide its own texture system nonetheless adheres to OIIO::TS's API, so that TS, if it doesn't have any important per-thread concepts, can just return a dummy pointer for TS::get_perthread_info(), and again, all will work fine.

This is a pretty minimal change, and a renderer is free to ignore the issue entirely (with very little penalty). There's some complexity on the OIIO side, and on the Arnold side (for us), but very little in OSL itself with regard to this issue. I think removing the last bit of it from OSL would more than make up for it with forcing extra complexity on the renderer implementation. So I'd like to stand by this review.

lgritz:
So, ok with this?

fpsunflower:
LGTM

PR #478, opened 2015-03-09
More texture handle tweaks (including RendererServices compat change)

lgritz:
I decided that the RendererServices methods that take texture handles should also still be passed the filenames. That really simplify things for some renderers that have cases internally where they need both.

Sorry about the compatibility break. There's been more churn on this matter in the past couple weeks than I like, but then again, this is the master branch where we say APIs are subject to change.


fpsunflower:
LGTM

PR #479, opened 2015-03-10
Fixed a gcc "possible undefined operation" warning.

est77:


lgritz:
Ick! Ha, obvious typo. I will merge this, thanks for the fix.

PR #481, opened 2015-03-16
Shader params that are arrays of unspecified length

lgritz:
You could previously declare OSL functions that have parameters that are arrays of unspecified length (like 'float foo[]'), but there was not an effective way to make shader parameters like that.

This patch, mostly the work of Max Liani from Animal Logic (with a little touch-up from LG), more or less fully enables shader parameters that are arrays of unspecified length. When an instance value is provided, which must itself be of definite length, it fixes the length of the shader parameter. Or, within a shader group, when an upstream shader's output parameter (of definite length) is connected to a downstream shader's parameter of unspecified length, it fixes the length. This provides some interesting new flexibility to write shaders that take arrays as parameters, without knowing the maximum length when you are writing or compiling the shader.

It's still a little dicey to alter one of these parameters for "re-rendering". Even if you mark it as "lockgeom=0", once the length becomes fixed, it's fixed, and you can't (without redeclaring the shader group and starting over) change its length again. So beware of situations like that.

Max - I hope you approve of my changes. The overall logic is still as you had it, I just fixed a couple edge cases to address some bugs that I noticed when I made the test cases, and added lots of comments.

### DIFF
--- a/src/liboslexec/instance.cpp
+++ b/src/liboslexec/instance.cpp
@@ -151,27 +151,14 @@ void *
 ShaderInstance::param_storage (int index)
 {
     const Symbol *sym = m_instsymbols.size() ? symbol(index) : mastersymbol(index);
-    TypeDesc t = sym->typespec().simpletype();
-    if (t.basetype == TypeDesc::INT) {
-        return &m_iparams[sym->dataoffset()];
-    } else if (t.basetype == TypeDesc::FLOAT) {
-        return &m_fparams[sym->dataoffset()];
-    } else if (t.basetype == TypeDesc::STRING) {
-        return &m_sparams[sym->dataoffset()];
-    } else {
-        return NULL;
-    }
-}
 
-
-
-const void *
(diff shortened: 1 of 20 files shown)


src/liboslexec/instance.cpp
```diff
@@ -241,31 +255,72 @@ ShaderInstance::parameters (const ParamValueList &params)
...
+                // that was an array of indefinite size. Magic! The trick
+                // here is that we need to allocate paramter space at the
+                // END of the ordinary param storage, since when we assigned
+                // data offsets to each parameter, we didn't know the length
+                // needed to allocate this param in its proper spot.
+                ASSERT (valuetype.arraylen > 0);
+                // Store the actual length in the shader instance parameter
+                // override info.
+                so->arraylen (valuetype.arraylen);
+                // Allocate space for the new param size and the end of its
```
aconty: typo: at the end

aconty:
LGTM

PR #482, opened 2015-03-19
Bug fix with closure code generation

lgritz:
I'm really not sure why this didn't ever come up earlier, but I found a minor bug in codegen for closure construction, wherein testing whether the passed argument matches the prototype used a more exact test than was necessary (== versus equivalent()), and if a vector and color temporary were coalesced in just the right way, it would say it was a bad argument type passed.

Also improved a nearby error message.

Also, while inspecting diffs of recent checkins to see what might have exposed this problem, noticed that the recent array checkin introduced an off-by-one error in oslc argument checking. Unrelated to this issue, and thus far unreported! But fixing it anyway.


fpsunflower:
LGTM

PR #483, opened 2015-03-20
pow fixes

lgritz:
* Update testsuite and refs to reflect simultaneous change in OIIO fmath.h
  fast_safe_pow to give higher accuracy to pow(x,1.0) and pow(x,2.0). (https://github.com/OpenImageIO/oiio/pull/1094)
* A couple more helpful special cases in the pow constant folding.


aconty:
What if it is not constant? Do we care?

lgritz:
We fixed the "non-constant" case (via OIIO::fast_safe_pow) to also special-case 1.0 and 2.0.

aconty:
Oh, nice. LGTM

lgritz:
It was spooky to have pow(x,1) != x or for pow(x,2) != x*x. I think that for other powers, even integer powers, it's ok to have a few ULPs of inaccuracy. But 1 and 2 are kind of special -- both very common, and likely to be noticed if their results are not exact.

(This is all because the underlying 'pow' implementation we use has a teensy bit of inaccuracy, in exchange for a HUGE improvement in performance. Not appropriate for avionics or life support systems, but fine for shaders.)

PR #484, opened 2015-03-20
More uniform path for python binary retrieval

lgritz:
Starting an executable python script with

    #!/usr/bin/env python

is preferable to

    #!/usr/bin/python

especially for those using a custom python that differs from that in /usr/bin. So once and for all, switch all the test scripts.


aconty:
LGTM, I always use that

PR #485, opened 2015-03-24
Fix subtle optimizer bug

lgritz:
Fix subtle optimizer bug -- under some very specific circumstances, an output parameter whose default value comes from "init ops" but which actually got an instance value, and was also connected to a downstream layer, could be eliminated but incorrectly, by passing along to the downstream shader the constant value it would have been assigned by default, rather than the constant instance value.

In addition to the bug fix itself, I also added some additional debugging aids.


aconty:
LGTM

PR #486, opened 2015-03-26
Constant folding sincos and matrix-from-floats

lgritz:
* Constant folding of the matrix constructor from 16 floats
* Constant folding of sincos.
* Noticed that the implementation of the vector version of sincos did not use fast_sincos underneath, as the other versions did, so fixed that.
* Along the way, it was helpful to change RuntimeOptimizer::turn_into_new_op() so that it takes all three arguments, instead of operating by keeping the result the same as before. It's only used in a handful of places, but had to touch those to add the extra argument.


fpsunflower:
LGTM

PR #487, opened 2015-03-26
Restore ability to use MCJIT (though not yet default)

lgritz:


lgritz:
Amended this pull request with more MCJIT-related work:

Primarily, remove all build-time MCJIT options. Instead, it automatically compiled in MCJIT support if it's using a new enough LLVM. Additionally, there is a "llvm_mcjit" runtime ShadingSystem attribute (defaulting to off). So now the logic is:

* LLVM < 3.3 : use old JIT
* LLVM 3.3-3.5 : use old JIT by default, but use MCJIT if "llvm_mcjit" attribute was set to nonzero.
* LLVM >= 3.6 : always use MCJIT (no more old JIT)

I don't yet recommend turning on "llvm_mcjit" -- there are some performance issues that I'm continuing to work through. But this makes it easier for me to experiment, in the hopes of coming up with a superior MCJIT solution.

fpsunflower:
LGTM

PR #488, opened 2015-04-01
More MCJIT speedups

lgritz:
MCJIT is a lot slower than old JIT. Part of that is inherent, and not addressed by this patch. But part is because MCJIT doesn't do lazy compilation of the bitcode in the module -- it does it all, and because we stuff a ton of things in the module (via llvm_ops.cpp - a large library of functions the shader might need), compiling the whole module is a large fixed per-shader-group cost.

This patch combats this in two ways:

1. A whole ton of stuff that is used rarely, is too big to inline, or is so expensive that inlining won't help appreciably, is moved OUT of llvm_ops.cpp and statically compiled, then just referenced by ordinary function call in the IR. The only point of having them written out in the module is to give them the opportunity to be inlined. So we remove the stuff where that's pointless.

2. Of what remains, we carefully traverse the module's functions and for each one, change its linkage designation from ExternalLinkage to LinkOnceODRLinkage. This has the effect of allowing it to still compile (or inline) if used within the module, but to understand that it won't be referenced from outside the module, and so it's free to be discarded if not actually used.

So these GREATLY reduce the "fixed overhead" of using MCJIT to very very close what we had for old JIT with lazy compilation.

(Unfortunately, we still need to compile the shader, and MCJIT is slower than old JIT, so for big shading groups, it takes quite a bit longer to JIT it. I'll address that separately with some more work. And also for this reason, old JIT is still used by default for LLVM versions old enough to still support it.)


src/liboslexec/opmatrix.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+OSL_NAMESPACE_ENTER
+namespace pvt {
+
+
+
+// Matrix ops
+
+OSL_SHADEOP void
+osl_mul_mm (void *r, void *a, void *b)
```
fpsunflower: Some of these matrix ops seem like they would be good candidates for inlining or even manual generation.

fpsunflower:
A few of these moves seem like they would affect old JIT as well. I assume the before and after benchmarks on old JIT stays comparable?

src/liboslexec/opmatrix.cpp
lgritz: Right. This is not the final word here. I'm still exploring whether any of these should be moved back to llvm_ops.cpp, or generated as LLVM IR directly (rather than as function calls), and also, for many of them, whether they should be replaced with SSE ops when on the right platform. (Matrix multiply is a good candidate for that.)

lgritz:
I've seen very little time difference when using old JIT, as this stuff has moved around. Within the timing noise. I think that the old JIT was faster at compiling stuff in llvm_ops.cpp, but worse at inlining it (I think the lazy bitcode stuff interfered with the inlining), and those two effects very nearly cancel each other out.

fpsunflower:
LGTM

PR #492, opened 2015-04-10
Fixed two bugs in generating code for closure optional keyword parameters

lgritz:
1. The test for matching types of the keyword parameters were too strict.  It used '==' test for exact type match, should have been equivalent() so that all the 3-vector types would be treated the same. This is because at this point in code generation, temporaries whose lifetimes do not overlap have been coalesced -- so maybe a point and color were merged, for example.

2. In the event of a keyword mismatch, it skipped the code generation for filling in that slot, leaving it uninitialized. Probably not good. So now we explicitly zero out the optional parameter for a keyword that is unknown.


aconty:
LGTM

PR #493, opened 2015-04-25
Fix bug with merging of instances

lgritz:
When two instances were being examined for possible merging, and
* a parameter was not actually used anywhere inside the shader at all
* but one or both instances have that parameter connected to downstream nodes
* the two nodes have difference instance values for the parameter
(Luckily, this is a very weird, unusual, particular circumstance.)

We weren't quite careful enough to not merge in those circumstances.
Tighten up some of the tests and preconditions for merging.

### DIFF
--- a/src/include/osl_pvt.h
+++ b/src/include/osl_pvt.h
@@ -617,6 +617,14 @@ public:
     bool everread () const  { return lastread() >= 0; }
     bool everwritten () const { return lastwrite() >= 0; }
     bool everused () const  { return everread() || everwritten(); }
+    // everused_in_group is an even more stringent test -- not only must
+    // the symbol not be used within the shader but it also must not be
+    // used elsewhere in the group, by being connected to something downstream
+    // or used as a renderer output.
+    bool everused_in_group () const {
+        return everused() || connected_down() || renderer_output();
+    }
+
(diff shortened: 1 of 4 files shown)


src/include/osl_pvt.h
```diff
@@ -617,6 +617,14 @@ class Symbol {
     bool everread () const  { return lastread() >= 0; }
     bool everwritten () const { return lastwrite() >= 0; }
     bool everused () const  { return everread() || everwritten(); }
+    // everused_in_group is an even more stringent test -- not only must
+    // the symbol not be used within the shader but it also must not be
+    // used elsewhere in the group, by being connected to something downstream
+    // or used as a renderer output.
+    bool everused_in_group () const {
```
aconty: Could it be that the symbol is connected down but then never used there?

lgritz: Sure. But there's not much to do about that if we haven't yet gotten to the use analysis in the later layer. But, a bit later, if it's discovered that the downstream layer doesn't use that parameter, the connection will be severed, and there will be a second chance post-optimization for layers to be merged.

aconty: makes sense

aconty:
LGTM

PR #494, opened 2015-04-28
fix for rhel7 at SPI

cmstein:
Build changes for rhel7 at SPI


lgritz:
LGTM, same comments as the oiio review

PR #495, opened 2015-04-29
Refactor methods to retrieve output symbol locations.

lgritz:
ShadingSystem::get_symbol() is what renderers call to retrieve the
memory location of renderer outputs, immediately after executing a
shader.  But it was more expensive than we ideally wanted.  Most of the
expense was in the name lookups, so we observe that if we can separate
the name resolution (expensive, but once per group) from the retrieval
of the address (once per shader execution, but cheap), we give renderers
a path to reduce per-shade overhead expense.

So we break it into two new methods: find_symbol(), which given a group
and the names will return an opaque pointer for the symbol that can be
reused for the lifetime of the shader group; and
symbol_address(ctx,sym), which retrieves the actual memory address of
the symbol for a particular shader execution of a context (inexpensive).
The old get_symbol() is now implemented in terms of these new
methods.

There is some fairly obvious refactoring and const-ifying to make this
happen.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -47,6 +47,7 @@ typedef ShaderGroupRef ShadingAttribStateRef; // DEPRECATED name
 struct ClosureParam;
 struct PerThreadInfo;
 class ShadingContext;
+class ShaderSymbol;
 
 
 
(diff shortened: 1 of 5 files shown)


aconty:
LGTM

PR #496, opened 2015-05-05
Vararray defaults

dheckenberg:
This changeset is to properly handle defaults for vararray parameters.  Without the patch, the following cases won't won't work/may crash if the vararray initializer has length > 1:
- oslinfo   (see testsuite/oslinfo-arrayparams)
- OSL arraylength() (see testsuite/vararray-connect, testsuite/vararray-param)
- OSL array access (see testsuite/vararray-default)

I also tightened up the parser grammar to error if a vararray output is specified (see testsuite/oslc-err-outputparamvararray).  This should probably be noted explicitly in the language spec.

A small patch it also required to OIIO for the TypeDesc struct.

### DIFF
--- a/src/include/osl_pvt.h
+++ b/src/include/osl_pvt.h
@@ -205,7 +205,15 @@ public:
 
     /// Returns the length of the array, or 0 if not an array.
     ///
-    int arraylength () const { return m_simple.arraylen; }
+    int arraylength () const { return abs(m_simple.arraylen); }
+
+    /// Is this a variable length array?
+    ///
+    bool is_varlen_array() const { return m_simple.arraylen < 0; }
+
+    /// Number of elements
+    ///
+    int numelements() const { return (m_simple.arraylen == 0) ? 1 : arraylength(); }
 
     /// Alter this typespec to make it into an array of the given length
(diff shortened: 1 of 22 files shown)


src/include/osl_pvt.h
```diff
@@ -205,7 +205,15 @@ class TypeSpec {
 
     /// Returns the length of the array, or 0 if not an array.
     ///
-    int arraylength () const { return m_simple.arraylen; }
+    int arraylength () const { return abs(m_simple.arraylen); }
```
fpsunflower: Per my comments on the OSL side -- would it make sense to instead keep the OIIO semantics the same, but add a field in here to indicate the default varlen size?

fpsunflower:
LGTM to the changes, aside from the discussion on where the concept of "default" length really belongs. This might just be a matter of extra docs or comments to clarify my mental model.

lgritz:
See #497 for my proposed alternate patch (which is still mostly Daniel's work, but I found a set of changes that let it be entirely self-contained in OSL land without any need for modifications to OIIO or semantic-bending use of TypeDesc).

lgritz:
Closing -- this was superseded by #497

PR #497, opened 2015-05-06
Daniel Heckenberg + LG = vararray defaults without OIIO changes

lgritz:
This is a proposed modification of #496 keeps track of the number of initializers in the Symbol structure itself, without the need for abusing TypeDesc or requiring any changes on the OIIO size.

(Aside: I wish GitHub had a way to propose modifications to PRs, on the PR itself. Or some kind of "multiple contributor PR" where people other than the original author can amend the patch. But it doesn't.)

If you guys like this, I will squash it into a single commit before merging. But I kept it separate for review so it's easy to see exactly what I changed compared to Daniel.

### DIFF
--- a/src/include/osl_pvt.h
+++ b/src/include/osl_pvt.h
@@ -203,9 +203,26 @@ public:
     ///
     bool is_array () const { return m_simple.arraylen != 0; }
 
+    /// Is this a variable length array, without a definite size?
+    bool is_unsized_array () const { return m_simple.arraylen < 0; }
+
+    /// Does this TypeSpec describe an array, whose length is specified?
+    bool is_sized_array () const { return m_simple.arraylen > 0; }
+
     /// Returns the length of the array, or 0 if not an array.
+    int arraylength () const {
+        DASSERT_MSG (m_simple.arraylen >= 0, "Called arraylength() on "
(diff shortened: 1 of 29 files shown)


src/include/osl_pvt.h
```diff
@@ -207,6 +207,14 @@ class TypeSpec {
     ///
     int arraylength () const { return m_simple.arraylen; }
 
+    /// Is this a variable length array?
+    ///
+    bool is_varlen_array() const { return m_simple.arraylen < 0; }
+
+    /// Number of elements
+    ///
+    int numelements() const { return std::max (1, arraylength()); }
```
fpsunflower: Both `arraylength()` and `numelements()` would be invalid to call if `is_varlen_array()` were `true` right? Maybe add a `DASSERT`?

lgritz: Yes, I like that. I'll add DASSERT to catch those cases where it would be misleading.

lgritz:
Pushed an update that adds some DASSERTs to ensure that we don't have subtle bugs. Ran a full testsuite in DEBUG mode and chased down the assertions one by one, now fully passes in debug mode!

I also renamed is_varlen_array to is_unsized_array, I like that terminology better.

fpsunflower:
LGTM

PR #498, opened 2015-05-08
Fix name spelling

darkhorse64:
I am French :-)


lgritz:
Ouch, so sorry! Bad fingers, bad!

lgritz:
fixed in master and RB-1.6

PR #499, opened 2015-05-08
Refactor of track_variable_lifetimes for vastly improved performance

lgritz:
For some really complex shaders, we profiled track_variable_lifetimes to be an important bottleneck in shader runtime optimization time.

Part of this function was an outer loop over every instruction, and any time it found a loop instruction, it would then loop over EVERY symbol, looking for symbols that were modified within the loop body, and marking the symbol in such a way as to expand its lifetime to fill the whole loop (because its value may be needed the next iteration).

This refactor still loops over all instructions, but we use a simple stack to keep track of any nested loops we are a part of. For each instruction, it looks at just the symbols that are arguments of that instruction, and for those, extends its lifetime to fill any loops the instruction is part of.

This changes the work from `number_of_insructions*k1*total_number_of_symbols` (where k1 is the fraction of instructions that are loop instructions) to `number_of_instructions*k2` (where k2 is the number of symbol arguments per instruction, almost always 2 or 3). I'm seeing about a 50x reduction in amount of time spent inside this function, for very complex shaders.

### DIFF
--- a/src/liboslcomp/oslcomp.cpp
+++ b/src/liboslcomp/oslcomp.cpp
@@ -105,8 +105,11 @@ namespace pvt {   // OSL::pvt
 
 
 OSLCompilerImpl *oslcompiler = NULL;
-
 OIIO::mutex oslcompiler_mutex;
+
+static ustring op_for("for");
+static ustring op_while("while");
+static ustring op_dowhile("dowhile");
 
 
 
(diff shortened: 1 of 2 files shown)


src/liboslcomp/oslcomp.cpp
```diff
@@ -1073,83 +1076,69 @@ OSLCompilerImpl::track_variable_lifetimes (const OpcodeVec &code,
...
+            // (the only arg) as used for the duration of the loop!
             ASSERT (op.nargs() == 1);  // loops should have just one arg
             SymbolPtr s = opargs[op.firstarg()];
+            int loopcond = op.jump (0); // after initialization, before test
+            int loopend = op.farthest_jump() - 1;   // inclusive end
             s->mark_rw (opnum+1, true, true);
-            s->mark_rw (op.farthest_jump()-1, true, true);
+            s->mark_rw (loopend, true, true);
+            // Also push the loop bounds for this loop
+            loop_bounds.push_back (std::pair<int,int>(loopcond, loopend));
```
fpsunflower: `std::make_pair` is a bit less verbose I think.

fpsunflower:
LGTM

src/liboslcomp/oslcomp.cpp
lgritz: Yes, I'll do that. 

make_shared is C++11, make_unique is C++14, I'd forgotten that make_pair was old school.

Some days, still working in C++98 makes me feel like I may as well be using Fortran 77.

PR #500, opened 2015-05-08
Remove unused insert_useparam and add_useparam from oslcomp

lgritz:
For a very long time now, we have not used the 'useparam' stuff on the oslc compiler side at all, instead doing it all as part of the runtime compilation and JIT. So remove the remaining clutter from liboslcomp.


fpsunflower:
LGTM

PR #501, opened 2015-05-08
Avoid locking during pointcloud read

fpsunflower:
Hide any unsafe access by hiding the Partio particle object behind some access methods for read and write access respectively

A small test from production renders roughly 20x faster.

### DIFF
--- a/src/liboslexec/pointcloud.cpp
+++ b/src/liboslexec/pointcloud.cpp
@@ -41,11 +41,10 @@ namespace { // anon
 
 #if USE_PARTIO
 
-class PointCloud {
-public:
-    PointCloud (ustring filename, Partio::ParticlesDataMutable *partio_cloud);
+struct PointCloud {
+    PointCloud (ustring filename, Partio::ParticlesDataMutable *partio_cloud, bool write);
     ~PointCloud ();
-    static PointCloud *get (ustring filename, bool read=true);
+    static PointCloud *get (ustring filename, bool write = false);
 
     typedef boost::unordered_map<ustring, shared_ptr<Partio::ParticleAttribute>, ustringHash> AttributeMap;
     // N.B./FIXME(C++11): shared_ptr is probably overkill, but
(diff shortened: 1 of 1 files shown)


src/liboslexec/pointcloud.cpp
```diff
@@ -294,7 +301,7 @@ RendererServices::pointcloud_get (ShaderGlobals *sg,
         return 0;
     }
     spin_lock lock (pc->m_mutex);
-    Partio::ParticlesDataMutable *cloud = pc->m_partio_cloud;
+    Partio::ParticlesDataMutable *cloud = pc->write_access();
```
aconty: Remind me why do we need write access here?

fpsunflower: Oops, I don't think we do - should be a `read_access()`.

fpsunflower:
Fixed `pointcloud_get`.

aconty:
LGTM

src/liboslexec/pointcloud.cpp
```diff
@@ -41,8 +41,7 @@ namespace { // anon
 
 #if USE_PARTIO
 
-class PointCloud {
-public:
+struct PointCloud {
```
lgritz: How come the switch from class to struct?

lgritz:
LGTM

src/liboslexec/pointcloud.cpp
fpsunflower: Because most of the fields are public?

fpsunflower:
Updated the review, really passes all unit tests now ...

src/liboslexec/pointcloud.cpp
lgritz: Generally speaking, I use struct only if it's something that could be done in a C struct -- a collection of publicly-accessible data fields that clients are expected to access directly (and, optionally, simple constructors/destructors).

Once it starts looking OO -- anything private, inheritance, methods other than ctr/dtr, etc. -- then I usually call it a class.

Maybe that's old school thinking? I suppose in C++ the only real difference is that it's default public rather than default private?

fpsunflower: Yeah that's the only difference, but I can change it back - its not a big deal.

lgritz: Not something I'm going to lose sleep over either. I was mainly just wondering why you did it.

PR #503, opened 2015-05-11
Big speedup loading large shaders -- look up symbols with hashmap, not linear search

lgritz:
Profiling was showing OSOReaderToMaster::instruction_arg as a hot spot when loading large shaders. This is because of a linear search to find symbol index from its name. I'm sure this made sense for short test shaders I was using when first bringing OSL online, but now we have immense shaders. Switch to a hash map for these lookups.

This results in over a 3x speedup of shader load time (which now is reported in the stats), on a scene with our production shaders.


fpsunflower:
LGTM

PR #505, opened 2015-05-20
Critical bug fix -- error in keeping track of which layers have been run

lgritz:
When a shader needs the value of a connected parameter, it runs the upstream layer it's connected to, if it hasn't been run already.  But, to keep a bzillion of these checks from occurring needlessly, it also tries to keep track (when building the LLVM ops) of which ones certainly have already been called, so that it isn't generating code to constantly check it at runtime. We were very subtly getting this wrong.  It's crucial not to do this for things pulled by init ops. They are inherently conditional, and also the m_layers_already_run set hadn't been cleared yet when they were run! So certain particular constructs of layer connections and init ops could result in it incorrectly skipping the execution of a layer by thinking it had already been run, even though it hadn't.

### DIFF
--- a/src/liboslexec/llvm_gen.cpp
+++ b/src/liboslexec/llvm_gen.cpp
@@ -167,6 +167,9 @@ BackendLLVM::llvm_run_connected_layers (Symbol &sym, int symindex,
         const Connection &con (inst()->connection (c));
         // If the connection gives a value to this param
         if (con.dst.param == symindex) {
+            // already_run is a set of layers run for this particular op.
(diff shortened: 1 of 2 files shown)


fpsunflower:
LGTM

I know its tricky to setup a small repro case, but given how sneaky the bug was - it definitely could use a unit test ...

lgritz:
Yeah, I'm trying. Given how critical this fix is, though, I'm thinking of trying to get this merged, and then double back and try to get a minimal repro case. The production case I'd been working with is enormous (but definitely fixed!).

PR #506, opened 2015-05-20
Keep track of layers that are the unused duplicate after an instance merge

lgritz:
The usual easy checks for if something is unused -- no code and no symbols -- aren't reliable in this case.

### DIFF
--- a/src/liboslexec/instance.cpp
+++ b/src/liboslexec/instance.cpp
@@ -58,7 +58,7 @@ ShaderInstance::ShaderInstance (ShaderMaster::ref master,
       m_layername(layername),
       m_writes_globals(false), m_run_lazily(false),
       m_outgoing_connections(false),
-      m_renderer_outputs(false),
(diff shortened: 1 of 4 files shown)


lgritz:
The reason that this is important is that depending on the order of certain optimizations versus instance merging, a discarded merged instance might look like one that needs to run unconditionally (because it seems to do something, yet has no outgoing connections), when in fact it should not be run at all. Eek!

aconty:
LGTM

lgritz:
Turns out the first version of this patch failed a couple unit tests. I've updated the PR with the minor fixes.

aconty:
What was the fix?

lgritz:
The special case in compute_run_lazily where it needs to make sure the last layer in the group isn't marked as lazy.

lgritz:
For some reason, it only came into play for 3 Arnold testsuite entries that involved rerendering.

aconty:
LGTM

PR #507, opened 2015-05-21
Improve some debugging printfs, assertions, and the like.

lgritz:
Odds and ends I found while debugging recently, but these are not the actual fixes.


fpsunflower:
LGTM

PR #508, opened 2015-05-21
C++11 work: Define OSL_BUILD_CPP11 to be 1 if OSL was built with C++11.

lgritz:
Note that this is subtly different than OSL_USING_CPP11, which is 1 if C++11 is being used at that very moment. (They will be the same for OSL, but may be different while building another package that uses OSL and/or its public headers, if one package was built with C++11 and the other not.)


aconty:
LGTM

lgritz:
I amended this PR with another set of fixes -- on our new OS at work, I got some compiler warnings only when in C++11 mode. It's extremely straightforward.

aconty:
LGTM, is c++11 more strict about int casts?

lgritz:
I don't think the standard prescribes warnings, only errors. So I'm not sure this issue is strictly dictated by C++11. But I only got this particular warning when using -std=c++11.

PR #509, opened 2015-05-21
Refactor adding symbols to be more clear and simplify

lgritz:
Minor refactor of runtime optimizer internals. This encapsulates some code that's repeated a few times into a single add_symbol() call. Also, that call makes sure to temporarily mark a newly added symbol as used potentially at any time so any choices based on lifetime analysis will be extra conservative until the next time that lifetimes for all symbols are re-analyzed (which happens with each optimization pass over the code).

### DIFF
--- a/src/include/osl_pvt.h
+++ b/src/include/osl_pvt.h
@@ -629,6 +629,18 @@ public:
         m_lastwrite  = std::max (m_lastwrite, lw);
     }
 
+    // Mark the symbol as always being read (and, if write==true, also
+    // that it's always written). Then is for when we don't know when
+    // it's read or written, but want to be sure it doesn't look unused.
+    void mark_always_used (bool write=false) {
+        m_firstread = 0;
+        m_lastread  = std::numeric_limits<int>::max();
+        if (write) {
+            m_firstwrite = 0;
+            m_lastwrite  = std::numeric_limits<int>::max();
+        }
+    }
+
     int firstread () const  { return m_firstread; }
     int lastread () const   { return m_lastread; }
     int firstwrite () const { return m_firstwrite; }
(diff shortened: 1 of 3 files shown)


src/include/osl_pvt.h
```diff
@@ -629,6 +629,18 @@ class Symbol {
         m_lastwrite  = std::max (m_lastwrite, lw);
     }
 
+    // Mark the symbol as always being read (and, if write==true, also
+    // that it's always written). Then is for when we don't know when
```
fpsunflower: typo: This is for ?

src/include/osl_pvt.h
```diff
@@ -629,6 +629,18 @@ class Symbol {
...
+    // it's read or written, but want to be sure it doesn't look unused.
+    void mark_always_used (bool write=false) {
```
fpsunflower: `write` stays false in all the uses so far?

So the newly added symbols always only read from, never written to?

lgritz: For the current uses, it doesn't matter. It just needs to look used, doesn't matter if read or written. For generality and future use, I implemented the function to also have the ability to mark it as written, though I don't have a specific use case at this time.

src/include/osl_pvt.h
lgritz: You're right. I think I had to read that about 10 times wondering what you were talking about before I saw the mistake.

src/include/osl_pvt.h
lgritz: The important bit here is that I noticed that the old code (in addition to being needlessly repeated) added the symbol but didn't mark it read or written at all. It doesn't seem to lead to any bugs at the moment, but some optimizations do look at when symbols are used, so I feared a source of future bugs. So I wanted to be sure that any time we added a symbol, it at least knew it was used, and with an expansive lifetime, so nothing could go wrong. The lifetime (and whether it reads or writes) will be tightened up at the end of the current pass, when everything is reanalyzed.

fpsunflower:
LGTM

PR #510, opened 2015-05-22
Add a small benchmark command line option to oslinfo

fpsunflower:
This can be handy to profile how long a given set of shaders take to load, and identify the slowest ones (by just piping the output to sort).

### DIFF
--- a/src/oslinfo/oslinfo.cpp
+++ b/src/oslinfo/oslinfo.cpp
@@ -43,6 +43,7 @@ Sony Pictures Imageworks terms, above.
 
 #include <OpenImageIO/strutil.h>
 #include <OpenImageIO/argparse.h>
+#include <OpenImageIO/timer.h>
 
 #include "OSL/oslquery.h"
 using namespace OSL;
@@ -51,6 +52,7 @@ using namespace OSL;
 static std::string searchpath;
 static bool verbose = false;
 static bool help = false;
+static bool benchmark = false;
 static std::string oneparam;
 
 
@@ -144,6 +146,7 @@ print_metadata (const OSLQuery::Parameter &m)
 static void
 oslinfo (const std::string &name)
 {
+    OIIO::Timer t(benchmark);
     OSLQuery g;
     g.open (name, searchpath);
     std::string e = g.geterror();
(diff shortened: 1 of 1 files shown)


src/oslinfo/oslinfo.cpp
```diff
@@ -235,6 +244,7 @@ main (int argc, char *argv[])
                 "-h", &help, "Print help message",
                 "--help", &help, "",
                 "-v", &verbose, "Verbose",
+                "--bench", &benchmark, "Benchmark shader loading time",
```
lgritz: My only concerns are related to this line. I'm not sure it's is a dealbreaker, but I want to bring it up and see what you think.

I worry a bit that "benchmark shader loading time" might give the (incorrect!) impression that this is a reliable way to benchmark how long it would take the shader to load in ShadingSystem. Maybe if it is described as "benchmark OSLQuery read time of the shader" it would be more clear?

Looking at other command line tools within OSL (and OIIO), when there is an option to print timing or memory information about what it's doing, it's never called --bench, it's always called --stats or --runstats. (I apologize for the inconsistency there. These days I prefer --runstats to distinguish stats about the running program from stats about the input,)

A single change may address both of these issues. Calling it --runstats and describing it as "print run statistics (time, etc.)" might both be consistent with related tools as well as be very clear that all you're doing is timing the operations of this program, not necessarily shader load time in a renderer.

fpsunflower: That's totally valid. I can change to `--runstats`.

How hard would it be to doctor the benchmarking to also simulate reading the shader into the renderer so we can benchmark that more easily?

lgritz: If you want to time what happens in ShadingSystem, then time in ShadingSystem, not OSLQuery. I think you could already accomplish this with

    testshade --stats myshader

and then the stats will already break out the time to read the shader master.

Apologies -- this is the one spot where I used "--stats". We should change that to "--runstats" for consistency. In contrast, oiiotool has both, because it's a separate operation to want statistics about the input images (--stats) versus time/memory/etc/ statistics about the running of the program itself (--runstats). We'll do a pass to make that consistent.

And I like --runstats better than --bench for two reasons: (1) it can incorporate things other than times; and (2) a proper benchmark would probably run lots of times and report the median or fastest time or whatever, whereas runstats just means tell me what happened on this one ordinary run.

lgritz:
LGTM with the name change.

fpsunflower:
Merged.

PR #511, opened 2015-05-25
Improve oso parsing time, especially OSLQuery/oslinfo

lgritz:
In an interactive application, we were hoping to eliminate the fractional second hiccup that occurred when the app had to use OSLQuery to parse an oso file on disk in order to comb through the metadata to build an appropriate UI. It seemed especially slow when the shader parameters had metadata containing very long strings.

I'm pleased to report that the changes in this PR give us around a 10x speed improvement for our big shaders, and up to 100x improvement for doctored shaders that we modified specifically to tickle this slowdown with the string metadata.

The changes are:

* Change (both in oslquery and oslc) our use of flex: instead of using the FlexLexer C++ wrapper, just use the usual straight C interface. The C++ wrapper uses IOstreams, I think that's why it was slow. Changing to C vastly speeds things up. That's 90% of the improvement.

* Change the way the lexer and parser dealt with hints. It was confusingly parsing the hints themselves into their constituent parts, then reasembling them into a big string to pass to OSOReader::hint(). Silly! So the lexer itself has been modified to just consider the entire hint to be one lexical object and pass it directly. Also simplified the way the lexer handles quoted strings generally to avoid creating so many intermediate ustrings.

* Some messing around with string_view instead of raw string pointers and ustrings for hint text.


fpsunflower:
Does the switch to the plain C interface of flex mean we put any symbols outside the namespace? Or are they all just private functions?

LGTM to the changes otherwise -- very nice speedup!

lgritz:
I'm going to guess that the flex-generated symbols using the C++ wrapper were also not inside our namespace. Also, the C++ wrapper I don't think prevented the C interface from being built underneath all along. So, this affair certainly shouldn't make the problem any worse.

I think the speedup may be even better than reported. I hadn't incorporates your timing patch, so those numbers I cited on our internal ticket were just total runtime of oslinfo, and the fact that the fast shader and slow shader reported the same time probably indicates that it's nearly all overhead of running the binary.

The story of this is amusing. I did not originally suspect IOstreams (if it's indeed that) to be a problem. Rather, I was reading the flex manuals, and there are some table size vs speed options that it kept saying were not available if you are using the C++ mode. So I switched to the C APIs only so that I could try those options. And it was a huge PITA that I almost abandoned a couple times. But once I finally got it building and working, suddenly it was 100x faster. I never even got around to trying out the options I was originally looking at. (I guess I could, but the time is already almost negligible, and I"m not sure how I'd accurately measure any further improvements.)

fpsunflower:
Cool to hear. We can worry about the symbols in other review, was just curious.

lgritz:
There's an option -- which we use -- to change symbol name prefix. That's how we can have an oso parser and an osl parser in the same binary. (yyparse becomes osoparse and oslparse, etc.) We might be able to do some preprocessor-fu to somehow incorporate the custom namespace and version. Let's tackle that separately.

lgritz:
I can't wait to hear how this speedup addresses your Katana-side hiccup.

PR #512, opened 2015-05-26
Improve uninitialized value detection mode ("debug_uninit")

lgritz:
Found a source of "false positives" -- we shouldn't do a check for
uninitialized values of the arguments to a "useparam" pseudoinstruction,
since by definition the parameters named may be uninitialized before the
useparam action itself copies the values from connected layers.

In the process, beef up the uninitialized value error message to point
not only to the OSL source file and line, but also to the specific group
and layer, and (mainly for me in debugging the feature) which
instruction it is.


fpsunflower:
LGTM

PR #514, opened 2015-05-27
Harmonize the way all binaries use --stats versus --runstats

lgritz:
--stats if applicable, prints statistics about the program inputs.

--runstats prints runtime statistics about the operation of the program itself (timing, memory, etc.)

oslinfo (and oiiotool) used this convention already, but testrender and testshade used --stats to print run statistics.


fpsunflower:
LGTM

PR #515, opened 2015-05-28
shade_image() utility runs a shader group to generate every pixel of an ImageBuf

lgritz:
At Chris's suggestion, this utility function encapsulates the kind of thing that testshade does (specifically, the shade_region, setup_shaderglobals, and save_outputs functions) -- running a shader group on each pixel of an image region and storing the designated output variables in an ImageBuf.

testshade can use this code path if you use the new --shadeimage flag, but this is not yet the default, because some behaviors change in a minor way and thus a full switchover would require a bunch of  ref images to be updated.

This makes it easy to write... other things... that have similar functionality of using shaders to generate a rectangular array of pixel or voxel values.


fpsunflower:
LGTM

My earlier suggestion was actually more related to the functionality of taking a small expression and turning into a small shader on the fly. But this is handy too.

lgritz:
Oh, I see. Yeah, I'll encapsulate that building block, too.

PR #516, opened 2015-05-31
'make USE_NINJA=1' uses Ninja back end rather than make

lgritz:
This greatly speeds up build times, especially for incremental and
null builds.

Full (non-ccache, post-make nuke) build improves from 0:47s to 0:43).
Full build, fully ccached, improves 14.7 to 11.6.
Null build (nothing changed) improves from 2.4s to < 0.2s.

Identical change on the OIIO side can be found here:
https://github.com/OpenImageIO/oiio/pull/1158

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,9 @@ ifdef PROFILE
 endif
 
 MY_MAKE_FLAGS ?=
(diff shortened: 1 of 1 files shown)


aconty:
LGTM.

PR #520, opened 2015-06-05
Make much less verbose builds, when not VERBOSE

lgritz:
Conform to CMake convention of setting mypackage_FIND_QUIETLY.
Do some pkgconfig with QUIET.
Fix some warnings that were clutting output.

When you build with VERBOSE=1, it's still as verbose as always.


lgritz:
The amount of console output diminishes further when using CMake 3.x -- it seems 2.8 still has a lot of code on its side that generates needless console output, which is better guarded in 3.x by appropriate "if" statements.

lgritz:
This also includes some other minor changes just to sync up some parts of the build scripts between oiio and osl, where equivalent functionality was expressed in different ways. A little bit of xxdiffing later and they are somewhat closer to each other.

PR #529, opened 2015-06-12
Add ShadingSystem::optimize_group to force a group to optimize/JIT

lgritz:
This can occasionally be handy to make it happen at a particular time, and not necessarily only as a side effect of shading.


fpsunflower:
LGTM

PR #530, opened 2015-06-12
Misc refactoring of testshade

lgritz:
Just realized that there was already a Filesystem call that does what was performed longhand here.

### DIFF
--- a/src/testshade/testshade.cpp
+++ b/src/testshade/testshade.cpp
@@ -100,7 +100,7 @@ inject_params ()
 {
     for (size_t p = 0;  p < params.size();  ++p) {
         const ParamValue &pv (params[p]);
-        shadingsys->Parameter (pv.name().c_str(), pv.type(), pv.data(),
+        shadingsys->Parameter (pv.name(), pv.type(), pv.data(),
                                pv.interp() == ParamValue::INTERP_CONSTANT);
     }
 }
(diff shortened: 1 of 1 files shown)


src/testshade/testshade.cpp
```diff
@@ -390,15 +390,7 @@ action_groupspec (int argc, const char *argv[])
...
-        std::ifstream in;
-        OIIO::Filesystem::open (in, groupspec);
-        groupspec.clear();
-        while (in.good()) {
-            std::string line;
-            std::getline (in, line);
-            groupspec += line + "\n";
-        }
-        in.close ();
+        OIIO::Filesystem::read_text_file (groupspec, groupspec);
```
fpsunflower: It seems a bit dangerous to send in a string view to a string that is being mutated (though the implementation details of this function make it ok I think).

lgritz: It doesn't modify the string until after it's done with the filename.

The contents of Filesystem::read_text_file work precisely like the lines I removed, except for using rdbuf rather than getline, which means the new way is faster.

lgritz: It does lead to an interesting rule of thumb though: if you write a function that takes a mutable string and a string_view, it is dangerous to use the string_view after you mutate the string, just in case somebody passed a view of that same string.

Though, this is a similar kind of weird you could get if you passed in one const std::string&, and another std::string& (meant to be modified). You won't have memory issues, but you can have "oops, the string I thought was read-only changed half way through the function."

fpsunflower:
Yeah, this particular usage is fine. Maybe worth pointing out in the implementation/comments for `read_text_file`?

LGTM!

PR #532, opened 2015-06-17
export shade_image symbol

est77:
It's needed when the cmake HIDE_SYMBOLS options is enabled, to use the new OIIO OSL plugin. Probably on windows too.


lgritz:
Of course, that makes sense. I will merge this.

lgritz:
Merged. (With extremely minor formatting changes.)

Let me ask you, though (everyone...), would this function be better as a method of ShadingSystem, rather than a standalone function?

est77:
Thanks!. It works now. I posted a pic of a procedural OSL texture inside a SeExpr on appleseed 
to OSL's mailing list.

I think I prefer shade_image as a free function. It's really a convenience function and not part of the interface clients use to integrate OSL into their apps.

PR #534, opened 2015-06-18
Fix bugs when deserializing shader group strings with unsized arrays.

lgritz:
When an unsized array parameter comes in (like: param int[] 1 2 3 4) we
don't actually know the parameter type yet, because the shader itself
hasn't been declared. So, first of all, handle the "int[]" part without
ending up with an uninitialized length, duh. But then, we can't
preallocate at that point, because we don't know how many ints there
will be, so grow the array as we read them, and then reset the type to
the right length that we actually saw.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -478,7 +478,8 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             texture-width texture-withderivs texture-wrap
             trailing-commas transform transformc trig typecast
             unknown-instruction
-            vararray-connect vararray-default vararray-param
+            vararray-connect vararray-default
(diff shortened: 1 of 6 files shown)


fpsunflower:
LGTM

fpsunflower:
Seems like it would be worth adding a few tests?

Do we allow the 0 length case? (for example `param int[] ;`)

lgritz:
I'm trying to construct a test that would fail before.

0 length case turns into [1] effectively -- and is then treated like any other case where fewer values are supplied than the declaration would require, and pads with zeroes.

lgritz:
Updated with test case!

fpsunflower:
LGTM

PR #535, opened 2015-06-19
Various cleanups on testrender and add "both" mode to microfacet closure

fpsunflower:
Standardize testrender pdf calculation (there was a mix of pdf vs. invpdf in there).

Add a mode to microfacet that lets reflection and refraction be described in a single closure (which is advantageous for sampling).


lgritz:
LGTM

fpsunflower:
Merged.

PR #536, opened 2015-06-19
Fix subtle optimizer bug with constant folding of format()

lgritz:
Undetected for quite some time, it seems that the constant folding of format() involved inserting an op, and the insert_code() call omitted a vital parameter that would instruct it to recompute the read/write ranges of symbols after the insertion. This could lead to a cascade of bad things later, and in particular could cause it to make mistakes where a symbol assignment might be removed if it (incorrectly) thought it would not be subsequently read.

It must have needed a very precise set of circumstances -- we've had this bug in the code for over a year, it never was reported as a bug, and we only discovered it when for other reasons we resurrected a scene and a version of our shader library from a couple years ago and found it had some weird errors. Somehow none of our current shaders seem to trigger it.

The real fix for the bug is just the one-line change at 1266 of constfold.cpp.

But along the way, I made a number of changes to help track down the problem and to make the code less error prone in the future:

* Changed the call sequence of insert_code to make the recompute_rw_range paramter required, not optional, and also made it (and also the 'relation' parameter) be typed enums rather than bool and int, respectively. So instead of calling ```insert_code (..., true, 1, ...)```, instead now you'd call ```insert_code (..., RecomputeRWRanges, GroupWithNext, ...)```. I like this idiom of eschewing bool parameters in favor of enums, and am starting to use it more frequently in new code.

* Improved the optimizer debug output so that the explanations of how ops were changed from one instruction to another, print the full argument list of the pre-changed version.

* Changed the error output for debug_uninit when uninitialized values are found, to say which shader it happened in, which argument to the instruction and the type (in addition to the name) of the argument that appears to be uninitialized.

### DIFF
--- a/src/liboslexec/builtindecl.h
+++ b/src/liboslexec/builtindecl.h
@@ -210,7 +210,7 @@ DECL (osl_dict_value, "iXiXLX")
 DECL (osl_raytype_name, "iXX")
 DECL (osl_range_check, "iiiXXi")
 DECL (osl_naninf_check, "xiXiXXiXiiX")
-DECL (osl_uninit_check, "xLXXXiXiXiXXii")
+DECL (osl_uninit_check, "xLXXXiXiXXiXiXii")
 DECL (osl_get_attribute, "iXiXXiiXX")
 DECL (osl_bind_interpolated_param, "iXXLiXiXiXi")
 DECL (osl_get_texture_options, "XX");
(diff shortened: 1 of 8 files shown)


fpsunflower:
LGTM!

Nice catch -- glad it turned out to be an actual bug. I think the new "enum" style args is a definite improvement.

PR #537, opened 2015-06-22
Add query for which getattribute() calls a shader group might make.

johnhaddon:
New shader group queries for "num_attributes_needed", "attributes_needed", "attribute_scopes", "unknown_attributes_needed". These follow the convention for the existing texture and closure queries.

The motivation for this is to allow OSL to be used in another non-rendering context in Gaffer - we have a prototype which is already massively outperforming Gaffer's existing expression engine, and this will allow further features and performance enhancements.

I would like to add unit tests for this - any advice as to how to go about that? On the same subject, I'm actually seeing 97 failures when I do "make test", both on master and on this branch, but I'm not sure how to dive in closer to see how/why things are failing - is more verbose output available somewhere? Presumably there's just a problem with my setup rather than them being genuine.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -263,6 +263,22 @@ public:
     ///   ptr userdata_offsets       Retrieves a pointer to the array of
     ///                                 int describing the userdata offsets
     ///                                 within the heap.
+    ///   int num_attributes_needed  The number of attribute/scope pairs that
+    ///                                are known to be queried by the group (the
+    ///                                length of the attributes_needed and
+    ///                                attribute_scopes arrays).
+    ///   ptr attributes_needed      Retrieves a pointer to the ustring array
+    ///                                containing the names of the needed attributes.
+    ///	                               Note that if the same attribute
+    ///                                is requested in multiple scopes, it will
+    ///                                appear in the array multiple times - once for
+    ///                                each scope in which is is queried.
+    ///   ptr attribute_scopes       Retrieves a pointer to a ustring array containing
+    ///                                the scopes associated with each attribute query
+    ///                                in the attributes_needed array.
+    ///   int unknown_attributes_needed  Nonzero if additonal attributes may be
+    ///                                  needed, whose names will not be known
+    ///                                  until the shader actually runs.
     ///   string pickle              Retrieves a serialized representation
     ///                                 of the shader group declaration.
     /// Note: the attributes referred to as "string" are actually on the app
--- a/src/liboslexec/oslexec_pvt.h
+++ b/src/liboslexec/oslexec_pvt.h
@@ -185,8 +185,22 @@ struct UserDataNeeded {
     }
 };
 
+// Struct defining an attribute needed by a shader group
+struct AttributeNeeded {
+    ustring name;
+    ustring scope;
 
+    AttributeNeeded (ustring name, ustring scope = ustring())
+        : name(name), scope(scope) {}
 
+    friend bool operator< (const AttributeNeeded &a, const AttributeNeeded &b) {
+        if (a.name != b.name)
+            return a.name < b.name;
+        if (a.scope != b.scope)
+            return a.scope < b.scope;
+        return false;  // they are equal
+    }
+};
 
 // Prefix for OSL shade up declarations, so LLVM can find them
 #define OSL_SHADEOP extern "C" OSL_LLVM_EXPORT
@@ -1359,9 +1373,12 @@ private:
     std::vector<TypeDesc> m_userdata_types;
     std::vector<int> m_userdata_offsets;
     std::vector<char> m_userdata_derivs;
+    std::vector<ustring> m_attributes_needed;
+    std::vector<ustring> m_attribute_scopes;
     std::vector<ustring> m_renderer_outputs; ///< Names of renderer outputs
     bool m_unknown_textures_needed;
     bool m_unknown_closures_needed;
+    bool m_unknown_attributes_needed;
     atomic_ll m_executions;          ///< Number of times the group executed
     atomic_ll m_stat_total_shading_time_ticks; ///< Total shading time (ticks)
 
--- a/src/liboslexec/runtimeoptimize.cpp
+++ b/src/liboslexec/runtimeoptimize.cpp
@@ -58,7 +58,8 @@ static ustring u_nop    ("nop"),
                u_pointcloud_write ("pointcloud_write"),
                u_isconnected ("isconnected"),
                u_setmessage ("setmessage"),
-               u_getmessage ("getmessage");
+               u_getmessage ("getmessage"),
+               u_getattribute ("getattribute");
 
 
 OSL_NAMESPACE_ENTER
@@ -2811,10 +2812,12 @@ RuntimeOptimizer::run ()
 
     m_unknown_textures_needed = false;
     m_unknown_closures_needed = false;
+    m_unknown_attributes_needed = false;
     m_textures_needed.clear();
     m_closures_needed.clear();
     m_globals_needed.clear();
     m_userdata_needed.clear();
+    m_attributes_needed.clear();
     for (int layer = 0;  layer < nlayers;  ++layer) {
         set_inst (layer);
         if (inst()->unused())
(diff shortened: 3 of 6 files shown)


lgritz:
This LGTM!

I can't think of anything to complain about, and definitely a useful feature and apt addition to our tracking of textures and closures. 

Do you want to just merge into master, or also back-port to 1.6?

I'd love to hear (maybe on osl-dev?) more about the use for Gaffer expressions, in particular how you actually present the interface or APIs. Dunno if you've been following some of the recent additions I made -- making a simple call that lets you run an OSL shader to generate a rectangle of pixel values, and an OIIO plugin that runs OSL to make pixels -- but these are similar themes and I've been thinking that the next step is to make a general eval() that's easy to use for other apps (probably good for the same kind of stuff as SeExpr, but with the benefit of actually being OSL and exactly matching the identical calculations in shaders). Maybe you've already done most of this work, or at the very least can advise on do's and don't's from the perspective of somebody who has already grappled with the issues.

As for tests...

In testshade.cpp, in test_group_attributes(), you'll see how we query and print the texture, closures, and globals. I think you should just add one more clause there to print the attributes (which you can test by hand)... and then maybe let me take it from there and I'll construct an automated test that makes sure the right things are printed for all similar features (attributes, texture, closures, globals), since we seem to lack a test for any of those at the moment.

You shouldn't get 97 test failures. I get 0! There's probably a single thing that's preventing a large swath of them from working for you. When you 'make test', at the end it should list the tests that failed. Pick one -- for the sake of argument, let's say it's aastep (and let's also supposed your "architecture" is "linux"). So

    cd build/linux/testsuite/aastep   (or any failed one, preferably a simple one)
    ../runtest

Now what does it say? The console output should at least help you distinguish between a crash, a mismatch of the resulting images, or a mismatch of captured output (the out.txt files don't match). Then we can proceed from there.

johnhaddon:
Great! A backport to 1.6 would be much appreciated - I've been working on master but that forced me up to OIIO master too, and deploying that at IE might be a bit of a headache. I can make a separate backport pull request once we're happy with this one.

I had already added the attribute output to testshape.cpp - does that look OK or is there more I should add?

I've done a little digging into the test failures, and a huge number of them seem to be down to failing to find `bin/idiff`. I can fix that by editing runtest.py to use `oiio_app( "idiff" )` at line 201- I'm pretty sure that's the right thing to do based on the other code in there. But I also have to modify `oiio_app()` itself to ignore the empty OPENIMAGEIOHOME passed in by whatever it is that launches this command :

```Command: "/usr/bin/env" "TESTSHADE_OPT=0" "OPENIMAGEIOHOME=" "python" "/home/john/dev/OpenShadingLanguage/gafferBuild/testsuite/runtest.py" "/home/john/dev/OpenShadingLanguage/gafferBuild/testsuite/aastep"```

Any thoughts on how that should be fixed? Is ignoring an empty OPENIMAGEIOHOME in runtest.py acceptable, or would the correct thing be to fix up that `/usr/bin/env` call, wherever it is originating from?

johnhaddon:
And here's a bit (probably too much) of background on what I'm up to....

Gaffer does all its computations using a dependency graph of nodes and plugs. Computations are performed by pulling on a plug in a particular context and getting a value back. A context is basically a dictionary containing arbitrary key/value pairs that the node can use as it sees fit. It usually contains at least the time to compute at, but also items like the tile index and channel name in the case of image processing, or object name in the case of scene processing. Any plug can be computed at any time in any context and from any thread.

As part of this we have an Expression node, which is intended to support multiple languages provided an engine is available for that language. The expression node takes care of managing the connections to input and output plugs on the external nodes referenced by the expression, and the engine is responsible for the following :

 - Parsing the expression
 - Returning the names of all input and output plugs referenced by the expression
 - Returning the names of all context variables referenced by the expression
 - Executing the expression, and setting output plug values from the result

 Our first expression engine was for python, and because python provides the ability to parse itself using the `ast` module, it's pretty easy to do the parsing part. We simply traverse the AST look for assignments and reads of the form `parent["NodeName"]["plugName"]` to determine what plugs are being referenced and reads of the form `context["variableName"]` to determine the contexts. When we want to execute the expression, we just build dictionaries with the plug and context values the expression requested, call `eval()`, and fish the output values from the dictionaries. This all works well, and because python is so general and we have bindings for everything in Gaffer/Cortex, you can do a lot with a python expression. But Python's not a quick language, and it's threading performance is dire.

That performance problem was our main motivation for implementing an OSL expression engine. Getting a prototype up and running has turned out to have been pretty straightforward, but has involved a bit more text-munging than might be ideal. We don't want the user to be aware they're writing a shader, and we want them to reference plugs and context variables with a natural syntax like so :

`parent.transform.translate.x = sin( parent.transform.translate.y ) * context( "frame" )`

But to actually make a shader OSL can parse and execute, we munge that into something like this :

`
float context( string name )
{
	float result = 0;
	getattribute( "gaffer:context", name, result );
	return result;
}

shader theExpression(
	float _parent_transform_translate_y = 0,
	output _parent_transform_translate_x = 0
)
{
	_parent_transform_translate_x = sin( _parent_transform_translate_y ) * context( "frame" );
}
`

Currently I'm just using some really dumb regular expressions to spot the references to plugs, and that's going to need some beefing up so it doesn't get confused by the contents of comments and strings. But the context references we can now get straight from the horse's mouth using the new "attributes_needed" query, and we can even catch unknown context queries, which are disallowed because Gaffer needs to know a bit about what's going on up front to maintain a computation cache. As a side note, I did first try building struct parameters to mimic the natural node/plug hierarchy, but OSL didn't seem quite ready for that, and as it turns out we need to munge the names anyway in case a plug name clashes with an OSL reserved word.

When we execute the expression, we use lockgeom=0, and a custom RendererServices which implements get_userdata() to retrieve the input values from the referenced plugs. And we implement get_attribute() to read from the context. Both the plugs and the context are provided via a simple little RenderState struct. I'm about to try using lazy_userdata to avoid evaluating inputs that don't end up being referenced at runtime, which will be nice.

This all currently looks like this :

https://github.com/johnhaddon/gaffer/blob/oslExpression/src/GafferOSL/OSLExpressionEngine.cpp

So really, given the fact that OSL is primarily designed for another purpose, this hasn't been too hard at all. The main thing I'd like to improve is the text munging side of things. As well as relying on naive regexes, I'm also doing nothing to remap compilation errors back to the original expression source, so the line numbers are all off. It'd be great if OSL could provide some assistance there, but I'm not sure what it would look like - maybe an `OSLCompiler::compile_expression_buffer()` call that assists with the preprocessing? That's not particularly straightforward though, because the knowledge that a plug has been requested would be in the OSL parser (basically it's any reference to an undeclared variable), but the knowledge of what type that variable should be auto-declared as (float/color etc) is on the Gaffer side, when we lookup the plug type by name. So its all quite entangled.

There's also a fair bit of boilerplate in RendererServices that seems unnecessary in this context, and you have to think hard about how you're going to map your particular use case to parameters, userdata, attributes and closures. When you look at it abstractly, you've got a fixed set of statically typed inputs/outputs (shader parameters), and a varying set of dynamically typed inputs (attributes) and outputs (closures). So I wonder if there might be an ExpressionServices abstraction that could present that more intuitively?

Ultimately though, I expect different folks and different host applications might have very different requirements. Now I've got my head round it, I really like having all the flexibility that RendererServices provides. And I'm becoming pretty convinced that OSL has a bright future as an expression language in a whole load of contexts.

lgritz:
John, #538 should address the issue of OSL trunk not building against OIIO 1.5. That was unintentional.

johnhaddon:
Thanks Larry - it's much appreciated.

dheckenberg:
This looks nice, John.

And I'd like to add my support for an elegant way to compile expressions directly, with error reporting etc.  We've done similar work in our scene description for mapping expressions to a shader with attribute bindings for variables etc.

johnhaddon:
Hi Larry. Just wanted to check that you're not waiting on anything else from me before merging? If so, let me know and I'll get on it...

lgritz:
No, sorry, John, it just fell through the cracks while I was looking into other issues.
I'm running tests now and will merge as soon as they're done.

lgritz:
Merged into master and RB-1.6.

johnhaddon:
Great - thanks Larry.

PR #539, opened 2015-06-24
Bug fix serializing shader groups with "unsized arrays"

lgritz:
If asked the serialize the group before it's been optimized & jitted,
information about the concrete size of instance values of "unsized" arrays
will be in the SymOverrideInfo, not yet copied to the instance's copy
of the Symbol itself. So get it from the right place.


fpsunflower:
LGTM

lgritz:
Minor amendment -- I also needed to get the data offset from the instance override info. Added this to the review. Now we're cooking with fire!

PR #541, opened 2015-06-25
Bug fix with unconditional running of earlier layers.

lgritz:
Subtlety: When generating code for the "group entry" (last) layer, which
is also responsible for executing all earlier layers that are marked as
needing to run unconditionally (non-lazily), it is important that the
group entry layer initialize its input parameters BEFORE it runs the
earlier unconditional layers and copies the upstream outputs to its own
inputs. Of course. Because, duh, if it first runs the upstream layer and
copies the upsteam outputs to its own input, THEN zeroes out or
otherwise initializes its inputs, it will overwrite the value set by the
earlier layer.


fpsunflower:
LGTM

PR #542, opened 2015-06-25
Enforce LLVM 3.5+'s need for C++11 with helpful build errors.

lgritz:


aconty:
LGTM, you could also just force USE_CPP11 when LLVM >= 3.5

lgritz:
I'm thinking that since enabling C++11 has far-reaching implications for people, I want throwing that switch to be explicit rather than have them surprised by our silently changing which C++ standard we are using just because we find a more recent copy of LLVM (which they may not know makes incompatible ABI demands).

aconty:
yeah, makes sense

PR #543, opened 2015-07-02
Fix for uninitialized static array in case of multithreaded access (vs2012)

UbisoftMotionPictures:
It seems that vs2012 does not guarantee that local static member are fully initialized when multiple thread access to the function. This commit fixes the problem we had with OSL arbitrarily not recognizing global symbols in shaders.


lgritz:
LGTM, I will merge this immediately.

lgritz:
Merged into both RB-1.6 and master.

PR #544, opened 2015-07-02
Support for closure arrays in shader inputs

UbisoftMotionPictures:
In RB-1.6 shaders with closure arrays inputs will compile but will not load at runtime (see test case)


lgritz:
Very nice!

LGTM, I will merge immediately.

lgritz:
Merged into RB-1.6 and master.

Before merging, I removed the out.opt.txt file from your patch, since it matched out.txt exactly. It's only necessary to have separate output references for opt and no-opt in cases where the output is expected to differ based on optimization level. (This can happen occasionally in cases where certain sequences of optimizations, even if they do not alter final numerical results of a shader, may change the order in which printf's are executed.)

PR #545, opened 2015-07-02
Fix for range checking asserts in vs2012 debug builds

UbisoftMotionPictures:
This code triggers a range checking assert in our vs2012 debug builds, also potential undefined behaviour.


lgritz:
LGTM, I will merge this.

lgritz:
Merged into both RB-1.6 and master.

PR #547, opened 2015-07-16
Keep and report stats on texture calls and handles

lgritz:
I did this to help me debug something (which turned out not to be a bug), but it's still a useful item to keep in the stats output.


fpsunflower:
LGTM

PR #548, opened 2015-07-16
Optimize away unused optional arguments to noise.

lgritz:
Suppose you have a call to noise:

    C = noise (noisetype, P, "bandwidth", val);

Let's suppose that noisetype is a shader parameter. We're passing the optional "bandwidth" paramter just in case noisetype is "gabor" at runtime. But if it's not (let's say it's "perlin"), we're needlessly computing whatever is in 'val'. In fact, "gabor" is currently the only noise type that uses any of the optional token/value arguments to noise.

This patch culls any optional noise arguments during runtime optimization, if the noise type can be deduced to be anything other than "gabor".

### DIFF
--- a/src/liboslexec/constfold.cpp
+++ b/src/liboslexec/constfold.cpp
@@ -2511,7 +2511,8 @@ DECLFOLDER(constfold_noise)
     Symbol *Name = rop.opargsym (op, arg++);
     ustring name;
     if (Name->typespec().is_string()) {
-        name = Name->is_constant() ? *(ustring *)Name->data() : ustring();
+        if (Name->is_constant())
+            name = *(ustring *)Name->data();
     } else {
         // Not a string, must be the old-style noise/pnoise
         --arg;  // forget that arg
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM

Did this improve performance much on our noise shaders?

lgritz:
I didn't even bother to measure, Chris, because I think it would take an unusual case to make it show up in the timing.

When it came time to JIT the post-runtime-specialized code, we were already skipping the step of copying the arguments into the noise options struct and passing it to the noise call, if it could establish at that time that the noise type could not have been "gabor".

This patch further augments the constant folding of noise to eliminate those arguments entirely in the middle of runtime specialization. This tightens things up just a tad more by eliminating any data dependencies on those values, and since this is in the middle of the optimization passes, subsequent optimization work may be able to exploit the disuse of those values. For example, if one of those optional arguments was the result of an expensive computation or was connected to an earlier layer (whose lazy evaluation may only have been triggered by the need for the value in the noise call), we may expect a performance improvement

Although I'll sleep better knowing this is now as tight as it can be, I'm not sure that we have any unusual cases like that in our shaders currently.

PR #550, opened 2015-07-27
Group deserialization: allow quoting of shader and layer names

lgritz:
I wrote a group spec by hand and accidentally quoted the shader name. Took me a couple awkward minutes to figure out the error message I was getting.


aconty:
LGTM

lgritz:
This actually ended up causing problems, so I backed it out. Haven't decided yet whether it's worth fixing, or I'll just abandon it.

PR #551, opened 2015-08-20
Remove deprecated SS::state()

lgritz:


src/liboslexec/shadingsys.cpp
```diff
@@ -205,14 +205,6 @@ ShadingSystem::ConnectShaders (string_view srclayer, string_view srcparam,
 
 
 
-ShaderGroupRef
-ShadingSystem::state ()
-{
-    return m_impl->state();
```
fpsunflower: Is there one more to remove? (the one inside `m_impl` ?)

lgritz: Ah, yes.

lgritz: pushed update

PR #553, opened 2015-09-06
Execution order refactor, and per-layer explicit execution order

lgritz:
Execution order refactor, and per-layer explicit execution order
    
Let's review what happens when you call
    
    shadingsys->execute (contextptr, shadergroup, shaderglobals);
    
This executes the given shader group on the context, with a specific set of shader globals. The group may be composed of many individual layers (shader nodes), and execution proceeds from first to last layer, executing those that must be run unconditionally and skipping those that can be run lazily. Of course, if a layer executes, it may trigger lazy execution of other not-yet-run layers. The unconditional layers are the layers that set global variables (like Ci or P), the layers whose outputs are connected to designated renderer outputs (AOVs and the like), and layers that have no connected outputs at all (under the theory that this must be the "root" node).

This all remains the same with this patch, but *additionally* there is another option where you can explicitly declare the "entry points" of a shader group,

    std::vector<const char *> entrynames;  // fill with layer names
    shadingsys->attribute (shadergroup, "entry_layers",
                           TypeDesc(TypeDesc::STRING, nentries),
                           &entrynames[0]);

and then execute the group layer by layer:
    
    shadingsys->execute_init (contextptr, shadergroup, shaderglobals);
    shadingsys->execute_layer (contextptr, shaderglobals, layernumber);
    ...
    shadingsys->execute_cleanup (contextptr);
    
The execute_init binds and prepares the context, execute_layer will run each layer (identified by number, and must correspond to one of the declared entry layers of the group). As before, other layers will execute lazily when their results are "pulled" by connections. However, if you are going to take the approach of manually calling layers, it's all up to you -- no layers (whose results aren't pulled downstream) will run unconditionally, even if they appear to set global variables or appear to be connected to renderer outputs.

The "all in one shot" approach still works just fine, and is probably the usual way of doing things. But for renderer that need even more control over which layers are executed and when, this provides an alternate calling sequence that gives more flexibility.

Some things that had to be done along the way:
    
* It used to be that the "last" layer was the presumed sole group entry point (or "root"), and its code was the one that did certain   initialization and the calling of the unconditionally-executed layers.   But now, the app can call any (declared entry) layer, in any order, so   we can't count on that initialization working. Instead, we now generate   an additional initialization function, which is called by execute_init.
    
* For lazy evaluation, it used to be that the *caller* of an earlier layer was responsible for marking it as run (so it wouldn't be run twice). But now nodes can get called by the app, so the "this layer ran" flag is set by the layer itself rather than the caller.
    
* For the reason immediately above, no layer's LLVM IR code is truly      empty; at the vest least, it marks the "this layer ran" flag. So the      prior test for "this whole group does nothing" no longer can count on      simply checking the post-LLVM optimized IR for being empty but for a      return statement. This forced us to move the 'group does nothing'      detection entirely out of LLVM generation and into the time of our      runtime optimization.
    
* Groups must track their declared list of possible externally-called      entry points.
    
* Augment testshade to be able to declare and call explicit entry      layers.
    
* New test: testsuite/layers-entry tests both the default and explicit      layer execution models, with a test carefully designed to have      different layer execution orders for the two calling strategies.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -446,7 +446,7 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             geomath getattribute-camera getsymbol-nonheap gettextureinfo
             group-outputs groupstring
             hyperb ieee_fp if incdec initops intbits isconnected
-            layers layers-Ciassign layers-lazy
+            layers layers-Ciassign layers-entry layers-lazy
             layers-nonlazycopy layers-repeatedoutputs
             logic loop matrix message mergeinstances-nouserdata
             metadata-braces miscmath missing-shader
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -416,7 +416,8 @@ macro ( TESTSUITE )
         # Run the test unoptimized, unless it's a "render_*" or "oslinfo_*"
         # test, which we don't bother testing unoptimized.
         if (NOT _testname MATCHES "^render" AND
-            NOT _testname MATCHES "^oslinfo")
+            NOT _testname MATCHES "^oslinfo" AND
+            NOT EXISTS "${_testdir}/OPTIMIZEONLY")
           set (_env TESTSHADE_OPT=0 OPENIMAGEIOHOME=${OPENIMAGEIOHOME})
           add_test ( NAME ${_testname}
                      COMMAND env ${_env} ${_runtest} )
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -179,9 +179,13 @@ public:
     /// Set an attribute for a specific shader group.  Return true if the
     /// name and type were recognized and the attrib was set. Documented
     /// attributes are as follows:
-    ///    string[] renderer_outputs   Array of names of renderer outputs
-    ///                                (AOVs) specific to this shader group
-    ///                                that should not be optimized away.
+    ///    string[] renderer_outputs  Array of names of renderer outputs
+    ///                                 (AOVs) specific to this shader group
+    ///                                 that should not be optimized away.
+    ///    string[] entry_layers      Array of names of layers that may be
+    ///                                 callable entry points. They won't
+    ///                                 be elided, but nor will they be
+    ///                                 called unconditionally.
     ///
     bool attribute (ShaderGroup *group, string_view name,
                     TypeDesc type, const void *val);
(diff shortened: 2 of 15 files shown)


aconty:
LGTM

fpsunflower:
How would one grab the output of a given layer after having executed it? Should this go through `renderer_outputs`?

lgritz:
Yeah, the usual mechanism:

1. After declaring the group, say which are outputs with attribute(group,"renderer_outputs",...).

2. For each output of a group, you can do a find_symbol(group, layername, symbolname) to get a opaque pointer for the symbol (good for all shades, all contexts, of that group -- so only do the name lookups once).

3. For each shade, after execution, retrieve the address where the output is actually stored using symbol_addres (contextptr, symbolptr), where symbolptr is the thing returned by find_symbol. This is very inexpensive just a couple pointer indirections, since all the hard parts (like the symbol name resolution) were in find_symbol.

fpsunflower:
From the renderer's perspective - I think it would be nicer to focus the API around those symbol handles (which would also further emphasize the two are meant to be used together).

Otherwise one would need to maintain a (redundant?) mapping of which layer to execute for each symbol. In particular -- one wants the layer by name, the other wants the layer by id. The numbering of layer ids seems like an implementation detail that should be hidden from the APIs.

Hopefully that information is already handy inside the symbol handle somehow and this could be a minor tweak to what you have already.

Basically instead of `execute_layer(layer_id)` I want to run: `execute_for_output(symbol_handle)` which internally would run: `execute_layer(symbol_handle->layerid)`.

lgritz:
I can do that!
Stay tuned.

fpsunflower:
Cool. Otherwise, I assume you've done benchmarks that show the code reorganization doesn't affect the current "default" path right?

If anything, it seems like moving the "tagging" of when a layer was run to that particular layer seems like it should remove a bit of bloat if a layer has many downstream connections.

lgritz:
So far I haven't noticed any perf increase (and have no reason to suspect there would be one -- I think it could only be if I really screwed something up), but I plan to benchmark on a few more production scenes before a final merge. I guess moving the tag to the callee could help if a node is connected to several downstream nodes. I'm not sure how often the "fan-out" will be large enough for this to add up to more than a few bytes.

lgritz:
I pushed an amendment to the patch that adds two new alternate execute_layer methods: in addition to specifying the layer by numeric index, you can specify it by name, or by ShaderSymbol* (which will run whichever layer is associated with the output named by the symbol).

I'm not done with constructing the testshade test for this yet, but thought I'd push the implementation so you could see what I've done and review the code part and let me know if this API will be satisfactory.

lgritz:
All that's really happening here is that it's dereferencing a (newly added) field in the ShaderSymbol that contains the index of the layer that the symbol belongs to. Since the "call by layer number" needs to exist anyway, at least underneath, I don't see why not expose it, and I can think of situations or apps that already know the layer numbers and shouldn't have to do a pointless lookup of a symbol.

fpsunflower:
Since the per layer calls only work in tandem with manual reads of output symbols, it seems the app almost has to have them lying around. I was just thinking it would save an extra function call. Since at the moment both are public I don't think it can inline.

LGTM on the patch overall though, this is just a minor detail.

lgritz:
Unit test added, minor corrections made :-), tests all work.

I stand by this review. Only pending running production benchmarks to be absolutely sure nothing has slowed down on the old default path.

PR #554, opened 2015-09-10
Closure cleanup

aconty:
Leave only the closure declarations needed by testrender
and testshade.

### DIFF
--- a/src/shaders/stdosl.h
+++ b/src/shaders/stdosl.h
@@ -470,17 +470,13 @@ string concat (string a, string b, string c, string d, string e, string f) {
 
 // Closures
 
+closure color emission() BUILTIN;
+closure color background() BUILTIN;
 closure color diffuse(normal N) BUILTIN;
+closure color oren_nayar (normal N, float sigma) BUILTIN;
 closure color translucent(normal N) BUILTIN;
-closure color reflection(normal N, float eta) BUILTIN;
-closure color reflection(normal N) { return reflection (N, 0.0); }
-closure color refraction(normal N, float eta) BUILTIN;
-closure color dielectric(normal N, float eta) BUILTIN;
-closure color transparent() BUILTIN;
-closure color microfacet_ggx(normal N, float ag, float eta) BUILTIN;
-closure color microfacet_ggx_refraction(normal N, float ag, float eta) BUILTIN;
-closure color microfacet_beckmann(normal N, float ab, float eta) BUILTIN;
-closure color microfacet_beckmann_refraction(normal N, float ab, float eta) BUILTIN;
+closure color phong(normal N, float exponent) BUILTIN;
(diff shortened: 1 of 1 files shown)


src/shaders/stdosl.h
```diff
@@ -488,73 +484,12 @@ closure color microfacet(string distribution, normal N, float alpha, float eta,
...
-closure color emission(float outer_angle) BUILTIN;
-closure color emission() BUILTIN;
+closure color reflection(normal N, float eta) BUILTIN;
+closure color reflection(normal N) { return reflection (N, 0.0); }
+closure color refraction(normal N, float eta) BUILTIN;
+closure color transparent() BUILTIN;
 closure color debug(string tag) BUILTIN;
-closure color background() BUILTIN;
 closure color holdout() BUILTIN;
-closure color subsurface(float eta, float g, color mfp, color albedo) BUILTIN;
```
lgritz: I feel like some kind of subsurface should be there.  

And what happened to the volume closures (absorption, isotropic, henyey_greenstein)? Where they never here? Ugh.

fpsunflower: This closure is super old - I don't think its been supported since we moved away from the interpreted code path.

aconty: But what kind of subsurface? Are there two renderers doing it the same way? I just left what's used by testrender and testshade.

aghiles: Even super old, we are actually using it. Do you have a proposition for a standard subsurface closure or we should just add our own ?

aconty: I think you should just add your own, that's what we do.

fpsunflower: Volume closures seem like good ones to keep on this side of the fence.

Ideally anything in this public header should be reflected in `testrender` though ...

aghiles: Understood. 

Although I think that this very simple closure declaration was quite useful. A simple/trivial implementation could be a BRDF like the one proposed in the original BSSRDF article.

aconty: I'm still voting for leaving what's used by testrender, as an example. I don't think there are any shaders that can claim to be render portable. Not for now, and it certainly doesn't look like a decision we can make in this thread (and be useful).

fpsunflower: I'm ok with keeping `subsurface()` in here for now.

The main goal for this was to just tidy up some of the more embarrassing sony renderer specific closures that were lying around.

We can defer to more discussion on the mailing list about what the base set of portable closures should really be. For now I would rather err on the side of having too few than too many.

aconty: Ok, I amended the commit putting back subsurface since it is 3 votes in favor.

fpsunflower:
LGTM

aconty:
merged

PR #555, opened 2015-09-11
Closure keyword refactor

fpsunflower:
This revamps how optional closure arguments are passed to the renderer. Instead of building a dynamic array filled with string, value pairs -- the renderer must provide a field in the parameter struct representing the closure.

This means that there is no per-shading point overhead on the renderer side to extract such arguments anymore.


aconty:
LGTM

lgritz:
Don't merge yet, I'm still reviewing

src/liboslexec/closure.cpp
```diff
@@ -76,7 +76,7 @@ print_component_value(std::ostream &out, ShadingSystemImpl *ss,
     else if (type == TypeDesc::TypeVector)
         out << "(" << ((Vec3 *)data)->x << ", " << ((Vec3 *)data)->y << ", " << ((Vec3 *)data)->z << ")";
     else if (type == TypeDesc::TypeString)
-        out << "\"" << ((ustring *)data)->c_str() << "\"";
+        out << "\"" << ((ustring *)data)->string() << "\"";
```
lgritz: What's this about?

lgritz:
LGTM. It's a nice simplification.

src/liboslexec/closure.cpp
fpsunflower: `c_str()` can return `NULL` (which happens to be the default value for any string keywords). Outputting `NULL` is documented as undefined behavior for c++ streams. Since this is not performance critical code, the string version seemed safer.

lgritz: Ah, good catch. Though it occurs to me that you could also use

  ...  << *((ustring *)data) << ...

ustring knows how to write itself to a stream (safely!) without needing ->string() or ->c_str().

fpsunflower: Good call, I didn't realize that overload was there. I'll amend this on my next pass through the code.

PR #556, opened 2015-09-11
Fresnel functions that return value

danilopeixoto5:
I propose new functions to return separately Fresnel reflectance (fresnel_reflection and fresnel_refraction). Many OSL users, such as Blender, have implemented proper functions just like these.


lgritz:
I'm nervous about the re-implementation of the old "compute both" fresnel as separate calls to fresnel_reflection and fresnel_refraction -- keeping in mind that fresnel_refraction is implemented by another call to fresnel_reflection. Might there be circumstances in which this leads to redundant computation that the compiler doesn't notice and eliminate?

I'm fine with adding the new fresnel_reflection and fresnel_refraction, but I think that the old fresnel() should not be changed, as it is the most efficient means of computing both when you need both.

lgritz:
Also, we will need a CLA from you before we can merge this.

danilopeixoto5:
I modified the Fresnel functions again relying on your comment:
```c++
void fresnel(vector I, normal N, float eta, output float Kr, output float Kt, output vector R, output vector T);
void fresnel(vector I, normal N, float eta, output float Kr, output float Kt);
float fresnel(vector I, normal N, float eta);
float schlick(vector I, normal N, float r);
```
`schlick` can be useful to set parameters such as "Reflectance at Normal" in Arnold aiStandard shader.

aghiles:
Not sure to understand what schlick has to do with "reflectance at normal". Simplifying the fresnel for such a scenario (for example the equation 22 from the GGX) gives a very simple:

      reflectance at normal = [(eta-1)*(eta-1)] / [(eta+1)*(eta+1)];

danilopeixoto5:
`schlick` is really inconvenient except for fast computing that isn't my intention.

danilopeixoto5:
May current `fresnel` be controlled with a "Reflectance at Normal" control?

aghiles:
Given a reflectance at normal, you can find out the ETA and you can use that in any fresnel. Anyway, this is unrelated to the task at hand.

danilopeixoto5:
So, `schlick` is really inconvenient except for fast computing that wasn't my intention.

PR #557, opened 2015-09-11
Merge type and id fields in ClosureColor

fpsunflower:
These two fields were redundant with each other. Merging them lines things up a bit more nicely internally (closure data starts at offset 16 instead of 20).

Also added some convenience methods to reduce the need for verbose casts in client code.


aconty:
LGTM

lgritz:
LGTM2

PR #560, opened 2015-09-22
Change logic of lazy evaluation -- unconnected layers are now lazy.

lgritz:
New logic: the only layers unconditionally run are (1) the LAST layer
(presumed to be the "root"), (2) any layer that sets a designated
renderer output, (3) any layer that sets a global, but only if
lazyglobals is off (it's on by default now). Everything else is run
"lazily", that is, only if and when its connected outputs are needed
downstream.

The old logic additionally considered a layer to be unconditional (i.e.,
not lazy) if it has NO outgoing connections, on the theory that is must
be the root, and also lazyglobals was not turned on by default. That
seemed like a recipe for inefficient shader evaluation, and it doesn't
look like any renderer needs or wants this.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -116,7 +116,9 @@ public:
     ///    int lazylayers         Evaluate shader layers only when their
     ///                              outputs are first needed (1)
     ///    int lazyglobals        Run layers lazily even if they write to
-    ///                              globals (0)
+    ///                              globals (1)
+    ///    int lazyunconnected    Run layers lazily even if they have no
+    ///                              output connections (1). For debugging.
     ///    int lazy_userdata      Retrieve userdata lazily (0).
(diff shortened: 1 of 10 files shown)


src/include/OSL/oslexec.h
```diff
@@ -116,7 +116,9 @@ class OSLEXECPUBLIC ShadingSystem
     ///    int lazylayers         Evaluate shader layers only when their
     ///                              outputs are first needed (1)
     ///    int lazyglobals        Run layers lazily even if they write to
-    ///                              globals (0)
+    ///                              globals (1)
+    ///    int lazyunconnected    Run layers lazily even if they seem to
```
fpsunflower: typo: to *have* no ... ?

lgritz: fixed

lgritz:
fixed comment typo

lgritz:
Chris says LGTH

PR #561, opened 2015-09-24
Fix tinyformat wchar build error on Windows

brechtvl:
`boost::filesystem::path.c_str()` returns a `wchar_t*` on Windows (at least in Boost 1.58). OIIO tinyformat doesn't support `wchar_t` by default unless you define `TINYFORMAT_ALLOW_WCHAR_STRINGS`, which OSL doesn't do.

This uses `OIIO::Filesystem::filename()` instead to get an `std::string`.

### DIFF
--- a/src/liboslcomp/ast.cpp
+++ b/src/liboslcomp/ast.cpp
@@ -30,12 +30,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <string>
 #include <sstream>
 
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM. We should be going through the Filesystem utilities rather than having any bare boost::filesystem references, if possible. I will merge this.

lgritz:
Merged into master and RB-1.6. Thanks for the fix.

I spotted a couple other uses of boost::filesystem. I'll fix those and prepare a separate PR.

PR #562, opened 2015-09-25
Replace bare boost::filesystem with strings & OIIO::Filesystem.

lgritz:
This had been the intent for some time, but a few references to
boost::filesystem had escaped our notice.


aconty:
LGTM!

PR #564, opened 2015-09-29
oslc bug fix for code generation of struct parameters copied whole.

lgritz:
When a shader param that was a struct had its value initialized by copying    another struct WHOLE, it neglected to set the "method" name for the init    ops for each field. This made it incorectly chalk up the init ops to the    struct param itself (which is just a dummy placeholder) instead of the    fields, and under some circumstances this could simply leave the struct    fields uninitialized.

It was fine all along if the struct param's fields were initialized one    by one -- in that case, the init ops were correctly associated with each    field individually. It was only the case of initializing by copying    another whole struct that was problematic.

The solution is to pass a flag to codegen_assign_struct indicating if we    are generating that code for a parameter initialization, in which case    it will do the proper calls to associate the init ops with the right    variable. This is just like we already did for the field-by-field case    handled in codegen_struct_initializers.

The fix to the main bug consists of the changes to `ast.h` and `codegen.cpp` (particularly line 575ff). The test of it is the material in `testsuite`. The rest are just some miscellaneous improvements to debugging (and a typo or two) that I found along the way in the quest to track this down.

### DIFF
--- a/src/liboslexec/runtimeoptimize.cpp
+++ b/src/liboslexec/runtimeoptimize.cpp
@@ -1827,7 +1827,7 @@ RuntimeOptimizer::eliminate_middleman ()
                             dsym = downinst->mastersymbol(c.dst.param);
                         const Symbol *usym = upinst->symbol(upstream_symbol);
                         if (! usym)
-                            usym = downinst->mastersymbol(upstream_symbol);
+                            usym = upinst->mastersymbol(upstream_symbol);
                         ASSERT (dsym && usym);
                         std::cout << "Removed " << inst()->layername() << "."
                                   << s.name() << " middleman for " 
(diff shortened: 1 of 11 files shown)


src/liboslexec/runtimeoptimize.cpp
```diff
@@ -1827,7 +1827,7 @@ RuntimeOptimizer::eliminate_middleman ()
                             dsym = downinst->mastersymbol(c.dst.param);
                         const Symbol *usym = upinst->symbol(upstream_symbol);
                         if (! usym)
-                            usym = downinst->mastersymbol(upstream_symbol);
+                            usym = upinst->mastersymbol(upstream_symbol);
```
aconty: Was this another bug?

lgritz: Yes. Completely asymptomatic, but I fixed it when I was in the vicinity and noticed it.

aconty:
LGTM

PR #565, opened 2015-10-02
Allow testshade to have st ranges other than 0-1. (For more flexible tests)

lgritz:
This makes it match how OIIO's "testtex" can set st ranges.

Also, I made -res a synonym for -g, I seem to always get that wrong, and now I can't remember why I would have used -g instead of -res.


fpsunflower:
LGTM

I guess it was `g` for grid?

lgritz:
yeah, probably grid.

PR #566, opened 2015-10-09
New OSL builtins for patterns: linearstep(), smooth_linearstep()

lgritz:
`linearstep(edge0, edge1, x)` linearly interpolates from 0 to 1 as x goes from edge0 to edge1.

Here is a picture of `linearstep(0,1,x)`:

linearstep

-----

`smooth_linearstep(edge0, edge1, x, eps)` is exactly linear (identical to linearstep) on [edge0+eps,edge1-eps] but has smooth ramp on and off between [edge0-eps,edge0+eps] and [edge1-eps,edge1+eps].

Below, linearstep(0,1,x) is blue, smoothstep(0,1,x) is green, smooth_linearstep(0,1,x,0.1) is red:

smoothlinearstep

`smooth_linearstep` has some interesting properties:
* It is exactly linear in much of its domain, unlike `smoothstep` which is always curvy.
* Its slope is 1 in the linear range, exactly matching `linearstep`, unlike `smoothstep` which is "steeper" than linear interpolation.
* But it does this be having its tails extend slightly beyond the edge0 and edge1 values.

### DIFF
--- a/src/doc/languagespec.tex
+++ b/src/doc/languagespec.tex
@@ -61,7 +61,7 @@ Language Specification
 Editor: Larry Gritz \\
 \emph{lg@imageworks.com}
 }
-\date{{\large Date: 6 May, 2015 \\
+\date{{\large Date: 9 Oct, 2015 \\
 %~ (with corrections, 2 Oct 2014)
 }
 \bigskip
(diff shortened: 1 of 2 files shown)


lgritz:
By the way, I made those graphs with OSL!  I will polish it a bit and then publish the graphing shader.

fpsunflower:
LGTM

PR #568, opened 2015-10-20
Make certain string ops robust to NULL arguments.

lgritz:
We're getting crashes, very hard to reproduce so I'm not 100% sure this is related, but there is circumstantial evidence that the crash happens inside osl_endswith_iss. Looking at the code, I can see how NULL arguments could be trouble to this and some related functions, so I'm beefing up handling of those cases.


fpsunflower:
LGTM

lgritz:
Cross fingers that it is the source of my doesn't-fail-in-the-debugger problem. :-)

PR #569, opened 2015-10-29
Fixes for OIIO 1.5 compatibility.

lgritz:
I'd inadvertently allowed a build break and a possible bug creep into
OSL 1.7 (dev) that depended on new features of OIIO 1.6 (dev). This
patch restores OSL trunk so that it will build correctly against the
stable release branch of OIIO which is currently 1.5.


aconty:
LGTM

PR #571, opened 2015-10-29
smooth_linearstep: Ensure that it's as efficient as possible when eps == 0

lgritz:
Try to make sure that when eps is detected to be constant and 0, it's as
close as possible to being as efficient as linearstep, by rearranging
the various clauses of the 'if'.

### DIFF
--- a/src/shaders/stdosl.h
+++ b/src/shaders/stdosl.h
@@ -430,11 +430,11 @@ float smooth_linearstep (float edge0, float edge1, float x_, float eps_) {
     float eps = eps_ * width_inv;
     float x = (x_ - edge0) * width_inv;
     float result;
-    if      (x < -eps)    result = 0;
-    else if (x < eps)     result = rampup (x+eps, 2.0*eps);
-    else if (x < 1.0-eps) result = x;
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM, but have you been able to measure a difference from this change?

fpsunflower:
(The original code was nicer to read)

lgritz:
I haven't measured, but I know what code will be generated underneath. Any way you slice it, the small transition regions should be the least common cases it hits, and therefore at the end so that common cases have fewer condition evaluations and branches. And when eps == 0, especially, this ordering avoids executing pointless condition checks.

lgritz:
One reason why measuring is hard is that I think there are two wildly different use cases here:

1. x is usually in the linear region, it's primarily used as a linear ramp that happens to not have sharp corners.

2. x is usually outside the edge regions entirely (often << 0 or >> 1) -- i.e. it's a "fuzzy edge" .

This means that there is no single most efficient ordering for all use cases. But for either of these, you want the ramp transition regions to be lowest priority.

fpsunflower:
Makes sense.

PR #572, opened 2015-10-29
linearstep & smooth_linearstep: handle degenerate edge0==edge1 case.

lgritz:
When edge0 == edge1, these functions should behave like step(edge0,x).

Also add a testsuite that exercises these two new functions, including the degenerate cases.


fpsunflower:
LGTM

PR #575, opened 2015-11-03
Record oslc options in comment in the oso file

lgritz:
Makes debugging a little easier for me, I don't need to wonder what command line options (defines or include directories, etc.) were used.


fpsunflower:
LGTM

PR #576, opened 2015-11-04
A collection of optimizer debugging aids

lgritz:
1. Instead of hard-coding 10 passes over each layer when optimizing, let it be a SS attribute. (This is for me, I don't expect or want users to mess with it.)

2. Improved optimizer debugging messages -- refactored in a way that made it easier to have the little "I changed instruction FOO to instruction BAR because ..." debug messages to note which shader file and line that corresponds to.

3. Make an easy way for the JITer to insert code that will generate a runtime error or warning.

4. When encountering a request to make a closure that the renderer doesn't know about, instead of being a JIT-time error, do the thing from (3) above to make it a shader runtime error, which of course will only trigger if it actually tries to execute that line. This removes some spurious error messages in certain circumstances where we make closures obsolete but still want to run old shaders that reference them.


fpsunflower:
LGTM

Number 4 in particular will be nice to have and avoid having us register dummy closures.

PR #578, opened 2015-11-04
isconstant(expr) returns 1 if the expression can be turned into a constant

lgritz:
isconstant(expr) returns 1 if the expression can be turned into a constant
at runtime (with full knowledge of the shader group's parameters values and
connections).

This is primarily a debugging aid for advanced shader writers to verify
their assumptions about what expressions can end up being constant-folded
by the runtime optimizer. Also an aid for me in debugging that the runtime
optimizer is behaving as I expect.

### DIFF
--- a/src/liboslexec/llvm_gen.cpp
+++ b/src/liboslexec/llvm_gen.cpp
@@ -3583,6 +3583,19 @@ LLVMGEN (llvm_gen_luminance)
 
 
 
+LLVMGEN (llvm_gen_isconstant)
+{
+    Opcode &op (rop.inst()->ops()[opnum]);
+    ASSERT (op.nargs() == 2);
+    Symbol &Result (*rop.opargsym (op, 0));
+    ASSERT (Result.typespec().is_int());
+    Symbol &A (*rop.opargsym (op, 1));
+    rop.llvm_store_value (rop.ll.constant(A.is_constant() ? 1 : 0), Result);
+    return true;
+}
+
+
+
 LLVMGEN (llvm_gen_functioncall)
 {
     Opcode &op (rop.inst()->ops()[opnum]);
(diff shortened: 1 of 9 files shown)


src/liboslexec/llvm_gen.cpp
```diff
@@ -3583,6 +3583,19 @@ LLVMGEN (llvm_gen_luminance)
...
 
 
+LLVMGEN (llvm_gen_isconstant)
+{
+    Opcode &op (rop.inst()->ops()[opnum]);
+    ASSERT (op.nargs() == 2);
+    Symbol &Result (*rop.opargsym (op, 0));
+    ASSERT (Result.typespec().is_int());
+    Symbol &A (*rop.opargsym (op, 1));
+    rop.llvm_store_value (rop.ll.constant(A.is_constant() ? 1 : 0), Result);
```
fpsunflower: Should this function return something different if its constant here vs. constant at runtime-optimization time?

Otherwise you might think its properly flagged as constant when in reality it still isn't able to fold it for some other reason?

fpsunflower: Or maybe it doesn't matter -- this is just testing if OSL knows its a constant, not the status of what optimization passes did with that information ...

lgritz: I don't think there's a distinction to be made here. It is "constant" if and only if the runtime analysis/optimizer figures out that it's a constant.

Sometimes we figure this out very early, sometimes we may make various optimizations before realizing it. 

Any time we are trying to do a direct constant-fold on the call, we can turn it into a simple "1" if at that point it's known already to be constant. If it's not marked as constant yet, we don't know that it won't be discovered to be constant later, so we can't fold a "0".

When all other optimization is done and it's time to generate LLVM IR, then we generate a 1 or 0 based on whether or not it's known to be constant. It may be that in fact, there is no optimization that could possibly have turned something into a constant that happened after the constfold but before this point, so it will always be 0. But I'm not sure, and it may change in the future, so it seems like the safe (and in no way expensive) approach is to check again rather than assume that it's non-constant at this point.

lgritz: I should clarify, isconstant(x) is trying to tell you whether the expression is a *known* (to the optimizer) constant value. It is NOT telling you if the expression is *uniform* (having the same value at all points being shaded), which is not the same as knowing its actual value with certainty.

fpsunflower:
Makes sense. LGTM

PR #579, opened 2015-11-04
Fix namespace conflict after addition of recent OIIO::debug()

lgritz:


lgritz:
never mind. I think this just shows I need a better name on the other side.

ignore this.

PR #581, opened 2015-11-16
C++11 compatibility clean-up

lgritz:
Use OSL_CPLUSPLUS_VERSION rather than OSL_USING_CPP11. The former is better for long-run distinguishing between 03, 11, 14, 17, and so on, whereas the latter is shortsightedly binary.


aconty:
LGTM. I understand I still need to set USE_CPP11 when compiling with clang 3.5, right?

lgritz:
No, you can compile the code base with clang 3.5 just fine, with either C++03 or 11.

But if you build OSL against LLVM 3.5, though, you must use C++11 because LLVM's API headers now require it.

I hope that's clear. The distinction is between which compiler you use to build OSL (just as you'd choose which gcc, if you used gcc), versus which LLVM you link against for OSL's JIT facilities.

PR #582, opened 2015-11-16
Use boost::thread_specific_ptr, don't rely on it in OIIO namespace.

lgritz:
The reason for this is that I want to take the inclusion of boost::tsp out of the OIIO thread.h public header file. These are the only places in OIIO that we use it.


aconty:
LGTM

PR #583, opened 2015-11-17
SS attribute 'connection_error' controls whether bad connection is a true error

lgritz:
This was helpful for me in a debugging situation.


fpsunflower:
LGTM

PR #584, opened 2015-11-17
Refactor of optimize_ops and debugging info

lgritz:
This doesn't change behavior at all (except for debugging messages), but it's a refactor that is necessary for what's coming next.

This is made of 4 commits, and it may be helpful to view the diffs for each of the commits separately, as they are independently very understandable, but as an aggregate are a bit of a mess. (But they do go together, so it's silly to break them into four sequential PRs.)

*  Move the guts of ShaderInstance::print() to RuntimeOptimizer:printinst(), since the only place it was called was by RuntimeOptimizer and I wanted to change what it prints in a way that needs some RuntimeOptimizer internals. In particular, when printing the instructions, it now will make it clear which instructions are the heads of basic blocks, something that was helpful for me for some other stuff coming down the road.

* Refactor optimize_ops to split out the optimization of assignments into a separate method, and also slightly different calling of peephole2. This is just splitting one big method into two pieces.

* Further simplify optimize_ops by moving the 'empty if' handling into constfold_if.

* Refactoring of debugging output by funnelling it all through debug_opt rather than having individual output to cout. The main reason this is helpful is that debug_opt appends the debugging messages with the stamp of the original source file and line. That makes it a lot easier for me to look at the debug output and understand how individual optimizations relate to shader source code lines.


aconty:
LGTM!

PR #586, opened 2015-11-21
Support OSLQuery from a given bytecode

Nazg-Gul:
This is handy feature to have for applications which has OSL in-memory bytecode
cache but still needs to run OSLQuery on such shaders without creating temporary
files on the disk.

Currently only explicit query.open_bytecode() call is supported, this is because
it's not really possible to distinguish filepath from memory buffer being passed
to the constructor.

### DIFF
--- a/src/include/OSL/oslquery.h
+++ b/src/include/OSL/oslquery.h
@@ -97,6 +97,14 @@ public:
     bool open (string_view shadername,
                string_view searchpath = string_view());
 
+    /// Get info on the shader from it's compiled bytecode.  Return
+    /// true for success, false if the shader could not be found or
+    /// opened properly.
+    /// Meant to be called from an app which caches bytecodes from
+    /// it's own side and wants to get shader info on runtime without
+    /// creating a temporary file.
+    bool open_bytecode (string_view buffer);
+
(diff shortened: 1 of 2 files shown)


src/include/OSL/oslquery.h
```diff
@@ -97,6 +97,14 @@ class OSLQUERYPUBLIC OSLQuery {
     bool open (string_view shadername,
                string_view searchpath = string_view());
 
+    /// Get info on the shader from it's compiled bytecode.  Return
+    /// true for success, false if the shader could not be found or
+    /// opened properly.
+    /// Meant to be called from an app which caches bytecodes from
+    /// it's own side and wants to get shader info on runtime without
+    /// creating a temporary file.
+    bool open_bytecode (const std::string &buffer);
```
lgritz: Let's change that to a string_view, gives more flexibility to the caller.

lgritz:
LGTM, with the suggestion I made.

Nazg-Gul:
Ok, hopefully the patch was successfully updated to use string_view.

lgritz:
Yeah, looks good, thanks! I will merge this.

PR #589, opened 2015-12-01
Change isconnected() to reveal down-connections

lgritz:
Old: isconnected(param) returns 1 if the param is connected to an upstream
layer, otherwise 0.

New: isconnected(param) returns 1 if the param is connected to an upstream
layer, 2 if connected to a downstream layer, 3 if both, 0 if neither.

### DIFF
--- a/src/liboslexec/runtimeoptimize.cpp
+++ b/src/liboslexec/runtimeoptimize.cpp
@@ -2353,10 +2353,9 @@ RuntimeOptimizer::resolve_isconnected ()
                 ASSERT (fieldsymid >= 0);
                 s = inst()->symbol(fieldsymid);
             }
-            if (s->connected())
-                turn_into_assign_one (op, "resolve isconnected() [1]");
-            else
-                turn_into_assign_zero (op, "resolve isconnected() [0]");
+            int val = (s->connected() ? 1 : 0) + (s->connected_down() ? 2 : 0);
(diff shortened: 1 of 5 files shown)


src/liboslexec/runtimeoptimize.cpp
```diff
@@ -2353,10 +2353,9 @@ RuntimeOptimizer::resolve_isconnected ()
                 ASSERT (fieldsymid >= 0);
                 s = inst()->symbol(fieldsymid);
             }
-            if (s->connected())
-                turn_into_assign_one (op, "resolve isconnected() [1]");
-            else
-                turn_into_assign_zero (op, "resolve isconnected() [0]");
+            int val = (s->connected() ? 1 : 0) + (s->connected_down() ? 2 : 0);
```
aconty: I guess it cannot be connected() and connected_down() at the same time. Maybe connected() should be connected_up(), but I'm not suggesting you rename the whole thing now. Maybe make this assignment a bit clearer like

val = s->connected() ? 1 : (s->connected_down() ? 2 : 0) : 0;

but I guess it's a matter of taste.

lgritz: It can, in fact, be both upconnected and downconnected.

aconty: Oh, sorry, I misunderstood

aconty:
LGTM

PR #590, opened 2015-12-02
Fix compiler warnings related to int vs size_t

lgritz:
Have I ranted lately about how much I hate size_t?


aconty:
LGTM

Agreed about size_t, but I wonder what a better solution would have been for generic containers.

lgritz:
I think the ISO committee is sick of it too, and wishes it could be simply 'int' (64 bits, of course). Making STL size() methods return size_t is now recognized as a big headache, and only exists as a historical artifact of the brief period where processor word size was 32 bits and you might want to be able to represent every memory address up to the full 2^32 bytes it could address (not just the 2^31 max that an int can hold).

PR #592, opened 2015-12-02
Travis-CI setup for OSL

lgritz:
Modeled after what I did recently for OIIO. Henceforth, all pull requests and pushes to the main OSL repo will cause builds to be kicked off for OSX and Linux, with a few different build options each. When somebody makes a PR, even before merging we can be reasonably certain it doesn't break the build, yay.

It's not 100% finished, but it's working and useful. The biggest unresolved issue is that I have to manually exclude 4 tests (out of almost 300) that don't match their output images exactly (and I'm sure are fine), and I haven't yet figured out how to recover the files it generates so I can commit it as an alternate reference image. But I'll sweep that under the rug for now and come back to it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,105 @@
+# .travis.yml for OSL
+
+language: cpp
+sudo: false
+compiler:
+    - clang
+    - gcc
+os:
+    - linux
+    - osx
+#osx_image: xcode7
+dist: trusty
+
+# Do a default build (C++03, sse2), an sse4.2 build, and a C++11 build.
+# Also try with both gcc 4.8 and 4.9 (on Linux; on OSX these will be
+# excluded).
+env:
+    matrix:
+      - WHICHGCC=4.8 OIIOBRANCH=master BUILD_FLAGS=""
+      - WHICHGCC=4.8 OIIOBRANCH=master BUILD_FLAGS="USE_CPP=1"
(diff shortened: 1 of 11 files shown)


fpsunflower:
It bugs me a bit that the broken tests are testrender related. I would love to find out where the errors are if you can retrieve the images.

LGTM otherwise.

MrMormon:
Great. Could this lead to a build script in the future for Unix-like users?

lgritz:
@fpsunflower : I'm trying to figure out how to retrieve output of failed tests so we can figure out what's wrong.

@MrMormon : What do you mean by "build script"? How is that different than the make/cmake we already have?

MrMormon:
Cmake still requires manual setup of LLVM and environment variables, for example.

MrMormon:
Thankfully I do have access to a pre-compiled Linux binary, but it's only thanks to Blender which is still stuck with OSL 1.5.12-9 for its Cycles renderer.

lgritz:
@MrMormon Let's redirect discussion off this closed tickets and to #593.

lgritz:
For anybody following along who has Windows experience, I'd very much like to set up CI for Windows via Appveyor (both for OSL and OIIO). It would be great for somebody with more Windows experience than me, and especially with Appveyor knowledge, to give it a stab.

PR #596, opened 2015-12-08
Fix potential SIGFPE from divide by zero in shaders.

lgritz:
The implementation in LLVM_Util we were relying on for safe div and mod    turned out to have the property that they generated safe results (e.g.,    x/0 returns 0, not inf), but still was capable of generating div-by-zero    exception if the app in which OSL is embedded was trapping SIGFPE.

Instead, switch to a truly safe implementation that only does the div    or mod if the denominator is nonzero. Also enhance it by doing a simple    unchecked div or mod any time the denominator is known to be a nonzero    constant (no need to have the conditional in that case).


src/liboslexec/llvm_gen.cpp
```diff
@@ -610,19 +610,28 @@ LLVMGEN (llvm_gen_div)
...
 
         if (deriv) {
             // Division of duals: (a/b, 1/b*(ax-a/b*bx), 1/b*(ay-a/b*by))
             ASSERT (is_float);
-            llvm::Value *binv = rop.ll.op_make_safe_div (type, rop.ll.constant(1.0f), b);
+            llvm::Value *binv;
+            if (B.is_constant() && ! rop.is_zero(B))
+                binv = rop.ll.op_div (rop.ll.constant(1.0f), b);
+            else
+                binv = rop.ll.call_function (safe_div, rop.ll.constant(1.0f), b);
```
fpsunflower: Its probably that I'm not familiar enough with the code, but can you explain in what cases it will generate a function call vs. generate LLVM IR directly? (ie: the `make_safe_div` call below).

lgritz: If the denominator is a nonzero constant, it will just do a plain div. Otherwise, it will call the safe_div function that tests for zero, etc. That function itself will generally be inlined by the time LLVM optimization is done. It's just easier to call that inlineable function rather than fiddle with making conditionals and blocks here.

src/liboslexec/llvm_util.cpp
```diff
@@ -1428,37 +1452,41 @@ LLVM_Util::op_bool_to_int (llvm::Value* a)
...
 llvm::Value *
-LLVM_Util::op_make_safe_mod (TypeDesc type, llvm::Value *a, llvm::Value *b)
+LLVM_Util::op_make_safe_mod (llvm::Value *a, llvm::Value *b)
 {
-    if (type.basetype == TypeDesc::FLOAT) {
+    if (a->getType() == type_float() && b->getType() == type_float()) {
         llvm::Value *mod = builder().CreateFRem (a, b);
         llvm::Value *zero = constant (0.0f);
         llvm::Value *iszero = builder().CreateFCmpOEQ (b, zero);
         return builder().CreateSelect (iszero, zero, mod);
```
fpsunflower: I guess I was asking about this function -- in what circumstances is this one called? It looks like it generates the "safe div" manually.

lgritz: Right now, this function is not called by OSL, and it is "safe result" but not "SIGFPE-safe". Maybe I should remove it entirely.

lgritz:
So, you ok with this?

lgritz:
removed and merged

PR #602, opened 2016-01-07
Use GCC attribute for the methods which are considered deprecated

Nazg-Gul:
This helps applications to update their OSL API use earlier without need to
read full release or commit logs. Could also be handy to find use of methods
which are gonna to be deprecated in OSL itself.

Currently only GCC-like compilers are supported, but it's easy to support
more compilers when needed.

Applications which doesn't want this attribute to be used can define empty
OSL_DEPRECATED macros in the build system.

This commit also updates some internal use of API.

### DIFF
--- a/src/include/OSL/oslconfig.h
+++ b/src/include/OSL/oslconfig.h
@@ -130,5 +130,18 @@ using OIIO::string_view;
   using boost::weak_ptr;
 #endif
 
+#ifndef __has_attribute
+#  define __has_attribute(x) 0
+#endif
+
+#if __cplusplus >= 201402L || (__cplusplus >= 201103L && __has_attribute(deprecated))
+#  define OSL_DEPRECATED(msg) [[deprecated(msg)]]
+#elif defined(__GNUC__)
+#  define OSL_DEPRECATED(msg) __attribute__((deprecated(msg)))
+#elif defined(_MSC_VER)
+#  define OSL_DEPRECATED(msg) __declspec(deprecated(msg))
+#else
+#  define OSL_DEPRECATED(msg)
+#endif
 
 OSL_NAMESPACE_EXIT
--- a/src/include/OSL/oslcomp.h
+++ b/src/include/OSL/oslcomp.h
@@ -41,7 +41,7 @@ namespace pvt {
 
 class OSLCOMPPUBLIC OSLCompiler {
 public:
-    /// DEPRECATED -- it's ok to directly construct an OSLCompiler now.
+    OSL_DEPRECATED("It's ok to directly construct an OSLCompiler now")
     static OSLCompiler *create ();
 
     OSLCompiler (ErrorHandler *errhandler=NULL);
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -42,8 +42,10 @@ OSL_NAMESPACE_ENTER
 class RendererServices;
 class ShaderGroup;
 typedef shared_ptr<ShaderGroup> ShaderGroupRef;
-typedef ShaderGroup ShadingAttribState;       // DEPRECATED name
-typedef ShaderGroupRef ShadingAttribStateRef; // DEPRECATED name
+OSL_DEPRECATED("Use ShaderGroupRef instead")
+typedef ShaderGroup ShadingAttribState;
+OSL_DEPRECATED("Use ShaderGroupRef instead")
+typedef ShaderGroupRef ShadingAttribStateRef;
 struct ClosureParam;
 struct PerThreadInfo;
 class ShadingContext;
@@ -75,11 +77,12 @@ public:
                    ErrorHandler *err=NULL);
     ~ShadingSystem ();
 
-    /// DEPRECATED -- it's ok now to just construct a ShadingSystem.
+    OSL_DEPRECATED("It's ok now to just construct a ShadingSystem")
     static ShadingSystem *create (RendererServices *renderer=NULL,
                                   TextureSystem *texturesystem=NULL,
                                   ErrorHandler *err=NULL);
-    /// DEPRECATED -- it's ok now to just destroy a ShadingSystem.
+
+    OSL_DEPRECATED("It's ok now to just destroy a ShadingSystem")
     static void destroy (ShadingSystem *x);
 
     /// Set an attribute controlling the shading system.  Return true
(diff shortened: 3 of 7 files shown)


Nazg-Gul:
This is more an experiment and request for comments, but found such changes to be handy in Blender. Hopefully we can generalize it further and make handy for everyone.

lgritz:
Yes, I really like this!

I don't think you needed the `#ifndef OSL_DEPRECATED` -- if oslconfig.h is the only place we define it, and it contains a `#pragma once`, then it can't already be defined.

I would offer the following minor suggestion, however, to help move us toward future standards and catch all of the compilers:

```
#ifndef __has_attribute
#  define __has_attribute(x) 0
#endif

#if __cplusplus >= 201402L || __has_attribute(deprecated)
#  define OSL_DEPRECATED [[deprecated]]
#  define OSL_DEPRECATED_MSG [[deprecated(msg)]]
#elif defined(__GNUC__)
#  define OSL_DEPRECATED __attribute__ ((deprecated))
#  define OSL_DEPRECATED_MSG OSL_DEPRECATED
#elif defined(_MSC_VER)
#  define OSL_DEPRECATED __declspec(deprecated)
#  define OSL_DEPRECATED_MSG OSL_DEPRECATED
#else
#  define OSL_DEPRECATED
#  define OSL_DEPRECATED_MSG
#endif
```

lgritz:
For my suggestion, I think for all platforms to work, you want the OSL_DEPRECATED at the front of a declaration:

```
OSL_DEPRECATED void foo();
```

And the version I suggested with a message can sometimes be helpful (at least for C++14):

```
OSL_DEPRECATED_MSG("foo() calls should be replaced by bar()")
void foo();
```

Nazg-Gul:
The idea behind `#ifndef OSL_DEPRECATED` was to allow some application to force disable this deprecation warnings, so nobody gets annoyed by this change. But i'm all for always have those deprecation attributes enabled actually.

Updated the patch with suggestions from you, also fixed definition of `OSL_DEPRECATED_MSG` for non-C++14. Only thing i'm not sure currently is where the proper place for `OSL_DEPRECATED` attribute in the typedefs. Seems both `OSL_DEPRECATED typedef Foo Bar;` and `typedef Foo Bar OSL_DEPRECATED` works with GCC. Any strong opinion which notation to prefer?

lgritz:
Oh, ok, that's fine with the `#ifndef OSL_DEPRECATED`. I just misunderstood why you had that.

lgritz:
My only preference for whether OSL_DEPRECATED goes in front or behind is that we do the one that also works for MSVC.

Nazg-Gul:
Think now with additional `OSL_DEPRECATED_MSG` it should be done differently a bit. Something like, having `#ifndef OSL_NO_DEPRECATION_WARNING blabblah`. But that's becoming nasty a bit. If you're fine with always having those warnings, let's just move forward without extra complexity here. We can always add such fine-tuning defines later when someone _really_ needs them.

src/include/OSL/oslconfig.h
```diff
@@ -130,5 +130,22 @@ using OIIO::string_view;
   using boost::weak_ptr;
 #endif
 
+#ifndef __has_attribute
+#  define __has_attribute(x) 0
+#endif
+
+#if __cplusplus >= 201402L || __has_attribute(deprecated)
+#  define OSL_DEPRECATED [[deprecated]]
+#  define OSL_DEPRECATED_MSG [[deprecated(msg)]]
```
lgritz: I think you are missing the (msg) part here!

Nazg-Gul:
That's a good catch, didn't have C++14 handy to test and overviewed missing bit!

Tested deprecation warning and typedef in MSVC -- it's required to have it in front of typedef. It should be now reflected in the commit.

lgritz:
It failed Travis tests and so I tried your patch on my machine and it needs a minor adjustment.

Turns out that some Clang versions, when compiled in old C++03 mode, think __has_attribute(deprecated) is there, but because `[[ ]]` is C++11 syntax it still gives an error. Ugh. So the following fixes that:

```
#if __cplusplus >= 201402L || (__cplusplus >= 201103L && __has_attribute(deprecated))
```

What that means is, if using C++14 (in which ``deprecated`` is standard), or if it's at least C++11 and supports the attribute, then use the standard way. Otherwise fall back to the GCC-specific or MS-specific methods.

Also, I noticed here http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3760.html that both gcc and MSVS support messaged versions. So I think that the full setup we want is:

```
#if __cplusplus >= 201402L || (__cplusplus >= 201103L && __has_attribute(deprecated))
#  define OSL_DEPRECATED [[deprecated]]
#  define OSL_DEPRECATED_MSG(msg) [[deprecated(msg)]]
#elif defined(__GNUC__)
#  define OSL_DEPRECATED __attribute__((deprecated))
#  define OSL_DEPRECATED_MSG(msg) __attribute__((deprecated(msg)))
#elif defined(_MSC_VER)
#  define OSL_DEPRECATED __declspec(deprecated)
#  define OSL_DEPRECATED_MSG(msg) __declspec(deprecated(msg))
#else
#  define OSL_DEPRECATED
#  define OSL_DEPRECATED_MSG(msg)
#endif
```

I really like this patch, by the way, and after we merge it here I will put the equivalent into OIIO as well.

lgritz:
By the way, another alternative -- I'm not advocating, just giving you the chance to think about it before we commit -- is to make OSL_DEPRECATED(msg) be the one and only version, always use the message (which for some deprecated things we'll just use "" if there's nothing helpful to say) and do away with the message-less versions.

Any opinions on that?

fpsunflower:
I'd be in favor of always requiring a message.

lgritz:
Yeah, the more I think about it, the more I like the idea of using the message version only. Not only does that allow us to put pressure on ourselves to provide an explanation, but that explanation should also say when it was deprecated.

    OSL_DEPRECATED("Use foo() instead. This was deprecated in OSL 1.6.")

Nazg-Gul:
Think it's indeed good idea to always have a message. Updated the patch which includes changes from Larry, also made `OSL_DEPRECATED()` to always require message. Some messages could have better wording tho.

lgritz:
LGTM. I will merge, perhaps with some minor edits to the messages. Thanks for the contribution!

lgritz:
Merged.

PR #603, opened 2016-01-07
Remove has_userdata method from RendererServices

fpsunflower:
None of the internals of OSL ever call this method


lgritz:
Travis says it fails to build!

Did you forget to check in a file?

fpsunflower:
Indeed! Somehow I lost one of the removals when moving the patch around the repos on my machine.

Yay for automated testing!

fpsunflower:
Updated the patch.

lgritz:
I swear, for any future project, setting up a .travis.yml file is the very first thing I'll do after 'git init'. In a very short time it has gone from "why would I need that" to "can't imagine doing anything without it."

lgritz:
LGTM

PR #605, opened 2016-01-08
Add string indexing and string hashing functions to stdosl.h

fpsunflower:
These two small functions will be useful from an upcoming contribution to `oslutil.h`.

### DIFF
--- a/src/liboslexec/constfold.cpp
+++ b/src/liboslexec/constfold.cpp
@@ -1053,8 +1053,46 @@ DECLFOLDER(constfold_strlen)
     if (S.is_constant()) {
         ASSERT (S.typespec().is_string());
         int result = (int) (*(ustring *)S.data()).length();
-        int cind = rop.add_constant (TypeDesc::TypeInt, &result);
+        int cind = rop.add_constant (result);
         rop.turn_into_assign (op, cind, "const fold strlen");
+        return 1;
+    }
+    return 0;
+}
+
+
+
+DECLFOLDER(constfold_hash)
+{
+    // Try to turn R=hash(s) into R=C
+    Opcode &op (rop.inst()->ops()[opnum]);
+    Symbol &S (*rop.inst()->argsymbol(op.firstarg()+1));
+    if (S.is_constant()) {
+        ASSERT (S.typespec().is_string());
+        int result = (int) (*(ustring *)S.data()).hash();
+        int cind = rop.add_constant (result);
+        rop.turn_into_assign (op, cind, "const fold hash");
+        return 1;
+    }
+    return 0;
+}
+
+
+
+DECLFOLDER(constfold_getchar)
+{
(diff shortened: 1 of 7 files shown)


lgritz:
LGTM!

I'll add these to the spec after it's merged.

Do you think we should add constant-folded versions of these? Or too much of a rare case to worry about?

fpsunflower:
I think we should add the folders for completeness. Will keep that on my list.

I'm going to at least write a few unit tests.

lgritz:
The low-level string funcs should be tested in testsuite/string/test.osl

Once you've done the magical thing that this is intended for, I would also encourage you to put a new testsuite entry that demonstrates it. (like the one you sent this morning, though low-res is fine)

fpsunflower:
Updated with constant folding and unit tests.

src/liboslexec/constfold.cpp
```diff
@@ -1062,6 +1062,44 @@ DECLFOLDER(constfold_strlen)
 
 
 
+DECLFOLDER(constfold_hash)
+{
+    // Try to turn R=strlen(s) into R=C
```
lgritz: Comment is wrong

src/liboslexec/constfold.cpp
```diff
@@ -1062,6 +1062,44 @@ DECLFOLDER(constfold_strlen)
...
+        return 1;
+    }
+    return 0;
+}
+
+
+
+DECLFOLDER(constfold_getchar)
+{
+    // Try to turn R=strlen(s) into R=C
```
lgritz: Comment is wrong

src/liboslexec/constfold.cpp
```diff
@@ -1062,6 +1062,44 @@ DECLFOLDER(constfold_strlen)
...
 
+DECLFOLDER(constfold_hash)
+{
+    // Try to turn R=strlen(s) into R=C
+    Opcode &op (rop.inst()->ops()[opnum]);
+    Symbol &S (*rop.inst()->argsymbol(op.firstarg()+1));
+    if (S.is_constant()) {
+        ASSERT (S.typespec().is_string());
+        int result = (int) (*(ustring *)S.data()).hash();
+        int cind = rop.add_constant (TypeDesc::TypeInt, &result);
```
lgritz: It's easier to

    int cind = rop.add_constant (result);

It knows result is an int. You only need the (TypeDesc,void*) version when it's not a simple type.

lgritz:
I think you'll also want to update the out.txt output for that test.

lgritz:
Other than my minor comments and the missing reference output, LGTM.

fpsunflower:
Fixed, updated and squashed commits

fpsunflower:
Just because I want to practice my git skills this morning -- update the unit test to properly grab the last character of the string

lgritz:
LGTM

PR #606, opened 2016-01-08
Add utility function that displays strings

fpsunflower:
Based on the previous review that added `getchar(string, index)` and `hash(string)`, add a new utility function in `oslutil.h` that lets a shader draw a string.

This is convenient for displaying text on objects in raster space for example.


lgritz:
This is insane and I love it.
LGTM

testsuite/draw_string/run.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
+
+command += testshade ("-g 512 512 -od uint8 -o Cout out.tif test")
+outputs = [ "out.txt", "out.tif" ]
+# expect a few LSB failures
+failthresh = 0.004
+failpercent = 0.05
```
lgritz: A few tests need slightly higher fail thresholds, but if this passes without these two lines, please take them out.

fpsunflower:
Yup it seems to pass fine without the thresholds. Let's see how it does on the remote hosts ...

fpsunflower:
Passed the CI tests as well. Merged.

PR #612, opened 2016-01-19
Testsuite organization overhaul -- no more src or ref in build

lgritz:
The old way was that when CMake runs for the first time (when it builds your make or ninja files), it copies everything from testsuite/ into build/PLATFORM/testsuite/, and then when you run tests, it uses all the info (source data, scripts, and reference output) in the build area.

That was fine usually, but if you were developing or modifying a test itself, it was a big PITA, because changing something in testsuite/mytest would not be reflected in the build/PLATFORM/testsuite/mytest that had already been copied. It was the source of many frustrations and errors. And beware modifying something in the build area and then blowing it away, not remembering to copy it to the source testsuite/ area.

This overhaul NO LONGER COPIES source data or reference output en masse to build at all. Instead, a combination of selective copies (usually just .osl files), links and references to the canonical files are used to make the build area only the temporary location of test output itself. This sure does make everything easier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -445,11 +445,15 @@ add_subdirectory (src/doc)
 #########################################################################
 # Testing
 
-# Make a copy of the testsuite into the build area
-if (DEFINED CMAKE_VERSION AND NOT CMAKE_VERSION VERSION_LESS 2.8)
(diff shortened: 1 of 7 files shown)


fpsunflower:
LGTM

PR #613, opened 2016-01-20
Disallow quoted strings from spanning lines (just like you can't in C).

lgritz:
This was never intentionally allowed; it was just a bug in the lexer's
regex describing a quoted string. The compiler would accept it silently,
but in the process forget the line count and botch subsequent error
messages.

### DIFF
--- a/src/liboslcomp/osllex.l
+++ b/src/liboslcomp/osllex.l
@@ -70,11 +70,8 @@ FLT2            {DIGIT}*\.{DIGIT}+{E}?
 FLT3            {DIGIT}+{E}
 FLT             {FLT1}|{FLT2}|{FLT3}
  /* string literal */
-STR     \"(\\.|[^\\"])*\"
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM

fpsunflower:
Would it be hard to support something like C++11's raw strings?

lgritz:
I don't think it would be hard. Would it be better to allow line-spanning strings, but just be sure to get the line counting correct?

Would you want the R"" syntax of C++, or should we just allow all strings to span lines if people really want? Are there bugs that this would mask?

fpsunflower:
I think making it explicit with the r"" syntax would be easiest and less likely to be invoked by accident. But then again multi-line hardcoded strings in shader code are so unlikely that its not clear how much benefit there is.

lgritz:
Let's put the raw string feature off until somebody requests it and has a compelling use case.

PR #618, opened 2016-02-06
oslc bug -- getmatrix failed to note that the matrix argument is written

lgritz:
This could lead to a cascade of optimization errors at runtime.
I have no idea why this never showed up as a problem before today.

In addition to fixing it on the oslc side, add a hacky fix to make sure    that old compiled shaders are retroactively fixed as they are loaded at    runtime.

### DIFF
--- a/src/shaders/stdosl.h
+++ b/src/shaders/stdosl.h
@@ -52,6 +52,7 @@
 // Declaration of built-in functions and closures
 #define BUILTIN [[ int builtin = 1 ]]
 #define BUILTIN_DERIV [[ int builtin = 1, int deriv = 1 ]]
+#define BUILTIN_NONSTANDARD_RW [[ int builtin = 1, int rw = 1 ]]
 
 #define PERCOMP1(name)                          \
     normal name (normal x) BUILTIN;             \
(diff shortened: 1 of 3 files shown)


aconty:
LGTM

src/shaders/stdosl.h
```diff
@@ -530,7 +531,7 @@ int iscameraray () { return raytype("camera"); }
 int isdiffuseray () { return raytype("diffuse"); }
 int isglossyray () { return raytype("glossy"); }
 int isshadowray () { return raytype("shadow"); }
-int getmatrix (string fromspace, string tospace, output matrix M) BUILTIN;
+int getmatrix (string fromspace, string tospace, output matrix M) BUILTIN_NONSTANDARD_RW;
```
fpsunflower: The definition of `getmatrix` already contains a marked `output` parameter. Maybe not for this review, but would it be possible to directly use this information instead of having to manually flag the argument in the compiler?

lgritz: That's a good point, it should be able to figure it out on its own. I'll look into that.

PR #620, opened 2016-02-11
For stdosl.h funcs marked builtin, auto-detect funny read-write cases.

lgritz:
As Chris points out in #618, the fact that params are marked as output
is all we need to know that function's parameters are not all read-only
(duh). That seems a little more foolproof than requiring a separate notation
in stdosl.h.


fpsunflower:
LGTM!

Does this mean a few more builtins could be moved into `stdols.h` ?

lgritz:
It does! But I wanted to get on with other things, there didn't seem to be any particular payoff from doing that now.

PR #626, opened 2016-03-09
Don't test or set "layer ran bits" on last layer.

lgritz:
This shouldn't change any behavior. It's just removing some pointless code that checks if a layer has been run before, but there's no need to do that on the root layer which by definition is run exactly once.

Also be sure that the entry point doesn't use the "fast" calling convention.

This is just some stuff I noticed in the process of scrutinizing the code we were generating.


fpsunflower:
LGTM

PR #627, opened 2016-03-10
LLVM-related refactoring

lgritz:
I'm in the middle of a concerted effort to bring us into a more modern era of LLVM versions. This set of patches is not a complete solution, but contains several intermediate steps in that direction.

* Remove all support for LLVM 3.3 and earlier. This gets rid of lots of confusing `#if` clauses and simplifies the structure of the code. Henceforth, LLVM 3.4 will be the minimal version we support for OSL 1.8.

* A number of changes that allow us to build cleanly against LLVM 3.6. Although it correctly runs the llvmutil_test unit test, it fails all the testsuite tests involving OSL itself (against LLVM 3.6). But at least it compiles.

* Remove runtime selection of MCJIT for LLVM versions that support both MCJIT and old JIT. This simplifies the logic in some places, by cleanly dividing LLVM 3.4/3.5 + old JIT, versus LLVM 3.6+ with MCJIT and no old JIT.

What this nets us is a simplification of code that will make the next steps easier. As of now (in master), we fully support and work properly with LLVM 3.4 and 3.5, and the next task will be to get some later version (maybe 3.6, or we may skip all the way to 3.7+ORC JIT, we'll see) fully working.

This whole discussion is about OSL master (what will some day be 1.8). The production release (OSL 1.7) will continue to work with LLVM 3.2 and 3.3 (though please note that I have not personally tested with either for quite some time, though I also have not purposely changed anything that would cause them to break).

### DIFF
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -86,10 +86,6 @@ public:
     void debug (int d) { m_debug = d; }
     int debug () const { return m_debug; }
 
-    /// Use MCJIT?
-    void mcjit (int on) { m_mcjit = on; }
-    int mcjit () const { return m_mcjit; }
-
     /// Return a reference to the current context.
     llvm::LLVMContext &context () const { return *m_llvm_context; }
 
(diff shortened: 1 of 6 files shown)


fpsunflower:
LGTM

PR #628, opened 2016-03-10
llvmutil_test enhancement: more comprehensive tests and benchmarks

lgritz:
* JIT two different functions, one of which uses vec3. One reason this
  is nice is to expose any bugs that only crop up with more than one thing
  is JITed, and the existance of multiple LLVM_Util's.

* Add a test that makes a function with a HUGE amount of unnecessary IR,
  and JITs it repeatedly, so that we can verify memory usage and have a
  simple time benchmark.


fpsunflower:
LGTM

PR #629, opened 2016-03-10
Make an option to have OSL isconnected() return 1 for userdata parameters

lgritz:
This adds a new ShadingSystem attribute, "userdata_isconnected". The
default of 0 is the old/usual behavior that most applications want.
However, setting this attribute to 1 will cause the OSL built-in
function isconnected(paramname) to return 1 not only for connected
parameters, but also for parameters that may be bound to userdata
(i.e., those that have lockgeom=0).

The use case for this involves a renderer implementation that can mix
C++ and OSL nodes within a shading group. The outputs exported by the
C++ nodes propagate their values to the downstream OSL node via
get_userdata(), hijacking the mechanism that is generally used geometric
primitive variable interpolation. But from the point of view of the
downstream node, they should semantically appear to be a connection.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -113,6 +113,9 @@ public:
     ///    int lazyunconnected    Run layers lazily even if they have no
     ///                              output connections (1). For debugging.
     ///    int lazy_userdata      Retrieve userdata lazily (0).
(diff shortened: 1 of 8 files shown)


aconty:
LGTM

PR #632, opened 2016-03-26
Restrict find_program LLVM to LLVM_DIRECTORY

stefanbuettner:
Hey there,

when building Blender on Debian testing, I stumbled over this.

LLVM can be installed on a system without clang.
So, if LLVM_DIRECTORY pointed to a locally installed llvm/clang version but the system had LLVM installed without Clang, until now the script would find the system LLVM and no Clang and the build would fail.
However, since LLVM is globally searched for if the hinted attempts fail, it should not hurt to only search in the hinted paths, would it?

And a question: Why is *llvm-config* (lines 194-207) only called if LLVM_DIRECTORY was not set?

Cheers,
Stefan

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -180,9 +180,9 @@ endif()
 
 # try to find llvm-config, with a specific version if specified
 if(LLVM_DIRECTORY)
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM, I will merge this fix.

As for your question about LLVM_DIRECTORY... I don't know. Maybe that's a mistake? Should I change

    if(NOT LLVM_DIRECTORY OR EXISTS ${LLVM_CONFIG})

to

    if(EXISTS ${LLVM_CONFIG})

???

lgritz:
merged.

PR #635, opened 2016-03-28
Warn for common misstep of forgetting constructor name.

lgritz:
shader test ( output color Cout = 0 )
    {
        vector x = (1, 2, 3);
    }

See where this goes wrong? This code will make x be (3,3,3).  If you
don't know why, think about how the comma operator works in C, C++, and
related languages.

The user almost certainly meant

    vector x = vector (1, 2, 3);

With this change, the original (buggy) code will get the following
warning output:

    test.osl:3: warning: Comma operator is very confusing here. Did you mean to use a constructor: x = vector(...)?

### DIFF
--- a/testsuite/oslc-warn-commainit/test.osl
+++ b/testsuite/oslc-warn-commainit/test.osl
@@ -1,0 +1,5 @@
+shader test ( output color Cout = 0 )
+{
+    vector x = (1, 2, 3);
+    Cout = (u, v, 0);
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -547,6 +547,7 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             oslc-err-format oslc-err-intoverflow
             oslc-err-noreturn oslc-err-outputparamvararray oslc-err-paramdefault
             oslc-err-struct-array-init oslc-err-struct-dup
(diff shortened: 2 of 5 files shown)


lgritz:
By the way, I am open to suggestions for better wording of the warning that would make it more clear what's going wrong.

testsuite/oslc-warn-commainit/test.osl
```diff
@@ -0,0 +1,5 @@
+shader test ( output color Cout = 0 )
+{
+    vector x = (1, 2, 3);
```
fpsunflower: Will the warning also work with `{` `}` or is that case already an error?

fpsunflower:
The wording seems fine to me.

testsuite/oslc-warn-commainit/test.osl
lgritz: That's already a syntax error.

I'm thinking about allowing aggregate assignment in that way, but it's a much bigger grammatical and implementation change than simply warning about the one thing people seem to commonly mistake.

fpsunflower: Makes sense.

LGTM!

lgritz: Specifically, what it currently says is:

    test.osl:4: error: Can't use '= {...}' initializer except with arrays or struct (x)

Issue #233, opened 2013-02-12
Returning struct from nested function assertion


brechtvl:
Compiling this
```
struct Test { float f; };

shader test()
{
    Test some_function()
    {
        Test a;
        return a;
    }

    Test b = some_function();
}
```

results in src/liboslcomp/typecheck.cpp:1241: Failed assertion '0': Don't know how to decode type code '49'

If some_function is not nested it still gives a compile error but no assertion is hit. I can try to fix this but I'm wondering if I should try to make it throw an error for this case or if returning structs should actually work?

lgritz:
This has been fixed.

PR #636, opened 2016-04-07
Allow functions to return a struct.

lgritz:


lgritz:
Fixes #233

fpsunflower:
LGTM

Are there any subtleties around nested structs or arrays that would warrant a few more tests?

lgritz:
Probably. I'll set up a test for that.

lgritz:
Added a complicated test with structs of structs, arrays in structs, arrays of structs in structs.

Believe it or not, it ran correctly the first time!

fpsunflower:
Nice!

PR #637, opened 2016-04-14
Fix oslc incorrectly emitting init ops for constructor color(float).

lgritz:
We're talking about a construct like:

    shader foo (color c = color(1), ...)

It got it right if it was

    color c = color(1,1,1)

In both cases, it properly reduced the default value to a static constant
value, but the bug was that it ALSO (unnecessarily) had extra init ops
for that parameter. That would be eliminated in runtime optimization, the
shader certainly ran correctly, but it threw off oslinfo, which figures
that if init ops are present, it can't figure out what the default values
were.


aconty:
LGTM

PR #640, opened 2016-04-26
Overhaul of nested struct handling for parameters

lgritz:
This enables nested structs (structs containing other structs) to be useful as shader parameters, with initialization working as expected with nested `{ {...}, {...} }` notation. Tricky!

I don't honestly expect anyone to understand this patch. It took me days to fully understand my own code as it existed before, the part that I replaced with this overhaul. Suffice it to say that the new code passes the full testsuite, including the new test that exercises deeply nested structures with parameter initializers. Previously, deeply nested structures had no viable way of being used as parameters because it couldn't figure out the notation to fully initialize their defaults (shader parameters, unlike locals, MUST be given defaults).


fpsunflower:
LGTM

Is it possible to nest an array of structs? Or is that a case that isn't fully working yet?

lgritz:
I'm not expecting testing of arrays of structs to work currently.

PR #642, opened 2016-05-28
Helper modes for trying to profile noise and pointclouds

lgritz:
Is your frame expensive, and you have a hunch it might be because you are using so much noise()? How can you figure that out?

This patch provides some new helper modes for ShadingSystem:

1. When the "profile" attribute is set to nonzero, it counts the number of noise calls and reports the total in the stats.

2. When the "no_noise" attribute is set to nonzero, all noise calls will be replaced with an inexpensive function that just returns 0 (for signed noise varieties) or 0.5 (for unsigned noises). You can then time your render in this mode versus with the regular noise calls, and the different should give you an estimate of how much time was spent doing noise math.

3. When the "no_pointcloud" attribute is set to nonzero, all pointcloud_get and pointcloud_search calls will take an early out and return that no points were found. This can also be hepful for trying to estimate how much time was being spent related to pointcloud lookups.


lgritz:
Updated, I decided I liked the attribute name "no_noise" better than my original choice of "nullnoise".

lgritz:
Updated again, added "no_pointcloud" attribute that shortcuts the pointcloud search functions, for various debug and profile purposes.

fpsunflower:
LGTM

PR #649, opened 2016-06-24
[bug fix] Allow compiling against Boost v1.61 and above

dracwyrm:
This patch can be applied to RB-1.7 branch directly with no modifications.

New versions of Boost use std::mutex when using a cpp11 compatible
compiler.

FILE: boost_1_60_0/boost/pool/detail/mutex.hpp
#if !defined(BOOST_HAS_THREADS) || defined(BOOST_NO_MT) || defined(BOOST_POOL_NO_MT)
  typedef null_mutex default_mutex;
#else
  typedef boost::mutex default_mutex;
#endif

Was changed to:

FILE: boost_1_61_0/boost/pool/detail/mutex.hpp
#if !defined(BOOST_HAS_THREADS) || defined(BOOST_NO_MT) || defined(BOOST_POOL_NO_MT)
  typedef null_mutex default_mutex;
#else
#if defined (BOOST_NO_CXX11_HDR_MUTEX)
  typedef boost::mutex default_mutex;
#else
  typedef std::mutex default_mutex;
#endif
#endif

The Boost Wave functions use default_mutex, so defining
BOOST_NO_CXX11_HDR_MUTEX allos the old method of boost::mutex
to be used, which is what OSL was originally written to use.

TODO: Allow files to be linked against std::mutex if using
a cpp11 compatible compiler.

Closes bug 648

### DIFF
--- a/src/include/OSL/oslconfig.h
+++ b/src/include/OSL/oslconfig.h
@@ -52,6 +52,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #  define OSL_CPLUSPLUS_VERSION  3 /* presume C++03 */
 #endif
 
+// Boost v1.61 uses std::mutex instead of boost::mutex
+// This forces the use of boost::mutex which compiles
+#define BOOST_NO_CXX11_HDR_MUTEX 1
+
 // Symbol export defines
 #include "export.h"
 


lgritz:
Unfortunately, I believe this patch is not correct.

I think the problem is that you are compiling OSL with C++11, but your installed Boost was built using C++03 (and thus used templates based on boost::mutex because its C++ was too old to have std::mutex available).

This patch is essentially saying "when compiling OSL, even though I may be using C++11 at the moment, use the old Boost mutex instead of std::mutex."

But this approach is not a general solution, and I believe it will fail for people who have a Boost that was built with C++11, because then you will be asking for boost::mutex but linking against a boost that was compiled to expect std::mutex.

The correct solution is to use a Boost and OSL that match in which language standard they are using.

lgritz:
Any response to my comments about this?
As I said, I believe that it's simply a mismatch between standard level when compiling boost versus OSL -- either both should be C++11, or neither should be.

dracwyrm:
Hi, sorry. Yeah. It turns out that Gentoo compiled Boost with C++98 and OSL was trying to compile as C++11, so it didn't work right. After some hacking, I was able to compile Boost with C++11 and OSL worked fine.

Sorry for the confusion. I shall close this.

PR #651, opened 2016-06-30
Fix subtle bug in optimize_ops that could seg fault.

lgritz:
The optimize_ops method had a convenience reference to the op being
examined in the loop.  We carefully do a reserve() first to make sure
that there was enough room for one insertion so that the reference would
still be valid if the constant-folding function has to add an op.  But
this relied on the assumption that no folder would insert more than one
op as it did its work, which was always true (once).

Oops, much later, a new way to speed up "mix" ops had a particular corner
case in which up to 7 new ops might be inserted, and I never went back
and fixed the now-broken assumption here.

The fix here is to change the reference to the op (which cannot be
changed dynamically) into a pointer to the op, which can be re-pointed
after calling the folding function to make sure that it's valid and
pointing to the (potentially new) correct position as it continues on in
the loop. So it is no longer sensitive to changes to the vector
allocation that might happen inside the folder.

I haven't changed this code for a long time. It's shocking that no
problems cropped up from this bug until this week! I guess it's not a
very common "corner case".

### DIFF
--- a/src/liboslexec/runtimeoptimize.cpp
+++ b/src/liboslexec/runtimeoptimize.cpp
@@ -2044,20 +2044,17 @@ RuntimeOptimizer::optimize_ops (int beginop, int endop,
     for (int opnum = beginop;  opnum < endop;  opnum += 1) {
         ASSERT (old_num_ops == num_ops); // better not happen unknowingly
         DASSERT (num_ops == inst()->ops().size());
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM

Good catch! Also surprised that we didn't run into this earlier ...

PR #656, opened 2016-07-06
Bug fix: incorrect instance merging when unspecified-length arrays differed

### LINKED ISSUE #655: optimization breaks variable length array parameters
If several layers in a network are instances of the same shader and that shader has a variable length array parameter, trying to set a different value for the two instances will fail. The same value will be used by both. This does not happen at O0 osl optimization level. This archive [simple_surface.tar.gz](https://github.com/imageworks/OpenShadingLanguage/files/350164/simple_surface.tar.gz) is a fairly minimal reproduction case:
```
> testshade --group shadergroup
0.000000 0.000000 2.000000

> testshade -O0 --group shadergroup
1.000000 0.000000 1.000000
```

lgritz:
Because the altered data offset wasn't transferred from the instance
overrides to the instance's full symbol table, it was comparing the
default values rather than the actual values, so two instances that
should not have been merged, were merged.

Also fixed a (harmless) bug in ShadingSystemImpl::ShaderGroupBegin,
where the serialized group is parsed, where it was double-calling
Parameter() to add the parameters.

Fixes #655

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -543,7 +543,8 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             layers layers-Ciassign layers-entry layers-lazy
             layers-nonlazycopy layers-repeatedoutputs
             linearstep
-            logic loop matrix message mergeinstances-nouserdata
+            logic loop matrix message
+            mergeinstances-nouserdata mergeinstances-vararray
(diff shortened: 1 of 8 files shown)


fpsunflower:
LGTM

lgritz:
Passes all my renderer tests also. Merging into master and RB-1.7.

PR #659, opened 2016-07-21
SS and group options "exec_repeat" can cause shader execution to repeat.

lgritz:
The default is 1, which is the old behavior: each call to
ShadingSystem::execute() will run the shader group once.

How is this useful? Suppose you have a very complicated app like a
renderer, and you're curious how much total time is attributable to the
actual execution of the shader. Each individual shader execution may be
so quick that putting timers around the calls may throw off the overall
time of the app (it takes time to retrieve the time, so by measuring
you are changing the timing characteristics; Heisenberg would
be proud).

So you set exec_repeat to 2, running each shader TWICE per call to
execute(). If you time this and compare it to shading the usual 1 time
per execution, the difference should be the total time spent shading for
an ordinary run. (But NOT including texture I/O time, since the "first
shade" will include any I/O that is triggered by needing non-resident
tiles, which will then already be resident for the "second shade." But
the OIIO TextureSystem stats should also tell you the amount of I/O
time).

Setting exec_repeat=0 will skip texture execution. I'm not sure how that's
helpful, since it changes appearance, duh.

You can set the exec_repeat for the whole ShadingSystem, and also override
it for individual groups.

Beware that a few things with side effects (printf, error calls, and
pointcloud_write) may change overall rendering behavior by doing them
multiple times. But everything else should be safe and yield the same
rendered results, if I've done this all correctly.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -96,6 +96,7 @@ public:
     ///    int profile            Perform some rudimentary profiling (0)
     ///    int no_noise           Replace noise with constant value. (0)
     ///    int no_pointcloud      Skip pointcloud lookups. (0)
+    ///    int exec_repeat        How many times to run each group (1).
     /// 2. Attributes that should be set by applications/renderers that
     /// incorporate OSL:
     ///    string commonspace     Name of "common" coord system ("world")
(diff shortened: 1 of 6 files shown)


lgritz:
Any opinions on this?

fpsunflower:
Assuming the extra logic isn't slowing down the regular case at all, LGTM

lgritz:
Too small to time it.

PR #660, opened 2016-08-02
Fixed OSL include support for UTF-8 file on Windows

ElaraFX:
This is caused by boost wave which is using ifstream by default. In this fix, I copied the implementation of load_file_to_string from boost source code and adapted it to support UTF-8.

The only thing that needs to be merged actually is at https://github.com/imageworks/OpenShadingLanguage/pull/660/files#diff-ab270a5694f4df87bfdc2be02f8a4347R201

I don't really have time to separate it from other changes, sorry for that.


lgritz:
I need to take a closer look, but the basics look ok. 

It would be really helpful when submitting changes like this to rebase atop the current master before submitting the PR.

lgritz:
Did you intend this to be a PR against RB-1.7, rather than against master? Because looking at the current master, I can see that a lot of the changes are already present (using OIIO::Filesystem::open, for example).

lgritz:
This has languished for a long time without a reply or rebase. Closing.

PR #662, opened 2016-08-03
Typo bug transpose() vs transposed().

lgritz:
Imath::M44f.transpose() transposes in place and returns the ref,
whereas transposed() returns the transpose without altering the original.

Not a symptomatic bug because we only did this on the result of an
inverse(), which was a temporary, so didn't hurt to mess with it in place.
But better to correct this code to properly reflect our intent.

### DIFF
--- a/src/liboslexec/llvm_ops.cpp
+++ b/src/liboslexec/llvm_ops.cpp
@@ -642,14 +642,14 @@ OSL_SHADEOP void osl_transformn_vmv(void *result, void* M_, void* v_)
 {
    const Vec3 &v = VEC(v_);
    const Matrix44 &M = MAT(M_);
-   M.inverse().transpose().multDirMatrix (v, VEC(result));
+   M.inverse().transposed().multDirMatrix (v, VEC(result));
 }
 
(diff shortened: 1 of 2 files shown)


src/liboslexec/llvm_ops.cpp
```diff
@@ -642,14 +642,14 @@ OSL_SHADEOP void osl_transformn_vmv(void *result, void* M_, void* v_)
 {
    const Vec3 &v = VEC(v_);
    const Matrix44 &M = MAT(M_);
-   M.inverse().transpose().multDirMatrix (v, VEC(result));
+   M.inverse().transposed().multDirMatrix (v, VEC(result));
```
fpsunflower: Would be even nicer to just have `multDirMatrixT` so you don't need to flip the matrix at all.

But I guess this function is rarely called ...

lgritz: This is Imath::M44f, it's a method. So I can't just add multDirMatrixT without also breaking symmetry in a weird way.

If only Bjarne had won the member-vs-function battle!

fpsunflower: Indeed ... LGTM!

PR #663, opened 2016-08-04
Fix OSL compiler assert when indexing into closure type

brechtvl:
Minor issue, when compiling this shader with `oslc` or an application embedding the OSL compiler it crashes:
```
shader test()
{
    closure color A;
    A[0] = 0;
}
```

```
src/include/osl_pvt.h:314: failed assertion '! is_closure() && "Don't call this if it could be a closure"'
```

From: https://developer.blender.org/T49016


lgritz:
LGTM!

PR #665, opened 2016-08-12
Implement pow(triple,triple)

lgritz:
Previously we had pow(float,float) and pow(triple,float).

The docs certainly implied that the second argument could be a triple
if the first argument was a triple.


fpsunflower:
LGTM

PR #666, opened 2016-08-23
Fix variant of spline() that didn't compute derivatives properly.

lgritz:
Specifically, the case where the knots are colors (without derivs), the
abscissa float with derivs.

(And add a regression test.)


fpsunflower:
LGTM

PR #667, opened 2016-08-23
Add ability to have derivs everywhere.

lgritz:
New SS attribute "force_derivs", when nonzero, will store and compute
derivatives for all float-based symbols, even if dependency analysis shows
they are not needed.

This is handy for quantifying how good we are at recognizing when we need
derivs (answer: very good, only a small percentage of post-optimized syms
end up with derivs) and also for debugging suspected derivative problems
(by essentially forcing derivs on almost everywhere).

Also amend stats report how many symbols carry derivatives.


lgritz:
I pushed an updated with a new name for the attribute, that I liked better. No other changes.

fpsunflower:
LGTM

PR #668, opened 2016-08-30
Allow optimization of shader groups with known raytype bits.

lgritz:
A new ShadingSystem::optimize_group() method takes bit fields describing
which ray types are known to be true and known to be false (not being
specified in either means it's indeterminite at optimize time, will only
be known at execution time).

When compiled in this manner, the runtime optimizer will constant-fold
calls to raytype(name) when the name matches those in the "in" and "out"
lists you have supplied.

This allows an app to specialize a shader for particular ray types (such
as a highly optimized version just for shadow rays).

There is also a new group attribute query "raytype_queries", which reveals
the bit pattern of any raytype() calls mde by any of the shaders within
the group.

As an example, the basic logic for specializing shadow shades and non-shadow
shades looks like this:

    // Make the shader group, however you want
    ShaderGroupRef group = shadingsys->ShaderGroupBegin ("groupname", ...);
    ...
    ShaderGroupEnd ();

    // A second reference-counted shader group starts just referring to the
    // same group.
    ShaderGroupRef group_shad = group;

    // Figure out if the shader group contains any raytype("shadow") queries.
    int rayqueries = 0;
    shadingsys->getattribute (group.get(), "raytype_queries", rayqueries);

    std::string serialized;
    int shadowbit = raytypebit(ustring("shadow"));
    if (rayqueries & shadowbit) {
        // The shader group asks (or might ask) for raytype("shadow"). Let's
        // assume it has different behavior for shadow rays and specialize.
        // First, make another group from the same serialized description.
        shadingsys->getattribute (group.get(), "pickle", serialized);
        group_shad = shadingsys->ShaderGroupBegin ("groupname", "surface", serialized);
        ShaderGroupEnd ();
        // Now specialize the shadow-specific group for shadow rays only:
        shadingsys->optimize_group (group_shad.get(),
                                    shadowbit,   /* ray types on: shadow */
                                    ~shadowbit); /* ray types off: everything else */
        // And specialize the original group for anything but shadow rays:
        shadingsys->optimize_group (group_shad.get(),
                                    0,           /* ray types on -- anything */
                                    shadowbit);  /* ray types off -- shadow */
    } else {
        // The shader group definitely does not check for raytype("shadow"), so
        // there's no point in specializing. Just stick to one version of the
        // shader. It will compile automatically in a generic way whenever we
        // first call shadingsys->execute () with that group. We don't need to
        // do anything special.
    }

    ... shading time: ...
    if (it's a shadow ray)
        shadingsys->execute (context, *group_shad, shaderglobals);
    else /* not a shadow ray */
        shadingsys->execute (context, *group, shaderglobals);

The bad news is that it takes a bit more memory (you're making extra
shader groups) and more time to compile (you're optimizing twice for
many shaders). Nonetheless, I'm finding that I come out ahead.

On my production tests, I'm seeing between 2% and 15% speedup -- very
dependent on the scene, as well as the structure of your shaders. YMMV.

The whole scheme assumes that your shaders take significant shortcuts for
shadow rays, based on explicitly using "if (raytype("shadow")) ...".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -561,7 +561,7 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             oslinfo-metadata oslinfo-noparams
             osl-imageio
             printf-whole-array
-            raytype reparam
+            raytype raytype-specialized reparam
             render-background render-bumptest
             render-cornell render-furnace-diffuse
             render-microfacet render-oren-nayar render-veachmis render-ward
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -294,6 +294,7 @@ public:
     ///                                  until the shader actually runs.
     ///   int num_renderer_outputs   Number of named renderer outputs.
     ///   string renderer_outputs[]  List of renderer outputs.
+    ///   int raytype_queries        Bit field of all possible rayquery
     ///   int num_entry_layers       Number of named entry point layers.
     ///   string entry_layers[]      List of entry point layers.
     ///   string pickle              Retrieves a serialized representation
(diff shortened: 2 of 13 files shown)


fpsunflower:
LGTM

fpsunflower:
I'd like to make an additional suggestion (which builds upon what you already have).

Very few closures in the renderer apply to shadow rays. Really only `transparent()`, `absorption()` and maybe one or two others.


What if we could tag a closure with a raymask that indicates which raytypes it is valid for?

Then at optimization time, you could simply generate 0 if the closure isn't relevant for the particular ray mask being JIT'ed.

This means you could write a trivial shader like:
```c
shader simple(color C, color O) {
  Ci = O * C * diffuse(N) + (1 - O) * transparent();
}
```

And get the same performance as if you had written the more verbose:
```c
shader simple(color C, color O) {
   if (raytype("shadow"))
     Ci = (1 - O) * transparent();
   else
     Ci = O * C * diffuse(N) + (1 - O) * transparent();
}
```

PR #669, opened 2016-09-10
liboslnoise -- C++ library with noises that match OSL

lgritz:
This patch makes a new library, liboslnoise (with corresponding
oslnoise.h), that just contains the various noise functions from OSL,
exposed in a way that can be easily called from other applicaitons.

For our first stab at this, only the perlin noises (signed and unsigned)
and cellnoise are exposed. We have varieties that take 1-, 2-, 3-, and 4-D
domain coordinates, and return either float or vector (3D) noise values.

At this point, we DO NOT expose the varieties that take derivatives, nor
the periodic perlin noises, nor the gabor and simplex noise varieties,
or varieties that directly take or return our simd types.  We may add
some or all of these in the future, but I thought that we should gague
reaction to what we have so far and make sure we're on the right track.

Although we produce a liboslnoise.so, the fact is that it only contains
the implementations of the gabor and simplex noises, so for now at
least, for the sake of using just the perlin and cell varieties, you can
treat this as a header-only library, it's not necessary to link to
liboslnoise.

Added a unit test oslnoise_test, that verifies correct functionality,
that the liboslnoise works without the rest of oslexec, and does timing
benchmarks on the noise functions.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,10 @@ Language, standard libary, and compiler changes (for shader writers):
 * Standard library additions/changes:
 
 API changes, new options, new ShadingSystem features (for renderer writers):
+* New library: `liboslnoise` and header `oslnoise.h` expose OSL's noise
+  functions to C++ apps. Currently, only signed and unsigned perlin, and
+  cellnoise, are exposed, but other varieties and options may be added if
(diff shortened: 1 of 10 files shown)


fpsunflower:
LGTM

PR #671, opened 2016-09-21
Bug fix: incorrect dropping of optional texture arguments

### LINKED ISSUE #670: texture() parameters optimization broken
There is a bug with (at least) the "swidth" and "twidth" parameters of the texture() shadeop. If they are set from a parameter and that parameter has a default or set value of 1, any connection to the parameter is ignored. It looks somewhat like there is an optimizer which pre-fills the texture opts and is not aware of connections. It is easy enough to work around by making the default 1.000001 in the shader.

I'm attaching a test case for testrender. It is a simple ramp of the filter width over the image so the texture should be sharp at the top and blurry at the bottom. You can use any texture you have lying around with some detail. Uncomment the Parameter in the xml to see what it should look like without the bug. It is broken in 1.7.2 and in the current git head.

[OSL_width_connection.tar.gz](https://github.com/imageworks/OpenShadingLanguage/files/480394/OSL_width_connection.tar.gz)

lgritz:
There was a well-meaning optimization: dropping the setting of TextureOpt
fields when the value was known to overwrite the field with the default
value. But of course this should only happen when the value is CONSTANT,
it shouldn't happen just because the value passed is a param and the
parameter's default value is the option default value -- because the
param may be connected to an earlier layer that passes a different
value.

The test is fixed now, the optimization only kicks in when the value
passed is definitely a constant. (Unconnected params should have been
turned into consts before this code is encountered, so it should still
provide the optimization all the places where it was originally intended
to work.)

Fixes #670.

Added a test to the testsuite to verify the fix.


fpsunflower:
LGTM

PR #673, opened 2016-09-27
New getattribute queries for shader, layer, and group name.

lgritz:
getattribute ("shader:shadername", dst)
    getattribute ("shader:layername", dst)
    getattribute ("shader:groupname", dst)

And of course they constant fold, so there is no runtime cost.

Beware: the layer names could be screwy if you are using the duplicate
instance merging optimization. If you are depending on layer names for
debugging, you may want to turn off just that optimization (set
ShadingSystem attribute "opt_merge_instances" to 0) while debugging.


fpsunflower:
LGTM

PR #674, opened 2016-09-30
Be sure to propagate C++ standard flags to the compilation of llvm_ops.cpp

lgritz:
We weren't properly transferring --std=c++11 from the directives to compile OSL's implementation, to the command line that compiles the llvm bitcode from llvm_ops.cpp. They should match.

This doesn't change any of the default for whether we use C++11, it just correctly uses the same version of the standard when comping llvm_ops.cpp as the rest of the software.

This mimics the way we propagate the -m flags (for SSE, AVX, etc.) to the compilation of the bitcode.


lgritz:
Wait, don't review this yet, it's slightly more complicated thanks to some bits in OpenEXR's headers that are not C++11 safe. I need to propagate some warning suppression as well as this flag.

lgritz:
OK, I think I have it set up correct now, ready to review.

aconty:
LGTM!

src/liboslexec/llvm_ops.cpp
```diff
@@ -91,7 +91,15 @@ examples), as you are just coding in C++, but there are some rules:
 */
 
 
+// Some gcc versions on some platforms seem to have max_align_t missing from
```
fpsunflower: Where is `max_align_t` used?

lgritz: We don't use it in our code. It's referenced in some standard headers. The conflict comes from the setup on Travis (and I assume this can be the case for external users as well) compiling  with llvm_ops.cpp with clang 3.4 but having gcc6 as the main system compiler. I think some of the logic clang uses to figure out which headers it can get from the system compiler and which it supplies itself ends up somehow getting a combination in which this definition is referenced but not provided. I'm assuming it will be resolved when we upgrade to newer clang/llvm versions, which I'm actively working on (between fires).

lgritz: It might also be broken only in particular gcc releases. In any case, that's just a safeguard to make sure it's defined properly on those platforms where somehow it doesn't pick it up. Because it's in a private namespace before any headers are included, it will be subsumed by the one in std, on systems that provide it properly.

This is all a "C++11 growing pains" problem exacerbated by the fact that we're stuck on such an old clang.

Working on that.

fpsunflower: Makes sense. LGTM!

lgritz:
I should add:

This was necessary to get into OSL before we throw the big switch on OIIO to make it C++11 only. In the process of testing that, I discovered that all along we had been compiling llvm_ops.cpp into bitcode in C++03 mode, even when compiling the rest of OSL and/or OIIO as C++11. Never noticed, didn't care... until an OIIO header comes along that actually generates an error when used in C++03 mode. So I just want this codebase to be robust to that before it hits the ground. This patch merely ensures that when you compile the rest of OSL with C++ 03 or 11, the same standard version is used to compile llvm_ops.cpp to bitcode with clang.

PR #675, opened 2016-10-03
Bug fix: matrix param initialized with matrix(space,1) left out init ops

lgritz:
When generating the static construction values, it needed to recognize
that case and set the "completed" flag to false, so that it would know
that it couldn't be constructed statically and needed init ops emitted
for that paramter.


fpsunflower:
LGTM

PR #676, opened 2016-10-06
Better runtime optimization of some matrix constructors.

lgritz:
matrix(spaceA,spaceB) already optimized to a constant matrix if spaceA and spaceB were both constant names and did not refer to motion-blurred transformations.

New case: if spaceA and spaceB alias to the same symbol, then we know this is the identiy matrix, even if they are not known constant values, and even if they refer to a motion-blurred space.

matrix(float0, float1, ... float15) already turned into a constant matrix if all the values were constant.

New case: matrix(float0) can be turned into a constant matrix if its value is constant (it's usually 1, but it doesn't have to be).

### DIFF
--- a/src/liboslexec/constfold.cpp
+++ b/src/liboslexec/constfold.cpp
@@ -1845,31 +1845,36 @@ DECLFOLDER(constfold_triple)
 
 DECLFOLDER(constfold_matrix)
 {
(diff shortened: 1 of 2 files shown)


fpsunflower:
LGTM

PR #678, opened 2016-10-07
Fix oslnoise link errors when building static OSL libraries.

brechtvl:
With static libraries the indirect dependency is not picked up automatically, so explicitly add it to `testshade` and `testrender`.

### DIFF
--- a/src/liboslexec/CMakeLists.txt
+++ b/src/liboslexec/CMakeLists.txt
@@ -28,12 +28,10 @@ if (NOT BUILDSTATIC)
 endif ()
 
 # oslnoise symbols used in oslexec
-if (NOT BUILDSTATIC)
-    LIST(APPEND liboslexec_srcs
-        ../liboslnoise/gabornoise.cpp
-        ../liboslnoise/simplexnoise.cpp
-        )
-endif ()
+LIST(APPEND liboslexec_srcs
+   ../liboslnoise/gabornoise.cpp
+   ../liboslnoise/simplexnoise.cpp
+   )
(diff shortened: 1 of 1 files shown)


lgritz:
I'm confused. liboslnoise is not needed by testshade and testrender. The noise functions are all in liboslexec. The liboslnoise is redundant, only for the benefit of apps that want *only* the noise functions without needing an entire OSL runtime and LLVM.

brechtvl:
There is this in `src/liboslexec/CMakeLists.txt`:
```
if (NOT BUILDSTATIC)
    LIST(APPEND liboslexec_srcs
        ../liboslnoise/gabornoise.cpp
        ../liboslnoise/simplexnoise.cpp
        )
endif ()
```

So for static builds, the noise functions are currently not included in liboslexec. If they are included users have to ensure they link only liboslexec or liboslnoise to avoid duplicate symbols. I'm not sure which is preferable.

brechtvl:
It would a little easier for existing users to include the noise functions in liboslexec, as existing build scripts wouldn't need any changes to add a new liboslnoise library.

So just removing that `if (NOT BUILDSTATIC)` may be a better fix.

lgritz:
Oops. Yes, I think you are correct. The error is that those two source files should be added to liboslexec regardless of whether it's being build as a static or dynamic library.

brechtvl:
Ok, I've updated the patch to do that.

lgritz:
LGTM, I will merge.
Thanks for the fix.

PR #683, opened 2016-10-13
Suppress clang 3.9 warnings

lgritz:
Equivalent to the change we made in OIIO on this same topic: https://github.com/OpenImageIO/oiio/pull/1529


aconty:
LGTM

PR #684, opened 2016-10-13
oslc better err reporting for use of non-function symbol as function.

### LINKED ISSUE #681: Variable names as a function crashes oslc
Hi,

Originally was reported to the Blender's bug tracker [1] but this issue can be reproduced with the current oslc (compiled from 5c8c973).

Easy to reproduce:

Create test.osl with the following content:

```
void bug(float cos){cos(1);} 
shader test(){}
```

and try to compile it using oslc.

[1] https://developer.blender.org/T49597

lgritz:
See the new testsuite example. This crashed oslc before!
Fix it up to not crash (by marking the unfound function's symbol as NULL),
and also improve the resulting error message.

Fixes #681


aconty:
LGTM

PR #685, opened 2016-10-13
Opt: triple(space,const,const,const) can get fully folded to a constant

lgritz:
Opt: triple(space,const,const,const) can get fully folded to a constant if the space name is known to be equivalent to "common".


aconty:
LGTM

PR #686, opened 2016-10-14
Add optional "errormessage" parameter to texture lookup functions.

lgritz:
OSL's texture(), texture3d(), and environment() functions now take a new
optional argument, "errormessage", followed by string variable
reference. A successful texture lookup will result in storing the empty
string ("") in the variable.

When this option is employed, a failed lookup or other error, such as a
missing texture, will store a non-empty string in the errormessage
variable (usually a descriptive error message, but will be "unknown" if
no message is forthcoming from the TextureSystem, as may be the case if
it's suppressing duplicate error messages from the same file). If the
"errormessage" option is passed, this will be the ONLY way an error is
handled for this call; it will not also forward any errors to the
renderer, assuming that if the shader asks for the message, the shader
intends to fully handle the situation.

If no "errormessage" optional param is passed, behavior is as it always
was: a texture lookup error will be passed to the renderer through the
usual error reporting mechanisms.

The reason for this change is UDIM -- a UDIM set may be sparse, with
some sections of uv space not having concrete texture files that exist.
We don't want those to spew error messages, and our typical way of
checking by doing a query like gettextureinfo(texname,"channels",chans)
is not reliable because it will return the channels for ANY match, doesn't
help you for particular uv ranges.

With this, you can imagine a shader gracefully (and more or less
efficiently) handling a sparse UDIM set with logic that looks like:

    int channels;
    if (gettextureinfo (texturefile, "channels", channels)) {
        string err;
        C = texture (texturefile, u, v, "errormessage", err);
        if (err != "") {
            // texture set exists, but no file for this UDIM section!
            C = missing_udim_color;
        }
    } else {
        // truly nonexistant texture set
        C = error_color;
    }

Unfortunately, this breaks some compatibility with the RendererServices
API. (That's ok, this is for master only, and will not be backported
to release branches because I just can't figure out how to do it without
an API change.) Renderers integrating OSL will need to add new
RendererServices methods texture(), environment(), and texture3d(), which
have an additional ustring* parameter for the error message destination.
Eventually, we'll get rid of the old versions. Allowing both for now will
limit the number of immediate breakages until the renderer authors get
the new calls squared away.


aconty:
LGTM

lgritz:
Please note that if all you want to do for missing UDIM regions is substitute a default color and suppress error messages entirely, then I recommend using the existing "missingcolor" (and "missingalpha") parameters, and not using "errormessage". The "errormessage" handling is intended for cases where you want more complex behavior than simply substituting a color.

PR #689, opened 2016-11-23
Change partio search

lgritz:
* Everybody: break partio finding into separate FindPartio.cmake
* SPI only: change Partio version/location, update our ref output (order of points returned by the search has changed, harmless, but doesn't match the old ref output)


cmstein:
LGTM!

Issue #480, opened 2015-03-14
Support LLVM 3.6 


svenstaro:
I know LLVM 3.6 breaks in all kinds of ways because the classical JIT has been removed and now only MCJIT is in there. This is entirely unsupported by OSL right now because the JITMemoryManager is now also gone. I suppose this isn't going to happen to quickly but I think we can use this issue to track that nonetheless.

mdriftmeyer:
This is sad. LLVM 3.6.2 just got RCd and August LLVM 3.7 OpenMP rolls, but OpenShading is stuck in the past crippling a lot of apps and features, not the least of which is Blender.

aghiles:
Not sure what you mean by "crippling a lot of apps" here. Certainly not crippling our app. It would be crippling if OSL shaders would run much faster with MCJIT that they do with the current JIT. Would that be the case ?

lgritz:
It's true that OSL doesn't currently build against LLVM 3.6. I'm working on fixing this ASAP (made a lot of progress on it today, prompted by this ticket).

For those not in the know, here's why it has not been high priority.

* None of the major VFX and animation studios are yet on a C++11 toolchain. They are basically forced into this state by the need to be link compatible with certain 3rd party apps such as Maya and Houdini. Since LLVM/Clang 3.5 requires C++11 to build, that means that none of the major studios -- or software vendors supplying them -- has any incentive or ability to use LLVM newer than 3.4. For now. It is hoped that for 2016, we'll all upgrade to C++11, but for now, nobody in VFX is using it.

* Starting with LLVM 3.6, the "old JIT" has been removed completely, forcing us to use MCJIT. In our experiments, MCJIT is much slower than old JIT, to a point where it's really a big problem for us with lots of complex shaders in a production environment. So even if our toolchain allowed us to use LLVM 3.6 (which it doesn't), at the moment we would very much want to hold back to 3.5 for performance reasons. I'm working on this (quite honestly, very slowly, since it's not a big problem until LLVM 3.6, and we can't even use 3.5 in production yet).

I'm not proud of any of this, but it is what it is. I thought it was worth explaining that it's actually not completely stupid that we are targeting an older version of LLVM.

[...]

PR #693, opened 2017-01-20
Fixes to build with modern LLVM.

marsupial:
All tests passing against LLVM 3.4 and 3.9.
Also have a patch to use the object caching features of the MCJIT that speeds it up dramatically.

### DIFF
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -636,11 +636,10 @@ LLVM_Util::setup_optimization_passes (int optlevel)
 #if OSL_LLVM_VERSION >= 37
 
     m_llvm_func_passes = new llvm::legacy::FunctionPassManager(module());
-    //m_llvm_func_passes->add(new llvm::DataLayout(m_llvm_exec->getDataLayout()));
+    llvm::legacy::FunctionPassManager &fpm = (*m_llvm_func_passes);
 
     m_llvm_module_passes = new llvm::legacy::PassManager;
-    llvm::legacy::PassManager &mpm (*m_llvm_module_passes);
-    //mpm.add (new llvm::DataLayout(m_llvm_exec->getDataLayout()));
+    llvm::legacy::PassManager &mpm = (*m_llvm_module_passes);
 
 #elif OSL_LLVM_VERSION >= 36
 
@@ -675,7 +674,6 @@ LLVM_Util::setup_optimization_passes (int optlevel)
 #endif
 
     if (optlevel >= 1 && optlevel <= 3) {
-#if OSL_LLVM_VERSION <= 34
         // For LLVM 3.0 and higher, llvm_optimize 1-3 means to use the
         // same set of optimizations as clang -O1, -O2, -O3
         llvm::PassManagerBuilder builder;
@@ -684,10 +682,6 @@ LLVM_Util::setup_optimization_passes (int optlevel)
         // builder.DisableUnrollLoops = true;
         builder.populateFunctionPassManager (fpm);
         builder.populateModulePassManager (mpm);
-#else
-        // FIXME -- should we have the equivalent for LLVM >= 35?
-#endif
-
     } else {
         // Unknown choices for llvm_optimize: use the same basic
         // set of passes that we always have.
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -63,7 +63,12 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <llvm/IR/LegacyPassManager.h>
 #include <llvm/Support/TargetRegistry.h>
 
-#include <llvm/Bitcode/ReaderWriter.h>
+#if OSL_LLVM_VERSION < 40
+#  include <llvm/Bitcode/ReaderWriter.h>
+#else
+#  include <llvm/Bitcode/BitcodeReader.h>
+#  include <llvm/Bitcode/BitcodeWriter.h>
+#endif
 #include <llvm/Support/ManagedStatic.h>
 #include <llvm/Support/MemoryBuffer.h>
 #include <llvm/ExecutionEngine/GenericValue.h>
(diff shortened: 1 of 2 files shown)


lgritz:
This is very exciting!

I'm simultaneously working on an upgrade-to-LLVM3.9 patch (including transition to ORC JIT), but I've honestly been having a lot of trouble getting it fully working.

I would be extremely pleased if this patch was able to get us to a point where we could efficiently use LLVM3.9 with MCJIT, and then I can build on top of it to get a full ORC JIT transition.

I'll need some time to evaluate this, but I'm very hopeful about it. 

Please get a fully signed CLA to opensource@imageworks.com as soon as you can; it's an absolute prerequisite to merging any contributed code to this project.

marsupial:
Signed and sent.
I haven't tested it that thoroughly yet, though as said all tests are passing.
The caching makes a ton of difference though still seems to be 10% slower.
Don't have any intensive scenes to test if that scales or exceeds the 3.4 Jit.
If you had any you could pass along or point to (Prman & Arnold should be available).

dracwyrm:
If it helps, I was able to compile OSL with this patch and to test OSL scripts in Blender 2.78a with LLVM 3.9. Before there were crashes, but now there are none when it renders, so this does appear to work with simple scripts. However, I haven't tried more complicated ones yet.

If you want scripts to test, I found these when I did a Google search, so they may help validate if this patch is truly working:
https://github.com/sambler/myblendercontrib/tree/master/online_mat_lib/material-library/bundled/cycles/scripts
https://github.com/sambler/osl-shaders

Cheers.

lgritz:
Fixes #480

lgritz:
It took a bit of cmake script cleanup on my end, but I'm able to verify that this patch allows OSL to build against LLVM 3.4, 3.5, and 3.9, on both Linux and OSX, and it passes both the OSL testsuite as well as the one for our in-house renderer.

I'm still working on production-size benchmarks, but this is passing a big hurdle already crossed. I'll try to get a full approval for this by the end of the week.

src/liboslexec/llvm_util.cpp
```diff
@@ -570,9 +696,11 @@ LLVM_Util::setup_optimization_passes (int optlevel)
         mpm.add (llvm::createVerifierPass());
         // Simplify the call graph if possible (deleting unreachable blocks, etc.)
         mpm.add (llvm::createCFGSimplificationPass());
+#if OSL_LLVM_VERSION <= 36
```
lgritz: Why is this optimization disabled for newer releases?

lgritz:
OK, let me summarize where we are with this:

The code looks very good. Other than the one question I asked about a disabled optimization, it all looks correct and I don't have any requests to alter it. I do, however, have a bunch of additional changes I made to the cmake and travis files, necessary to get your changes to build cleanly on all the platforms I use as well as CI, and I'll commit those fixes simultaneously with yours.

I've been able to build on OSX and Linux, using LLVM 3.4, 3.5, 3.9, all passing the OSL test suite as well as the test suite for our internal renderer. I haven't tried the intermediate versions (3.6, 3.7, 3.8) but I see there are what look like the appropriate `#ifdef`s so it wouldn't surprise me if they worked. But since I was already prepared to skip all the way to 3.9 soon, I'm not terribly inclined to do much work to verify or fix 3.6-3.8. 

I've tested a number of production frames and have the following findings:

* When building against LLVM <= 3.5, performance and memory is identical to what we had before, which of course is as we expect/hope, since you're still using old JIT in that case. But good to confirm, and it means that no matter what other pros/cons we have, it's "safe" to merge this because all people need to do to avoid losing any performance is just stay on LLVM 3.4/3.5 (while we continue to improve things for LLVM >= 3.9).

* When building against LLVM3.9 (MCJIT), I'm seeing slightly more memory use, but not enough to care about (maybe another 50MB RSS size on a several-GB render. I'm not inclined to sweat this, especially considering that we want a future step to be to move from MCJIT to ORC JIT, which as I understand it is better about compiling lazily and being able to throw things away when it's done with them.

* With LLVM 3.9 / MCJIT, the compilation is MUCH SLOWER. Across several scenes, I'm consistently seeing the total time we spend doing LLVM JIT and optimize increase by 2-3x! In the context of a multi-hour render, it's a tiny percentage of time, not a big deal. And I'm seeing some evidence (rerendering now to be sure, I'll report back) that the generated code sometimes runs faster for some shaders/scenes, so long frames may even go faster at full resolution. But it's going to hurt for short interactive renders. 

Nonetheless, I think it's a step forward to merge this now. It can't hurt anyone who sticks with 3.4/3.5. People who want to move to 3.9/4.0 right away **CAN**, knowing that it's probably worse performance for interactive renders, but approximately neutral for long full-res renders, and that once it's merged, you can be sure we will be looking for ways to optimize it, as well as switching to ORC JIT to see if that helps.

I'm letting my long benchmarks all run again, but I'm currently thinking of merging this (plus my changes) on Monday, if there are no objections.

Great contribution!

src/liboslexec/llvm_util.cpp
marsupial: `createScalarReplAggregatesPass` doesn't exist anymore, or if it does I cannot find it.

Updated with a comment and moved the **#ifdef** a line lower (where it should have been to begin with).

dracwyrm:
Just out of curiosity, there are no changes in the rendered frames between versions of LLVM, like when comparing them, they are exactly the same? I've read how newer versions of LLVM have increased the quality of other programs due to the new techniques of doing things.

Also, would it be possible to note in the Documentation that 3.9/4.0 may run slower on certain conditions as of right now? I know that will change with future enhancements.

Thanks.

lgritz:
The rendered frames are identical.

lgritz:
Some more details on this overhead vs performance tradeoff with this patch:

As I said, as long as you build against LLVM 3.4 or 3.5, you will get no change in performance. So for the rest, I'm only talking about if you upgrade to LLVM 3.9.

I tried 3 production frames (#1 from a recent past animated feature, #2 from an in-progress animated feature, #3 from an in-progress VFX show). In call cases, below, the total render times are wall clock times while rendering with all cores, and the "LLVM" time is total summed over all cores (hard to translate to wall clock time, but try dividing by 12 to get a rough idea).

    Scene  3.5/old  3.9/MCJIT
    -----  ------   -------

     1
            4:44 ->    9:21  LLVM-related time (across all threads)  and +60MB
           55:31 ->   56:37  Wall time, 64 samples per pixel
         3:28:58 -> 3:27:47  Wall time, 256 samples per pixel

     2
            4:35 -> 10:03    LLVM-related time (across all threads)  and +33MB
         1:28:32 -> 1:23:56  Wall time, 64 samples per pixel
         5:50:29 -> 5:43:15  Wall time, 256 samples per pixel

     3
           12:14 -> 29:09    LLVM-related time (across all threads)  and +120MB
           10:37 -> 11:27    Wall time, 144 samples per pixel
           27:00 -> 26:23    Wall time, 400 samples per pixel



So you can see the effect I mean -- the resolution-independent LLVM related processing approximately doubles (or slightly more), which is going to be very noticeable for quick, interactive or low-quality renders. But for multi-core-hour full-quality frames, it seems slightly faster across the board. 

So, I think we should merge this, then work on ways to speed up the compilation. If the compilation overhead is going to wreck your day, just stay on LLVM 3.4/3.5 for a while longer as we work on this.

lgritz:
OK, I have merged this into master. Let me know if anything breaks!

PR #696, opened 2017-01-31
New OSL built-in: select(a,b,cond)

lgritz:
This is like `mix(a,b,x)`, except that cond is treated like a boolean, so the components of `result` are always either exactly the corresponding components of `a` or `b`, never anything in between.

This is equivalent to per-component `(cond?a:b)`
or equivalently mix `(a, b, step(0,abs(cond)))`
but the underlying implementation does not need conditionals or multiplies.


src/shaders/stdosl.h
```diff
@@ -165,6 +165,20 @@ color  mix (color  x, color  y, color  a) BUILTIN;
 color  mix (color  x, color  y, float  a) BUILTIN;
 float  mix (float  x, float  y, float  a) BUILTIN;
 #endif
+normal select (normal cond, normal x, normal y) BUILTIN;
```
fpsunflower: The docs have "cond" as the last argument?

lgritz: Oops, I dithered on the argument order a few times, changing this back and forth. In the end, I thought it made the most sense to be in the same order as mix(), but I may not have fixed it everywhere.

lgritz:
updated

fpsunflower:
LGTM

PR #698, opened 2017-02-02
Use llvm::getLazyBitcodeModule in newer LLVM.

marsupial:
Using `llvm::getLazyBitcodeModule` instead of `llvm::parseBitcodeFile` seems to provide a significant speed bump. The comment about why `parseBitcodeFile ` was used seems no longer valid.

I've only tested on OS X, so if it still is an issue on other platforms could one try the steps mentioned on line `519`?

### DIFF
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -170,7 +170,7 @@ public:
     void setup_optimization_passes (int optlevel);
 
     /// Run the optimization passes.
-    void do_optimize ();
+    void do_optimize (std::string *err = nullptr);
 
     /// Retrieve a callable pointer to the JITed version of a function.
     /// This will JIT the function if it hasn't already done so. Be sure
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -109,6 +109,11 @@ namespace pvt {
     typedef llvm::SectionMemoryManager LLVMMemoryManager;
 #endif
 
+#if OSL_LLVM_VERSION < 40
+    typedef std::error_code LLVMErr;
+#else
+    typedef llvm::Error LLVMErr;
+#endif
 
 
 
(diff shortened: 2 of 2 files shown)


lgritz:
Trying this out...

lgritz:
It passed the OSL testsuite immediately, but had crashes in my production frames and our renderer testsuite. Hitting assertions deep in LLVM.

I'm still not 100% sure why this was never a problem before, but I tracked down that it only seemed to happen when we optimized and JITed what turned out to be empty shaders. So PR #699 is my fix for that -- take a shortcut for empty shaders without invoking LLVM opt+jit. I'm not sure why we weren't taking that shortcut before, or exactly why it crashes, but whatever. Passes OSL testsuite, renderer testsuite, and my selection of production assets.

I'm running extensive benchmarks tonight. I'll report results tomorrow, and if they all turn out ok, I'll merge this patch.

marsupial:
I was seeing assertions as well before adding the call to **materializeAll**.
The stack was roughly:
```
LLVM_Util::do_optimize
  llvm::PassManager
   llvm::SomePassIForget
    llvm::Function::front()
      llvm::iplist -> "Called front() on empty list!"
```
Which would seem to be inline with an empty function (no BasicBlocks) as the culprit.

lgritz:
Yes, that's roughly what the assertion was. Looked like it had to do with no basic blocks somewhere. I'm not quite sure how that happens, because even for an empty shader, there should be one basic block (containing the 'return'). But whatever, it seems to only show up with completely empty shaders, and we can catch that case upstream.

lgritz:
Benchmark results (times are just for the LLVM related stuff, not total runtime of the renders):

Scene 1:  9:21 -> 5:15  -44% !!! (versus 4:44 for old jit)

Scene 2:  10:03 -> 9:13  -10%  (vs 4:35 for old jit)

Scene 3:  29:09 -> 30:21  +4% (vs 12:14 for old jit)

So it's not an across the board win on all scenes, but the things sped up seem to outweigh the things slowed down, so I will incorporate patch.

LGTM

lgritz:
Merged by hand

PR #699, opened 2017-02-03
Skip LLVM JIT for empty shaders.

lgritz:
A good optimization in any case, but some of the new code triggered
assertions in LLVM that only seemed to happen for empty shaders.


fpsunflower:
LGTM

PR #700, opened 2017-02-03
Fix subtle issue[s] with FunctionPass.

marsupial:
This might affect binary compatibility by changing LLVM_Util's layout.
Other than that there should be no behavioral change (by default).

Full commit message:
```
Fix issue with user added FunctionPass not being run and avoid allocating a
FunctionPassManager unless it is actually used.

If someone added a pass to LLVM_Util::setup_optimization_passes with opt level 0,
it would never be run. If a PassManagerBuilder was used, everything was fine.

Cleanup LLVM_Util::setup_optimization_passes and consolidate the LLVM
version detection macros.
```

### DIFF
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -487,6 +487,7 @@ public:
 private:
     class MemoryManager;
     class IRBuilder;
+    class PassManager;
 
     void SetupLLVM ();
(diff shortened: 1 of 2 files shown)


lgritz:
Closing. This section of code has all been changed in the mean time.

Issue #702, opened 2017-02-06
Memory leak


marsupial:
Am I missing something or is c8e38df now causing the llvm::MemoryBuffer to be leaked?

fpsunflower:
The code as it was didn't compile on our flavor of Linux (against LLVM 3.5). So it seemed most prudent to roll it back to what it was.

Are you sure about the ownership semantics here?

marsupial:
The code you added seems to only be for LLVM-3.5. I have no idea who owns it on 3.5.

Everything above is returning a _unique_ptr_ from `getMemBuffer` and takes one as an argument to `getLazyBitcodeModule` (to avoid this issue I presume).

So if `getMemBuffer` returns a unique_ptr, which is then converted to a raw pointer, isn't it actually worse and returned _unique_ptr_ is actually a temporary therefore deleting after assignment?

marsupial:
I guess the standards committee thought of this, it won't even compile anymore!
I'll submit a PR.

marsupial:
#703

lgritz:
Fixed by #703

PR #703, opened 2017-02-06
Add #ifdef for changes in c8e38df to allow building on LLVM > 3.5 again.

marsupial:


lgritz:
LGTM, thanks for the fix. I will test on my end and merge if there are no problems.

lgritz:
Fixes #702

marsupial:
Hold off on this for a minute.
I might have been a bit hasty...and this won't fix the leak in 3.5...

marsupial:
From what I gather reading though 3.5, c8e38df did introduce a leak.
Unless there is an objection I'm going to update this to fix the leak as well as compilation errors for 3.6+.

lgritz:
Yes,  please. 


On February 6, 2017 12:34:59 PM PST, marsupial <notifications@github.com> wrote:
>From what I gather reading though 3.5, c8e38df did introduce a leak.
>Unless there is an objection I'm going to update this to fix the leak
>as well as compilation errors for 3.6+.
>
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/imageworks/OpenShadingLanguage/pull/703#issuecomment-277805036

-- 
Larry Gritz
lg@larrygritz.com

marsupial:
Ok, updated. I'd certainly build and test against 3.5 as I haven't.
There was a possible leak, though the chances of it were probably small.

lgritz:
To make it still work for LLVM 3.4 / C++03, you need to change

    #if OSL_LLVM_VERSION < 40
        typedef std::error_code LLVMErr;
    #else
        typedef llvm::Error LLVMErr;
    #endif

to

    #if OSL_LLVM_VERSION >= 35
    #if OSL_LLVM_VERSION < 40
        typedef std::error_code LLVMErr;
    #else
        typedef llvm::Error LLVMErr;
    #endif
    #endif

because std::error_code is not present prior to C++11.

AFTER we branch RB-1.8, then we can be totally free with the C++11 idioms.

marsupial:
Yes `llvm::Error`, `llvm::error_code`, and `std::error_code` was horrifically annoying me last week.
I'm not sure how that relates to these changes though, or you want me to add something on top?

lgritz:
As far as I can tell, those error codes are only used for LLVM 3.5 and beyond (which are C++11), so I just enclosed that section in one more layer of 

    #if OSL_LLVM_VERSION >= 35

It would be handy if you did it on your end, but I can apply it on my end if you want. (I did so and was ready to merge, before you added another change; each time you push an update I have to re-apply it.)

lgritz:
I'll apply it on my end again, for the sake of expediently getting this merged.

marsupial:
Let's just keep this as is confined to the leak and failure to build > 3.6 from the previous commit?

Here's how I wound up solving the problem in unrelated code. The problem was that 
LLVM 3.4 no std::error_code, llvm::error_code
LLVM 3.5+ std::error_code
LLVM 3.9/4.0 both use std::error_code but 4.0 changes the usage of llvm::Error for some calls

```
#if OSL_LLVM_VERSION <= 34
typedef llvm::error_code LLVMErr;
typedef llvm::error_code std_error_code;
#else
typedef std::error_code std_error_code;
# if OSL_LLVM_VERSION >= 40
typedef llvm::Error LLVMErr;
inline bool error_string (llvm::Error err, std::string *str) {
    if (err) {
        if (str) {
            llvm::handleAllErrors(std::move(err),
                      [str](llvm::ErrorInfoBase &E) { *str += E.message(); });
        }
        return true;
    }
    return false;
}
# else
typedef std::error_code LLVMErr;
# endif /* OSL_LLVM_VERSION >= 40 */
#endif /* OSL_LLVM_VERSION <= 34 */

inline bool error_string (const std_error_code &err, std::string *str) {
    if (err) {
        if (str) *str = err.message();
        return true;
    }
    return false;
}
```

lgritz:
So, would you like me to merge this one now, and then tackle the LLVM 3.4 fix separately? Is that what you mean?

I just want the smallest possible fix for LLVM 3.4, because as soon as we branch, master is going to have the LLVM 3.4 support taken away.

lgritz:
Exactly which LLVM versions are you able to test against on your end?

lgritz:
With your latest patch here, I'm still getting build errors for LLVM 3.5 (but works for 3.4 and 3.9).

lgritz:
llvm_util.cpp:534:41: error: no viable conversion from 'llvm::MemoryBuffer *' to 'std::unique_ptr<llvm::MemoryBuffer>'
        std::unique_ptr<llvm::MemoryBuffer> buf =
                                            ^

marsupial:
Ugh! Think if I move that to a construction rather than assignment it should work.

lgritz:
yes, it does!
I'll just do that on my end

lgritz:
OK, with that change, I'm getting clean builds on OSX with llvm 3.4, 3.5, 3.9, and on Linux with 3.4, 3.5, 3.9, and 4.0.

Merging as soon as I see non-disastrous results from the TravisCI builds!

marsupial:
Updated here as well. I've noticed Travis not passing for any PR, but master seems to go fine.
It's a weird issue with building ImageIO on Linux, failing for -ldl.

For what it's worth I write/compile against 3.9 (against a 3.9 debug build tree).
Compile and run the tests against 4.0 release install.
Every now and then I run the changes against a 3.4 release install.

So before I submit a PR it's passing all tests on OS X for 3.4, 4.0; and definitely compiles against 3.9.

lgritz:
Don't worry about the Travis, I have a set of changes to address that.

marsupial:
Oh, are they available somewhere? I just started copying/pasting some Travis stuff I have so I can stop running the tests locally.

lgritz:
Not yet. Working on it. The code base keeps shifting around underneath me!

lgritz:
Your current #703/this patch (with my minor fixes on top of them) pass all the Linux flavors on Travis now.

The OSX Travis builds are messed up for an entirely different reason.

Merging...

lgritz:
MERGED!

No more code changes will be merged until I branch, working on that now.

PR #704, opened 2017-02-07
C++11: switch to C++11 baseline

lgritz:
Also require CMake >= 3.2.2


lgritz:
master only!  OSL 1.8 releases will remain buildable by C++03.

fpsunflower:
LGTM!

PR #705, opened 2017-02-08
Fix recent breakage against LLVM 3.5

lgritz:
The last couple checkins botched some stuff, was leading to crashes
on real frames even though testsuite had passed. For the LLVM 3.5
case, it wasn't right.

1. The call to materializeAll() in do_optimize seems suspicious
for LLVM 3.5, taking it away solves crashes, so at Chris Kulla's
suggestion we're changing from #if OSL_LLVM_VERSION >= 35 to
> 35, so it only does the materializeAll for MCJIT cases.

2. LLVM_Util::module_from_bitcode was a mess. Impossible to follow.
Instead of a line-by-line #ifdef madness, I refactored it into a
much simpler <= 3.5 section and a >= 3.6 section, even if there was
some code repetition between the two. Also, I hereby declare that
we will not support LLVM 3.6, which further simplifies the logic.

3. As a result of that refactor, it was more obvious that another change,
the call to buf.release() was never there before and probably represents
some kind of double free. The older (pre-Marsupial) code for 3.4 & 3.5
in fact had a comment stating "don't delete buf, the module has taken
ownership of it."

### DIFF
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -483,31 +483,38 @@ LLVM_Util::module_from_bitcode (const char *bitcode, size_t size,
     if (err)
         err->clear();
 
-    // Keep uninitialized so compiler will error if not set through macros below
-    llvm::Module *m;
-
-#if OSL_LLVM_VERSION <= 34
-
+#if OSL_LLVM_VERSION <= 35 /* Old JIT vvvvvvvvvvvvvvvvvvvvvvvvvvvv */
     llvm::MemoryBuffer* buf =
         llvm::MemoryBuffer::getMemBuffer (llvm::StringRef(bitcode, size), name);
 
-    m = llvm::getLazyBitcodeModule (buf, context(), err);
-
-#else // > 34
-
-# if OSL_LLVM_VERSION <= 36
-    typedef llvm::ErrorOr<llvm::Module*> ErrorOrModule;
-# elif OSL_LLVM_VERSION < 40
-    typedef llvm::ErrorOr<std::unique_ptr<llvm::Module> > ErrorOrModule;
+    // Create a lazily deserialized IR module
+    // This can only be done for old JIT
+# if OSL_LLVM_VERSION >= 35
+    llvm::Module *m = llvm::getLazyBitcodeModule (buf, context()).get();
 # else
+    llvm::Module *m = llvm::getLazyBitcodeModule (buf, context(), err);
+# endif
+    // don't delete buf, the module has taken ownership of it
+
+#if 0
+    // Debugging: print all functions in the module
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM

@marsupial seemed to think there was a memory leak there before, which is quite possible although it seems safer to maintain the old behavior for now. Hopefully this codepath won't be around for too much longer anyway.

lgritz:
I'm going to bet that if there was a memory leak, we would have noticed it in production, we've been using LLVM 3.5 for a while, therefore the code was probably either correct, or any leak was truly inconsequential.

lgritz:
@marsupial, I sincerely apologize -- you have a couple outstanding PRs and I'm sure more in-progress work that will probably be heavily merge conflicted with my refactor of module_from_bitcode, and require a bit of trouble on your part to rebase on top of it. But I hope that when you see what I've done here, you will also recognize that it's a helpful cleanup and will make it easier for us both to continue moving forward, by cleanly separating the old JIT from MCJIT parts of that function into completely disjoint blocks of text.

marsupial:
I mentioned in the PR the chances of the leak are small, but it is there:
If there is an error creating the llvm::Module, then it can't have taken ownership of the buffer, and you wind up owning/leaking it.

There was a slight conflict, but pretty trivial.
But should I really be rebasing to this or is master where the Orc work would go?

lgritz:
But doesn't the code as you had it free it after it's been transferred? It said

    if (m) buf.release();

Is it possible you meant

    if (!m) buf.release();

??

But anyway, the way it is now, it's doing exactly what it had been doing before, which nobody was complaining about. And when we eventually retire support for LLVM 3.5, this whole section will disappear, so it won't matter if there was an edge case that had a leak under an error condition.

lgritz:
So the current status is, 

* 3.4, 3.5 behave as they always did
* 3.9, 4.0 work, but their JIT is slower than 3.4/3.5 (though the code generated runs a bit faster, so it's a tradeoff).

This is currently the same in both 1.8 and master, though they are replicated in the two branches, so they may diverge.

I would say, we should not ever change the LLVM code again in the 1.8 branch. Unless there's a critical correctness error, I mean. 

Master is where we will try to continue improving the JIT (and other) performance, including a switch to ORC JIT. So for any work you're doing on ORC, please target only master.

But, like I said, to the extent that your ORC JIT work may also touch the same function I just patched, I may have made it no longer cleanly merge, and you'll have to do a little work to apply whatever in-progress ORC work you have to the new master.

marsupial:
No you want to release (not reset) when a module was created, because it know owns.
When no module you want ~unique_ptr to clear the memory.

Either way if there was no module you're probably in for a lot more hurt than leaking...

PR #706, opened 2017-02-08
Use std::shared_ptr

lgritz:
Now that we are C++11 minimum, we can use std::shared_ptr directly without needing to "wrap" it. A previous commit removed the clause of an `#if` that alternately defined it using boost::shared_ptr for C++03, so all that was left was pointless use of the wrapper.

This is for master only, will not be backported to RB-1.8, which must still remain compatible with C++03.


fpsunflower:
LGTM!

PR #707, opened 2017-02-09
C++11 and Boost changes

lgritz:
I've waited a long time for this flurry of adjustments to 'master', now that we are safely assuming a base of C++11 minimum, Boost minimum 1.55 per VFXPlatform's guidelines (as far back as CY2015), and also our stated intention for OSL master/1.9 to require a minimum OpenImageIO version of 1.7 (which is the current supported release, so seems safe).

In the process, a whole bunch of our Boost use disappears, in favor of C++11 std and modern OIIO wrappers:

* BOOST_FOREACH -> C++11 "range for"
* boost::unordered_map and unordered_set -> std
* boost string algorithms -> OIIO Strutil utilities
* boost::bind and ref -> std
* boost::filesystem -> OIIO Filesystem utilities

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ Dependency and standards changes:
   OIIO 1.7 or newer. You may find that 1.6 is still ok, but we are not doing
   any work to ensure that.
 * CMake >= 3.2.2
+* Boost >= 1.55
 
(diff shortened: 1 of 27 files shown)


fpsunflower:
LGTM!

PR #708, opened 2017-02-09
OIIO dependency fixes

lgritz:
* Fancy new FindOpenImageIO.cmake, and related cleanup
* Bump OIIO minimum to 1.7
* Remove old #if's related to older OIIO versions

This is master only, it will not be backported to OSL 1.8.


fpsunflower:
LGTM!

PR #709, opened 2017-02-09
More boost removal

lgritz:
New changes in this patch:

* boost::intrusive_ptr -> OIIO::intrusive_ptr (we could've done this a while back probably, but it's trivial now that we've bumped master's minimum OIIO version)

* boost thread, bind -> std + OIIO thread utils

After this patch is applied, our ONLY direct boost use in OSL is:
* boost::wave (for oslc's C preprocessor)
* boost's flat_map and flat_set containers (header only, fast)
* boost's thread-specific pointer (very handy, and equivalent is not yet in std::thread)
* boost::regex (only because std::regex is incompletely implemented prior to gcc 4.9, so we'll remove this next year if VFXPlatform moves to gcc 4.9 or beyond)

That's getting pretty minimal, and none of it is in the public APIs.

### DIFF
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -31,8 +31,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <cstdio>
 #include <fstream>
 #include <cstdlib>
-
-#include <boost/thread.hpp>
+#include <mutex>
 
 #include "oslexec_pvt.h"
 #include "OSL/genclosure.h"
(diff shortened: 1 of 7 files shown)


src/liboslexec/shadingsys.cpp
```diff
@@ -2692,28 +2692,23 @@ ShadingSystemImpl::optimize_group (ShaderGroup &group,
...
-
-
 void
 ShadingSystemImpl::optimize_all_groups (int nthreads, int mythread, int totalthreads)
 {
     // Spawn a bunch of threads to do this in parallel -- just call this
     // routine again (with threads=1) for each thread.
     if (nthreads < 1)  // threads <= 0 means use all hardware available
-        nthreads = std::min ((int)boost::thread::hardware_concurrency(),
+        nthreads = std::min ((int)OIIO::Sysutil::hardware_concurrency(),
```
fpsunflower: Is there a difference with `std::hardware_concurrency()` ?

fpsunflower:
LGTM

src/liboslexec/shadingsys.cpp
lgritz: No, just a wrapper dating from before C++11. I guess I can skip the middleman here and to straight for std::thread::hardware_concurrency().

lgritz:
Updated slightly, using std::thread::hardware_concurrency as suggested (no need for the OIIO wrapper), but also backed off a bit on some of the thread launching, because I'd inadvertently used a couple idioms that are only in OIIO master and not in the OIIO 1.7 release that I claim this should be compatible with. (Yay for TravisCI -- one entry in my build matrix explicitly tests against the release OIIO rather than master, telling me immediately if I break that.)

fpsunflower:
LGTM!

PR #711, opened 2017-02-10
Refactor finding of LLVM

lgritz:
Split into a proper FindLLVM.cmake, require at least 3.4 (for now).

### DIFF
--- a/src/cmake/modules/FindLLVM.cmake
+++ b/src/cmake/modules/FindLLVM.cmake
@@ -1,0 +1,111 @@
+# Module to find LLVM
+#
+# This module defines the following variables:
+#  LLVM_FOUND       - True if llvm found.
+#  LLVM_VERSION     - Full LLVM version
+#  LLVM_INCLUDES    - where to find llvm headers
+#  LLVM_LIBRARIES   - List of LLVM libraries to link against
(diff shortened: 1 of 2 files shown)


src/cmake/modules/FindLLVM.cmake
```diff
@@ -0,0 +1,111 @@
...
+#                      OUTPUT_STRIP_TRAILING_WHITESPACE)
+# endif ()
+
+# shared llvm library may not be available, this is not an error if we use LLVM_STATIC.
+if ((LLVM_LIBRARY OR LLVM_LIBRARIES OR LLVM_STATIC) AND LLVM_INCLUDES AND LLVM_DIRECTORY AND LLVM_LIB_DIR)
+  if (LLVM_STATIC)
+    # if static LLVM libraries were requested, use llvm-config to generate
+    # the list of what libraries we need, and substitute that in the right
+    # way for LLVM_LIBRARY.
+    execute_process (COMMAND ${LLVM_CONFIG} --libfiles
```
sambler: Surprised this has been like this for so long.

To get the list of static libs we need `llvm-config --libfiles --link-static`

lgritz: Thanks for the tip. Submitted as https://github.com/imageworks/OpenShadingLanguage/pull/1070

PR #715, opened 2017-02-23
Use clang's C preprocessor as alternative to Boost Wave

lgritz:
Boost Wave is a bit of a pain, in particular it can have problems if compiling OSL with C++11 (as is mandatory now in master) but link against a Boost that was compiled without the C++11 options.

At Chris Kulla's suggestion, we can switch to using the C preprocessing faciliies from clang's internals. Ordinarily, LLVM/clang would be much to painful to add just for this, but since we're already tied to LLVM as a dependency, it's really no additional burden.

So here we are. There's a build time cmake option USE_BOOST_WAVE, which defaults to off, but if problems are discovered with the clang cpp, that gives is a Plan B. Eventually, when we're completely confident in the way I've done it with clang, we may remove the Wave option entirely.

This is not 100% working. It seems to work fine if you are using clang as your compiler, or if you are using gcc 4.8.x, or if you are building against LLVM 3.4/3.5.  But the combination of using gcc >= 4.9 to compile and linking against LLVM >= 3.9 is resulting in crashes with complaints about double-freeing something. Surely it's our error, making a mistake about who owns what allocation or something, but I'm not sure what's wrong and I'm tired of holding up this patch for all the people and cases that it can help. So for now, for the combinations that don't run properly, it automatically falls back to using Boost Wave. We will remove this special case as soon as we figure out why it's failing in that case. (As stated earlier, you can also force use of boost wave with USE_BOOST_WAVE=1.)

Also made some changes to the Travis build matrix to cover various cases of combinations of compiler and LLVM version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,6 @@ addons:
       - cmake-data
       - gcc-4.8
       - g++-4.8
-      - gcc-6
-      - g++-6
       - libboost1.55-all-dev
       - libtiff4-dev
       #- llvm-3.9-dev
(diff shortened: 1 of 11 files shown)


lgritz:
Amended the PR to fall back on Boost Wave if the necessary clang components aren't found (let's consider them an optional dependency) and updated the README to explain this.

aconty:
LGTM!

lgritz:
OK, I'm going to merge this as it currently is, it seems to work when it works, and fall back to boost wave for the cases where it doesn't work. But I'm still investigating why it crashes with some compiler combinations, and as soon as I have it figured out, I will submit a PR with that patch (which I bet will be fairly minor).

PR #717, opened 2017-02-28
Simplify how we put raw source code into .oso comments

lgritz:
Use OIIO::Filesystem::read_text_file to read the whole file at once
instead of doing piecemeal file reads.

This eliminates an occasional buffer overrun as well.


fpsunflower:
LGTM

PR #718, opened 2017-02-28
Fix subtle use-after-free error in Dictionary::dict_find

lgritz:
It was incorrect to take a reference to an element of the vector
m_nodes, and then to keep using the reference inside the loop in
which the vector is grown (thus possibly invalidating the memory
that the reference points to).

Found by address sanitizer!


fpsunflower:
LGTM

PR #719, opened 2017-03-08
oslc with clang cpp: capture preprocessor errors properly

lgritz:
We were not setting up "diagnostics" properly, which led shaders
with certain errors in their use of the C preprocessing to hit
assertions in the clang code.


fpsunflower:
LGTM

lgritz:
Also will backport to release branch

PR #720, opened 2017-03-09
Fix inconsistency in normalize between Dual and float versions

fpsunflower:
This could lead to NaNs or Infs on very small arguments.


lgritz:
LGTM

PR #721, opened 2017-03-09
Fix clang assertion / linkage error

lgritz:
On some systems, sometimes, when linking against LLVM <= 3.5 and using
clang internals for C preprocessing, we've seen assertions coming from
llvm::cl::OptionCategory::registerCategory, which we aren't even
intentionally using.

This email explains the issue:

 http://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20140203/203968.html

Seems pointless to fight what appears to be an awkward mis-design in
older LLVM. The problem was fixed some time between 3.5 and 3.9.

So let's sweep it under the rug by falling back to boost wave when using
older LLVM.

### DIFF
--- a/site/spi/Makefile-bits
+++ b/site/spi/Makefile-bits
@@ -16,9 +16,6 @@ ifeq ($(SP_OS), rhel7)
     NINJA := ninja
     MY_CMAKE_FLAGS += -DCMAKE_MAKE_PROGRAM=${NINJA}
 
-    # Temporary bug fix, having problems with clang cpp, revert to Boost wave.
-    USE_BOOST_WAVE ?= 1
-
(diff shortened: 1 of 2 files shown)


fpsunflower:
LGTM!

lgritz:
I added an amendment that includes a bit of rearrangement to make sure the full wave-vs-clang logic is handled properly in the BOOST_CUSTOM case (as we do at SPI).

lgritz:
Oy, apologies. I just moved a couple clauses out of an "if", but because the indentation changes, the diffs on GitHub makes it look like a lot more change than it really was. I wish GH had an "ignore whitespace changes" diff mode.

fpsunflower:
LGTM

PR #722, opened 2017-03-11
More robust handling of exceptions to "useless op elision."

lgritz:
OSL's runtime optimizer tries to eliminate operations whose results are never used (people don't write shaders like this, but as we optimize and rewrite the shader's ops, many ops may reveal themselves as having no effect).

But, we don't do this with built-ins that have no written arguments at all (such as printf or setmessage), because it figures that having no immediate outputs means it's there purely for the side effects.

But -- OH NO! -- it turns out that trace() is an exception to the rule. It has a writeable output (the return value), but it ALSO has an important side effect (it sets up trade state that may later be queried by getmessage("trace")). So if you wrote a trace call and did not assign its return value to something that got used, the trace call might be eliminated entirely, which would totally bork your later getmessage queries about that traced probe ray.

So sorry.

This fix formalizes the notion that certain ops have important stateful side effects, and it will not apply the "useless op elision" optimization to those instructions.

### DIFF
--- a/src/liboslexec/oslexec_pvt.h
+++ b/src/liboslexec/oslexec_pvt.h
@@ -169,7 +169,7 @@ struct OpDescriptor {
         : name(n), llvmgen(ll), folder(fold), simple_assign(simple), flags(flags)
     {}
 
-    enum FlagValues { None=0, Tex=1 };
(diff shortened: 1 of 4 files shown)


fpsunflower:
LGTM!

PR #724, opened 2017-03-17
Add OSL_LLVM_PREFIX option for prefixing names of functions invoked by LLVM

mattyjams:
When OSL_NAMESPACE is used to wrap the C++ API in a namespace, it may also
be desirable to ensure that the C free functions invoked by LLVM are
disambiguated as well. This can be enabled by providing a string value for
OSL_LLVM_PREFIX. This string will be prefixed to all free functions that may
be invoked via LLVM so that the right version of OSL API finds the right
version of the free function. If no prefix is supplied, an empty one is used.

This change adds an OSL_LLVM_PREFIX build option for make and cmake, and it
adds macros for prefixing C strings and tokens with that prefix, as well as
building op function names that include the prefix. It also wraps all existing
functions that get invoked via LLVM with these macros to ensure that they are
prefixed properly if OSL_LLVM_PREFIX is set.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,6 +115,11 @@ if (OSL_NAMESPACE)
 endif ()
 message(STATUS "Setting Namespace to: ${OSL_NAMESPACE}")
 
+if (OSL_LLVM_PREFIX)
+    add_definitions ("-DOSL_LLVM_PREFIX=${OSL_LLVM_PREFIX}")
+endif ()
+message(STATUS "Setting OSL LLVM prefix to: ${OSL_LLVM_PREFIX}")
+
 include (util_macros)
 include (externalpackages)
 include (flexbison)
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,10 @@ ifneq (${NAMESPACE},)
 MY_CMAKE_FLAGS += -DOSL_NAMESPACE:STRING=${NAMESPACE}
 endif
 
+ifneq (${OSL_LLVM_PREFIX},)
+MY_CMAKE_FLAGS += -DOSL_LLVM_PREFIX:STRING=${OSL_LLVM_PREFIX}
+endif
+
 ifneq (${HIDE_SYMBOLS},)
 MY_CMAKE_FLAGS += -DHIDE_SYMBOLS:BOOL=${HIDE_SYMBOLS}
 endif
@@ -359,6 +363,7 @@ help:
 	@echo "      USE_BOOST_WAVE=1         Force boost wave rather than clang preprocessor"
 	@echo "  OSL build-time options:"
 	@echo "      NAMESPACE=name           Wrap OSL APIs in another namespace"
+	@echo "      OSL_LLVM_PREFIX=prefix   Apply prefix to names of OSL functions invoked by LLVM"
 	@echo "      USE_FAST_MATH=1          Use faster, but less accurate math (set to 0 for libm defaults)"
 	@echo "      OSL_BUILD_TESTS=0        Don't build unit tests, testshade, testrender"
 	@echo "      USE_SIMD=arch            Build with SIMD support (choices: 0, sse2, sse3,"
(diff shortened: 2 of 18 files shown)


dankripac:
A big +1 from me on this pull request!

I've wanted to find the time for this due to issues with us running our own OSL plug-in (often at newer versions) alongside the shipped built-in Pixar OSL shader.
We've had crashes emanate from the c functions inside the pixar OSL implementation with our shader being upstream in the stack trace.

fpsunflower:
This makes sense and seems like a pragmatic way to solve this (even if its a bit verbose).

Why did you opt for adding `OSL_LLVM_PREFIX` vs. just using `OSL_NAMESPACE`?

mattyjams:
Thanks @dankripac! That's exactly the situation I was aiming to address with this change.

@fpsunflower: Yeah, the addition of the macros does make the function definitions a bit cumbersome. I tried to name them something sensible to at least make them a little easier to read. And I ended up adding `OSL_LLVM_PREFIX` rather than re-using `OSL_NAMESPACE` mainly to make the new functionality opt-in, and since I thought people might prefer the flexibility of having the two forms of "namespacing" separable. Since the C++ namespacing has been around for a while, some users may want to continue to use just that and not be forced into changing the symbols for the LLVM-invoked functions. I'm definitely open to other ideas though if you think it's better to just combine the two.

Thanks for taking a look!

dankripac:
Hey there, just wondering what the status of this PR is?
It would be great if this mechanism (which I have not had time to test sorry) made it into the main stream of OSL releases.

lgritz:
Sorry, I think this came in when I was on vacation a few weeks back and accidentally slipped through the cracks. I will take a closer look...

lgritz:
I totally understand and approve of the general idea here.

I'm wondering about several points:

1. The big batch of macros that you put in oslversion.h.in -- is this necessary to be in a public header? Is there ever a case where it's necessary for client (i.e. renderer) to invoke this macro? Or maybe we can put it in osl_pvt.h which is geared for internal declarations?

2. So verbose! Is it too much to ask for a shorter name? Like, instead of OSL_LLVM_PREFIXED_TOKEN, maybe just OSL_PREFIXED()?

3. In the absence of a definition of OSL_LLVM_PREFIX, should it default to OSL_NAMESPACE to match the rest? (And double-default to "OSL" if that is also not defined?)

4. Should we eliminate the osl_ that we already have and just let that be part of the prefixing?

Oh, ok, wait, let's step back and maybe think a little simpler.

Currently, if OSL_NAMESPACE is set, it makes a double namespace `${OSL_NAMESPACE}::OSL` and if OSL_NAMESPACE is not defined, it makes a single namespace OSL.

Now that I'm looking at this, I kind of don't like the fact (and had forgotten) that it's dependent on the builder inserting versions. I'm wondering if we should revise the order a bit to always ensure a version. I'm thinking that we want the nested namespaces to look like ${OSL_NAMESPACE}_v${VERSION}::OSL.
We can come back to that.

So perhaps for those C linkage symbols, we can prefix them like OSL_PREFIXED(mul_closure_color) and that will turn it into OSL_vXY_mul_closure_color. The OSL_LLVM_PREFIX can be separately set (really, is that even necessary), but if not, then it defaults to LLVM_NAMESPACE and the version appended.

Would you mind if I take this PR and massage it just a bit to show you what I mean?

mattyjams:
Hey Larry!

Yeah, that's sounding pretty good to me. Feel free to tweak the PR however you like!

Tucking these macros away where they won't be publicly exposed makes sense. They mostly ended up here out of convenience and proximity to the namespace macros which are loosely related. I don't expect any client code would/should ever need to touch them directly.

And yeah, sorry for the verbosity. I opted for that in hopes of making it clearer what each one did, but it did end up being a bit much.

I was aiming to be minimally invasive and make it mostly a bolt-on thing with this first pass just to get the idea across, but I'll definitely defer to you as to what's best in terms of integrating it with other features.

Thanks again for checking it out!

lgritz:
A revised patch is proposed as #731 

See what you think about that. Your basic idea is intact, but I did some renaming and moving around, and also took the opportunity to do some revision to the way the C++ namespaces work as well (including automatically incorporating the major and minor version into the namespace) and making the "C" linkage prefix match the C++ namespace.

The comments on that PR should explain it all.

mattyjams:
Sounds good! I'm taking a look at #731 now. We can close out this PR in favor of that one.

lgritz:
#732 is the one we chose to adopt, and was merged into master, RB-1.8, and RB-1.7.

PR #725, opened 2017-03-18
Shader specialization API change

ElaraFX:
An API change to allow specialized (surface/shadow/displacement) shader groups to defer its `optimize_group` to its first `execute`.

In our case, we always construct tons of specialized shader groups. But many of them are not even needed because we have user specified flags to avoid some shader calls. So deferring `optimize_group` helps a lot to save shader optimization cost, and to make `optimize_group` calls multi-threaded.

In practice, we found this change give us 10-100X faster scene pre-processing.

Sorry, I don't really have time to re-base this change, please just refer to the commit below to see if it's useful:

https://github.com/imageworks/OpenShadingLanguage/commit/6c8b93ee6e005201fa05c7402e827467c18fa5f8


lgritz:
It's too hard to understand what has changed.

You can submit it as a PR versus our RB-1.7, that will be a much lesser set of irrelevant changes versus master. I can easily review and test it against RB-1.7 and then just cherry-pick from there to master.

lgritz:
I'm a little unclear on why this is necessary. You don't actually need to call optimize_group ever if you don't want. You can just call execute(), and it will do the optimize/JIT if it hasn't bee done already.

Oh, I think I see. The problem is that optimize_group lets you pass the ray flags to specialize to a particular kind of ray/shader, but execute does not. So you've added a set_raytypes to allow you to pre-declare that, so that they are known up whenever the optimize happens. Yeah, I totally see why that would be handy. Good idea.

But you have *replaced* optimize_group() with set_raytypes(). But there are times when we might want to call optimize_group explicitly and know it is finished. An example would be if a renderer really does want to immediately compile shaders up front (perhaps while you have idle threads to do the work) so that you never have one thread blocking while another thread is JITing, which can happen if you are fully lazy -- you could easily have two rays (threads) that simultaneously hit objects with the same material and thus both need to do the lazy JIT at the same time.

Another reason we shouldn't do it that way is that changing the name of an existing API call would break link compatibility, which we have pledged not to do within release branches (1.7 or 1.8, though it's ok to break APIs in master if necessary). It's fine to add a new non-virtual method to the class though, that won't break working code or link compatibility.

I think this is a good idea and you're very close. What I recommend is that you edit slightly to leave the original optimize_group as it was, and make set_raytypes a new, additional method. Then it can work either way -- forced immediate, or fully lazy. Then take that one commit (I think your changes are just in that one commit, right?) and cherry pick it into a new branch off master, and submit that as your PR. Or else, I think you could also just tell GH when you submit a PR that it's versus RB-1.7 rather than versus the default of master (and I can easily forward-port it to 1.8 and master by cherry-picking). Either of those choices would yield a PR that will be small and easy to review.

lgritz:
You don't need to submit a new PR. If you are the submitter, click the "Edit" button at the top (on the right of the main title of this PR in bold). When you do that, right below the edit box where you could re-edit the title of the PR, you will see it say "<you> wants to merge 1 commit into 
base: [master] from ..." and the [master] is a pull-down menu you can just switch to RB-1.7, if that's the branch you were working from when you added this feature.

fpsunflower:
I noticed the same asymmetry in the API as well. You can pre-optimize with ray-type specialization but can't do lazy optimization with ray-type specialization.

I was actually contemplating making the same change.

lgritz:
@ElaraFX did you have any thoughts on my comments? It's a fairly minor operation to turn set_raytypes into an addition rather than a replacement to optimize_group. But if you simply don't have the time to revisit this, I could do it for you very easily. Up to you.

lgritz:
Any further thoughts on this, @ElaraFX ?
I like your idea, but I want one or the other of us to turn it into a strict API addition, rather than changing the existing call sequence.

fpsunflower:
I am going to submit a patch that does what we discussed above that should cover @ElaraFX 's needs as well.

lgritz:
Closing, this was fixed by #733

PR #727, opened 2017-03-23
Fix linkage problem with some tests unnecessarily linking liboslcomp

lgritz:
liboslexec contains all the functionality of oslcomp. You never need
both. It used to be harmless to link against both, though, until we
included clang C preprocessing functionality. Clang has some components
that cause trouble if linked twice, and so we run into that both
liboslexec and liboslcomp contain statically linked clang components.


fpsunflower:
LGTM

That's the same set of changes I did on my end.

lgritz:
OK, will merge as soon as the CI tests come back.

PR #728, opened 2017-03-31
clang-tidy support and include file fixes

lgritz:
Similar to recent changes to OIIO:

* Build scripts modified to support easy running of clang-tidy.

* Minor fixes of deprecated C header includes, updated to proper C++ headers.

* Lots of fixes making a uniform approach to including headers within our own project. In particular, all public headers are included as <OSL/foo.h> and not as "OSL/foo.h" or "foo.h".


fpsunflower:
LGTM

PR #729, opened 2017-04-03
Use C++11 vector::emplace_back() where appropriate.

lgritz:
Now that we have C++11 as a minimum, we can use this handy feature that
grows a vector by one slot (uninitialized) and the constructs a new
object in that location. In contrast, push_back() constructs the object
first, then grows the vector with a default constructed object, they
copies (or if you're lucky moves) the new object into the vector
position. Maybe the compiler will figure it all out, maybe not.  Using
emplace_back guarantees that there are no needless copies, moves, or
pointless constructions.

Thanks, clang-tidy, for making this a 5-minute change!


fpsunflower:
LGTM!

PR #731, opened 2017-04-09
Namespace overhaul

lgritz:
This is a continuation/replacement for PR #724 

First, instead of a nested namespace (like `OSL_custom::OSL`), only have one, which now incorporates the     major and minor version (not automatically folding the version in was an oversight I wish to correct now). So, for example `namespace OSL_v1_9`.   The base name (the `OSL` part of it) is customizeable  by the `OSL_NAMESPACE` cmake variable, which defaults    to OSL. Individual sites or products may override, and    are advised to if they do anything nonstandard to their OSL build. Just like before, there is a `namespace OSL = the wrapped namespace`, so client applications can just say `OSL::foo`, and it's really referring to the full versioned and/or customized name that the linker sees.
    
Then I layered @mattyjams Matt Johnson's patch (and revised it somewhat), which provides proper namespacing for the "C" entry points that are exposed for LLVM code to call, something which we had overlooked before but which is better to be properly namespaced.
    
Matt proposed a good fix for this as PR #724, which LG revised    and amended here. Matt's design is basically intact -- a series of    clever macros prepends the namespace (including the version, and    overrideable at build time) to all the symbols in llvm_ops.cpp and that    are in C++ but declared extern "C" so that they can be linked to by the    JITed code. In many cases, they already were osl_foo, so now we turn    that "osl_" into the proper prefix that matches the namespace fully.
    
For this final PR, Larry amended Matt's patch in the following major    ways:
    
* Just use the OSL_NAMESPACE for the prefixing of the LLVM-callable    functions as well. LG thinks that adding a second set of naming is    unnecessarily complex if nobody has a good reason for needing it (we can    always add that flexibility later).

* Simplified, documented, and renamed the set of macros for the sake of    clarity, brevity, and style, as well as to mesh well with the previous    commit which changed the way the C++ namespaces work.

* Use the namespace to replace the previous "osl_*" naming convention    rather than prepend to it (e.g., Matt's OSLNAMEPACE_osl_foo becomes just    OSLNAMESPACE_foo).

* Matt went for minimal changes by doing the append inside llvm_util's    call_function, but Larry wanted to keep that totally generic, so call_function was    largely restored to its previous state and all the many places that call    it now utilize the macro -- this is a more extensive change, but I think    ultimately preserves the independence and generality of the LLVMUtil    class. But the spirit and overall design of Matt's proposal remain as    the bones of this patch.


dankripac:
Hi Larry, would it be possible to get a back-port of this (assuming it goes through) to 1.8 branch - and hopefully 1.7 branch as this is what we are currently on in production?

fpsunflower:
LGTM

Just want to mention the alternative design which would be to hide all functions with "C" linkage and register them "manually" with LLVM. This would avoid name clashes but also avoid the need to use `dlsym` to find symbols. I think we tried this a while back - but I don't remember what the conclusion was.

lgritz:
Oh that's interesting, I'd kind of forgotten that. Let me give that a try, it would be a lot less intrusive.

lgritz:
Dan: yes, I'll try to backport.
I'm checking out Chris Kulla's suggestion... if that pans out, it will be much easier to backport, and I think it should be just as safe for linkage.

lgritz:
Chris is looking good here. A **substantially** smaller patch will serve to completely hide all those `extern "C"` symbols.

At least, on Linux and OSX.  I'm really not sure if it will work properly on Windows, I don't have a good way to test it. I'm not 100% sure if it's necessary on Windows anyway. 

Let me do a little testing, and I will post YET ANOTHER proposed patch, and you guys can weigh in on what you like best.

lgritz:
Yes, please see #732, I think that is a much simpler approach.

PR #732, opened 2017-04-10
Namespace overhaul (take 2)

lgritz:
History: Matt Johnson @mattyjams proposed #724 as a solution to the LLVM-discoverable symbols not being properly namespaced. LG @lgritz revised and amended, countering with #731. Then Chris Kulla @fpsunflower pointed out a much simpler solution, presented in this patch.

(Sorry, Matt, this version pretty much loses all of your work, but boy is it a lot simpler. I think that moving forward, all those special macros would have been an ongoing burden and common source of bugs, this fixes it all easily and is much friendlier to future additions and changes to the code, I think.)

The new, revised description:

1.  Instead of a nested namespace (like OSL_custom::OSL), only have one,    which now incorporates the major and minor version (not automatically    folding the version in was an oversight I wish to correct now). So, for    example namespace OSL_v1_9. The base name (the OSL part of it) is    customizeable by the OSL_NAMESPACE cmake variable, which defaults to    OSL. Individual sites or products may override, and are advised to if    they do anything nonstandard to their OSL build. Just like before, there    is a namespace OSL = the wrapped namespace, so client applications can    just say OSL::foo, and it's really referring to the full versioned    and/or customized name that the linker sees.

2. Hide symbol visibility for the extern "C" osl_* functions LLVM needs.

   We never previously mangled or hid the many functions OSL needs to call    by address, which it figures out with dlsym calls, and therefore those    functions needed "C" linkage (no C++ mangling) and external visibility.    But that was the weak point for completely hiding the library symbols    to avoid clashes between versions or multiple OSL copies.

   In this patch, we change the visibility of all those symbols to local    (which we also ensure is hidden by default) and since that would make    the dlsym fail, we explicitly make LLVM aware of the mapping from    llvm::Function* to callable address.


fpsunflower:
This seems even better! LGTM

Do we need to do anything to remove `dlsym` from the picture entirely? Or is that still needed?

lgritz:
I think that by explicitly giving it the address of every callable function, there is no longer any situation where it will rely on the dlsym. The dlsym calls will fail anyway, now that the symbols are hidden from the linker.

fpsunflower:
Can you see if adding `EE->DisableSymbolSearching()` still works? (I think that tells it never to call `dlsym`)

lgritz:
Oh, didn't know about that! Sure, I'll try.

brechtvl:
I've tried using `DisableSymbolSearching()` before, but we found that it causes issues with large arrays (and probably more cases). In particular with an array variable of length > 32, LLVM would generate a call to `memset` and doing a lookup for that would fail.

lgritz:
Thanks, Brecht. I'm inclined to leave that alone. I don't see what the harm is, considering that all our symbols are now hidden and we give it the addresses. Anything it still needs to search for with dlsym, is obviously something we haven't given it, so I wouldn't want to mess that up in unexpected ways.

brechtvl:
Another subtle issue here that we've encountered is that LLVM tries `dlsym` before the manually specified function mapping (at least in LLVM 3.5).

This means an older OSL library which does not hide its symbols can still interfere with a new library that does hide them, unless the OSL function names are different. I did not find an LLVM method to reverse this lookup order.

lgritz:
@brechtvl -- yeah, it's not perfect. The fact that we had older versions not fully namespaced or hidden is unfortunate. But this patch is simple enough that I'm a lot more confident about backporting it, so that at least NEW releases of old branches will be safe. And new branches will be entirely safe moving forward. It will certainly not be any more broken than it was before; definitely a step forward.

brechtvl:
Agreed. I'm only mentioning it in case people still encounter issues after this change.

mattyjams:
Hey Larry!

No problem with ditching the stuff I did. I'm all for a simpler solution, and this is indeed *way* simpler.

I've been testing this out this afternoon and it's been working great. My symbol name clashing issues are still resolved in this most recent revision.

Sorry that I probably caused you more work than I contributed, but thanks very much for taking this on!

lgritz:
That's great news, @mattyjams. I want to do a little more testing tomorrow. This passed my OSL unit tests and the CI builds, but I didn't get a chance to build it into our full renderer and do a whole testsuite there. I'll do that in the morning, and if it passes the renderer testsuite for me, I'll merge it, backport to the last couple release branches for @dankripac, and close the other tickets.

Yay, I really like this change. It's been nagging at me for a while that the symbol hiding didn't extend to those `extern "C"` functions. This is a nice cleanup of that issue.

lgritz:
Merged into master.

Also backported the "C" symbol hiding to RB-1.8 and RB-1.7. Note that the top-level HIDE_SYMBOLS option defaults to OFF in those branches, ON in master. It felt risky to change release branch defaults for important build modes with link implications.

dankripac:
Thanks so much for this Larry, will try it out today!
Yay

fpsunflower:
I'm happy to report this small change appears to have made runtime optimization roughly 10% faster! (Reduction is mainly in LLVM optimize time)

lgritz:
Fascinating! I had my mind focused so much on simply blocking the visibility of the symbols, it never occurred to me to benchmark. But in retrospect, of course it makes sense that it might save time to pre-declare the functions without it needing to do any searching function-by-function (let alone with dlsym, who knows what that cost is).

dankripac:
Hi @lgritz , just checking. If I look at RB 1.7 (or 1.8) I don't see these changes?
You mentioned backporting above but maybe I'm missing something?

lgritz:
?? The fix that hides the osl symbols have been pushed to the head of both RB-1.7 and RB-1.8 in the main imageworks repo. I haven't created new release tags, I usually do official releases at the start of each month, this didn't seem critical enough to cut a separate release.

I did not backport the commit that changed the C++ namespace. That infected some cmake files that had diverged in master and made a backport tricky. I prefer the new namespacing, but the old one worked (in contrast to the llvm symbol hiding, which just didn't work before), so I didn't see the need to work very hard to backport that portion.

CMakeLists.txt
```diff
@@ -111,10 +112,9 @@ endif ()
 
 
 # Set the default namespace
-if (OSL_NAMESPACE)
-    add_definitions ("-DOSL_NAMESPACE=${OSL_NAMESPACE}")
-endif ()
-message(STATUS "Setting Namespace to: ${OSL_NAMESPACE}")
+set (PROJ_NAMESPACE "${${PROJ_NAME}_NAMESPACE}")
```
dankripac: Hey @lgritz, I was referring to some of these changes in the main CMakeLists.txt file. i.e creating PROJ_NAMESPACE. This doesn't seem to end up in the Release branch if I set github to say point at RB 1.7 and the download a ZIP of this. I've now also tried cloning, checking out RB-1.7 and it's the same? I'm probably doing something dumb, but I can't see any of these new lines from this pull request?

lgritz: Oh, oops, let me take a look. I may need to add that definition, as it's part of things that recently changed in master.

lgritz: Right, there were two commits that were part of this PR.

I did backport the change that hides the osl_* symbols that were `extern "C"` for LLVM's benefit, which were never properly hidden or namespaced before. (Though I'm nervous about changing the default to be hidden in the release branches; for 1.7/1.8, you'll want to explicitly build with HIDE_SYMBOLS=1.)

I am intentionally not backporting the changes to C++ namespaces for the things that were already namespace-protected. That sounded like a riskier thing to change for a release branch, and as well, the changes to support it involve some more widely-diverged sections of the cmake scripts that would be harder to backport correctly without errors.

dankripac: Ok no problem, it did dawn on me that the namespace config change would be a bit of a breaking back port change.
I'll try a build with HIDE_SYMBOLS turned on and see if this works for us.
Cheers

lgritz: I'm less worried about breaking things with the namespace change (though maybe I should be), but rather that there was a pretty significant refactoring of the cmake scripts in master after it diverged from 1.8, so there is no easy clean merge to backport just the namespace change in isolation. I was really hoping not to need to mess with the older cmake files again, except to fix something that was significantly broken.

PR #733, opened 2017-04-11
Add an API call to define raytypes_on and raytypes_off

fpsunflower:
This makes it possible to use this feature of the optimizer even when relying on lazy JIT of shader groups.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -557,15 +557,19 @@ public:
     /// data passed in via attribute("raytypes")).
     int raytype_bit (ustring name);
 
+    /// Configure the default raytypes to assume to be on (or off) at optimization
+    /// time for the given group. The raytypes_on gives a bitfield describing which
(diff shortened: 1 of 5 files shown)


fpsunflower:
This is related to #725, but should be backwards compatible as it is purely an API addition.

lgritz:
LGTM!

lgritz:
Oops, I missed something while reviewing. I should have warned you not to take away the ray parameters from optimize_group, because that breaks the API/linkage. I'll leave it this way in master, but the backport to 1.8 will include an amendment to put those parameters back, since we aren't allowed to break existing API calls in release branches.

fpsunflower:
I thought the contents of `oslexec_pvt.h` were all private?

lgritz:
Sorry, forget I said anything. Yes, I was looking at the pvt file. The public file oslexec.h only adds doesn't change existing calls. Never mind.

lgritz:
I'm just starting my coffee now, ok?

PR #735, opened 2017-04-11
Eliminate varargs from OSLCompilerImpl and ASTNode

lgritz:
Use type-safe variadic templates instead.

The implementation details get cleaner if you have a sufficiently new OIIO, so there are a couple `#if`s
for now, but eventually the old path can be removed.

Not a pressing issue, just some cleanup I finally got around to.


fpsunflower:
LGTM

PR #736, opened 2017-04-12
Move matrix transform shadeops out of llvm_ops and into opmatrix

fpsunflower:
This has the side benefit of avoiding the Imath exception handling in matrix inverse from making it into the bit code LLVM needs to process, reducing the amount of runtime symbol resolution required.

This also avoids some code duplication we had between these two files.


fpsunflower:
Oops, clearly something broke - this isn't passing the tests.

lgritz:
Don't worry if the CI test fails on OSX, that's a Travis issue I need to work out. But all the Linux tests should pass.

fpsunflower:
Bad copy-paste. This one should be good.

lgritz:
Still seems to be failing testsuite/vector in the CI tests

fpsunflower:
One more time ...

(I had forgotten to move out the declaration out of the bitcode portion, which only the `vector` test covered)

fpsunflower:
All tests are passing now (minus the os x ones).

lgritz:
I'm testing on my end now, will merge if it passes.

lgritz:
LGTM! Merging.

Issue #730, opened 2017-04-06
State of OSL


carsonbrownlee:
hello,
I've been looking over OSL and it looks very promising, however I've run into a few issues.  I built OSL on Centos 7 off of branch RB-1.8 and about half of the tests fail, and only about half of the "testsuite/render-" tests produce a non black image using testrender.  To compile, I needed to remove a line for adding the  MCJIT llvm library in case that affected anything.  I'm using llvm 3.4.2.
Is this indicative of the state of OSL (or at least the testrender/testshader apps) or specific to my build/dependencies?  

Would blender be a better starting point for assessing the stability and functionality of OSL?
Is there a place that maintains a set of supported shaders comparable to something like the vMaterial Library, or would the best place for prebuilt shaders be an external package like blender if they can export to OSL?

lgritz:
In what way do the tests fail, exactly? Crash? Incorrect result?

Which part of RB-1.8? The current head? A release tag?

Are you using llvm 3.4 because you're stuck with C++03? Or do you have the option to use a modern LLVM (beyond 3.4 does require C++11)?

carsonbrownlee:
adding partial output below.  I'm assuming incorrect output, but haven't delved deeper.

I checked out RB-1.8 branch, latest commit was: 

I'm using llvm 3.4 simply because that's what Centos' package manager has and it was listed as supported.  I can update it manually if that may fix these issues.  

`
$ make test |& tee out.txt
make[1]: Entering directory `/home/carson/git/OpenShadingLanguage/build/linux64'
make[2]: Entering directory `/home/carson/git/OpenShadingLanguage/build/linux64'
make[3]: Entering directory `/home/carson/git/OpenShadingLanguage/build/linux64'
make[3]: Entering directory `/home/carson/git/OpenShadingLanguage/build/linux64'
make[3]: Leaving directory `/home/carson/git/OpenShadingLanguage/build/linux64'
make[3]: Leaving directory `/home/carson/git/OpenShadingLanguage/build/linux64'
make[3]: Entering directory `/home/carson/git/OpenShadingLanguage/build/linux64'
[  0%] Built target CopyFiles
[  2%] Built target testshade_dso
make[3]: Entering directory `/home/carson/git/OpenShadingLanguage/build/linux64'
make[3]: Leaving directory `/home/carson/git/OpenShadingLanguage/build/linux64'
make[3]: Entering directory `/home/carson/git/OpenShadingLanguage/build/linux64'
[...]

PR #737, opened 2017-04-13
More robust finding of stdosl.h

lgritz:
There was a discrepancy in how stdosl.h is searched for, among the different ways that compiles are triggered (osl from command line, versus OSLCompiler from file, versus OSLCompiler from buffer). This patch unifies them, and expands with additional heuristics for finding the file.

Also falling out of this is that the build and test scripts are now much more robust to the user not setting $OSLHOME -- a top-level 'make test' will find it all in the build itself. Also, it fixes some cases where osl.imageio.so can't find stdosl.h, and thus would appear inexplicably broken.


lgritz:
Fixes #730 (inasmuch as it was related to a user having tests fail because they didn't know to set OSLHOME).

aconty:
LGTM

fpsunflower:
LGTM2

PR #738, opened 2017-04-13
Revert HIDE_SYMBOLS default back to OFF

lgritz:
I was getting nervous about having changed the default here, and in
preparing a similar change on OIIO was getting odd behavior on OSX.
So it seems wise to revert just this default value. Individual sites
may choose build with HIDE_SYMBOLS=1 to be extra careful, if it all
works for them. This does not change the fact that all of OSL is
in its own namespace.


fpsunflower:
I assume this doesn't impact the hiding of `OSL_SHADEOP` symbols?

lgritz:
Um... maybe never mind. I think I'm just confused.

PR #739, opened 2017-04-13
Streamline range checking

lgritz:
We noticed some cases where the range checking overhead was measurable.     The overhead can be cut roughly in half by making osl_range_check inline     in llvm_ops.cpp so that there is no function call involved in the check     itself, only in the uncommon case of when an error is detected does it     call osl_range_check_err which lives elsewhere.


src/liboslexec/llvm_ops.cpp
```diff
@@ -772,11 +772,37 @@ OSL_SHADEOP void osl_filterwidth_vdv(void *out, void *x_)
...
+
+
+
+OSL_SHADEOP int
+osl_range_check (int indexvalue, int length, const char *symname,
+                 void *sg, const void *sourcefile, int sourceline,
+                 const char *groupname, int layer, const char *layername,
+                 const char *shadername)
+{
+    if (indexvalue < 0 || indexvalue >= length) {
```
fpsunflower: Does it go any faster if you use: `unsigned(indexValue) < unsigned(length)` ?

lgritz: Though Compiler Explorer shows that it is generating more compact machine code, I can't seem to measure any improvements at all in practice. What do you recommend that I do? Clarity, or theoretical perf?

fpsunflower: Best leave it alone then. LGTM!

Out of curiosity, what kind of speedups does this patch give us?

lgritz: Laurent was benchmarking a particular shader in isolation (with testshade, not in a render) that went from 7.8 to 6.4 when range checking was disabled entirely. This patch lowers the cost to 7.1 (i.e. eliminating roughly half of the added expense of the range check). 

I tried two production scenes. One appears to speed up by about 1% overall, the other did not have a measurable difference.

PR #740, opened 2017-04-14
Fix the handling of alignment for closure structs

fpsunflower:
`ubsan` complained that our closure memory accesses were not properly aligned. The previous code wasn't tracking the actual `alignof` that the target struct wanted.

Due to how we pack the user's closure `struct` into the memory of `ClosureComponent`, we can't support arbitrary sizes. But I have at least added some error detection code to check that the user isn't trying to use a struct with wider than usual alignment requirements.

I've nuked the old `alignment` flag to `register_closure` because it wasn't working as intended and was somewhat hard to use (none of the example `test*` programs set it, nor did our renderer).

### DIFF
--- a/src/include/OSL/oslclosure.h
+++ b/src/include/OSL/oslclosure.h
@@ -120,11 +120,11 @@ struct OSLEXECPUBLIC ClosureColor {
 /// whatever type of custom primitive component it actually is.
 struct OSLEXECPUBLIC ClosureComponent : public ClosureColor
 {
-    Vec3   w;        ///< Weight of this component
-    char   mem[4];   ///< Memory for the primitive
-                     ///  4 is the minimum, allocation
-                     ///  will be scaled to requirements
-                     ///  of the primitive
+    Vec3 w;                     ///< Weight of this component
+    alignas(void*) char mem[8]; ///< Memory for the primitive, 8 is the minimum, allocation will be
+                                ///  scaled to requirements of the registered closure struct.
+                                ///  Alignment is forced to a void* to match likely alignement requirements
+                                ///  of the user's structs.
 
     /// Handy method for getting the parameter memory as a void*.
     ///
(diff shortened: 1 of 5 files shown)


src/include/OSL/oslclosure.h
```diff
@@ -120,11 +120,13 @@ struct OSLEXECPUBLIC ClosureColor {
...
-    Vec3   w;        ///< Weight of this component
-    char   mem[4];   ///< Memory for the primitive
-                     ///  4 is the minimum, allocation
-                     ///  will be scaled to requirements
-                     ///  of the primitive
+    Vec3 w;                 ///< Weight of this component
+    union {
+        char mem[8];        ///< Memory for the primitive, 8 is the minimum, allocation will be
+                            ///  scaled to requirements of the registered closure struct
+        void* force_align;  ///< This is just to force alignment of the struct to be at least that of a pointer
```
lgritz: Can you use alignas() or the OIIO_ALIGN macro to force alignment on the union?

lgritz:
Failing CI tests -- you are triggering a signed vs unsigned warning on gcc.

src/include/OSL/oslclosure.h
fpsunflower: Yeah that might be cleaner.

fpsunflower:
Addressed comments.

I wonder why clang wasn't picking up the signed vs. unsigned warning.

fpsunflower:
All CI tests passing now.

lgritz:
LGTM. I assume you have also run the renderer-side testsuite with these changes in OSL?

lgritz:
Semi-related PS. I did some work over the weekend on fixing the OSX builds on Travis so that they will pass also.

fpsunflower:
Yes. And the undefined behavior sanitizer is happier about closure access at runtime now.

PR #741, opened 2017-04-18
Allow shaders to be accidentally specified with trailing .oso

lgritz:
One way you could bump into this is if you

    testshade ... test.oso

it would complain about the shader not being found.


fpsunflower:
LGTM

PR #742, opened 2017-04-19
Fix OSX build problems on TravisCI, support make INSTALL_PREFIX=

lgritz:
Our TravisCI builds on OSX have been failing for a while and I finally
got a chance to track down the issues.

1. Slight adjustment to ensure that on OSX the rpath is set correctly.
(At least I think it's more "correct" now. Seems to fix some things.)

2. Make sure that on Travis we install the homebrew freetype package,
which is needed for the texture-udim test.

3. Make sure OIIO_LIBRARY_PATH is set to help the osl-imageio test, which
needs to find the osl plugin for OIIO.

4. Also simultaneously depends on a change to OIIO (analogous to the first
item here) to adjust its RPATH behavior slightly on OSX.

None of these things should affect non-OSX builds.

Also, as was recently done to OIIO, add to the wrapper Makefile the
ability to set INSTALL_PREFIX, which is passed along to cmake as
CMAKE_INSTALL_PREFIX. It defaults to the usual of ./dist/PLATFORM, but
this lets you override to, for example, instruct the build to put the
full installed files in another location. You could always do this if
invoking cmake directly, but it was tricky for those who preferred to
use the make wrapper.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,6 +84,8 @@ script:
     - make VERBOSE=1 $BUILD_FLAGS cmakesetup
     - make -j2 $BUILD_FLAGS
     - export LSAN_OPTIONS=suppressions=$PWD/src/build-scripts/nosanitize.txt
+    - export OSLHOME=$PWD/dist/$PLATFORM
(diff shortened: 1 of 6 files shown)


aconty:
LGTM

PR #743, opened 2017-04-20
Fix memory leak of ASTNode's in osl compiler.

lgritz:
Chris discovered that oslc was leaking ASTNode allocations, which was
interfering with some of his sanitizer runs.

Function declarations had the raw pointers squirreled away in the symbol
table without ever keeping a ref-counted version, so they were never
freed. So now we keep a vector of the ref-counted pointers for
ASTfunction_declaration in the OSLCompilerImpl, which all get freed when
the compiler destructs.

Quite independently, we were leaking the nodes associated with certain
metadata, they were just dropped on the floor by the add_meta() method,
which walked through what it was handed and used it to adjust things,
but never freed the pointers. This is fixed by changing add_meta to
force the raw pointers to be converted to ref-counted.

I added some code that for DEBUG builds only will track the construction
and destruction of ASTNodes, and upon program termination will print
messages if more nodes were created than destroyed.

### DIFF
--- a/src/liboslcomp/ast.cpp
+++ b/src/liboslcomp/ast.cpp
@@ -29,6 +29,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <vector>
 #include <string>
 #include <sstream>
+#include <functional>
 
 #include "osl_pvt.h"
 #include "oslcomp_pvt.h"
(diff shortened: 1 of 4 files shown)


fpsunflower:
LGTM

But would it not be simpler to manage these via a memory pool instead of reference counts?

It would likely be faster and less error prone (a single missing `ref` here can break the whole scheme). During compilation lots of small nodes need to be allocated one at a time, but they all get freed in one big batch at the end.

lgritz:
But there are a ton of different subclasses. I think I'd need a different pool for each.

fpsunflower:
If you can get all of them to have only POD members, you could use a single pool since you wouldn't care about running the destructors (just like we do for closures).

Not sure how easy of a change that is, I see that `ASTNode` has at least one `std::vector` in it ...

fpsunflower:
I can't tell if `children` in `ASTNode` is ever used with more than a handful of elements. It looks like many classes could just store their own `ASTNode`s directly and you could remove `children` from the base class maybe?

`AST_variable_declaration` has its own vector too, which might be hard to replace.

lgritz:
Difficult change.

fpsunflower:
Understood. Its certainly not something we need to worry about right now!

PR #745, opened 2017-04-23
testshade: fix to accept array of strings as parameter value

lgritz:
I discovered that while you can pass a parameter that's an int array, float array, etc., string arrays didn't quite work, they were treated as single strings.


aconty:
LGTM

PR #746, opened 2017-04-28
oslc: Add error detection of duplicate functions in the same scope.

lgritz:
Previously, if you for some reason defined two identical functions in
the same scope (including redefining a "built-in" function exactly in
the outermost scope), it would just silently accept it, and then you
wouldn't be sure which one would actually end up getting called.

Now it's an outright error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -267,7 +267,7 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             operator-overloading
             oslc-comma oslc-D
             oslc-err-arrayindex oslc-err-closuremul oslc-err-field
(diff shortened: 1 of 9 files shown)


aconty:
LGTM!

lgritz:
Before I merge, I want to throw this out there:

Is there anybody sufficiently scared about their shaders failing to compile (because they are riddled with duplicate function definitions) that they want to request that this be just a warning rather than an error? (And maybe later on, after everybody has fixed, we can upgrade to a full error?)

lgritz:
You know, as this sits with me, I'm feeling more like it should just be a warning for now.

lgritz:
Pushed an update that changed to a warning.

aconty:
LGTM

PR #747, opened 2017-04-29
Add ways to test OSL version:

lgritz:
* C++: define OSL_VERSION (10900 for 1.9.0), OSL_VERSION_MAJOR,
  OSL_VERSION_MINOR, OSL_VERSION_PATCH.

* osl source at compile time: oslc now pre-defines OSL_VERSION,
  OSL_VERSION_MAJOR, OSL_VERSION_MINOR, OSL_VERSION_PATCH.

* osl runtime now supports getattribute("osl:version"), which also
  get properly constant folded.

Note that this means that renderers ought to make sure their
implementation of RendererServices::get_attribute detects requests for
"osl:version" and in that case retrieve the integer value of
OSL_VERSION. But in practice, you will only notice a renderer's failure
to do this if you turn off runtime optimization and the constant folding
of the attribute doesn't occur.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,7 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             oslc-err-struct-array-init oslc-err-struct-dup
             oslc-warn-commainit
             oslc-variadic-macro
+            oslc-version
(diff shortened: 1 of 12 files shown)


fpsunflower:
LGTM!

PR #749, opened 2017-05-05
Improve efficiency for texture (..., "subimage", "")

lgritz:
When we can detect that the user is passing the empty string as subimage
name to a texture call, there's no need to have the useless call to
osl_texture_set_subimagename.


fpsunflower:
LGTM

PR #751, opened 2017-05-13
Struct constructors work like built-in type constructors.

lgritz:
You know how you can have an expression for a built-in type like color?

    foo = color(1,1,1)

or

    void func (color c) { ...}
    func (color(1,1,1);

It was really awkward to do this with structs.

So with this patch, for every struct, you can construct it by name
with a function-call-like syntax, where the arguments must be the same
types as the data members of the struct, in order.

For example:

    struct RGBA {
        color rgb;
        float a;
    };

    RGBA x = RGBA(color(1,1,1), 1);

    void func (RGBA x) { ... }
    func (RGBA(Cblah,alpha));

You don't need to explicitly declare the constructor like in C++, it
just exists.


lgritz:
So...

aconty:
LGTM!

PR #752, opened 2017-05-15
More flexibility in pointcloud_search/get retrieval

lgritz:
In the default RendererServices implementation of retrieving data from
point clouds via Partio, there had to be an exact match between the
attribute type of the point cloud file (as understood by Partio) and the
element type of the OSL array used to store the retrieved data.

Because OSL doesn't support 2D arrays, that means that you could retrieve
only data of the main non-array OSL types: int, float, 3-float (color,
point, vector, normal etc.).

But we had somebody create a point cloud that stored float[4] to store
quaternions (don't ask), and then was unable to retrieve it from OSL,
because there is no float[4] type.

So the solution is that we are loosening the type matching tests.
Simply stated, it's willing to match either a float or a float[n] in the
point cloud, when retrieving into a float array in OSL. If it's a
float[n], it will just pack them contiguously into the OSL array of
floats, and it's up to the shader to know what's expected and to unpack
and interpret it properly. An error will be reported if there is not
enough room in the array provided.

### DIFF
--- a/src/doc/languagespec.tex
+++ b/src/doc/languagespec.tex
@@ -4337,6 +4337,15 @@ Note that the named point cloud will be created, if it does not yet
 exist in memory, and that it will be initialized by reading a point
 cloud from disk, if there is one matching the name.
 
+Generally, the element type of the data arrays must match exactly the type
+of the point data attribute, or else you will get a runtime error. But there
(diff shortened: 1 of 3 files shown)


fpsunflower:
LGTM

PR #753, opened 2017-05-17
Operator overloading

lgritz:
This patch allows you to implement operator overloading in OSL -- for example, if you define a type like

    struct vector4 {
        float x, y, z, w;
    };

you can also define math operators in the following manner:

    vector4 __operator__add__ (vector4 a, vector4 b) {
        return vector4 (a.x+b.x a.y+b.y, a.z+b.z, a.w+b.w);
    }

and then use plain old `+` as you would for any built-in types:

    vector4 a, b, c;
    c = a + b;    // automatically calls __operator__add__(vector4,vector4)

It's fairly straightforward: for any of the usual unary and binary operators, if a function called `__operator__NAME__` is visible from the current scope and the types match, it is understood that the operator really means a function call to that special function. The valid `NAME` choices are add, sub, mul, div, neg, and so on. See the docs for full list, but it's pretty obvious.

With great power comes great responsibility! Use this wisely -- that is, sparingly, and only when the operator is exactly analogous to how it works with the built-in types. Same general guidance about using it in C++.


fpsunflower:
Cool! LGTM

lgritz:
The best part about this is that it lets people use struct + constructor + overloading to make some interesting types that look native, even when I think they would be silly to add to the core language spec.

luisbarrancos:
That's really cool!

greje656:
This is awesome!

PR #754, opened 2017-05-19
ShadingSystem::convert_value, allow float[3] <--> triple conversion.

lgritz:
This utility is used by renderers for getattribute and other
renderer-to-OSL data copying, and is meant to facilitate the
obvious set of copy conversions between slightly mismatched yet
sensible types.

Adding a clause that lets you retrieve a triple (point, color, etc.)
even if it was declared on the renderer side as float[3], and vice
versa.


fpsunflower:
LGTM

PR #755, opened 2017-05-27
Fix bug comparing structure types with equivalent() -- false positive

lgritz:
It was incorrectly saying that two different structs matched! The
problem was a subtle issue of the ordering of clauses in a complex
comparison -- we need to test the structure case first, because the
other clause makes the assumption that the types are simple and not
structures and will have a false positive in that case. An important way
this could manifest is that if you had two overloaded functions whose
arguments were structs, say foo(structA) and foo(structB), it was not
able to correctly distinguish between them for type checking.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,6 +237,7 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             draw_string
             error-dupes exit exponential
             function-earlyreturn function-simple function-outputelem
+            function-overload-struct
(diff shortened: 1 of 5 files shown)


aconty:
LGTM

fpsunflower:
LGTM2 -- sounds like its worth to add a unit test though.

lgritz:
Yeah, good idea, I'll add a unit test.

lgritz:
Test case added. Verified that it fails before this patch is applied, passes with the patch.

PR #756, opened 2017-05-27
Move TypeSpec::string and c_str from symtab.cpp to typespec.cpp

lgritz:
Just moving these methods from one file to another, so they live with the rest of the methods for the TypeSpec class.


aconty:
LGTM

PR #757, opened 2017-05-30
testshade improvements

lgritz:
A grab-bag of minor testshade improvements:

* Rename command line arguments `--scalest` and `--offsetst` to `--scaleuv` and `--offsetuv`, since what they do is change the u,v range, had nothing to do with anything named "s" or "t". (For compatibility, the old names also still work, but are no longer documented and are considered deprecated.)

* New option `--print` sends to the console the value of all saved outputs. This is useful if you are trying to unit-test or directly inspect the outputs of a shader, without needing 'printf' calls in the shader. Though it's probably not helpful to do this when using -g to produce big output images -- too much console output to be useful.

* When `-o` is outputting to a JPEG, GIF, or PNG file, automatically convert the linear shader output to the sRGB color space required of those file types. That makes it easier to use testshade to produce output images that are "web ready" without a separate color correction step.

* `--runstats` is more careful about not including the time of writing output images in the statistic of shader runtime.

* Change the `-od` option to be called `-d` to match the equivalent option in oiiotool and maketx. I'm not sure why it was ever `-od`, that's weird. Maybe at some point I thought `-d` was going to be something else, maybe debug. Anyway, for compatibility, the `-od` will still work, but it's considered deprecated and is no longer documented.

### DIFF
--- a/src/testshade/testshade.cpp
+++ b/src/testshade/testshade.cpp
@@ -78,6 +78,7 @@ static bool do_oslquery = false;
 static bool inbuffer = false;
 static bool use_shade_image = false;
 static bool userdata_isconnected = false;
+static bool print_outputs = false;
 static int xres = 1, yres = 1;
 static int num_threads = 0;
 static std::string groupname;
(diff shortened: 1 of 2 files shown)


fpsunflower:
LGTM

The output to sRGB part would also be nice for `testrender`.

lgritz:
Sure, of course, that can be arranged.

lgritz:
I looked at testrender, hoping to just copy and paste what I did for testshade. But that section of the code is quite different, so I'm going to merge this now do the testrender change as a separate patch.

PR #758, opened 2017-05-31
testrender: output jpg/gif/png in sRGB

lgritz:
Just like we recently did for testshade.

### DIFF
--- a/src/testrender/testrender.cpp
+++ b/src/testrender/testrender.cpp
@@ -34,6 +34,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <cmath>
 
 #include <OpenImageIO/imageio.h>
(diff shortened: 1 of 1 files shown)


src/testrender/testrender.cpp
```diff
@@ -655,7 +657,10 @@ int main (int argc, const char *argv[]) {
 
     double setuptime = timer.lap ();
 
+    // Local memory for the pixels
     std::vector<Color3> pixels(xres * yres, Color3(0,0,0));
+    // Make an ImageBuf that wraps it
+    ImageBuf pixelbuf (ImageSpec(xres, yres, 3, TypeDesc::FLOAT), &pixels[0]);
```
fpsunflower: I prefer `pixels.data()` these days.

fpsunflower:
LGTM!

src/testrender/testrender.cpp
lgritz: ok, I'll change that

PR #759, opened 2017-06-01
Fix missing faceforward() implementation

brechtvl:
These functions are declared as builtin in stdosl.h but not implemented anywhere, which leads to errors when executing the shader. This fix implements them in stdosl.h following the OSL specification.

Alternatively, these functions could be removed since no one seems to have missed them until now, I don't really care one way or the other.

### DIFF
--- a/src/shaders/stdosl.h
+++ b/src/shaders/stdosl.h
@@ -202,8 +202,14 @@ float distance (point a, point b, point q)
 }
 normal normalize (normal v) BUILTIN;
 vector normalize (vector v) BUILTIN;
-vector faceforward (vector N, vector I, vector Nref) BUILTIN;
-vector faceforward (vector N, vector I) BUILTIN;
+vector faceforward (vector N, vector I, vector Nref)
+{
+    return (dot(I, Nref) > 0) ? -N : N;
+}
+vector faceforward (vector N, vector I)
+{
+    return faceforward(N, I, Ng);
+}
 vector reflect (vector I, vector N) { return I - 2*dot(N,I)*N; }
 vector refract (vector I, vector N, float eta) {
     float IdotN = dot (I, N);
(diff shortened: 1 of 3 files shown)


lgritz:
Wow, how did we go for so long without anybody noticing?
LGTM, I will merge it shortly.

fpsunflower:
The code looks fine -- but why are these needed? I believe we require the normal to be on the same side as the incoming ray and expose the side we are on via `backfacing()` instead.

I would vote for removing these.

lgritz:
It doesn't have to be literal N. Can you think of no situation in which you might have some other vector that you wish to flip around using a traditional faceforward operation?

fpsunflower:
I guess I am mostly thinking of avoiding confusion. Someone reading the header only might not realize its not needed like in other shading languages?

If we want to include it, it might be better to avoid the name `N` since it clashes with the global symbol?

fpsunflower:
(actually `N` and `I` are both globals)


Should the 2 argument form call `faceforward(N, I, N)` instead of `faceforward(N, I, Ng)`? Its unclear which is more desirable if we are thinking of use cases that are not related to `backfacing()`.

lgritz:
I think it's right as it is. The 3-argument lets you use whatever vector you want, and the 2-argument version is implicitly the geometric normal.

If you want to argue that we should name the parameters something other than N and I, to avoid confusion with the globals of those names, I think that's a reasonable position.

dankripac:
oh we noticed, but we were naughty and implemented it in our own headers
and then didn't report it... and then forgot about it..
Sorry about that!

lgritz:
I'm going to take @dankripac's answer as empirical proof that these are useful and commit the change.
If somebody wants to argue that the choice of parameter names that we document for these functions is confusing, we can change the documentation/declaration. But I observe that somehow, this naming scheme does not seem to have significantly hindered the many prior shading languages that have the identical functions and parameter names.

PR #760, opened 2017-06-01
Refine structure type comparison

lgritz:
The prior method of comparing structures by examining them for identical
layout failed if you wanted to have two structures that purposely had
the same layout but wanted their semantics to be different (for example,
by having a function that would work on only one of them).

Change the struct type checking comparisons to check NAME as well as
struct layout. Also need to no longer mangle the name when reading in
oso.

While I was there, and adding a new test for this, I renamed a test
I added recently, to more closely match other related ones.


lgritz:
Derek reports that this fixes the issue he reported.

PR #761, opened 2017-06-01
Improve type checking errors with structs

lgritz:
Change the implementation of TypeSpec::string(), which returns the text
representation of a TypeSpec, so that it prints the name of the struct
rather than just its numerical designation.

An example of the old behavior, as applied to a type checking error
message:

    test.osl:10: error: No matching function call to 'max (struct 1, struct 1)'
    ...

And the new behavior:

    test.osl:10: error: No matching function call to 'max (struct vector4, struct vector4)'
    ...


fpsunflower:
LGTM!

PR #762, opened 2017-06-01
Fix oslc assertion when passing stucts with operator overloading.

lgritz:
The recent patch that allows users to specify operator overloading
for structs means that, in this particular spot, the AST node type
that might be passed as the argument to a function might be a binary
or unary expression -- those could never have returned a struct before,
so were not enumerated in the list of nodes of this conditional, causing
an assertion to hit.

### DIFF
--- a/src/liboslcomp/codegen.cpp
+++ b/src/liboslcomp/codegen.cpp
@@ -1732,7 +1732,9 @@ ASTfunction_call::codegen (Symbol *dest)
                 // If the formal parameter is a struct, we also need to
                 // alias each of the fields
                 if (a->nodetype() == variable_ref_node ||
-                    a->nodetype() == function_call_node) {
+                    a->nodetype() == function_call_node ||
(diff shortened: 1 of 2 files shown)


src/liboslcomp/codegen.cpp
```diff
@@ -1732,7 +1732,9 @@ ASTfunction_call::codegen (Symbol *dest)
                 // If the formal parameter is a struct, we also need to
                 // alias each of the fields
                 if (a->nodetype() == variable_ref_node ||
-                    a->nodetype() == function_call_node) {
+                    a->nodetype() == function_call_node ||
```
fpsunflower: LGTM

Maybe add this to the operator overloading unit tests?

lgritz: yeah, good idea

lgritz:
Pushed an update where I've added to the operator-overload test.

PR #765, opened 2017-06-05
Add color/vector-valued version of smoothstep

lgritz:


fpsunflower:
LGTM

How about `smoothstep(color a, color b, float x)` ?

lgritz:
That already works, thanks to the magic of type promotion!

PR #766, opened 2017-06-05
Add mix() for color closures

lgritz:
The mixing factor may be either a float or a color, and it blends the
color closures in the obvious way.


fpsunflower:
LGTM

fpsunflower:
Actually - does this need to be a builtin so that the short-circuit provisions of the other `mix` calls can apply?

lgritz:
That's probably a good idea, but do you mind if I merge this working (but not extra smart) version now, and then I'll put the smarter version on my list?

fpsunflower:
Sure - its always easier to add opcode specializations later than the other way around.

PR #767, opened 2017-06-06
Fix osl_range_check not found error with USE_LLVM_BITCODE=OFF.

brechtvl:
This broke in master in a9c21bdf, there's a runtime error for many OSL shaders when building with USE_LLVM_BITCODE=OFF.


lgritz:
LGTM

lgritz:

brechtvl:
Yes, it's only an issue in master.

PR #768, opened 2017-06-06
Fix IlmBase linker warnings with LINKSTATIC on Windows.

brechtvl:
Since d771ee8 OPENEXR_DLL is defined when LINKSTATIC=ON which I'm guessing was not an intentional change. This restores the behavior before the refactor, fixing a bunch of LNK4217 warnings when building with static IlmBase libraries.

The same fix is needed in OpenImageIO.


lgritz:
LGTM

lgritz:

brechtvl:
It is only an issue in master, created OIIO pull request here: OpenImageIO/oiio#1701

PR #769, opened 2017-06-06
Fix Windows build to not force enable /MT

brechtvl:
As part of Appveyor support, /MT was always enabled when BUILDSTATIC=ON. The only way to build with /MD now seems to be to patch the OSL source to remove this, whereas previously it was possible to build with custom `CMAKE_CXX_FLAGS_RELEASE` or similar flags.

Note that /MT and /MD control if the MSVCRT.lib system library is statically or dynamically linked, which is orthogonal to the choice of creating a static or shared OSL library.

This commit only enables /MT for Appveyor builds, restoring the previous behavior for regular builds.

The same issue exists in OpenImageIO.


lgritz:
LGTM

Are you going to submit the corresponding patch on the OIIO side, or did you want me to do that on my end?

lgritz:

brechtvl:
It is only an issue in master, created OIIO pull request here: OpenImageIO/oiio#1702

PR #771, opened 2017-06-10
Fix minor numerical precision problem with inversespline

lgritz:
We were having some troubles right at the boundary, which was easily
fixed if we changed the tests for trivial cases to be `<=` and `>=` instead
of `<` and `>` the boundary value.

Required also implementing `<=` and `>=` for Dual types.


fpsunflower:
LGTM

lgritz:
User who reported the error says this patch fixes his case.

PR #773, opened 2017-06-10
testshade: proper error for common mistake with order of --param vs shader

lgritz:
Example:

    $ testshade -param x 1.0 test
    x=1.0

    $ testshade test -param x 1.0
    ERROR: Pending parameters without a shader: x
    Did you mistakenly put --param after the shader declaration?


fpsunflower:
LGTM

PR #774, opened 2017-06-19
Fix bug in transitive assignment of indexed arrays.

lgritz:
This worked:

    float x, y, z;
    x = y = z;

for non-array values. But this hit an assertion:

    float f[3];
    f[0] = f[1] = f[2];


aconty:
LGTM

PR #775, opened 2017-06-19
hash() of numeric types, and hashnoise()

lgritz:
`int hash(T)` extends the existing integer-valued hash that we have for
strings, to accept arguments of float, 2 floats, triple, triple+float,
int.

`float_or_triple hashnoise(...)` is like cellnoise, but is discontinuous
everywhere (i.e., returns a different uncorrelated value for every
distinct real-valued input).


fpsunflower:
LGTM!

PR #780, opened 2017-07-14
Implement fprintf

lgritz:
Has always been in the docs. I swear I had this working at one point.
But it seems to be MIA.


src/liboslexec/opstring.cpp
```diff
@@ -213,6 +213,24 @@ osl_warning (ShaderGlobals *sg, const char* format_str, ...)
...
+OSL_SHADEOP void
+osl_fprintf (ShaderGlobals *sg, const char *filename,
+             const char* format_str, ...)
+{
+    va_list args;
+    va_start (args, format_str);
+    std::string s = Strutil::vformat (format_str, args);
+    va_end (args);
+
+    OIIO::mutex fprintf_mutex;
```
fpsunflower: Did you mean to declare this mutex as `static` ?

lgritz: Yes, I certainly did. :-)

lgritz: Pushed an update to fix this.

fpsunflower:
LGTM

PR #782, opened 2017-07-15
Overhaul MX shader generation to produce the actual osl files

lgritz:
Previously, the mx templates used tricky #define techniques to do
the type specialization. But now we do some string substitution
to generate the actual .osl for every type flavor of every mx shader,
then just directly oslc that.

In the process, mx_types goes away because there's nothing left in it
that was necessary. And I merged swizzle.h into mx_funcs.h, and also
needed to expand the various typed swizzling operations.


lgritz:
Updated after realizing that (1) I'd clobbered Adam and Derek's script in a way I didn't mean to (thanks, Adam, for augmenting this patch to put things back into sane state), and (2) I needed to touch a bunch more .mx files that still harbored relics of the #define approach I was trying to change.

I feel better about the state now.

Adam and Derek, I need you to look over the code very carefully to make sure I haven't broken anything.

lgritz:
In particular, Adam & Derek, please note the more extensive change I made to mx_funcs.h, and the resulting simplification of the structure of the noise, fbm, image, and triplanar nodes.

derekhaase:
LGTM.
I like the fbm refactor, much clearer vs those #ifdefs!

PR #784, opened 2017-08-11
fixes var types not substituting properly

amartinez-cg:
mx build script wasn't properly resetting the mx template after substitution. 
This change copies the template into the var_type scope for each new var_type


lgritz:
LGTM

PR #787, opened 2017-08-18
Make OSL work with clang5/llvm5

lgritz:
Very minor. This is the most minimal LLVM upgrade I can recall.

Note that the changes to Makefile-bits are site-specific to SPI.

### DIFF
--- a/site/spi/Makefile-bits
+++ b/site/spi/Makefile-bits
@@ -16,26 +16,18 @@ ifeq ($(SP_OS), rhel7)
 
     ## If not overridden, here is our preferred LLVM installation
     ## (may be changed as new versions are rolled out to the facility)
-    LLVM_DIRECTORY ?= /shots/spi/home/lib/arnold/rhel7/llvm_4.0_final/
+    LLVM_DIRECTORY ?= /shots/spi/home/lib/arnold/rhel7/llvm_4.0_final
 
     # A variety of tags can be used to try specific versions of gcc or
     # clang from the site-specific places we have installed them.
-    ifeq (${COMPILER}, clang342)
-        MY_CMAKE_FLAGS += \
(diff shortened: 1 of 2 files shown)


site/spi/Makefile-bits
```diff
@@ -16,26 +16,18 @@ ifeq ($(SP_OS), rhel7)
 
     ## If not overridden, here is our preferred LLVM installation
     ## (may be changed as new versions are rolled out to the facility)
-    LLVM_DIRECTORY ?= /shots/spi/home/lib/arnold/rhel7/llvm_4.0_final/
+    LLVM_DIRECTORY ?= /shots/spi/home/lib/arnold/rhel7/llvm_4.0
```
fpsunflower: `llvm_4.0_final` is the right one. The symlink still appears to be pointed as 4.0rc2b

fpsunflower: (of course we may as well just update the `symlink` since thats what we added it for)

lgritz: Yeah, that's what I was thinking, but I forgot to do it. I'm out today, can you fix the link?

lgritz: And similarly, I installed llvm_5.0_RC2, with a llvm_5.0 link that points to it. But in a couple weeks when a final one comes out, I'll just update the link rather than change all the build files.

fpsunflower: The danger with the link is that changing it will "live change" any older builds that point to those files.

Seems a bit risky. I think I prefer to keep things explicit.

lgritz: OK, fair enough. I will update the PR.

lgritz:
Updated with the SPI portions reverting a change in directory for where to find llvm 4.0.

fpsunflower:
LGTM

PR #790, opened 2017-08-18
OSL runtime constant folding of % (modulus)

lgritz:
This was never done as a runtime optimization, just never got around to it because % is so rarely used in shaders. But once somebody pointed it out, it was just a few minutes of work.

No new tests need to be added, since % is already tested in the testsuite. If the constant folding doesn't work, testsuite/arithmetic will break.


fpsunflower:
LGTM

PR #791, opened 2017-08-29
Fix mod constant folding to handle floats, oops

lgritz:


src/liboslexec/constfold.cpp
```diff
@@ -308,25 +308,30 @@ DECLFOLDER(constfold_mod)
...
+            int cind = rop.add_constant (a % b);
+            rop.turn_into_assign (op, cind, "const % const");
+            return 1;
+        } else if (A.typespec().is_float() && B.typespec().is_float()) {
+            float a = A.get_float();
+            float b = B.get_float();
+            int cind = rop.add_constant (fmodf(a, b));
+            rop.turn_into_assign (op, cind, "const % const");
+            return 1;
+        }
```
fpsunflower: How about `vec% float` ? That seems to be handled on the llvm_gen side as well.

lgritz:
Revised.

After some digging, I see that for some time now, any code compiled with oslc will have a mod() implementation provided in stdosl.h and thus doesn't need to be folded here. But very old .oso may hit this, and can't guarantee that both args are int. So remove the assertion, but don't bother optimizing for that case because it can't happen with newly compiled code.

fpsunflower:
LGTM!

PR #794, opened 2017-10-05
Allow an (optional) slightly more relaxed policy when declaring parameters

fpsunflower:
This allows a color to be declared as float[3] for example. This may be useful when declaring shaders from systems that do not explicitly keep track of the source data type.


aghiles:
At first view it seems like a nice addition. But frankly, I feel this will not be a real service for users and might lead to sloppiness and mistakes down the road.

fpsunflower:
We figured some may feel that way, that's why the default is still to typecheck exactly.

You have to opt-in to the sloppy behavior.

lgritz:
LGTM. I think Aghiles' concerns are already addressed by the implementation. You won't see any change unless you take action to set the "relaxed_param_typecheck" attribute to nonzero.

aghiles:
Perfect ! Thanks guys for your continued work on OSL.

PR #796, opened 2017-10-16
transformc overhaul

lgritz:
Convert transformc() from strosl.h-implemented to C++.

Rather than have it aware of just a few color hard-coded color space
names, allow it to fall back on OpenColorIO (via ColorConfig/ColorProcessor
functionality exposed by OIIO).

This requires OIIO 1.9 and some recent changes in order to be fully
functional. The corresponding OIIO patch is here:
https://github.com/OpenImageIO/pull/1788

Currently, this does not compute derivatives correctly for transformc()
results. I want to go back to this, but it's not trivial because OCIO
doesn't have entry points that return partial derivatives. I don't think
it's an emergency for any practical cases -- you would only notice if you
used the *results* of a transformc call as texture coordinates.


src/liboslexec/llvm_gen.cpp
```diff
@@ -1596,6 +1596,26 @@ LLVMGEN (llvm_gen_transform)
...
+    Opcode &op (rop.inst()->ops()[opnum]);
+    ASSERT (op.nargs() == 4);
+    Symbol *Result = rop.opargsym (op, 0);
+    Symbol *From = rop.opargsym (op, 1);
+    Symbol *To = rop.opargsym (op, 2);
+    Symbol *C = rop.opargsym (op, 3);
+
+    llvm::Value *args[] = { rop.sg_void_ptr(),
+        rop.llvm_void_ptr(*C), rop.ll.constant(C->has_derivs()),
+        rop.llvm_void_ptr(*Result), rop.ll.constant(Result->has_derivs()),
```
fpsunflower: Maybe add a warning if `Result->has_derivs()` is true? Just so that we have breadcrumbs to follow when things go wrong many years from now and we have all forgotten about the shortcut?

lgritz: How about... not years, I'll do it this week? I just wanted to get this stuff to production for a high-priority request. But I don't know how to do it for the OCIO case except by finite differences.

fpsunflower: Are you thinking of handling the HSL/HSV cases with real dual2 at least? Those are the cases that are more likely to end up needing derivatives and would be harder to get right in a finite differencing setup.

I'd honestly be fine with making it a warning (error?) too. It does seem highly unlikely for derivs to be needed here (although more unlikely scenarios have come up before ...).

lgritz: Yeah, I was going to use dual2 for all the "hard coded" ones.

lgritz: I don't want to think about how to do the warning (at JIT time? or shader runtime?). I'm inclined to commit this in its current state so we can get the builds done that we need, and then immediately go fix it and make it work properly with derivs, within the next few days.

fpsunflower: I was thinking JIT time, but your plan sounds good.


LGTM!

PR #797, opened 2017-10-16
Make relaxed parameter type checking less strict

fpsunflower:
We previously allowed float to vec3 conversions, and it turns out some workflows at SPI could lead to float parameters being passed in as int which we need to perserve.

However, we don't want to get too much more relaxed that this, so make all other cases errors because other mismatches should be flagged as bigger problems and not silently make the wrong picture.


lgritz:
LGTM

PR #798, opened 2017-10-18
ctransform full derivative support.

lgritz:
Using Dual2 for exact derivs of the hard-coded color transformations. Use finite differences to numerically differentiate the OCIO color transformations.

Along for the ride are a number of minor additions to Dual2 and friends.


fpsunflower:
LGTM!

PR #800, opened 2017-10-23
Fix improper parsing of string_view integer and try to validate it better.

marsupial:
## Description
Current code is passing **string_view::data()** to both **strchr** and **atoi**, which expect/require null terminated strings.

## Tests

Part 1 of 2 which allows connecting individual components of aggregate types.
Second part includes proper test against **ShadingSystemImpl::decode_connected_param**.

## Checklist:
- [x ] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [ x] I have previously submitted a [Contributor License Agreement (http://opensource.imageworks.com/cla/).

### DIFF
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -2532,7 +2532,11 @@ ShadingSystemImpl::decode_connected_param (string_view connectionname,
 
     c.type = sym->typespec();
 
-    if (!bracket.empty() && c.type.is_array()) {
+    // Early out redundant checks below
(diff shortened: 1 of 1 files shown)


lgritz:
I think your diagnosis of the problem is correct (good catch), and I believe your code is fine, but I think there may be a simpler solution. Let me take a quick stab at it and see which way you like better.

lgritz:
Counter-proposal presented as #805. Thoughts?

lgritz:
I merged the counter-proposal of #805, so closing this one.

PR #801, opened 2017-10-23
Allow individual components of aggregate types to be connected.

marsupial:
## Description
Connecting only the r,g, or b channels into another color is common in a lot shader networks.
Also a possible start to allow native _swizzling_ in the language.

## Tests

Yes, **testsuite/render-components**

## Checklist:

- [ x] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [x ] I have previously submitted a [Contributor License Agreement](http://opensource.imageworks.com/cla/).
- [x ] I have ensured that the change is tested somewhere in the testsuite (adding new test cases if necessary).
- [ x] My code follows the prevailing code style of this project.


aghiles:
Oh this one would be really helpful !! Maya's hypershade networks are full of these component->float connections. You can imagine how many intermediate shader nodes this work would save. Great stuff.

lgritz:
I'm sorry to have taken so long to evaluate this properly.

I like this. 

I see that the code changes are ONLY in llvm_instance.cpp and backend_llvm.{cpp,h}. Does that mean that up until now, the ShadingSystem::ConnectShaders accepted "foo[1]" or whatever to connect individual components to floats, but the implementation was simply wrong or incomplete? That's the way it looks to me; I'm surprised that hasn't come up as an issue for somebody.

There are a couple ASSERT calls for unsupported conditions in `BackendLLVM::llvm_assign_initial_value`. I wonder if maybe there needs to also be code in `ShadingSystemImpl::ConnectShaders` to catch those conditions at the time that the shader group is declared and report an error properly, rather than only find out by way of an ASSERT when we are generating IR. I'm asking this somewhat rhetorically; it can be done as a later review, I suppose, since what you have here is certainly an improvement over what we had before (which also would have failed in some funny way for that case).

lgritz:
Note to self: Lee Kerley reminds me to look into whether this can harm runtime optimization and recognition of constants in any way. I'll investigate.

lgritz:
Is it ok if I add some commits to your branch, in particular another unit test that reveals some problems?

marsupial:
Feel free to push whatever you like on top.

ShadingSystem::ConnectShaders used to fail parsing `connect A.outColor[1]` properly, but that was a deeper bug using C functions with a `string_view::c_str` and fixed in f785cc2.

As far as the asserts, I believe ShadingSystem::ConnectShaders is still unable to parse components of arrays `connect A.outColor[1][0]` and will error already, the assert is just validating that.

lgritz:
Well, it says "Add more commits by pushing to the PR/1 branch on marsupial/OpenShadingLanguage," but I can't figure out how to do that as I don't have permissions on your account. When I do the push, it just tells me "Authentication required: You must have push access to verify locks."

I've never tried to amend a PR with changes before. Is there a trick to it that I need to know? 

Did you check the box that says "allow edits from maintainers" when you created the PR?

marsupial:
Edits by maintainers are allowed. No clue how to do it.
This seems to state that you'd have to clone my fork and push to the branch there?
https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/

lgritz:
Ah, so maybe I can't push to your repo from mine (even if yours is added as a remote), but if I fully clone yours, then I can? Let me try that, stay tuned...

lgritz:
Aha, how is that?

I also rebased on top of the current master.

I added a new testsuite/connect-components that revealed that the individual component connections would hit assertions in certain optimizations. Your test avoided that, because you made the connection from the checker shader -- it didn't exercise the code that wanted to turn those partial connections into constants.

I fixed it myself, those are also included in the amendment. I didn't alter any of your existing code.

aghiles:
Great news ! Can't wait to try that with our HyperShade OSL exporter.

lgritz:
Use with extreme caution... there are a number of caveats to be aware of:

* Because we (SPI) don't use single component connections in practice (and won't, because Katana don't have a way to express them), I can't give the usual "this has been hardened by constant production use" promise that I usually am comfortable stating.

* The notion of where a parameter's value comes from (default, instance value, connection, geom) is tracked for the whole parameter, not for each individual component or array element. Connecting just a subset of the components seems to work in our unit tests, but maybe there are edge cases where it leads to weird behavior.

* It won't surprise me if minor buglets turn up in unanticipated edge cases that involve partial connections -- not all components overridden, or each component connected to different source shaders.

* As maybe you can see from the code, partial connections give up on trying to figure out a few optimizations that are able to happen when whole parameters are connected. 

If you see anything weird happening, please report (with a clear, minimal, reproduction case) so we can fix anything that's subtly broken.

marsupial:
If not obvious, seems the change of **testsuite/runtest.py** in e8f8d96 is breaking tests and should be:
`for filetype in [ "*.osl", "*.h", "*.oslgroup", ".xml" ] :`

lgritz:
oops, fixing now

aghiles:
Thank Larry for the heads up. We will be testing this and will report anything that doesn't work.

lgritz:
pushed fix for runtest.py

marsupial:
Whoops, The local branch I have adds tests automatically, and I just notices the PR doesn't explicitly run my original test.
You mind if I shuffle/squash this branch into the following:

1. Support connections between Symbol components.
2. Component connection tests. [squash all tests additions/fixes into here]
3. Debug aid: human readable Connection
4. Avoid errors for component connections when optimizing.
5. Additional runtime optimization I can currently see.

lgritz:
Sure. You can also add on top... when I merge for real, I'm going to squash all those individual commits into just one, so there's not a lot of value in carefully curating which changes are in which commit, unless we have reason to re-separate your contributions from mine.

marsupial:
Updated.
Last commit adds removal of init-ops for float->triple connections which seems like a valid thing to do?

lgritz:
Yes, I think your update is correct.

lgritz:
OK, no more touching!

I'll do some tests on my end to make sure it doesn't break anything in our production shaders, then merge if that all looks ok.

marsupial:
Not sure how to adjust the allowed variance of an output image from a **testrender** test, but that seems to be why the **render-components** is failing on Linux (while OS X is passing).

lgritz:
It seems that you recently removed these two lines from render_components/run.py:

-failthresh = 0.03   # allow a little more LSB noise between platforms
-failpercent = .5

lgritz:
I amended. Let's see if that fixes the CI tests.

marsupial:
Sadly not, maybe the output helps?

```
Comparing "out.exr" and "ref/out.exr"
320 x 240, 3 channel
  Mean error = 0.13774
  RMS error = 0.217738
  Peak SNR = 13.2583
  Max error  = 1.00195 @ (40, 196, R)  values are 0, 0, 0 vs 1.00195, 0.346436, 0.346436
  65449 pixels (85.2%) over 0.12
  72221 pixels (94%) over 0.06
FAILURE
Comparing "out.exr" and "ref/out.exr"
320 x 240, 3 channel
  Mean error = 0.13774
  RMS error = 0.217738
  Peak SNR = 13.2583
  Max error  = 1.00195 @ (40, 196, R)  values are 0, 0, 0 vs 1.00195, 0.346436, 0.346436
  65449 pixels (85.2%) over 0.12
  72221 pixels (94%) over 0.06
FAILURE
```

lgritz:
Oh that's weird. I'll take a closer look at the images.

lgritz:
Dammit, on my work Linux machine, I get a perfect match. 

Are you able to get the failure directly on your machine?

lgritz:
There's really no part of the test that is dependent on closures or 3D scenes, so there's no particular reason why it should be a "testrender" test (more expensive to run, and more finicky to compare cross-platform). Instead of fighting this, maybe we should just quickly morph it into a straightforward "testshade" test? I'm happy to give this a try, bet I could do it in minutes.

marsupial:
It's only a **testrender** because it was the only way I saw to test connections.
Only saw it could be done otherwise from looking at your other tests today.

All it is doing is testing the following connections work as expected:

colorParmA[1] -> colorParmB
colorParmA[2] -> colorParmB[2]

colorParmA[0] -> colorParmB[2]
colorParmA[1] -> colorParmB[1]
colorParmA[2] -> colorParmB[0]

redParm[0] -> blackParm[0]
redParm[1] -> blackParm[1]
redParm[2] -> blackParm[2]

marsupial:
Should be all good now.
Took the liberty of removing upstream.oso & downstream.oso which seem to have wandered in.

lgritz:
LGTM, merging.

PR #803, opened 2017-10-23
Dual refactor and generalization

lgritz:
In this patch, I completely overhaul the main class that implements our
dual arithmetic-based automatic differentiation. No more Dual2, with
hard-coded x and y partials. Now we have Dual<T,P>, templated on the
number of partial dimensions P. Now Dual2<T> is just a templated typedef
that specilizes to 2 dimensions.

In addition to general cleanup, this leads to several benefits:

1. It is now trivial to extend to 3 dimensions (dx, dy, dz) if we wanted
to use Dual for automatic differentiation of volumetric functions.

2. Dual<T,1> can be used for automatic differentiation of 1-D functions,
which doesn't have any applications within OSL that I'm aware of, but
makes it a whole lot more convenient to re-use outside of OSL, any time
you want to use A.D. with these techniques.

3. It sets us up for a potential future distribution of the Dual stuff
as a separate library, if it seems useful.  It's not 100% free of other
dependencies from OSL (& OIIO), but it's fairly close and would not be
hard to make it entirely self-contained (and header only). I don't need
to do that this week, but it's something I'd like to work toward.

I did double check this by benchmarking full renders, and have confirmed
that it doesn't affect performance at all.

### DIFF
--- a/src/include/OSL/dual.h
+++ b/src/include/OSL/dual.h
@@ -875,8 +875,8 @@ inline Dual<T,P> fast_exp(const Dual<T,P> &a)
 template<class T, int P>
 inline Dual<T,P> exp2 (const Dual<T,P> &a)
 {
-    // FIXME: std::exp2 is only available in C++11
-    T f = exp2f(float(a.val()));
+    using std::exp2;
+    T f = exp2(a.val());
     return dualfunc (a, f, f*T(M_LN2));
 }
 
@@ -892,8 +892,8 @@ inline Dual<T,P> fast_exp2(const Dual<T,P> &a)
 template<class T, int P>
 inline Dual<T,P> expm1 (const Dual<T,P> &a)
 {
-    // FIXME: std::expm1 is only available in C++11
-    T f  = expm1f(float(a.val())); // float version!
+    using std::expm1;
+    T f  = expm1(a.val());
     T df = std::exp  (a.val());
     return dualfunc (a, f, df);
 }
@@ -910,8 +910,8 @@ inline Dual<T,P> fast_expm1(const Dual<T,P> &a)
 template<class T, int P>
 inline Dual<T,P> erf (const Dual<T,P> &a)
 {
-    // FIXME: std::erf is only defined in C++11
-    T f = erff (float(a.val())); // float version!
+    using std::erf;
+    T f = erf (a.val());
     const T two_over_sqrt_pi = T(1.128379167095512573896158903);
     T df = std::exp (-a.val() * a.val()) * two_over_sqrt_pi;
     return dualfunc (a, f, df);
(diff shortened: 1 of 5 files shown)


fpsunflower:
I like the change overall. I spotted a few small things in reading the code - but they appear to be bug-compatible with the previous implementation.

Also: It might be good to address the C++11 FIXME's now that C++11 is a requirement (as a separate commit of course).

LGTM!

src/include/OSL/dual.h
```diff
@@ -287,533 +404,588 @@ inline Dual2<T> operator/ (const Dual2<T> &a, const T &b)
...
+inline Dual<T,P> erfc (const Dual<T,P> &a)
 {
     // FIXME: std::erfc is only defined in C++11
-    T erfca = erfcf (float(a.val())); // float version!
-    T two_over_sqrt_pi = -T(1.128379167095512573896158903);
-    T derfcadx = std::exp (-a.val() * a.val()) * two_over_sqrt_pi;
-    return Dual2<T> (erfca, derfcadx * a.dx(), derfcadx * a.dy());
+    T f = erfcf (float(a.val())); // float version!
+    const T two_over_sqrt_pi = -T(1.128379167095512573896158903);
+    T df = std::exp (-a.val() * a.val()) * two_over_sqrt_pi;
```
fpsunflower: Not really related to this review, but the '-' sign is missing here.

fpsunflower: EDIT: scratch that - i see it now.

src/include/OSL/dual.h
```diff
@@ -287,533 +404,588 @@ inline Dual2<T> operator/ (const Dual2<T> &a, const T &b)
...
-    return Dual2<T> (erfa, derfadx * a.dx(), derfadx * a.dy());
+    T f = erff (float(a.val())); // float version!
+    const T two_over_sqrt_pi = T(1.128379167095512573896158903);
+    T df = std::exp (-a.val() * a.val()) * two_over_sqrt_pi;
+    return dualfunc (a, f, df);
 }
 
-inline Dual2<float> fast_erf(const Dual2<float> &a)
+template<class T, int P>
+inline Dual<T,P> fast_erf(const Dual<T,P> &a)
```
fpsunflower: The old implementation only made the `fast_*` calls available for `T=float`. Do you think there is any value in supporting the fast implementations for doubles given they are all float internally?

lgritz: I was trying to specifically avoid hard-coding anything to be float only, unless I truly thought the only possible implementation was for single, scalar floats.

I want to keep all the Dual functionality maximally composable with the SIMD types (we allow `fast_exp(vfloatN)`, so why not allow a version with derivatives without any additional code?) as well as Imath types (if we add a `fast_exp(Vec3)` that operates per-component, this template would also give us a `fast_exp(Dual<Vec3>)`).

Using it with double will already be a build-time error, since no fast_blah(double) will be found most of the time.

fpsunflower: Ah I wasn't thinking of the SIMD types. Make sense as it is then.

`fast_blah(double)` doesn't exist but float is convertible to `double` so I think the compiler will accept it anyhow.

lgritz:
Oh, good catch on the FIXME's, I missed that. I will update that...

fpsunflower:
LGTM!

lgritz:
Just pushed one more -- made the unit test verify that, for some simple examples, the Dual derivatives match numerical derivatives using central difference method.

PR #805, opened 2017-10-25
Fixes to ShadingSystemImpl::decode_connected_param

lgritz:
@marsupial correctly points out that we are unsafely passing
a string_view.data() into both atoi and strchr, which is unsafe because
there's no guarantee that a string_view is null-terminated. Ugh!

He proposed a fix here: https://github.com/imageworks/OpenShadingLanguage/pull/800
and there's nothing wrong with it per se, but I suspected it could be done
more simply, so this PR is my stab at it, using the `OIIO::Strutil::parse_*`
functions.

### DIFF
--- a/src/liboslcomp/osllex.l
+++ b/src/liboslcomp/osllex.l
@@ -223,7 +223,7 @@ void preprocess (const char *yytext);
 
 
 {FLT}                   {
-                            yylval.f = atof (yytext);
+                            yylval.f = OIIO::Strutil::from_string<float>(yytext);
                             SETLINE;
                             return FLOAT_LITERAL;
                         }
--- a/src/liboslexec/constfold.cpp
+++ b/src/liboslexec/constfold.cpp
@@ -1239,7 +1239,7 @@ DECLFOLDER(constfold_stoi)
     if (S.is_constant()) {
         ASSERT (S.typespec().is_string());
         ustring s = *(ustring *)S.data();
-        int cind = rop.add_constant ((int) strtol(s.c_str(), NULL, 10));
+        int cind = rop.add_constant (Strutil::from_string<int>(s));
         rop.turn_into_assign (op, cind, "const fold stoi");
         return 1;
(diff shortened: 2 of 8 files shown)


lgritz:
This is a counter-proposal to #800

fpsunflower:
LGTM

marsupial:
Seems to play nicely here with #801.

FWIW **OIIO::Strutil::parse_int** also looks to have the same issue and may parse beyond the string_view's length.
`line 689:        p.remove_prefix (end-p.begin());` can throw with a std::string_view.

lgritz:
Oh, let me have a look at parse_int, too, then.

lgritz:
Ow, you are quite right. Will make a fix there...

lgritz:
parse_float as well...

lgritz:
I want to point out that this is a legit bug according to spec, but in practice in our code base the string_view's are always generated from a std::string or ustring or null-terminated char* (or a substring of one of those), so it's not a huge emergency "how did this ever work" kind of bug -- in our code, every string_view does point to (or within) a null-terminated string. So it can't crash, but it can generate wrong results if there's a string like "1234" and the string_view is referring to just the first 2 characters (should parse to 12, but will erroneously parse to 1234). I don't think we actually do this anywhere, but it's a nasty bug just waiting to happen, so I want to fix it.

lgritz:
I feel like this issue is very much related to the changes in https://github.com/OpenImageIO/oiio/pull/1785, so I think I'll address the primary problems with the parse functionality there.

marsupial:
The **strchr** bug was definitely creating bad results for me.

I agree the rest is a bit pedantic from what I can see in the context of OSL, but it's not to hard to imagine an exploit where OIIO (or some client of) tries parsing a well crafted string in some image header/metadata.

lgritz:
I agree totally!  I'm working on a robust fix for it all.

PR #810, opened 2017-11-14
Fix ClosureComponent to work with SSE alignment requirements

brechtvl:
## Description

In #740 the `alignment` parameter was removed from `register_closure`. This was originally added in bcaf5778, and Cycles still needs 16 bytes alignment for SSE types, while it is fixed to 8 bytes in OSL master now.

This patch uses the simple solution of enforcing 16 byte alignment for `ClosureComponent.mem`, under the assumption that in practice this is the most alignment you need on x86.

Alternatively the `alignment` parameter could be added back. Maybe in theory someone might care about 32 byte alignment for AVX. In practice I guess it's not so likely since alignment is only required for older CPUs that don't support unaligned loads well.

## Checklist:

### DIFF
--- a/src/include/OSL/oslclosure.h
+++ b/src/include/OSL/oslclosure.h
@@ -114,29 +114,26 @@ struct OSLEXECPUBLIC ClosureColor {
 
 /// ClosureComponent is a subclass of ClosureColor that holds the ID and
 /// parameter data for a single primitive closure component (such as
-/// diffuse, translucent, etc.).  The declaration leaves 4 bytes for
-/// parameter data (mem), but it's expected that the structure be
-/// allocated with enough space to house all the parameter data for
-/// whatever type of custom primitive component it actually is.
-struct OSLEXECPUBLIC ClosureComponent : public ClosureColor
+/// diffuse, translucent, etc.).
+///
+/// ClosureComponent itself takes up 16 bytes, and its allocation will be
+/// scaled to add parameters after the end of the struct. Alignment is
+/// set to 16 bytes so that 64 bit pointers and 128 bit SSE types in user
+/// structs have the required alignment.
+struct OSLEXECPUBLIC OSL_ALIGNAS(16) ClosureComponent : public ClosureColor
 {
     Vec3 w;                     ///< Weight of this component
-    alignas(void*) char mem[8]; ///< Memory for the primitive, 8 is the minimum, allocation will be
-                                ///  scaled to requirements of the registered closure struct.
-                                ///  Alignment is forced to a void* to match likely alignement requirements
-                                ///  of the user's structs.
 
     /// Handy method for getting the parameter memory as a void*.
-    ///
-    void *data () { return &mem; }
-    const void *data () const { return &mem; }
+    void *data () { return (char*)(this + 1); }
+    const void *data () const { return (const char*)(this + 1); }
 
     /// Handy methods for extracting the underlying parameters as a struct
     template <typename T>
-    const T* as() const { return reinterpret_cast<const T*>(mem); }
+    const T* as() const { return reinterpret_cast<const T*>(data()); }
 
     template <typename T>
-    T* as() { return reinterpret_cast<const T*>(mem); }
+    T* as() { return reinterpret_cast<const T*>(data()); }
 };
 
 
--- a/src/include/OSL/oslconfig.h
+++ b/src/include/OSL/oslconfig.h
@@ -131,4 +131,12 @@ using OIIO::string_view;
 #  define OSL_DEPRECATED(msg)
 #endif
 
+/// Work around bug in GCC with mixed __attribute__ and alignas parsing
+/// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69585
+#ifdef __GNUC__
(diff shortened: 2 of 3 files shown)


fpsunflower:
This seems like the most expedient fix, but it does mean the minimum size of a closure grows to 32 bytes instead of 24 (with 8 bytes being wasted on alignment in a way no one can use, even if they have weaker alignment requirements).

Looking at this again, I think the more correct way to fix this would be require closure structs to inherit from `OSL::ClosurePrimitive`. This should result in the same memory layout, but the compiler would be able to guarantee alignment itself. Then, because we store its `alignof` as well as `sizeof` when we register a closure -- it would be allocated correctly at runtime.

What do you think?  (It's obviously a much more intrusive change in terms of backwards compatibility)

brechtvl:
The size of `ClosureComponent` is reduced to 16 bytes in the updated patch, and then memory after it would be aligned to 16 bytes too.

Inheriting from `OSL::ClosureComponent` is not ideal for SA Arnold, where closures can be allocated by both OSL and C++ shaders. We can probably find some way to do it but it would make things more complicated.

src/include/OSL/oslclosure.h
```diff
@@ -118,25 +118,28 @@ struct OSLEXECPUBLIC ClosureColor {
...
-    /// Handy method for getting the parameter memory as a void*.
-    ///
-    void *data () { return &mem; }
-    const void *data () const { return &mem; }
+    /// Handy method for getting the parameter memory as a void*. Allocation
+    /// allocation will be scaled to requirements of the registered closure
+    /// struct. Alignment is forced to 16 bytes to match alignment requirements
+    /// of 64 bit pointers and 128 bit SSE types in user's structs.
+    static const size_t alignment = 16;
+    static size_t aligned_size() {
```
fpsunflower: I don't think this is needed, right ? If a type has `alignas(N)` then its `sizeof()` will be at least `N` (so that an array of said struct would have properly aligned entries).

src/include/OSL/oslclosure.h
```diff
@@ -118,25 +118,28 @@ struct OSLEXECPUBLIC ClosureColor {
...
+        return (sizeof(ClosureComponent) + alignment - 1) & ~(alignment - 1);
+    }
+
+    void *data () { return (char*)this + aligned_size(); }
```
fpsunflower: `(char*) (this + 1)` should work too I think.

fpsunflower:
This is even easier. I like the patch (modolo the notes above).

fpsunflower:
Longer term, letting user structs inherit from `OSL:ClosurePrimitive` does seem a bit more sound, though having reviewed the code - it does require a bunch of changes. We don't currently store `alignof` the various types like I thought we did.

brechtvl:
Simplified the code now as suggested.

I also added a workaround for a GCC bug, to fix the CI failures.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69585

src/include/OSL/oslclosure.h
```diff
@@ -118,25 +118,23 @@ struct OSLEXECPUBLIC ClosureColor {
...
-                                ///  of the user's structs.
 
-    /// Handy method for getting the parameter memory as a void*.
-    ///
-    void *data () { return &mem; }
-    const void *data () const { return &mem; }
+    /// Handy method for getting the parameter memory as a void*. Allocation
+    /// allocation will be scaled to requirements of the registered closure
+    /// struct. Alignment is forced to 16 bytes to match alignment requirements
+    /// of 64 bit pointers and 128 bit SSE types in user's structs.
```
fpsunflower: 64 bit pointers have `alignof == 8`

src/include/OSL/oslclosure.h
```diff
@@ -118,25 +118,23 @@ struct OSLEXECPUBLIC ClosureColor {
...
-    alignas(void*) char mem[8]; ///< Memory for the primitive, 8 is the minimum, allocation will be
-                                ///  scaled to requirements of the registered closure struct.
-                                ///  Alignment is forced to a void* to match likely alignement requirements
-                                ///  of the user's structs.
 
-    /// Handy method for getting the parameter memory as a void*.
-    ///
-    void *data () { return &mem; }
-    const void *data () const { return &mem; }
+    /// Handy method for getting the parameter memory as a void*. Allocation
```
fpsunflower: Typo: "Actual allocation" ?

src/include/OSL/oslconfig.h
```diff
@@ -131,4 +131,11 @@ using OIIO::string_view;
 #  define OSL_DEPRECATED(msg)
 #endif
 
+// Work around bug in GCC with mixed __attribute__ and alignas parsing
```
fpsunflower: Maybe leave a link to the gcc bug you found? (In the distant future maybe we'll be able to re-evaluate if its still needed)

brechtvl:
More detailed comments added now.

fpsunflower:
LGTM

lgritz:
LGTM2. Merging.

brechtvl:
Can this be backported to RB-1.9? We need it for Cycles.

lgritz:
yes, running tests then will merge it

lgritz:
merged into RB-1.9

PR #812, opened 2017-11-15
oslc: Fix param write analysis bug for token/value optional arguments

lgritz:
Functions that take optional argument lists in token/value form, like `texture` or `pointcloud_search` needed to mark just those value references as "writeable" that are (for example, the one following "alpha" for texture calls, but all of the optional value slots for pointcloud_search).

The logic on the oslc side here was broken -- it worked when the token was a string literal, but when it was a variable whose value wouldn't be known until runtime, it was failing to mark it as writeable, whereas the correct thing to do is to conservatively conclude that it *might* be written to (we don't know yet) and mark it as such.

The cascade of errors could lead to incorrect runtime optimization, if the only used output of the function call was one of these parameters that failed to be marked as an output, so so the whole function call might be subject to the "useless op elision" optimization wherein ops whose writes are never subsequently read can be eliminated entirely.

### DIFF
--- a/src/liboslcomp/typecheck.cpp
+++ b/src/liboslcomp/typecheck.cpp
@@ -946,9 +946,17 @@ ASTfunction_call::mark_optional_output (int firstopt, const char **tags)
        ASTNode *s = argvec[a].get();
        bool isoutput = false;
        // compare against output tags
-       if (s->typespec().is_string() && s->nodetype() == ASTNode::literal_node) {
-           for (const char **tag = tags; *tag && !isoutput; ++tag)
-               isoutput = isoutput || mark_all || (! strcmp (((ASTliteral *)s)->strval(), *tag));
+       if (s->typespec().is_string()) {
+           if (s->nodetype() == ASTNode::literal_node) {
+               // If the token is a string literal, see if it's one of the
+               // ones designated as an output slot.
+               for (const char **tag = tags; *tag && !isoutput; ++tag)
+                   isoutput = isoutput || mark_all || (! strcmp (((ASTliteral *)s)->strval(), *tag));
+           } else {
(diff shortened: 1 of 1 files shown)


src/liboslcomp/typecheck.cpp
```diff
@@ -946,9 +946,17 @@ ASTfunction_call::mark_optional_output (int firstopt, const char **tags)
...
+       if (s->typespec().is_string()) {
+           if (s->nodetype() == ASTNode::literal_node) {
+               // If the token is a string literal, see if it's one of the
+               // ones designated as an output slot.
+               for (const char **tag = tags; *tag && !isoutput; ++tag)
+                   isoutput = isoutput || mark_all || (! strcmp (((ASTliteral *)s)->strval(), *tag));
+           } else {
+               // If the token is not a literal, we don't know what it'll
+               // be at runtime, so mark it conservatively as possible output.
+               isoutput = true;
```
fpsunflower: Is `pointcloud_search` the only function affected? Or were other ones buggy as well?

Are there any functions that take optional args but don't have any optional outputs?

lgritz: They were all buggy in the same way. It could easily(*) have affected a texture() (and environment(), etc.) call if the "alpha" or "errormessage" optional parameter used something other than a string literal as its token.

But "easily" should be taken with a grain of salt -- it's a bit of a weird edge case for those not to be string literals. It's possible, but very unusual. So I'm not tempted to fret over the possibility that there might be values conservatively marked as "potentially written" that in fact are not.

The mark_optional_output function in question is only called for texture (family) and pointcloud_get/search, both of which have this property of potentially writeable variable references.

fpsunflower:
LGTM!

PR #813, opened 2017-11-16
oslc internals: Police up some uses of stderr -- use the error handler

lgritz:
Doesn't matter much from the command line, but when compiling in memory
with OSLCompiler, you don't want the errors leaking into stderr.

Sorry, a few tab/space changes snuck into the diff.


fpsunflower:
LGTM

PR #816, opened 2017-11-16
ShadingSystem option "allow_shader_replacement"

lgritz:
Default is 0, for old behavior in which it's an error for
LoadMemoryCompiledShader to try to load the same named shader more than
once.

If set to nonzero, this is allowed, replacing the old shader master.
This does not change any already declared shader groups, which
will continue to use the original definition at the time of their
declaration. But subsequent groups that reference the master will get
the latest one.


aconty:
LGTM

PR #817, opened 2017-11-17
Bump minimum OIIO dependency to 1.8 (master only)

lgritz:
OSL 1.9 will of course remain compatible with OIIO 1.7.


lgritz:
No objections -->  merge

This only affects master. The 1.9 release will continue to support OIIO 1.7.

PR #818, opened 2017-11-24
Various windows fixes.

LazyDodo:
-Fix compilation issue in dual.h with msvc2015 (constexpr not initializing all members is not allowed)
-Split llvm's platform libraries from a string into a list to prevent the linker seeing it as a single file to look for.
-Enable the use of clang instead of boost::wave for msvc
-Allow clang and llvm libraries and includes to live in a different directory.

## Description

Bumped into these while integrating into blender. 

## Checklist:

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -68,6 +68,7 @@ find_package (LLVM 3.9 REQUIRED)
 
 # ensure include directory is added (in case of non-standard locations
 include_directories (BEFORE SYSTEM "${LLVM_INCLUDES}")
+include_directories (BEFORE SYSTEM "${CLANG_INCLUDES}")
 link_directories ("${LLVM_LIB_DIR}")
 # Extract and concatenate major & minor, remove wayward patches,
 # dots, and "svn" or other suffixes.
@@ -109,6 +110,9 @@ if (GCC_VERSION)
     if (${GCC_VERSION} VERSION_LESS 4.9)
         set (_CLANG_PREPROCESSOR_CAN_WORK ON)
     endif ()
+endif ()
+if (MSVC)
+    set (_CLANG_PREPROCESSOR_CAN_WORK ON)
 endif ()
 if (USE_BOOST_WAVE OR (NOT CLANG_LIBRARIES)
     OR (NOT _CLANG_PREPROCESSOR_CAN_WORK))
--- a/src/include/OSL/dual.h
+++ b/src/include/OSL/dual.h
@@ -61,7 +61,7 @@ public:
 
     /// Default ctr leaves everything uninitialized
     ///
-    constexpr Dual () { }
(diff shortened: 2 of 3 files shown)


src/include/OSL/dual.h
```diff
@@ -61,7 +61,7 @@ class Dual {
 
     /// Default ctr leaves everything uninitialized
     ///
-    constexpr Dual () { }
+    OIIO_CONSTEXPR14 Dual () { }
```
lgritz: THAT'S a weird one! 

Do you have any idea why MSVC would think that an empty ctr isn't C++11-constexpr compliant?

lgritz:
LGTM. I'll merge once the CI tests finish.

Can I trouble you for a CLA?

src/include/OSL/dual.h
LazyDodo: known issue in msvc2015, the release notes *1 said they wanted to fix it in a future update, but it doesn't look like they ever did (i build with 2015 update 5) haven't tested with msvc2017 yet. 

*1 https://blogs.msdn.microsoft.com/vcblog/2015/04/29/c11-constant-expressions-in-visual-studio-2015-rc/

LazyDodo:
For a significant contribution I'd happily sign the CLA, for a couple of one liners fixing the build it sounds like a hassle.

lgritz:
The hard requirement to have all contributors sign a CLA is very nearly the only thing my employers (the sponsors of this project) have asked from us, in exchange for complete freedom to run this project 100% in public and for the benefit of the whole industry.

It's not a hassle... just print the CLA, sign it, take a picture with your phone, and email it to opensource@imageworks.com. Two minutes, tops.

src/cmake/externalpackages.cmake
```diff
@@ -68,6 +68,7 @@ find_package (LLVM 3.9 REQUIRED)
 
 # ensure include directory is added (in case of non-standard locations
 include_directories (BEFORE SYSTEM "${LLVM_INCLUDES}")
+include_directories (BEFORE SYSTEM "${CLANG_INCLUDES}")
```
lgritz: This line failed in the CI tests.
CMake doesn't like it if the string is empty?

LazyDodo:
The problem is not how easy it is, or how much time it'll take, it's the fact i'm signing a legal document with my personal information on it, while not knowing how this will be used, where it will be stored and how my personal information will be protected from abuse. With the recent spike in almost weekly breaches I'm just very cautious handing out this information.  I Personally don't feel I should give up this information just to get in what is essentially a build fix, sadly, for these reasons I respectfully have to retract this PR.

lgritz:
To answer your questions: it's not public, only kept on file by the legal department of Sony Pictures Imageworks. The only personal information on it is your email and address. There are no passwords, access codes, government identity numbers, or financial information of any kind. It's not clear to me how it can be abused.

PR #821, opened 2017-11-27
OSLQuery improvements

lgritz:
* OSLQuery::Parameter -- add copy/move constructors

* Make the ShadingSystem side of OSLQuery correctly report default vals.

### DIFF
--- a/src/include/OSL/oslquery.h
+++ b/src/include/OSL/oslquery.h
@@ -61,12 +61,12 @@ public:
     struct Parameter {
         ustring name;                    ///< name
         TypeDesc type;                   ///< data type
-        bool isoutput;                   ///< is it an output param?
-        bool validdefault;               ///< false if there's no default val
-        bool varlenarray;                ///< is it a varying-length array?
-        bool isstruct;                   ///< is it a structure?
-        bool isclosure;                  ///< is it a closure?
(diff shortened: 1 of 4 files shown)


src/include/OSL/oslquery.h
```diff
@@ -75,10 +75,10 @@ class OSLQUERYPUBLIC OSLQuery {
         std::vector<ustring> fields;     ///< Names of this struct's fields
         ustring structname;              ///< Name of the struct
         std::vector<Parameter> metadata; ///< Meta-data about the param
-        Parameter ()
-            : isoutput(false), validdefault(false), varlenarray(false),
-              isstruct(false), isclosure(false), data(NULL)
-        { }
+
+        Parameter () {}
```
fpsunflower: Is there any need to declare these? Don't you just want the "default" behavior?

lgritz: For the copy and move constructors, it's important that the `data` pointer refers to ITS OWN default field, not a straight pointer copy from the source. So default ctrs won't work.

lgritz: ok with this?

fpsunflower: Ah I missed that detail.

LGTM!

PR #822, opened 2017-11-27
Fix to compile OSL debug against OIIO-1.7.x.

markfinal:
## Description

Fixes edge case of building debug OSL 1.9 against OIIO 1.7. Affected Windows, Linux and macOS, as tested.
OIIO 1.8's atomic.h unconditionally includes <atomic>, and defines Strutil::printf.

I wasn't sure of the best alternative to Strutil::printf in OIIO 1.7, so went with just using printf as a fallback.
Also, I couldn't see a macro to generate an OIIO version number as an integer, so went for duplicating what's in oiioversion.h.in.

I haven't seen any replies to my post on OSL developers to see if anyone would use this combination, but I thought I'd submit this for completeness.

## Tests

Built debug on all three platforms. Tested render-cornell.

## Checklist:

- [N/A] I have updated the documentation, if applicable.

### DIFF
--- a/src/liboslcomp/ast.cpp
+++ b/src/liboslcomp/ast.cpp
@@ -30,6 +30,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <string>
 #include <sstream>
 #include <functional>
+#ifndef NDEBUG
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM, thanks, I will merge.

PR #823, opened 2017-11-29
Expose optional "anisotropic" parameter to the texture funtions in OSL.

lgritz:
This gives explicit control to the amount of anisotropy.  This consol
existed in the underlying OpenImageIO texture lookup options, but never
was exposed to direct control by the OSL shaders.


fpsunflower:
I think this one slipped through the cracks?

LGTM :)

fpsunflower:
Is this still something we want to integrate? Or should we drop it?

PR #825, opened 2017-12-01
VisualStudio 2015 compilation fix for 1.9

markfinal:
## Description

Occurred with the latest OSLQuery::Parameter changes.

The compilation error is

```
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\xutility(2316): error C2280: 'OSL_v1_9::OSLQuery::Parameter &OSL_v1_9::OSLQuery::Parameter::operator =(const OSL_v1_9::OSLQuery::Parameter &)': attempting to reference a deleted function
D:\dev\Thirdparty\OSL-1.9\src\include\OSL/oslquery.h(82): note: compiler has generated 'OSL_v1_9::OSLQuery::Parameter::operator =' here
```

The reason is because there is an explicit move constructor for
Parameter now, meaning an implicit assignment operator is deleted. See http://en.cppreference.com/w/cpp/language/copy_assignment

## Tests

Just tested compilation, but verified that the change is ok on my macOS and Linux builds.

## Checklist:

### DIFF
--- a/src/include/OSL/oslquery.h
+++ b/src/include/OSL/oslquery.h
@@ -79,6 +79,10 @@ public:
         Parameter () {}
         Parameter (const Parameter& src);
         Parameter (Parameter&& src);
+        // because a move constructor is defined, assignment operators will be implicitly deleted
+        // which is a problem for use in std::vectors
+        Parameter &operator=(const Parameter &) = default;
+        Parameter &operator=(Parameter &&) = default;
     };
 
     OSLQuery ();


src/include/OSL/oslquery.h
```diff
@@ -79,6 +79,10 @@ class OSLQUERYPUBLIC OSLQuery {
         Parameter () {}
         Parameter (const Parameter& src);
         Parameter (Parameter&& src);
+        // because a move constructor is defined, assignment operators will be implicitly deleted
+        // which is a problem for use in std::vectors
+        Parameter &operator=(const Parameter &) = default;
```
lgritz: Unfortunately, this is incorrect. The default assignment is wrong for the same reason we can't rely on the default copy constructor or mover.

It's my fault. I'll fix this for real and submit a patch.

lgritz:
Closing because this is incorrect. Sit tight, I will post a correct patch in a few minutes.

PR #827, opened 2017-12-01
Add missing OSLQuery::Parameter correct assignment operator

lgritz:
@markfinal, I think it needed full implementations in order to properly set the `data` field.


markfinal:
I did wonder if my PR might not be sufficient using the compiler defaults for the assignment operator.

I've patched in this PR, and test built on VisualStudio 2015, and all looks good in the build of OSL 1.9.

Thanks for the update!

PR #829, opened 2017-12-06
Pass the context's texture_thread_info to the render services texture callbacks instead of always passing NULL

fpsunflower:
The code was all prepared to do this, but for some reason we always passed `NULL`. This means the carefully pre-generated texture thread handles were not getting used.


lgritz:
oops

LGTM!

PR #830, opened 2017-12-07
Enable symbols in the JIT for the Intel Vtune profiler

lgritz:
This also bumps the version of llvm we're using at SPI, that part won't affect anyone else.


src/liboslexec/llvm_util.cpp
```diff
@@ -449,6 +450,17 @@ LLVM_Util::make_jit_execengine (std::string *err)
     if (! m_llvm_exec)
         return NULL;
 
+    // These magic lines will make it so that enough symbol information
```
fpsunflower: Well that's easy :+1: 

Let me get a benchmark run setup with 5.0.0 vs this patch.

fpsunflower: Performance tests are looking good. No measurable change.

LGTM

lgritz: You mean no measurable change on 5.0.1, the one where we built LLVM with the support for this enabled, right?

fpsunflower: Correct.

PR #831, opened 2017-12-08
Some bitcode and symbol debugging aids

lgritz:
* Add a new shadingsys option "llvm_optput_bitcode" that will dump the
bitcode for each group, even if the other debug options aren't turned all
the way on.

* bitcode now dumps in text as well as binary. (It saves a step of running
"llvm-dis" if your goal is just ot see the text in the first place, and also
it helps the llvm bug where sometimes it refuses to work because of mismatched
version strings.)

* Make sure that both in the bitcode output and other debug symbols,
we are more careful about being sure the shader group name is included,
not just the layer name, and also if a layer name is not specified to
ss->Shader(), make one up based on the shader master name and the
layer number.

### DIFF
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -474,7 +474,10 @@ public:
     /// file.  If err is not NULL, errors will be deposited there.
     void write_bitcode_file (const char *filename, std::string *err=NULL);
 
-    /// Convert a function's bitcode to a string.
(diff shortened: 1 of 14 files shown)


fpsunflower:
LGTM

lgritz:
Updated. I managed to break some tests the first time.

fpsunflower:
LGTM

PR #832, opened 2017-12-08
Adjust deprecations

lgritz:
Remove a method that has been deprecated since OSL 1.6, and mark some
methods that were deprecated since 1.8 with the OSL_DEPRECATED warning.


fpsunflower:
LGTM

PR #833, opened 2017-12-12
Fixes to group names and debug bitcode output

lgritz:
* Better guaranteed unique naming of missing names.

* When dumping ll/bc files for debugging to filenames derived from the
  group name, beware group names that have slashes in them! Also,
  sheesh, watch out for group names that are so long that they exceed
  the longest filename allowed (256 chars for most modern filesystems).

* Decided only the ll files are useful for debugging, I don't think we
  need the binary bc files, so I removed the bc output.

* Simplify ShaderGroup ctr by using C++11 data member defaults and
  delegating constructors.


fpsunflower:
LGTM

PR #834, opened 2017-12-18
Assignments to components of fields within a struct array are optimized away by the runtime optimizer

marsupial:
## Description
The following code leaves c[0].rgb[0] in an undefined state when the runtime optimizer is used:

```
struct custom { color rgb; };


custom c[1];
c[0].rgb[0] = 123;
printf("rgb: %g\n", c[0].rgb);
printf("rgb: %g\n", c[0].rgb[0]);

```
Patch add ability to have a test pass who's diff operation fails.
Fixes nothing related to the bug, but adds a test showing the behavior.

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -239,6 +239,7 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             blackbody blendmath breakcont
             bug-array-heapoffsets
             bug-locallifetime bug-outputinit bug-param-duplicate bug-peep
+            bug-struct-assign
             cellnoise closure closure-array color comparison
             compile-buffer
             component-range const-array-params const-array-fill
--- a/testsuite/bug-struct-assign/ref/out.txt
+++ b/testsuite/bug-struct-assign/ref/out.txt
@@ -1,0 +1,3 @@
+Compiled test.osl -> test.oso
+c[0].rgb: 1 2 3
+(-1 2 3)  == (-1 2 3)
--- a/testsuite/bug-struct-assign/run.py
+++ b/testsuite/bug-struct-assign/run.py
@@ -1,0 +1,4 @@
+#!/usr/bin/env python
+
+command = testshade("-g 1 1 test")
+failureok = failstage2 # diff should fail, bug still exists!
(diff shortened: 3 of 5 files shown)


lgritz:
Are you already working on a solution, or just giving us the test and assuming that I'll find the fix?

I'm truly not saying that in an accusatory way... I'm very happy to jump right on fixing this, but I don't want to do the work if it's a waste because you already are in the middle of it but haven't posted a PR yet.

marsupial:
I assume nothing, so have at it!
Was trying to fix it in **olsc**, but didn't get far and realized the runtime optimizer would probably still have to be addressed in case one loaded an **oso** compiled with an older version.

lgritz:
Yeah, I'm happy to take a stab at this.

marsupial:
Good luck. In case it not too obvious already, it looked to be  generating tmps for every access and not getting the aliasing or r/w boundary right.  Which cause the runtime optimizer to think is can be elided as a useless write.

c[0].rgb[0] = 123; ->
```
const	int	$const1	0		%read{0,17} %write{2147483647,-1}
temp	color	$tmp1	%read{2147483647,-1} %write{0,0}
const	color	$const2	1 2 3		%read{1,1} %write{2147483647,-1}
temp	color	$tmp2	%read{3,3} %write{2,2}
temp	float	$tmp3	%read{4,4} %write{3,3}
temp	float	$tmp4	%read{6,6} %write{4,4}
temp	color	$tmp5	%read{2147483647,-1} %write{5,6}
code ___main___
# test.osl:10
#     c[0].rgb = color(1,2,3);
	aref		$tmp1 c.rgb $const1 	%filename{"test.osl"} %line{10} %argrw{"wrr"}
	aassign		c.rgb $const1 $const2 	%argrw{"wrr"}
# test.osl:13
#     c[0].rgb[0] = -c[0].rgb[0];
	aref		$tmp2 c.rgb $const1 	%line{13} %argrw{"wrr"}
	compref		$tmp3 $tmp2 $const1 	%argrw{"wrr"}
	neg		$tmp4 $tmp3 	%argrw{"wr"}
	aref		$tmp5 c.rgb $const1 	%argrw{"wrr"}
	compassign	$tmp5 $const1 $tmp4 	%argrw{"wrr"}
```

PR #836, opened 2017-12-18
Add named component access and swizzling operations to vector types.

marsupial:
## Description
Accessing/reordering channels via rgb/xyz selectors makes for much more readable code.
Also porting GLSL code to OSL can be quite tedious if it make extensive use of swizzling.

## Tests
**testsuite/oslc-err-swizzle**
**testsuite/swizzle**

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,6 +296,7 @@ TESTSUITE ( aastep allowconnect-err and-or-not-synonyms arithmetic
             struct-isomorphic-overload struct-layers
             struct-operator-overload struct-return struct-with-array
             struct-nested struct-nested-assign struct-nested-deep
+            swizzle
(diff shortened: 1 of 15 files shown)


marsupial:
Appreciate the time spent integrating the other PR's over the weeks so if these two are low priority let me know.

You had previously also expressed interest in method call syntax, but less so in swizzling.
Component access was less clear,  but there is a bit of overlap here with #839 as:
(in my opinion) `this.r` is way better than `this[0]` (which [also in my opinion] should not be legal syntax).

lgritz:
Closing. (Sorry, Roman.)

I have an independent and simpler implementation coming soon for the parts of this feature that I like (and omitting the parts I don't like from a language design perspective).

PR #843, opened 2018-01-20
Fix regression where function overload candidates might not be reported

marsupial:
## Description
597b6b1 introduced a bug which might make the error message less verbose than it used to be.
This rectifies that.

## Tests
**oslc-err-funcoverload**

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -267,8 +267,8 @@ TESTSUITE ( and-or-not-synonyms aastep arithmetic array array-derivs array-range
             operator-overloading
             oslc-comma oslc-D
             oslc-err-arrayindex oslc-err-closuremul oslc-err-field
-            oslc-err-format oslc-err-funcredef oslc-err-intoverflow
-            oslc-err-noreturn oslc-err-notfunc
+            oslc-err-format oslc-err-funcoverload oslc-err-funcredef
+            oslc-err-intoverflow oslc-err-noreturn oslc-err-notfunc
             oslc-err-outputparamvararray oslc-err-paramdefault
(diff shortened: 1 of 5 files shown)


lgritz:
Something is wrong, it's failing the CI tests, no longer correctly compiling some of the MaterialX shaders.

marsupial:
Yes, was not using the proper constant.
Running on a separate CI before updating the PR.

marsupial:
Should be good to go.  Was actually trying to fix some other issues that are coming to light, but just reverted to simple fix for missing labels of some ambiguities.

marsupial:
Might be prudent to revert #837 (597b6b1).
Not sure how it got missed, but I'm seeing some behavior that is definitely incorrect.

lgritz:
So, just to clarify:

You think we should revert #837, but keep this one #843?

lgritz:
Or did you mean we should put both on hold until you sort it out?

marsupial:
I was thinking reverting #837 and rolling this (#843) into either a new PR or re-open #837.
If you're more interested in not reverting and moving forward with fixing the original problem I'll have a few more hours today and tomorrow (which should be enough to nail it down), but a back and forth might be necessary.

I'll do a further writeup of the problems I'm discovering in the next hour.

lgritz:
OK, since we know there's something wrong, I'll revert #837 in master, and you can close this PR, and then when you have it totally worked out, just post a new combined PR (or re-open 837). There's no rush on this, we've obviously lived with the existing ambiguous edge cases for a long time with nobody complaining.

PR #844, opened 2018-01-21
Fix poorly defined behavior when choosing certain function overloads.

marsupial:
## Description

OSL currently does not define how function overloads are chosen very well.
As such oslc can choose different overloads depending on the order of declaration.
This attempts to define how overloads are chosen in a more concrete way by scoring:

all arguments match without conversion
prefer spatial -> spatial conversion over spatial <-> color
prefer triple -> triple conversion over float -> triple promotion
use return type to resolve any possible ambiguities
more erring for unresolvable ambiguities

Adds more verbose warning/error messages showing location of ambiguous functions.
Adds **OSL_LEGACY_FUNCTION_RESOLUTION** environment variable so that changes in behavior can be ignored, reported, or reverted.

## Tests
**testsuite/function-overloads
testsuite/oslc-err-funcoverloads**

## Checklist:

### DIFF
--- a/src/liboslcomp/typecheck.cpp
+++ b/src/liboslcomp/typecheck.cpp
@@ -1160,14 +1160,23 @@ ASTfunction_call::typecheck_struct_constructor ()
 ///
 /// Score a set of polymorphic functions based on arguments & return type.
 ///
-/// When choosing which varaint to use precedence is given to arguments and the
-/// return type is used as a possible tie breaker in case of any ambiguity.
-/// The highest score for an argument is given when it is an exact match.
-/// If the argument needs to be coerced into a type OSL will prefer going
-/// from any triple-types to any other triple-type before promotion of a
-/// float to a triple.  When converting one triple to another triple, OSL will
-/// prefer to coerce from one of the spatial triples (vector/point/normal) to
-/// another spatial triple before coercion from a spatial triple to color.
+/// The idea is that every function with the same name (visible from the scope)
+/// is evaluated and 'scored' against the actual arguments.
+///
+/// An exact match of all arguments will have the highest score and be chosen.
+/// If there is no exact match, then each possible overload is given a score
+/// based on the number and type of substitutions or coercions they require.
+///
+/// Different types of coercions having different costs so that: int to float is
+/// scored relatively high, beating out all other coercions; spatial-triple
+/// to spatial-triple is a closer match than spatial-triple to color;
+/// and triple to triple is a closer match than float to triple.
+///
+/// If a single choice has a best score, it wins.
+/// If there is a tie (and only then), the return type is considered in the score.
+/// If there is still not a single winner using the return type, it is considered
+/// an error whose message will show all the high scoring possibilities that
+/// cannot be dis-ambiguated.
 ///
 /// Float to int coercion is scored, but is currently a synmonym for kNoMatch
 /// as the spec does not allow implicit float to int conversion.
--- a/src/liboslcomp/typecheck.cpp
+++ b/src/liboslcomp/typecheck.cpp
@@ -1180,9 +1180,11 @@ ASTfunction_call::typecheck_struct_constructor ()
 ///
 /// If a single choice has a best score, it wins.
 /// If there is a tie (and only then), the return type is considered in the score.
-/// If there is still not a single winner using the return type, it is considered
-/// an error whose message will show all the high scoring possibilities that
-/// cannot be dis-ambiguated.
+/// If there is still not a single winner then a function is chosen by ranking
+/// the possible return types, using the following precedence:
+///   float, int, color, vector, point, normal, matrix, string
+/// A warning is shown, printing which was function chosen and the list of all
+/// that were considered ambiguous.
 ///
 /// Float to int coercion is scored, but is currently a synmonym for kNoMatch
 /// as the spec does not allow implicit float to int conversion.
@@ -1220,6 +1222,8 @@ class CandidateFunctions {
     TypeSpec m_rval;
     ASTNode::ref m_args;
     size_t m_nargs;
+
+    FunctionSymbol* m_called; // Function called by name (can be NULL!)
 
     const char* scoreWildcard(int& argscore, size_t& fargs, const char* args) const {
         while (fargs < m_nargs) {
@@ -1352,9 +1356,15 @@ class CandidateFunctions {
         return argscore;
     }
 
+    void candidateHeader(const char* msg) const {
+        m_compiler->errhandler().message("  %s:\n", msg);
+    }
+
 public:
-    CandidateFunctions(OSLCompilerImpl* compiler, TypeSpec rval, ASTNode::ref args, FunctionSymbol* func) :
-        m_compiler(compiler), m_rval(rval), m_args(args), m_nargs(0) {
+    CandidateFunctions(OSLCompilerImpl* compiler, TypeSpec rval, ASTNode::ref args,
+                       FunctionSymbol* func) :
+        m_compiler(compiler), m_rval(rval), m_args(args), m_nargs(0),
+        m_called(func) {
 
         //std::cerr << "Matching " << func->name() << " formals='" << (rval.simpletype().basetype != TypeDesc::UNKNOWN ?  compiler->code_from_type (rval) : " ");
         for (ASTNode::ref arg = m_args; arg; arg = arg->next()) {
@@ -1371,27 +1381,28 @@ public:
         }
     }
 
-    void reportError(ASTNode* caller,
-                     std::string name = "", bool showArgs = true,
-                     bool candidateMsg = true,
-                     const char* msg = "No matching function call to") const {
+    void reportAmbiguity(ASTNode* caller, const ustring& funcname,
+                         bool candidateMsg = true,
+                         const char* msg = "No matching function call to",
+                         bool asWarning = true, bool showArgs = true) const {
+        std::string argstr = funcname.string();
         if (showArgs) {
-            name += " (";
+            argstr += " (";
             const char *comma = "";
             for (ASTNode::ref arg = m_args; arg; arg = arg->next()) {
-                name += comma;
-                name += arg->typespec().string();
+                argstr += comma;
+                argstr += arg->typespec().string();
                 comma = ", ";
             }
-            name += ")";
+            argstr += ")";
         }
-
-        caller->error ("%s '%s'", msg, name);
+        asWarning ? caller->warning ("%s '%s'", msg, argstr)
+                  : caller->error ("%s '%s'", msg, argstr);
         if (candidateMsg)
-            m_compiler->errhandler().message("  Candidates are:\n");
+            candidateHeader("Candidates are");
     }
 
-    void reportAmbiguity(FunctionSymbol* sym) const {
+    void reportFunction(FunctionSymbol* sym) const {
         int advance;
         const char *formals = sym->argcodes().c_str();
         TypeSpec returntype = m_compiler->type_from_code (formals, &advance);
@@ -1408,10 +1419,25 @@ public:
                      m_compiler->typelist_from_code(formals).c_str());
     }
 
-    std::pair<FunctionSymbol*, TypeSpec> best(ASTNode* caller, bool strict = 0) {
+    std::pair<FunctionSymbol*, TypeSpec>
+    best(ASTNode* caller, const ustring& funcname) {
         switch (m_candidates.size()) {
-            case 0:  return { nullptr, TypeSpec() };
-            case 1:  return { m_candidates[0].sym, m_candidates[0].rtype };
+            case 0:
+                // Nothing at all, Error
+                // If m_called is 0, then user tried to call an undefined func.
+                // Might be nice to fuzzy match funcname against m_compiler->symtab()
+                reportAmbiguity(caller, funcname,
+                                m_called != nullptr /*Candidate Msg?*/,
+                                "No matching function call to",
+                                false /*As warning*/);
+                for (FunctionSymbol* f = m_called; f; f = f->nextpoly())
+                    reportFunction(f);
+
+                return { nullptr, TypeSpec() };
+
+            case 1: // Success
+                return { m_candidates[0].sym, m_candidates[0].rtype };
+
             default: break;
         }
 
@@ -1426,17 +1452,64 @@ public:
                 ambiguity = candidate.rscore;
         }
 
-        if ((ambiguity != -1) || strict) {
+        ASSERT (c.first && c.first->sym);
+
+        if (ambiguity != -1) {
             ASSERT (caller);
-            reportError(caller, m_candidates[0].name(), false,
-                        !m_candidates.empty(), "Ambiguous call to");
+            if (true /*m_rval.simpletype().is_unknown()*/) {
+                // Ambiguity because the return type desired is unknown
+                //   float noise(point p)
+                //   color noise(point p);
+                //   float mix(color a, color b, float mx);
+                //   color mix(color a, color b, color mx);
+                //   mix(c0, c1, noise(P));
+
+                // Sort m_candidates, so the ranking code can be much more
+                // legible, and ambiguities will be reported in order they
+                // would be chosen.
+                std::sort(m_candidates.begin(), m_candidates.end(),
+                    [](const Candidate& a, const Candidate& b) -> bool {
+                        auto rank = [](const TypeSpec& s) -> int {
+                            if (s == TypeDesc::TypeFloat)
+                                return 0;
+                            if (s == TypeDesc::TypeInt)
+                                return 1;
+                            if (s == TypeDesc::TypeColor)
+                                return 2;
+                            if (s == TypeDesc::TypeVector)
+                                return 3;
+                            if (s == TypeDesc::TypePoint)
+                                return 4;
+                            if (s == TypeDesc::TypeNormal)
+                                return 5;
+                            if (s == TypeDesc::TypeMatrix)
+                                return 6;
+                            if (s == TypeDesc::TypeString)
+                                return 7;
+                            ASSERT (0 && "Unreachable");
+                            return std::numeric_limits<int>::max();
+                        };
+                        return rank(a.rtype.simpletype())
+                                 < rank(b.rtype.simpletype());
+                    });
+
+                // New choice is now front of the list
+                c = std::make_pair(&m_candidates.front(),
+                                   m_candidates.front().rscore);
+            }
+
+            reportAmbiguity(caller, funcname, false, "Ambiguous call to");
+
+            candidateHeader("Chosen function is");
+            reportFunction(c.first->sym);
+
+            candidateHeader("Other candidates are");
             for (auto& candidate : m_candidates) {
-                if (candidate.rscore >= ambiguity)
-                    reportAmbiguity(candidate.sym);
+                if (candidate.sym != c.first->sym)
+                    reportFunction(candidate.sym);
             }
         }
 
-        ASSERT (c.first);
         return {c.first->sym, c.first->rtype};
     }
 
@@ -1557,7 +1630,7 @@ ASTfunction_call::typecheck (TypeSpec expected)
     FunctionSymbol* poly = func();
 
     CandidateFunctions candidates(m_compiler, expected, args(), poly);
-    std::tie(m_sym, m_typespec) = candidates.best(this);
+    std::tie(m_sym, m_typespec) = candidates.best(this, m_name);
 
     // Check resolution against prior versions of OSL
     static const char* OSL_LEGACY = ::getenv("OSL_LEGACY_FUNCTION_RESOLUTION");
@@ -1571,10 +1644,10 @@ ASTfunction_call::typecheck (TypeSpec expected)
 
             m_compiler->errhandler().message ("  Current overload is ");
             !m_sym ? m_compiler->errhandler().message("<none>")
-                   : candidates.reportAmbiguity(static_cast<FunctionSymbol*>(m_sym));
+                   : candidates.reportFunction(static_cast<FunctionSymbol*>(m_sym));
             m_compiler->errhandler().message ("  Prior overload was ");
             !legacy ? m_compiler->errhandler().message("<none>")
-                       : candidates.reportAmbiguity(legacy);
+                       : candidates.reportFunction(legacy);
         
             if (strcasecmp(OSL_LEGACY, "use") == 0)
                 m_sym = legacy;
@@ -1593,19 +1666,6 @@ ASTfunction_call::typecheck (TypeSpec expected)
             typecheck_builtin_specialcase ();
         }
         return m_typespec;
-    }
-
-    // Ambiguity has already been reported.
-    if (!candidates.empty())
-        return TypeSpec();
-
-    // Couldn't find any way to match any polymorphic version of the
-    // function that we know about.  OK, at least try for helpful error
-    // message.
-    candidates.reportError(this, m_name.string(), true, poly != nullptr);
-    while (poly) {
-        candidates.reportAmbiguity(poly);
-        poly = poly->nextpoly();
     }
 
     return TypeSpec();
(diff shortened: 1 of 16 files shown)


lgritz:
I like the way this is going, it looks very good, but I have a couple requests.

1. First, can you write up a paragraph that in plain English spells out what you believe are the correct rules about the polymorphic coercion priorities, and what constitutes an ambiguity that will be an error? This is both so I can document it properly in the spec, but also so that I can more carefully review your code and try to ensure that it's doing what is intended.

2. When it couldn't find a match before, it would print an error like the following:
```
test.osl:11: error: No matching function call to 'funca ()'
    Candidates are:
        void funca (int, int)
        void funca (int)
```

I'd like to stick to this format if possible, I think it's extremely readable. In contrast, yours looks like:
```
test.osl:11: error: No matching function call to 'funca ()'
test.osl:3 candidate function:
    void funca (int, int)
test.osl:2 candidate function:
    void funca (int)
```

I find that (especially when there is a string of them) rather difficult to visually parse. Although I do  very much like that your way shows the files and lines. Maybe we can combine the two to make something even better?  Like this:
```
test.osl:11: error: No matching function call to 'funca ()'
  Candidates are:
    void funca (int, int)    [test.osl:3]
    void funca (int)    [test.osl:2]
```

marsupial:
Have no major preference, but I'm finding the following most readable, as using a tab does some easy column formatting and personally find the location/line number more important.
```
testsuite/oslc-err-funcoverload/test.osl:11: error: No matching function call to 'funca ()'
  Candidates are:
    testsuite/oslc-err-funcoverload/test.osl:3	void funca (int, int)
    testsuite/oslc-err-funcoverload/test.osl:2	void funca (int)
```

Let me know if you want it the way you suggest and whether to merge into the original commit (which will change the SHA1s) or just add the change on top.

lgritz:
Yes, that looks good.

I don't have a preference for merge (you mean amend?) or add on top. It's all the same to me in this case, since I will do a "squash and merge" to bring it into the master as a single commit.

marsupial:
Ok.  I'll just squash/rebase your requests into the first commit.
Thought maybe during review the SHA1's might be noted for validation/certification later.

marsupial:
Updated.
Tried to add the documentation requested in the 'docdeep' format...

src/liboslcomp/typecheck.cpp
```diff
@@ -1180,58 +1156,422 @@ ASTfunction_call::typecheck_struct_constructor ()
...
+/// <doc LegacyOverload>
+///
+/// Check how a polymorphic function variant was chosen in prior versions.
+/// Check is performed based on the env variable OSL_LEGACY_FUNCTION_RESOLUTION
+///
+/// OSL_LEGACY_FUNCTION_RESOLUTION      // check resolution matches old behavior
+/// OSL_LEGACY_FUNCTION_RESOLUTION=0    // no checking
+/// OSL_LEGACY_FUNCTION_RESOLUTION=err  // check and error on mismatch
+/// OSL_LEGACY_FUNCTION_RESOLUTION=use  // check and use prior on mismatch
+/// </script>
```
lgritz: You don't have to worry about the docdeep format for internals, that's just for certain external APIs that we want to make doc web pages about.

You don't want the `</script>` or `~~~~`. Those do something different than you think; they were specifically for fully quoting large code sections that followed.

src/liboslcomp/typecheck.cpp
```diff
@@ -1180,58 +1156,422 @@ ASTfunction_call::typecheck_struct_constructor ()
...
+    }
+
+    // Save the currently choosen symbol for error reporting later
+    FunctionSymbol* poly = func();
+
+    CandidateFunctions candidates(m_compiler, expected, args(), poly);
+    std::tie(m_sym, m_typespec) = candidates.best(this);
+
+    // Check resolution against prior versions of OSL
+    const char* OSL_LEGACY = ::getenv("OSL_LEGACY_FUNCTION_RESOLUTION");
```
lgritz: Is it weird or expensive to do this getenv and evaluation for every function call we typecheck? Maybe it's something we should do once in the OSLCompilerImpl at startup time?

Or maybe it doesn't matter because I think we will probably remove it soon. It's primarily so that we don't have to revert again if a minor problem is found.

lgritz:
Doc/intent clarification: So, the idea is that each function with that name (visible from the scope) is evaluated and scored against the actual arguments. An exact match is chosen automatically. If there is no exact match, then each possible overload is given a "score" based on the number and types of type of substitutions or coercions they require (with different types of coercions having different costs). If a single choice has a best score, it wins. If there is a tie (and only then), the return type is considered in the score. And then, if there is still not a single winner, it is considered an ambiguous case and an error message is printed showing all the possibilities it was choosing among.

Do I have that right?

And multiple coercions are just summed? So, let's say you have

    float func (float, color);
    float func (color, float);

and you call `func (int, int)`, that would be the error for ambiguous case, because both possible choices have the same score (each needs two substitutions, one is int->float and one is int->color)?

And as far as the cost of coercions, it's weighted so that spatial triple vs spatial triple is a closer match than spatial triple vs color, and that triple-to-triple is a closer match than float-to-triple, and int-to-float is even closer?

Am I getting it all right now?

src/liboslcomp/typecheck.cpp
```diff
@@ -1180,58 +1156,422 @@ ASTfunction_call::typecheck_struct_constructor ()
...
+/// another spatial triple before coercion from a spatial triple to color.
+///
+/// Float to int coercion is scored, but is currently a synmonym for kNoMatch
+/// as the spec does not allow implicit float to int conversion.
+///
+/// </script>
+/// ~~~~
+class CandidateFunctions {
+    enum {
+        kExactMatch     = 100,
```
lgritz: So higher scores are closer matches? Do you think that for future proofing (imagine a future with default arguments or ... style arguments) it might be better to have it the other way around? In other words, if 0 was perfect match, and each substitution had a "cost", summing the cost over the arguments? Because as you have it now (if I understand properly), you are summing score over the argument and aiming for high score, so "more arguments" naturally seems to score higher, whereas if perfect match is 0, then you are more directly optimizing (fewer errors) if you know what I mean.

I don't have a strong feeling about this one way or the other, but just pointing it out in case we want to reverse the sense of it now.

src/liboslcomp/typecheck.cpp
```diff
@@ -1180,58 +1156,422 @@ ASTfunction_call::typecheck_struct_constructor ()
...
+        kIntegralToFP   = 77,
+        kArrayMatch     = 44,
```
lgritz: I'm a little concerned about kArrayMatch, which as I understand it is when the function definition has a formal parameter array[], but the call has a specific length like array[5]. 

It seems to me that the substitution of exact array length for undefined length is *expected* and should have very high score (or low cost). 

As an example, let's say there are 3 parameters, two polymorphic varieties:

    void func (float a[], int b, int c)
    void func (float a[3], float b, float c)

and you call `float x[3]; int i; func (x, i, i);`

The score on the first one is 44+100+100==244, the score on the second is 100+77+77=254. So the second will be chosen. Does that match our intuition? I don't know, maybe it's fine? I'm just trying to find all the little things that might trip us up later. 

I believe that C++ looks for exact match first, then allows certain substitutions, but if more than one function can match with substitutions, it considers it an error (there is no ranking of substitutions like we are trying to do here). I think that's a little rigid and we want more flexibility for OSL, but I do see that they have maybe done a smart thing by not trying adjudicate that. I'm just thinking out loud here. Not advocating more rigidity. But if you wanted to argue that this whole effort is just pushing trouble farther down the road but it's going to come back to haunt us, I would certainly entertain that line of argument.

src/liboslcomp/typecheck.cpp
marsupial: Whatever the cost, you're right paying it multiple times is stupid.
Can easily move it to a static and incur the hit of a lock only?

src/liboslcomp/typecheck.cpp
marsupial: More arguments does lead to a higher score, but having too few or too many will force the score for that variant to 0.

I think scoring needs to go up to support variadic args (or at least it's unclear to me how to do the following in your model)
```
A. void printer(string fmt, ...);
B. void printer(string prefix, string fmt, ...);

printer("FORMAT", 2) // should call A
printer("PREFIX", "FORMAT") // should call B
```

src/liboslcomp/typecheck.cpp
marsupial: ArrayMatch isn't really that low its just exact matches and int->float are extremely high.
The reason the number isn't extremely high is to provide some space trying to figure out how to score an init list in the future.
```
A. void func (vector v, int b, int c);
B. void func (float a[3], int b, int c);

func({1,2,3}, 1, 2);  // Should this call A, B, or is it ambiguous?
```
##
I really like the new feature of preferring spatial->spatial vs spatial->color, past that I have no preference on whether ambiguity is made more strict, but might easily break what is totally legal now:

```
A. float mix(float a, float b, float c);
B. float mix(int   a, int   b, float c);

// Currently legal (calls B), but would become ambiguous
func(1, 2, 1);
```
##
I get the ambiguity of `func(x3, i, i)` you're describing but in the cases providing literals it seems to me quite un-ambiguous what the user is trying to do.
```
A. void func (float a[], int b, int c);
B. void func (float a[3], float b, float c);

int i; float f;

float x5[5];
func (x5, i, i) Should call A.
func (x5, f, f) Should call A.
func (x5, 2, 2) Should call A.

float x3[3];
func (x3, i, i) Should call B.
func (x3, f, f) Should call B.
func (x3, 2, 2) Should call B.
```
Seems to do what I would expect.

marsupial:
Removed incorrect doc tags and made the **getenv** call assign to a static.

You're doc/intent clarification seems accurate to me and if you like the wording better I can move it into **typecheck.cpp**

lgritz:
Yes, if you don't mind, I think it would be helpful to add my explanation to the comments. I think that will help somebody in the future who is coming across this for the first time and trying to discern what it does.

I think the static getenv is an improvement.

src/liboslcomp/typecheck.cpp
lgritz: OK, makes sense to me. We can always revise if we find problems in practice.

I will try these on our production shader library before merging, just to be sure there aren't any problems we haven't seen yet.

lgritz:
If you fail to match at all, you'll get an error message

    test.osl:3: error: No matching function call to 'pnoise (float, float, float, int, int)'
      Candidates are:
          ...

But if you have an ambiguous match, you get

    test.osl:3: error: No matching function call to 'pnoise'
      Candidates are:
          ...

Can you please make the error messages match, i.e. have the ambiguous error also spell out the actual arguments passed? I think that will make it a lot easier for people to track down and fix their errors.

lgritz:
I'm ok with that fixup being a subsequent PR after I merge this.

Let me finish testing what we have so far so I can merge it. Then we'll tackle any leftovers like that error message.

marsupial:
Idea being the ambiguity means there was no exact match so the arguments passed are being coerced into types that don't match what the functions was called with.

If more info is desired in the second case, it is trivial to toggle/remove the **showArgs** argument to **reportError** though.

lgritz:
I just added a commit that helps for an ambiguous case that pops up regularly in our shader library. 

It was figuring out `(vector)ambiguous_function()`, but not `vector(ambiguous_function())`, which I think many users will consider equivalent.

lgritz:
OK, I ran into a problem that I think needs modifications to the scoring system.

The noise and texture functions both come in varieties that differ only by return type, for example `float noise(point)` and `color noise(point)`.

If you have a use like this:

    x = func (noise(u))

Think of how the type checking is proceeding recursively. Basically, `func` will not select its polymorphic variety (which tells it the exact argument types it's receiving) until after it typechecks its arguments, which in this case is a function call to noise(). So when typechecking the noise call, it doesn't know which return type is preferred (because it hasn't typechecked func yet), so it's an ambiguous case EVERY TIME. 

This is a very common idiom, and I dread having every instance of it be an error until people edit it, and even then, the unambiguous formulation with a cast is more verbose than necessary.

I believe that the best (principle of least surprise) solution is to amend the scoring as follows:

*If there is a tie in the argument scoring, with the return type being being used as tie-breaker, but there is ambiguity in the return type also, <b>if one of the return types is float, choose that one</b>*.

That's effectively the behavior we had before, except that we implemented it by exploiting the order-dependency of the declarations, making sure that the float version had highest priority in the ordering.

We don't want order-dependency, but I think we still do want the "in cases of ambiguity, float is preferred over compound types" heuristic somehow.

Make sense? Do you think this is an easily achievable change?

marsupial:
What you're describing seems easy to get to, but would it be worth taking a moment to see if something could be worked out without adding the special case?

marsupial:
Not sure if I buy into the ammendmant.
In the following I think it is reasonable to assume the user is calling the color variants vs. promoting a float to color at the end.
```
float  funcb(float a) { printf("b.float\n"); return a; }
color  funcb(color a) { printf("b.color\n"); return a; }

shader test() {
    float f = funcb(noise(u));
    color c = funcb(noise(u)); // c should be {f,f,f} ?
}
```

[The joys of rediscovering why C++ did certain things!]

lgritz:
You are correct, there are definitely cases where a human would probably choose something other than float, but I fear that to get it "right" will be much more major surgery, and I'm not sure it's worth the effort.

Let's consider a slightly more complex case:

    float f = mix (afloat, bfloat, noise(P))

Previously, the noise declarations were purposely ordered in such a way that `float noise()` would be the first match it would find -- which is the only correct choice in this case. For the color `mix` case would also work if it selected float noise (i.e., `mix(color,color,float)`), though it's possible that the user intended color noise to do a per-channel mix selection. In that case, they would have to notice the wrong behavior and change it to `mix (c1, c2, (color)noise(P))`. But with your patch, it's ambiguous, so it's considered an error and won't compile, and that's going to trip a lot of people up.

But think about what it will take to get this "right." We humans see that the first two parameters to mix are floats, and so of all the umpteen mix varieties, only the one that takes a float for the 3rd argument is a possible match, so obviously the noise variety we want is the one that returns float.

In the current state of the code, the arguments to mix are typechecked recursively before we select the mix variety (the selection itself needs to know the argument types). We'd have to overhaul the scoring in a really nontrivial way to accommodate that the function polymorphism needs to know the argument types, but also the argument types might be selected by the available polymorphism!

And it gets worse. If instead of afloat and bfloat being constants or variables (whose type is discernible without recursive descent), what if they were functions, which themselves were polymorphic in return type?

It gets hairy fast.

So I'm arguing that at least for now, we resolve it by the rule that if the return type hint is Unknown, prefer a variety returning float (if it exists) rather than considering it an error. 

Maybe the more general approach (so we handle point/vector ambiguity as well) would be that when return type is being used as the "tie breaker" (between multiple argument lists being possible matches with identical scores), and multiple return types also might match (including when the return type hint is Unknown), don't consider it an ambiguity failure, always select one, but have a priority order (e.g., float, int, color, vector, point, normal, matrix, string). This is akin to what the old code did, but we'd be selecting based on a documented fixed priority list, rather than the error-prone dependency on order of declarations.

For that matter, I think that even for function argument ambiguity, rather than throw our hands up and error, it would be acceptable to me to instead *always* choose according to a priority list of types, but just have a warning to point out that it chose one of a number of possible matches and why.

lgritz:
The more I think about this, the more I believe that the original sin of the existing implementation was not that it always succeeded when faced with ambiguous choices, but rather that the heuristic was poorly documented and depended on the order of declaration of the polymorphic variants.

I think there is merit to always succeeding when possible, and that a true error should only occur if no match can be found at all. But the big improvements we can make are:
* A formal scoring system that we can document clearly and rely on.
* Declaration order independence.
* Clear warnings when we have chosen among ambiguous cases with highly similar scores.

So maybe instead of this error:
```
test.osl:15: error: Ambiguous call to 'funcb'
  Candidates are:
    test.osl:7	color func ()
    test.osl:7	vector func ()
    test.osl:6	float func ()
```

we should instead aim for a clear warning:

```
test.osl:15: warning: Ambiguous call to 'funcb()'
  Chose:
    test.osl:6	float func ()
  Other candidates were:
    test.osl:7	color func ()
    test.osl:5	vector func ()
  Use explicit type casts for arguments or return types to force a different selection.
```

and much like C++ compilers might have `-Werror`, we can give oslc an option for warnings to be treated as fatal errors, for people who don't want any warning-level ambiguities in their code to succeed.

marsupial:
On board for demotion to warning as long as an option exists for **-Werror**.
___
Feels like an important issue of a pretty subtle issue.
Should this discussion be opened up on OSL-dev to get some other viewpoints or agreements?
As you said the current behavior _is-what-it-is_ so spending a day or two getting it right costs little.
___
I'm still not totally convinced the amendment as proposed is necessarily a good solution. 

Explicitly casting to force a certain behavior is nice; but thinking of a GUI for shaders, seems rough to rely upon for users to get the behavior they want/expect.

I do understand you have a large library you don't want to break/mess with, but carrying that prior behavior forward, rather than taking some pain and dealing with the fact this was ill-defined (possibly even making wrong choices) doesn't seem all that fruitful.

If changing the proposed scoring system to account for final return-type in a more elegant way it's a bigger win, right?

Back to this example:
```
float  funcb(float a) { printf("b.float\n"); return a; }
color  funcb(color a) { printf("b.color\n"); return a; }

shader test() {
    color a = funcb(noise(u)); // c should be {f,f,f} ?
}
```
The prior behavior (and what you would like to set in stone) seems very wrong to me.
In the context of a **VEX** editor how would a user force `funcb((color)noise(u))`?
Why should they have to force it, when it seems obvious it's what they want?

I think getting this right is do-able as the initial type check occurs for the eventual assignment and propagating that forward seems possible (the initializer lists PR had some overlap in this area that was resolved pretty easily).
##
More complex/hairy examples such as `color b = mix (c1, c2, noise(P)); ` truly does seem ambiguous to me.
Reading it I have no idea what to expect or what the user was expecting and requires.
That I'm involved here means I may understand; but for everyone else, requires some pretty subtle knowledge.

Is it worth thinking that preferring float returns was not necessarily intended behavior and should not be relied on / go away in some future version (say OSL-2?) but fallback to that until then?

lgritz:
Let's do warning + option. (Though for expediency of not having this one PR drag on forever, let's make this PR solve the ambiguity and warn, and I'll do a separate one to handle the general -Werror case of warnings being treated as errors.)


> More complex/hairy examples such as color b = mix (c1, c2, noise(P)); truly does seem ambiguous to me.
> Reading it I have no idea what to expect or what the user was expecting and requires.
> That I'm involved here means I may understand; but for everyone else, requires some pretty subtle knowledge.

It is ambiguous, but I'm convinced that *usually* people want a single slider for a mix (even when mixing colors), and very occasionally would want to control the mixing on a per-channel basis. So I think it's a good guess that the noise should return a float, and if the shader author wants the other case, they can employ a cast.

Is that sensibility specific to mix, or does it generalize?  My intuition is that the "narrower" type (float versus color, say) is usually the correct one.  I'm not 100% sure, but I'm confident enough to think that the warning gets us out of the mess. (I certainly agree that the prior state of getting out of the ambiguity silently and without any warning, was a problem waiting to happen.)


> Is it worth thinking that preferring float returns was not necessarily intended behavior and should not be relied on

It was definitely intended behavior. It is not an accident that the ordering of declarations caused float-return to be favored over color-return in cases where it was ambiguous -- I ordered the functions in a particular way to ensure that happened. (And after 60 or 70 movies using OSL without it surfacing as a reported problem, I think it must be that it corresponds closely to user expectations.) But I very much admit and agree that this was poorly documented in the spec, and that it is better to implement this behavior in a centralized way with a rating system rather than implicitly and separately by careful ordering of each function's declarations, and to have super clear warnings when disambiguation is making choices that potentially is not what the user intended. We can definitely clean all that up.


> Should  this discussion be opened up on OSL-dev to get some other viewpoints or agreements?
> go away in some future version (say OSL-2?) but fallback to that until then?

Let's switch to warnings + disambiguation of return type (when no hint is available) with a prioritized type list. That preserves the old behavior, but avoids decl order dependency (and errors stemming from getting it wrong), and gives a lot more feedback about how and when ambiguities are being addressed.

Then let's see what people's reactions are to the new warnings they get. Will they say, "this is helpful, now I know where I must be explicit to get what I want", or will they say, "I never realized how frequently it was quietly botching what I intended, ick!" or will they say, "I would prefer ambiguity to be an error", or will they say, "I liked it better when it was quiet and just did what I thought was always the right thing anyway?"

What I'm hearing internally from the shading department here is mostly aversion to break anything that they've been using to great success for many years. I'm all for opening up wider discussion, but let's do it down the road when people have experience with the warnings enough to judge. In the mean time, I'm more concerned with getting some form of improvement to the situation, wrapping up this PR, and moving on to the many other outstanding issues (including other improvements from you that I want to concentrate on reviewing and merging).

lgritz:
1 & 2 : yes. But let's make sure the warning is clear about which of the ambiguous choices was actually used.

No, all the texture functions work like this, too, and so could user functions. It's not "just noise".

Yes, in our implementation, I think float `noise(P)` is the same value as `(color noise(P))[0]`. But color noise is 3x more expensive than float noise, so you definitely don't want to compute the color one just to use only one component.

marsupial:
Ok, should be possible tomorrow.
##
I wasn't suggesting `noise(P)[0]` as a replacement more that the following seems odd:
```
noise(P[0]) != noise(P)[0]
noise(P) == noise(P)[0]
```
**cell** & **hash** seem to use all 3 dimensions as a seed, but is that intentional?
I would think the following should be true for all **noise_type**
```
string ntype = <noise_type>;
noise(P) == color(noise(ntype, P.x), noise(ntype, P.y), noise(ntype, P.z));
```

lgritz:
>    noise(P) == color(noise(ntype, P.x), noise(ntype, P.y), noise(ntype, P.z));

Nope, noise doesn't work like that at all.

Think about it -- if you hold y and z constant, and move along x, all three components of your color noise ought to be changing, not just the red one!

Also, the way you write it, the component [0] of the "1D trace" of noise along the x axis would be exactly the same as the component [1] of the 1D trace along the y axis and component [2] along the z axis. That would imply a noise field with a strange rotational symmetry that would be very undesirable.

marsupial:
Updated.
Tested against `ctest -R func` locally, using CI for the full suite.
___
> noise(P) == color(noise(ntype, P.x), noise(ntype, P.y), noise(ntype, P.z));

Agree, I seem to have been dreaming of a language that tags component access with their dimensionality.

Though passing 3-dimensions for 1-dimensional noise and the prior equivalence / non-equivalence along x still seems odd.
No reason to belabor it though.

lgritz:
Just so you know where we are with this... as far as I'm concerned, this seems like a LGTM, but we're still testing against our production shaders and there are some differences that we want to 100% understand before merging. It will probably end up waiting until Monday to fully resolve, and then (assuming that we understand it and it can all be fixed by casting as indicated the warnings showing us what we might have misjudged) I will do the merge.

lgritz:
Update: haha, this seems to have caught some bugs in our shading library, where even the old behavior wasn't working quite like we assumed it was. :-) 

I'm just waiting for the final word from our shading department on their tests, but this looks good to go, with some interesting verifications that it's an improvement over the ambiguous rules we previously had.

marsupial:
Any interesting notes about what it was effecting?

lgritz:
Yeah, the problematic case looked like this:

    vector(noise(...))

Remember, somewhere in all this mess I fixed this case where the type constructor was not correctly passing down the hint about desired type to the typechecking of noise() (though the type *cast*, `(vector)noise(...)` did so properly).

So the old code was choosing float noise, and replicating it in all three channels of the vector -- which rather shockingly, was not noticed although was clearly not what the author of the shader intended.

It wasn't just one shader, either, it was a library function that we use for voronoi noise. So we are going to change the shader library code to use a `(float)` cast for now in order to preserve the (wrong) behavior so that we can merge the osl changes safely, and then later down the line when it's safe to potentially change appearance, remove the cast to get the vector-valued one we always wanted. But at this moment, we have films in progress that can't have any appearance changes.

Sorry about the long hard road on this PR. We build our production renderer using OSL directly from the GitHub master, so we are extra careful with certain kinds of changes that could mess up the shows in progress.

PR #845, opened 2018-01-23
Fix slightly broken overload resolution

lgritz:
When I committed #746, in which I made double declarations of user functions to be a warning, I changed some declarations to remove duplicates in stdosl.h, not realizing that there was some overloading resolution logic that was order dependent (as was so astutely pointed out and attempted to be solved by Marsupial in #837) and that I was inadvertently *depending* on the redundant declaration to get it right.

Ugh, so I broke it, then Marsupial's patch fixed it. Then Marsupial discovered some problems (not this!) with his patch and we reverted it, thus re-introducing the bug.

I'm reviewing another attempt to fix the order dependency problem once and for all (#844), but in the mean time, this fix to stdosl.h fixes the pow and fmod cases that are broken at the moment.

### DIFF
--- a/src/shaders/stdosl.h
+++ b/src/shaders/stdosl.h
@@ -96,8 +96,17 @@ PERCOMP2 (atan2)
 PERCOMP1 (cosh)
 PERCOMP1 (sinh)
 PERCOMP1 (tanh)
-PERCOMP2F (pow)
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM

PR #847, opened 2018-01-26
Fix compiler warning about possibly uninitialized variable (false positive)

### LINKED ISSUE #846: Release builds shouldn't use -Werror
On my platform, building head of git fails because of uninitialised variable warnings being treated as errors.

```
/home/dg/src/OpenShadingLanguage/src/liboslexec/lpeparse.cpp: In member function ‘OSL_v1_10::lpexp::LPexp* OSL_v1_10::Parser::parseSymbol()’:
/home/dg/src/OpenShadingLanguage/src/liboslexec/lpeparse.cpp:117:9: error: ‘iscustom’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
         if (iscustom) {
```

Please don't use -Werror for release builds --- warnings aren't standardised between different compilers, which means that as people upgrade their compilers to have better analysis and more warnings, stuff randomly stops building. (In this case, `iscustom` is being initialised via an output variable in a function call; I bet what's happened is that gcc 7.2.0, which is what I have, has started taking into account the function implementation rather than just assuming that because `iscustom` is being passed by reference then it must be initialised.)

lgritz:
Fixes #846

### DIFF
--- a/src/liboslexec/lpeparse.cpp
+++ b/src/liboslexec/lpeparse.cpp
@@ -106,7 +106,7 @@ Parser::buildStop(LPexp *etype, LPexp *scatter, const std::list<LPexp*> &custom)
 LPexp *
 Parser::parseSymbol()
 {
-    bool iscustom;
+    bool iscustom = false;
     ustring sym = parseRawSymbol(iscustom);
     if (m_ingroup) {
         if (sym == udot)


fpsunflower:
LGTM

PR #849, opened 2018-01-30
Appveyor CI support

marsupial:
## Description
Add Appveyor CI build, shown passing here: https://ci.appveyor.com/project/marsupial/openshadinglanguage-jbc1y/history

The only major changes occur in the first two commits.
1.  Printing a struct will now error at compile time.
```
struct test { float a; float b; };
test t;
printf("t = %g\n", t); // error: struct 'test' is not a valid argument
```

This was previously allowed, but judging from the test it was relying on undefined behavior and not really doing what one would expect.

2. boost::wave is escaping Window's path separators which caused the current work dir to not be stripped properly, which is now detected and fixed in **osllex.l** _preprocess_.

## Tests
Updated some for Windows, predominately focused on IO synching and newlines.
**texture-udim** is skipped as I wasn't successful getting OIIO to build with FreeType support.

## Checklist:


lgritz:
LGTM!

lgritz:
I want to thank you again for the work to get us up on Appveyor for CI on Windows.

Though I've already merged this, I have a couple questions that I'll ask as commit comments...

appveyor.yml
```diff
@@ -0,0 +1,166 @@
...
+        Start-FileDownload https://github.com/OpenImageIO/oiio/archive/Release-1.8.7.zip
+        unzip Release-1.8.7.zip; cd .\oiio-Release*\; md "$OSL_INT_DIR" -Force | Out-Null; cd "$OSL_INT_DIR"
+        cmake ..\.. -G "$GENERATOR" -DCMAKE_CONFIGURATION_TYPES="$env:CONFIGURATION" -DCMAKE_PREFIX_PATH="$DEP_DIR;$BOOST_ROOT" -DCMAKE_INSTALL_PREFIX="$DEP_DIR" -DFREETYPE_DIR="$DEP_DIR" -DBOOST_LIBRARYDIR="$BOOST_LIBRARYDIR" -DUSE_PYTHON=OFF -DOIIO_BUILD_TESTS=OFF -DUSE_NUKE=OFF
+        cmake --build . --config $env:CONFIGURATION --target INSTALL
+
+        # LLVM
+        cd $DWN_DIR
+        git clone --depth 10 --branch release_40 https://github.com/llvm-mirror/llvm.git
+        cd llvm; md "$OSL_INT_DIR" -Force | Out-Null; cd "$OSL_INT_DIR"
+        cmake ..\.. -G "$GENERATOR" -DCMAKE_CONFIGURATION_TYPES="$env:CONFIGURATION" -DCMAKE_PREFIX_PATH="$DEP_DIR" -DCMAKE_INSTALL_PREFIX="$DEP_DIR" -DLLVM_TARGETS_TO_BUILD="X86"
```
lgritz: I'm wondering why you chose to build LLVM from source, which seems to me like it would end up being the majority of time of the CI build and test, possibly pushing it over the time limit and terminating the job early.

Would it not be better to grab pre-built binaries from here?
  http://releases.llvm.org/download.html

appveyor.yml
```diff
@@ -0,0 +1,166 @@
...
+    $env:Path += ";$OSL_BUILD_DIR\src\oslinfo\$env:CONFIGURATION"
+    $env:Path += ";$OSL_BUILD_DIR\src\liboslcomp\$env:CONFIGURATION"
+    $env:Path += ";$OSL_BUILD_DIR\src\liboslexec\$env:CONFIGURATION"
+    $env:Path += ";$OSL_BUILD_DIR\src\liboslnoise\$env:CONFIGURATION"
+    $env:Path += ";$OSL_BUILD_DIR\src\liboslquery\$env:CONFIGURATION"
+
+    cmake --build . --config $env:CONFIGURATION
+
+    # Test OSL
+    ctest -C $env:CONFIGURATION -E "texture-udim"
```
lgritz: Why did you exclude texture-udim? What was going wrong or missing for that test?

marsupial:
The way it is set up is it will compile the necessary dependencies and then cache them.
After the first run all of the dependencies are then downloaded and only OSL will be compiled.
___
Currently OpenImageIO is also cached, but can be pretty easily be adjusted to CI against OIIO.master.

**texture-udim** is skipped because I couldn't get OIIO to find FreeType properly, and that test needs OIIO with FreeType support.

Using the pre-built **LLVM** is rough for two reasons.
- It is actually an installer, not an archive, which throughs up UI when installed.
- The installed product does not contain **llvm-config** which is necessary for CMake.

It is probably do-able to get CMake to recognize an installed LLVM properly, but as OpenEXR is not available as an archive, nor OIIO caching was going to happen anyway.

lgritz:
How about https://www.nuget.org/packages/Llvm.Libs ?

Wouldn't that save a ton of time?

I've tried a whole bunch of time and can't get the full build to happen within the 1 hour time limit. Almost all the time is building llvm from scratch.

marsupial:
Might work, but I could only see it installing headers & libs, no **llvm-config**.

179f16ce2b3eb2290976fe049ef847e0f87fbf15 tests if there is enough time to build OSL, otherwise proceeds only uploading the cache (and marks the build as failing so CI won't get stuck always building the cache and showing success).

If that still doesn't work or is too hackie, can try to ask Appveyor to up your allowed time (though thats probably worse for other contributors).

PR #853, opened 2018-02-02
Function redefinitions

marsupial:
## Description
OSL is currently warning about redeclaration which is totally legal.
The first commit fixes the issue and adds additional context as to where it was previously defined.
It also moves to an actual error as it was unclear given the test name whether it is wanted to be a warning or an error

The second commit moves it back to a warning, which I get as beneficial for debugging purposes.
But that does mean that part of the language spec is that functions are bound to the last definition which existed before the call site and tests against that.

```
point abs (point p) {
  printf ("Alternate abs!\n");
  return abs(p); // Should call original, not recurse
}

int redclfunc (int f) { return 0; }  // ok
int redclfunc (int f) { return 1; }  // fail

void subroutineA() {
    printf ("subroutineA.redclfunc = %d\n", redclfunc(32));
}

shader test() {
    printf ("abs(P)= %g\n", abs(P));
    subroutineA();
    printf ("redclfunc = %d\n", redclfunc(32));
}
```
Should lead to:
```
Alternate abs!
abs(P)= 0.5 0.5 1
subroutineA.redclfunc = 1
redclfunc = 2
```
## Tests
Updated **oslc-err-funcredef** and then moved to **function-redef**

## Checklist:


src/liboslcomp/ast.cpp
```diff
@@ -385,12 +385,31 @@ ASTfunction_declaration::ASTfunction_declaration (OSLCompilerImpl *comp,
...
+        for (FunctionSymbol *f = static_cast<FunctionSymbol *>(existing_syms);
+             f; f = f->nextpoly()) {
+            if (f->argcodes() == argcodes && f->scope() == current_scope) {
+                // If the argcodes match, only one should have statements.
+                // If there is no ASTNode for the poly, must be a builtin, and
+                // has 'implicit' statements.
+                auto other = static_cast<ASTfunction_declaration*>(f->node());
+                if (!other || (other->statements() || other->is_builtin())) {
+                    if (!EH) {
+                        EH = &m_compiler->errhandler();
```
lgritz: I've seen you use this idiom a few times, but I'm not sure it's necessary. You can just directly call

    m_compiler->error (" %s", blah);

or warning or message or whatever. You don't need to retrieve the errorhandler and call through it.

Oh, I see now that there is no m_compiler->message(). That's the missing piece. I will make that for you, simple.

lgritz: Possibly never mind. I see also that the compiler->error and friends want to take a source file and line as arguments. 

So your choice, but I think it may often be helpful to go through the compiler in case we want the compiler itself to do anything special upon error, beyond passing straight to the ErrorHandler.

lgritz: Ah, I see now, there are access points on the ASTNode itself, which figures out the source/line for you. Hang on, will make a PR to give you the missing calls.

lgritz: #854

marsupial: I'm explicitly using the error handler to avoid source/line info for the lines after the error:
```
test.osl:28: warning: Function 'string concat (string, string)' redefined in the same scope
  Previous definitions:
    test.osl:27
    built-in
```
Also OIIO::Filesystem::filename(other->sourcefile().string()) must be called to reduce the path to a filename, which I rather don't like but the test will break down if full-path is printed.

lgritz: After I merge #854, you can call m_compiler->warning, etc., and if you pass an empty filename (nullptr) it won't print the location information.

The advantage of using this rather than the error handler is:

The error handler is passed to the compiler by an application. It's meant to be called by the compiler to communicate the error messages back to the app. Example of when that's useful: if it's a GUI app and it wants to capture the error messages to display in a window. But the error handles doesn't actually process the error on behalf of the compiler, if you know what I mean. It's totally conceivable for the compiler to do additional things -- in fact, you can see that OSLCompilerImpl::error() sets m_err=true, after calling errhandler->error() to print/log the error to the app. If you called errhandler->error() directly, m_err in the compiler would never be set!

It just so happens that at this moment, the compiler's warning(), info(), and message() methods don't do anything "extra" other than pass the errors along to the errorhandler, but it could in the future, so I'd rather that all the compiler error messages go through the compiler itself, rather than directly to the app's error handler.

marsupial: I'm fine updating to call `m_compiler->message` for the list of priors, but just to be clear in case it's getting lost/overlooked:

`ASTNode::error`/`ASTNode::warning` methods are called directly for the initial error at line 401.
```
this->error ("Function '%s %s (%s)' redefined [edit...]\n  Previous definitions:");
=>
test.osl:28: warning: Function 'string concat (string, string)' redefined in the same scope
  Previous definitions:**
```
##
`ErrorHandler::message` is only called after that has occurred to handle formatting of the list of prior definitions at lines >405.
```
EH->message("    %s:%d\n");
=>
    test.osl:27
    built-in
```

lgritz: I understand. I think you should consider the `OSLCompilerImpl::error()` (or `ASTNode::error()` if you want the default behavior of naming the source file and line corresponding to that ASTNode) as the way to register an error (or warning, info, message) to the compiler. Whereas the ErrorHandler is the way the compiler communicates to the app. Calling ErrorHandler is bypassing the compiler, and I'm not sure there's ever a reason we specifically want to do that.

I think that given a choice, making a warning and then a bunch of subsequent (but separate) message calls to clarify it is perhaps not the best approach. It's probably better to add those extra messages to make one big warning (which can include `\n`'s to span multiple lines) and then just one `compiler->warning()` call with all of it. The reason is that you could imagine an app that is using its error handler to intercept the errors and warnings, and it would probably want all the messaging related to a particular warning or error to be in one discrete chunk. Once they are made as separate calls, it's hard for anything intercepting and responding to those calls to understand that they go together.

marsupial: If thats the case I can easily push a commit that batches all formatting into a string and then a single call  to `ASTNode::warning` without #854. (?)

I tend to assume IO will be buffered and avoid appending into a single buffer just to have it copied into another...but the benefits of that pattern here are for sure dubious.

lgritz: I would push it all into formatting of a string and then a single call to warning.

marsupial:
Updated.
Took a liberty in af713c6 to add a [possible] minor optimization of comparison order.

lgritz:
LGTM, I will merge this at the same time as the polymorphism one. Hopefully tomorrow.

PR #854, opened 2018-02-03
Add missing info and message reporting methods to ASTNode and OSLCompilerImpl

lgritz:


marsupial:
Not for nothing but might be worth refactoring all the redundant / boilerplate string formatting, ifdef, and assertions of `info, message, warning, error` routines to forward into one common function?

ast.h
```
  template<typename... Args>
  void show_message (int type, string_view format, const Args&... args) const {
        DASSERT (format.size());
#if OIIO_VERSION >= 10803
        show_message (type, OIIO::Strutil::format (format, args...));
#else /* DEPRECATE when OIIO minimum is at least 1.8 */
        show_message (type, OIIO::Strutil::format (format.c_str(), args...));
#endif
  }

  template<typename... Args>
  void error (string_view format, const Args&... args) const {
    show_message(0, format, std::forward<Args>(args)...);
  }

  template<typename... Args>
  void warning (string_view format, const Args&... args) const {
    show_message(1, format, std::forward<Args>(args)...);
  }

```
ast.cpp
```
void ASTNode::show_message(int type, string_view msg) {
 if (type == 0) m_compiler->error();
 else if (type == 1) m_compiler->warning();
 else if (type == 2) m_compiler->info();
 else if (type == 3) m_compiler->message();
}
```

lgritz:
I will need to refactor anyway when I bump up the OIIO version floor, to declutter these functions, will think about it then.

PR #855, opened 2018-02-06
Appveyor

marsupial:
## Description
Moving to an uncached OIIO created some failures not noticed due to them being skipped.
Get OIIO to build with FreeType.

## Tests
Enable **texture-udim** test on Windows.

## Checklist:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,9 @@
 clone_depth: 10
 version: '{branch}-{build}'
 
+# Disable Appveyor's test discovery which is not used
+test: off
+
 branches:
   only:
     - master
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -1,6 +1,9 @@
 clone_depth: 10
 version: '{branch}-{build}'
 
+# Disable Appveyor's test discovery which is not used
+test: off
```
lgritz: Just out of curiosity, what does this do?

lgritz:
LGTM, merging.

appveyor.yml
marsupial: Appveyor has a built-in discovery for some unit testing frameworks.
Hopefully saves a bit of time by disabling it since **ctest** is not one of them.

marsupial:
This is driving me nuts.
Seems to be timing out before building OSL completes now.
I'll watch if the next one squeaks by and push the **if statement** higher if not.
Maybe the load on Appveyor over weekend was lower allowing further progress?

lgritz:
Some of my appveyor builds complete, others don't. I assume it's about load.
The real solution is going to be to somehow not have to rebuild llvm from scratch every time.

PR #856, opened 2018-02-06
Minor fixes to recent changes.

marsupial:
I think the merge caused a renamed test to not be removed in CMake.
Also fixes Windows strcasecmp usage.


src/liboslcomp/typecheck.cpp
```diff
@@ -1643,6 +1643,10 @@ ASTfunction_call::typecheck (TypeSpec expected)
         auto* legacy = LegacyOverload(m_compiler, this, poly,
                                    &ASTfunction_call::check_arglist)(expected);
         if (m_sym != legacy) {
+            // Fix for Windows.
+#if defined(_WIN32) || defined(_WIN64)
+  #define strcasecmp _stricmp
```
lgritz: Actually, it's probably best to use Strutil::iequals()

lgritz:
LGTM, merging.

PR #857, opened 2018-02-07
Shader parameter metadata [[ int allowconnect = 0 ]]

lgritz:
If a shader input parameter has this metadata present and set to
nonzero, no incoming connections via ConnectShaders() will be accepted,
resulting in a runtime error if it is attempted.


fpsunflower:
LGTM

PR #858, opened 2018-02-08
Fix remaining undefined behavior from ambiguous function returns

marsupial:
Fixes [hopefully] all remaining undefined behavior from return type disambiguation.
Adds color closures, structs, and void to the defined precedence order of resolving via return type.
Errors when disambiguating via return type and more than one user struct is a match.

###
Opted for the order:
**float, int, color, vector, point, normal, matrix, string, color closure, struct, void**

Should be no behavioral change against prior PR except in cases where behavior was undefined.
(i.e. color closure, struct, void were all previously sorted last as well)


lgritz:
LGTM. I will merge this on Tuesday (I'm not at the office Monday, so don't want to merge and then not be there to fix if anything goes wrong).

PR #860, opened 2018-02-13
Fixes for LLVM 6.0

fpsunflower:
Fix some aspects of the MemoryManger interface that have changed in LLVM 6.0+. LLVM was technically source compatible, but because we release our memory managers via a global variable, we were being subjected to destruction order fiasco.

Work around this by implementing our own `MemoryMapper` (an exact clone of LLVM's).

There was also some confusing casting going on between our internal `MemoryManager` and the real llvm one which I refactored to be more clear.


lgritz:
LGTM

PR #861, opened 2018-02-13
OptiX integration

timgrant:
## Description

This patch adds Nvidia OptiX as an optional dependency in the OSL build system. OptiX support can be enabled by setting the option `USE_OPTIX=1` at build time.

It adds a new project (testoptix) that renders a simple scene intended to test basic functionality: 

- OptiX context setup
- CUDA-to-PTX compilation using NVCC
- OptiX kernel launch
- Image I/O

## Tests

testsuite/testoptix

## Checklist:


dictoon:
Amazing!

lgritz:
Awesome! It's great to see this getting rolled out, even though we all know this is just an initial toe in the water.

I'll need an executed CLA before I can merge this.

lgritz:
When I build, it gets most of the way through, but I am getting errors when linking testoptix about undefined reference to lzma stuff.  It says the references are from libOpenImageIO, but if I build OSL without USE_OPTIX=1, I don't get that complaint.

Does that ring a bell?

timgrant:
I haven't encountered that error. I've been using a locally-compiled version of OpenImageIO pulled from the current master on github (so, 1.9.2). I've tested building with clang 4.0 and gcc 5.4 and haven't experienced any linking issues. This has all been in Ubuntu 16.04.

I just tested using OIIO 1.8 and again did not experience any linking errors. My make options when building OSL were: ``LLVM_VERSION=4.0 USE_OPTIX=1``; when building OIIO, I didn't supply any extra arguments to make.

testoptix links libOpenImageIO just as testrender does. I can't imagine why the linking would fail. If you can tell me more about your build environment I can try to reproduce the issue.

lgritz:
We use OIIO straight from master, llvm 5.0, and clang to compile. This is CentOS7.

Cliff said that when he was linking against OptiX for another program he was working on (to test the denoiser), he got the same errors. Moved some things around to fix it, I think he said he he attached liboptix to his library rather than the end binary. I tried that for osl but haven't had luck yet.

I'll fiddle with it and let you know what I find.

timgrant:
What version of OptiX are you using?

lgritz:
Optix 5.0.0, Cuda 8.0

It seems to link properly if I just add "lzma" to the end of the TARGET_LINK_LIBRARIES in src/testoptix/CMakeLists

But when I run testoptix, I get

    NVRTC Compilation failed: NVRTC_ERROR_COMPILATION
    /usr/include/gnu/stubs.h(7): catastrophic error: cannot open source file "gnu/stubs-32.h"

We build our software with clang 5, but the system compiler is gcc 4.8.5, if that makes a difference.

timgrant:
It seems as though ``stubs-32.h`` is only included if ``__x86_64__`` is not defined. Could you try supplying that definition to NVRTC at ``testoptix.cpp:180``, below the other similar definition?

lgritz:
Oh, I see, at line 179 there is

    options.push_back ("-D__x86_64")

should that be changed to "-D__x86_64__"?  Is that a typo? Or are both definitions needed?

I see, if I remove the existing def (without the ending "__") I get a different set of errors in optix_internal.h. It needs both sets of definitions.

But with both, IT WORKS!

I am able to render an image with the three spheres, yay!

src/testoptix/testoptix.cpp
```diff
@@ -0,0 +1,397 @@
...
+    options.push_back (optixu_inc.c_str());
+
+    // NB: The 'gpu-architecture' option specifies the minimum compute
+    //     capability required to execute the generated PTX and may need
+    //     to be changed depending on the requirements of the input CUDA.
+    options.push_back ("--gpu-architecture=compute_35");
+    options.push_back ("--use_fast_math");
+    options.push_back ("--device-as-default-execution-space");
+    options.push_back ("--relocatable-device-code=true");
+    options.push_back ("-D__x86_64");
```
lgritz: This seems to also need "-D__x86_64__"

lgritz:
So what shall we do about the lzma issue?

It works on my end to add that to the library list in src/testoptix/CMakeList.txt. But I'm afraid of breaking other people by adding it. I certainly don't understand what would be the proper test for when it's needed and when it isn't.

src/cmake/externalpackages.cmake
```diff
@@ -208,3 +210,55 @@ endif ()
 
 # end Qt setup
 ###########################################################################
+
+###########################################################################
+# CUDA setup
+
+if (USE_CUDA OR USE_OPTIX)
+    if (NOT CUDAHOME AND NOT $ENV{CUDAHOME} STREQUAL "")
```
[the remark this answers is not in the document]
timgrant: Yes, I believe you are right. ``CUDAHOME`` was used in a prior incarnation of this file that wasn't using CMake's FindCUDA module, and it might be useful for selecting an alternative CUDA installation at configure time.

timgrant:
I'm glad that you're able to get things working with some minor workarounds. I'll try to get a CentOS machine up and running to help identify any distribution-specific issues, if that is indeed what we're seeing with lzma linking problem and the x86_64 flag.

src/testoptix/testoptix.cpp
timgrant: An alternative to using NVRTC would be using NVCC to compile the CUDA sources to PTX as part of the build process, and have testoptix load the PTX files at run time. It would remove some complexity from testoptix and avoid recompiling the CUDA sources each run (although the overhead is small for these simple cases). I can implement either approach.

lgritz: Well, remember that part of the purpose of testoptix is (ideally) to have a clear example that people can look over to figure out how to integrate OptiX+OSL into their own renderer. So I would advise choosing based on what you want to promote as best practices.

lgritz:
So where are we with this PR?

I have received the CLA, so I'm ready to merge it, if it's in a shape you like.

Were you able to figure out was was up with the linkage issues? Is that what's addressed with the offline PTX generation?

timgrant:
I'm still in the process of getting a CentOS7 build environment up and running, so I haven't been able to test out the changes. But I'm hoping that using NVCC and explicitly listing ``lzma`` as a dependency will resolve the linking issues.

lgritz:
I can try on my end, too, if I get a chance today, and let you know.

ee33:
NVCC is a little simpler, more recommended, and hopefully less sensitive to the compile environment.

Yeah, CentOS is a little finicky to set up, so if you could test, that’d be great.  But overall, this PR is ready from our side!

lgritz:
Yes, builds and runs for me now, out of the box! 

So, I declare this "LGTM" and I will merge this now. Then we can all move on to the next PR!

PR #863, opened 2018-02-15
Warning fixes and general cleanup to recently added typecheck code

lgritz:
The most important change here is to modify the practice making
multi-line warnings/errors by issuing a warning/error for the first line
followed by message() calls for the subsequent line.  Instead, we
generate those line-by-line messages as strings, and at the very end
issue a single warning or error call.

Also, directly call m_compiler->warning or error, don't bypass the
compiler by calling the app's error handler.

The reason both of those are important is that I have a separate change
I am ready to submit that is about giving a `#pragma nowarn` that can be
used to suppress warnings on the following line. This is totally borked
by warnings that span multiple calls and where most of their output is
in the form of message() and thus aren't easily recognized as part of
the warning.

This includes some general stylistic cleanup also in a few spots, some
preferences I have but but did not feel was worth nit picking line by
line in the original PRs.

### DIFF
--- a/src/liboslcomp/oslcomp_pvt.h
+++ b/src/liboslcomp/oslcomp_pvt.h
@@ -105,6 +105,8 @@ public:
     {
         ASSERT (format.size());
         std::string msg = OIIO::Strutil::format (format, args...);
+        if (msg.size() && msg.back() == '\n')  // trim extra newline
+            msg.pop_back();
         if (filename.size())
             m_errhandler->error ("%s:%d: error: %s", filename, line, msg);
         else
@@ -119,6 +121,8 @@ public:
     {
         ASSERT (format.size());
         std::string msg = OIIO::Strutil::format (format, args...);
+        if (msg.size() && msg.back() == '\n')  // trim extra newline
+            msg.pop_back();
         if (m_err_on_warning) {
             error (filename, line, "%s", msg);
             return;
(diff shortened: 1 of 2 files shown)


lgritz:
@marsupial, are you ok with this?

marsupial:
Minor thoughts (or differences from what I had):

Have **reportFunction** handle the newline rather than scattering formatting via `"\n" + reportFunction()`
Have **reportFunction** `return "    <none>"` when `sym == nullptr`
Drop OSL_LEGACY_FUNCTION_RESOLUTION=err/warn in favor of new **-Werror**.

First just encapsulates the formatting in the called function vs. caller a bit better.
The last two greatly reduce the ugliness of error reporting from OSL_LEGACY_FUNCTION_RESOLUTION.

marsupial:
**Actually important**

I believe typecheck.cpp@1952 should invert the first condition:

`if (!candidates.hadinitlist() && OSL_LEGACY && strcmp(OSL_LEGACY, "0"))`

lgritz:
The "\n" business -- your suggestion is how I first had it, then switched around to make it easier to avoid accidental double newlines. But you're right, it was more clear the other way. Let me patch it.

I'll check the 1592 thing you point out, too...

marsupial:
I'm pretty sure the correction on 1592 is accurate.)
The basic idea (per comment on 1590) is **skip** legacy checking if any argument is an init-list.

(but I am very tired right now)

lgritz:
Updated with your suggestions.

To fix the "\n" part, I had to modify the CompilerImpl::warning (et al) to chop off any stray \n that made it onto the message by the time it got there, so there wouldn't be an extra blank line. That seems reasonable, and saves us in any case where people do error("blah\n") without realizing that the implementation within error will also add a linefeed.

PR #864, opened 2018-02-16
`#pragma nowarn` suppresses warning on the following source line.

lgritz:


marsupial:
I'd really like for some headers in the future to be shared between OSL and C++ and using a namespace allows for less conflicts and more obvious warnings over them (i.e. I see a **huge** benefit in setting the precedent to do this):

`#pragma osl nowarn` or `#pragma osl(nowarn)`

lgritz:
I'm not sure how big is the subset is of functions that are useful and identically expressed in both language.

But I totally can get behind your rationale and am happy to change the syntax.  

I think I'm leaning toward your first suggestion, `#pragma osl blah`

lgritz:
Pushed an update. Now it's

    #pragma osl nowarn

Good suggestion.

fpsunflower:
LGTM

Issue #851, opened 2018-01-31
gcc vs clang testrender failures


BruceCherniak:
OSL (master head and RB-1.9) compiled with gcc fails several more tests than when compiled with clang.

### Problem
OSL built with gcc fails the following testrender scenes, whereas clang passes:
render-cornell
render-oren-nayar
render-veachmis
render-ward

Simply using the clang generated "testrender" executable in the gcc build tree is sufficient for the tests to run correctly.

**Expected behavior:** gcc and clang should produce the same test results.

**Actual behavior:** gcc produced "testrender" executable has several suspect failures.  Images produced by the gcc executable are far too dark.

### Steps to Reproduce
(gcc is the default compiler)
1. make; make test   [this produces 8 failures]
2. export MYCC=clang MYCXX=clang++
3. make nuke; make; make test   [this produces 4 failures]
4. save build/linux64/src/testrender/testrender
5. unset MYCC MYCXX
6. make nuke; make; make test   [produces original 8 failures]
7. replace build/linux64/src/testrender/testrender with saved version
8. make test   [the 4 render-* failures now pass]

### Versions

* OSL branch/version:  master and RB-1.9
* OS: ubuntu 17.10 and manjaro archlinux
* C++ compiler:  gcc 7.2 and clang 5.0
* LLVM version:  5.0
* OIIO version:  master

fpsunflower:
Have you tried opt vs. debug in both compilers? It would help discount an optimizer related issue.

BruceCherniak:
No change.  Debug build; clang still passes, gcc fails.

To further limit differences, I added USE_BOOST_WAVE=1 to the clang build so that both use the same preprocessor.  No change.

marsupial:
I'm seeing the same tests fail with VisualStudio-2017.
They do pass with VisualStudio-2015.

lgritz:
I've had a very hard time making these render tests produce pixel-identical images on all compiler versions and platforms. I still don't understand why. But I'm very certain that it's because of testrender sampling internals, and not with OSL itself. I can't justify spending more time tracking it down, and this is is of no real consequence, so we can just increase the comparison tolerance or skip those few tests on CI tests if necessary. Closing.

lgritz:
PS. If somebody really wants to track this down, I would welcome a PR to fix testrender so it absolutely guarantees identical output pixels on all platforms and compilers.

PR #865, opened 2018-02-17
Rework Appveyor's caching

marsupial:
## Description
Move away from Appveyor's internal caching which is a bit too opaque, fails often, and makes it difficult to test more than one setup/compiler on Windows.

This is accomplished by reserving a branch named **appveyor** which when built will upload the dependencies into GitHub's **Release** page.

Since the deps are built on the **appveyor** branch, changes to dependencies and building up the cache can be isolated from **master**.

This PR has a few steps before it should be integrated:
Generate an OAuth token for the imageworks Github account.
Encrypt the token here: https://ci.appveyor.com/tools/encrypt
Replace the current token on line 231 of appveyor.yml
Commit into imageworks/appveyor and make sure the dependencies are built/deployed
Commit into imageworks/master

Example of how/where the binary-dependencies are stored:
https://github.com/marsupial/OpenShadingLanguage/releases/tag/WindowsDependencies

## Tests
Per #851, disabled 4 tests on VisualStudio-2017
render-cornell
render-oren-nayar
render-veachmis
render-ward

## Checklist:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -201,7 +201,13 @@ build_script:
         $env:Path += ";$BOOST_LIBRARYDIR"
 
         cmake --build . --config $env:CONFIGURATION $OSL_BUILD_FLAGS
-        ctest -C $env:CONFIGURATION --output-on-failure
+
+        [Array]$OSL_TEST_FLAGS = "-C", $env:CONFIGURATION, "--output-on-failure"
+        if ($env:APPVEYOR_BUILD_WORKER_IMAGE -eq "Visual Studio 2017") {
+            $OSL_TEST_FLAGS += "-E"
+            $OSL_TEST_FLAGS += "render-cornell|render-oren-nayar|render-veachmis|render-ward"
+        }
+        ctest $OSL_TEST_FLAGS
     }
 
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,23 +23,23 @@ platform:
   # - x86
 
 configuration:
-  #- Debug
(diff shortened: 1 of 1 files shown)


lgritz:
This is really interesting!

So the basic idea is that the builds will try to retrieve precompiled versions of the expensive dependencies, presuming that they have been stored on the GH releases page for the imageworks/OpenShadingLanguage repo? And if not found will try to build from scratch (which we know is iffy, in terms of total time)? And the magical way to create or update the set of prebuilt packages is simply to force a build from the "appveyor" branch by the main account holder (imageworks, i.e. me)?

Do I have that right?

And to seed the process, I'll need to set up the encrypted token first, yes? 

And the safeguard is that the update of the binary assets will ONLY happen when when the appveyor job runs from my account?

marsupial:
Sounds right.
Though, I'm not sure the specifics of whether anyone with commit access to the repo or only the main account holder will trigger deployment (I'd assume the later).

lgritz:
Even if it's anybody with commit access, that's fine. I just want to be sure some other rando can't push a broken version.

lgritz:
OAuth token -- do I have to "register an application" first? If so, what should I put in the fields?

marsupial:
**Personal access tokens** -> **Generate New Token**
Deployment works with the **public_repo** (Access public repositories) checked.

lgritz:
I'm going to close this because our aim is to abandon appveyor in favor of using GitHub CI even on Windows.

PR #867, opened 2018-02-21
Improve oslc error reporting

lgritz:
For many grammatical constructs, the "source line" recorded in the AST
node is where the grammatical case ENDS. As an example,

    1:  shader foo (int holdout
    2:                    =
    3:                    1)
    4:  { }

There is a warning because 'holdout' masks another symbol at the same
scope. You would think the warning should be reported as occurring on
line 1, but it is reported for line 3, because that's where the clause
ends.

This will tend to happen for grammatical constructs that look like

     ANCHOR potentially-long-expression-that-may-span-lines

where any sensible person would want the warning associated this to be
on the line of the anchor, not wherever the long expression happens to
end.

In this patch, I carefully fix a whole lot of these grammatical
constructs to patch the 'sourceline' of the AST node they create with
the line that a particular anchor symbol is on.

There's a bit of extra work for ASTvariable_declaration, because there
are some errors (for duplicate symbols, etc.) that happen in the
constructor itself, so it's too late to return from the constructor and
then reset the source line; the misleading error has already been
issued.  So in that case, I added a line override parameter to the ctr
itself, mimicking the solution I implemented some time back for a
similar situation with function declarations.

### DIFF
--- a/src/liboslcomp/ast.cpp
+++ b/src/liboslcomp/ast.cpp
@@ -492,12 +492,19 @@ ASTvariable_declaration::ASTvariable_declaration (OSLCompilerImpl *comp,
                                                   const TypeSpec &type,
                                                   ustring name, ASTNode *init,
                                                   bool isparam, bool ismeta,
-                                                  bool isoutput, bool initlist)
+                                                  bool isoutput, bool initlist,
+                                                  int sourceline_start)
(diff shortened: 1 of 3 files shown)


lgritz:
Anybody have opinions on this?

fpsunflower:
LGTM!

PR #868, opened 2018-02-28
OSL-to-PTX compilation

timgrant:
## Description

This patch consists of modifications to the ``liboslexec`` source to enable the compilation of optimized ShaderGroups to PTX assembly suitable for execution on Nvidia GPUs. To optimize execution on the GPU, there have been changes to string handling userdata parameters.

### llvm_ops.cpp
To make the shadeops defined in ``llvm_ops.cpp`` available to CUDA devices, it was necessary to modify function declarations in ``llvm_ops.cpp``, ``dual.h`` , and ``dual_vec.h`` to include special decorators when compiling for CUDA.

These changes depend on modified versions of ``fmath.h`` from OpeneImageIO and ``ImathVec.h`` from OpenEXR. The changes to ``fmath.h`` are the subject of an open [pull request](https://github.com/OpenImageIO/oiio/pull/1888). The in case of ``ImathVec.h``, we have added a CUDA-friendly variant to the OSL source that is conditionally included when compiling for CUDA.

When OptiX-support is enabled, ``llvm_ops.cpp`` is compiled to LLVM bitcode, serialized, and embedded in the library, in the same manner as the CPU version. 

## Tests

The ``testoptix`` project has been updated to load and render scenes following the XML format used in the testsuite. The rendering is simplistic (flat shading, no lighting), and is intended to demonstrate basic shader execution and closure evaluation.

## Checklist:


src/include/OSL/oslconfig.h
```diff
@@ -56,17 +56,23 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "export.h"
 
 // All the things we need from Imath
-#include <OpenEXR/ImathVec.h>
+#ifdef __CUDACC__
+#include <OSL/ImathVec_cuda.h>
+#else
+#include <OSL/ImathVec_cuda.h>
```
lgritz: You sure this is right?

timgrant: Fairly sure it's not. I'll fix it.

lgritz:
ImathVec_cuda.h -- is the only thing you changed about this file the sprinkling of HOSTDEVICE all over the place? 

I'm just thinking aloud here... don't do anything... just trying to think if there's a way to skin this cat that doesn't involve cloning and altering OpenEXR headers... 

Is there a favored set of small matrix & vector types that Cuda/graphics interop programmers tend to use?

*Most* of places where we use Imath::Vec and Matrix classes (I want to say all, but I'm sure we inadvertently missed some spots), we try to use OSL::Vec3, OSL::Matrix44, etc., which are typedefed to be the Imath ones in oslconfig.h, but don't have to be. I'm wondering if a couple of handy `#if`s in oslconfig.h can solve a bunch of problems including the need to duplicate (and subtly change) Imath headers.

Maybe there isn't. Don't force things to be awkward. But replicating that file feels a little suboptimal to me.

This must crop up all the time with Cuda programming, right? People see a C++ header they want to use in their Cuda code. What do they do? Is the usual practice to completely replicate a header and sprinkle it with `__host__ __device__`? Or is there a more elegant way to deal with this pattern?

src/testoptix/CMakeLists.txt
```diff
@@ -1,37 +1,99 @@
...
+    # Command to turn the renderer library source file into LLVM bitcode .bc,
+    # then back into a C++ file with the bc embedded!
+    # This macro is set up to accept a .cu file, but could be adapted to accept
+    # a generic C++ file.
+    ADD_CUSTOM_COMMAND ( OUTPUT ${llvm_bc_cpp}
+      COMMAND ${LLVM_BC_GENERATOR}
+          "-I${OPTIX_INCLUDE_DIR}"
+          "-I${CUDA_INCLUDE_DIR}"
+          "-I${CMAKE_SOURCE_DIR}/cuda"
+          -D__CUDACC__ -D__CUDA_ARCH__=350
```
lgritz: I found that here, in the assembly of this command, I also needed:

    --cuda-path="${CUDA_TOOLKIT_ROOT_DIR}"

src/liboslexec/CMakeLists.txt
```diff
@@ -117,11 +117,48 @@ MACRO ( LLVM_COMPILE llvm_src srclist )
...
+        COMMAND ${LLVM_BC_GENERATOR}
+            ${LLVM_COMPILE_FLAGS}
+            "-I${CMAKE_CURRENT_SOURCE_DIR}"
+            "-I${CMAKE_SOURCE_DIR}/src/include"
+            "-I${CMAKE_BINARY_DIR}/include"
+            "-I${OPENIMAGEIO_INCLUDE_DIR}"
+            "-I${ILMBASE_INCLUDE_DIR}"
+            "-I${Boost_INCLUDE_DIRS}"
+            -DOSL_COMPILING_TO_BITCODE=1
+            -D__CUDACC__
```
lgritz: I found that here, in the assembly of this command, I also needed:

    --cuda-path="${CUDA_TOOLKIT_ROOT_DIR}"

src/testoptix/CMakeLists.txt
```diff
@@ -1,37 +1,99 @@
...
-            -m64 -arch sm_35 -ptx --std=c++11 -O3 --use_fast_math
+            -DOSL_USE_FAST_MATH=1
+            -m64 -arch sm_35 -ptx --std=c++11 -O3 --use_fast_math --expt-relaxed-constexpr
             "-I${CUDA_INCLUDE_DIR}"
             "-I${OPTIX_INCLUDE_DIR}"
             "-I${CMAKE_SOURCE_DIR}/cuda"
+            "-I${OPENIMAGEIO_INCLUDE_DIR}"
+            "-I${CMAKE_BINARY_DIR}/include"
+            "-I${PROJECT_SOURCE_DIR}/src/include"
             ${cuda_src} -o ${cuda_ptx}
```
lgritz: I found that here, it also needed

    "-I${ILMBASE_INCLUDE_DIR}"

because the ImathVec_cuda.h includes some other Imath files and needs to know where to find them.

lgritz:
In building this, I ran into the lzma problem again. Not sure how to solve elegantly. I was able to complete the build by adding "lzma" to the linkage of testoptix, but I fear that this may fix my work setup while simultaneously breaking other people (e.g. if they neither need nor have lzma).

So to get us unstuck for now, I suggest this:

1. In the main CMakeLists.txt, right after the USE_OPTIX option is declared:

        set (OPTIX_EXTRA_LIBS CACHE STRING "Extra lib targets needed for OptiX")

2. In src/testoptix/CMakeLists.txt, change the TARGET_LINK_LIBRARIES line to:

        TARGET_LINK_LIBRARIES ( testoptix oslexec oslquery ${CUDA_LIBRARIES} ${OPTIX_LIBRARIES} ${OPTIX_EXTRA_LIBS} ${OPENIMAGEIO_LIBRARIES} ${Boost_LIBRARIES} ${CMAKE_DL_LIBS} )

This changes nothing by default, but for anybody who runs into the lzma problem, we know what to tell them to get them unstuck. (And for me at work, I can easily just add `MY_CMAKE_FLAGS += -DOPTIX_EXTRA_LIBS=lzma` to `site/spi/Makefile-bits` to fix it for us.)

If it turns out that this is more commonly seen than just with whatever weirdo setup I have at SPI, we can revisit how to solve it more generally or autodetect the situation.

src/liboslexec/CMakeLists.txt
timgrant: Interesting. I think the lesson here is to take nothing for granted when it comes to build environments. I'll update the commands.

timgrant:
I must have inadvertently reverted the previous "lzma" change, so sorry for that. But your suggested fix probably generalizes a little better, so maybe it was all for the best. :smile:

ee33:
_>  Is the usual practice to completely replicate a header and sprinkle it with `__host__ __device__`?_
To my surprise, yes, it seems to be.

_> Is there a favored set of small matrix & vector types that Cuda/graphics interop programmers tend to use?_
Yes, CUDA comes with vector_types.h, which defines float3, int4, etc.  IIRC these supply basic methods like operator+= but not everything that's in Imath_Vec.h, like dot and cross.  We could build an ImathVec work-alike on top of float3 etc, but I expect we can get the same performance by just decorating ImathVec.  (Later when we look at performance, maybe we'll need to add alignment decorators, in addition to `__device__` decorators.)

Half floats will be another story, but I don't think OSL uses those (yet?).

src/include/OSL/ImathVec_cuda.h
```diff
@@ -0,0 +1,2246 @@
...
+#include <iostream>
+#include <limits>
+
+#if (defined _WIN32 || defined _WIN64) && defined _MSC_VER
+// suppress exception specification warnings
+#pragma warning(push)
+#pragma warning(disable:4290)
+#endif
+
+#ifdef __CUDACC__
```
lgritz: Perhaps this whole definition block should be surrounded by

    #ifndef HOSTDEVICE
        ...
    #endif 

just to save future grief of somebody who includes this after some other header that also defines it?

lgritz:
With all my suggested changes, I am able to get what I think is a clean build!

But I get a runtime error when running the testoptix test:

    $ testoptix -r 320 240 scene.xml out.exr
    ../../src/liboslexec/llvm_util.cpp:1583: failed assertion '0 && "LLVM Target is not initialized"

OH, I bet it's because we didn't build our llvm with ptx support!

Let me try that, I'll get back to you on this point.

timgrant:
Ah, that's an interesting wrinkle. Is there a build-time way to check for a supported backend?

lgritz:
I think it's probably sufficient to just turn that assertion into a fatal error, and assume that if the back end isn't found and it's looking for Cuda, that's the likely reason and explain it.

src/liboslexec/llvm_util.cpp
```diff
@@ -1530,6 +1544,109 @@ LLVM_Util::write_bitcode_file (const char *filename, std::string *err)
...
+    options.FloatABIType                           = llvm::FloatABI::Default;
+    options.AllowFPOpFusion                        = llvm::FPOpFusion::Fast;
+    options.NoZerosInBSS                           = 0;
+    options.GuaranteedTailCallOpt                  = 0;
+    options.StackAlignmentOverride                 = 0;
+    options.UseInitArray                           = 0;
+
+    llvm::TargetMachine* target_machine = llvm_target->createTargetMachine(
+        new_triple, "sm_35", "+ptx50", options,
+        llvm::Reloc::Static, llvm::CodeModel::Default, llvm::CodeGenOpt::Aggressive);
```
lgritz: FYI, this line worked when built against LLVM 5.0, but against llvm 6.0 it says that there is no member named 'Default' in namespace 'llvm::CodeModel'.

lgritz:
Strictly for my own edification, if one is already building with a fairly modern clang (or if it's required to be around to generate the compiled llvm bitcode for llvm_ops.cpp), what are the pros/cons to using clang versus nvcc for compiling Cuda?

lgritz:
@ee33, I think I forgot to reply to part of your message -- there is no 'half' in OSL.

marsupial:
FWIW `llvm-config --targets-built` will repot something like: `X86 NVPTX` if you want to error/warn in CMake.

src/liboslexec/llvm_util.cpp
timgrant: Ah, I see that the 'Default' enum value was removed in 6.0. I believe that ``CodeModel::Small`` should do the right thing in LLVM 4.0 through 7.0. I've tested it in 4, 5, and 6.

lgritz:
Although a series of snafus keeps me from fully testing this, the code looks good, it certainly has no ill effect for default builds, and I'll take your word that it works on OptiX. Certainly anything broken in the OptiX path can be patched as we continue this line of development. So let's merge this and move on to the next step.

Merging!

timgrant:
Sweet!

PR #876, opened 2018-03-24
specify the VERSION and SOVERSION when building library shared objects

mattyjams:
## Description

In this change, we specify the VERSION and SOVERSION properties of library
shared objects so that the SONAME in the shared object contains the OSL version
number suffix (major.minor). This also results in the build creating versioned
symlinks back to the installed shared objects. Code that links against the
libraries can then link against a specific version of OSL such that the correct
version can be dynamically linked at runtime.

This follows a similar pattern to the one used in OpenImageIO.

## Checklist:


mattyjams:
Hi all!

I wanted to put up this change for consideration. My main interest here was getting the SONAME of the built shared library objects to include the OSL version number suffix, and to get version number-suffixed symlinks generated by the build. I followed a pattern similar to what I found in OpenImageIO that does this already.

This addresses an issue for us where we're attempting to dynamically link multiple versions of OSL in the same process. The libraries that depend on OSL simply declare for example "liboslquery.so" as a dependency, so it's not currently possible for them to find the version that they were built against. That should be addressed with this change if that code is rebuilt, as its dependency should then include the version number suffix.

Let me know what you think when you have a chance! Thanks!

lgritz:
LGTM, thanks!

Do you need this backported to release branches?

mattyjams:
That's great, thanks very much Larry!

Yeah, if you're comfortable with it, we'd love for this to get into the 1.8 release branch, as that is currently the most recent release we have in active production.

lgritz:
Yes. Not a clean merge, but I ported by hand. Testing now, if it passes CI, I will merge it into RB-1.8 and cut an end-of-month release for you.

lgritz:
Merged the equivalent patch into RB-1.9 and RB-1.8.

mattyjams:
Sweet! Thanks again, Larry! :)

PR #877, opened 2018-03-26
OptiX noise functions

timgrant:
## Description

This round of changes adds support for most standard OSL noise functions. The idea is to compile and serialize opnoise.cpp and simplexnoise.cpp, just as is done with llvm_ops.cpp. Again, this mostly involves adding the ``OSL_HOSTDEVICE`` macro to functions in the relevant source files as needed, as well as some modifications to OIIO (merged in [#1905](https://github.com/OpenImageIO/oiio/pull/1905)).

The CMake build has been modified to support compiling and linking multiple "shadeops" sources. Additionally, the minimum CUDA version has been bumped to 8.0, and some additional (and experimental) checks have been added to hopefully detect incompatibility between CUDA and LLVM versions.

Not all noise functions are currently supported when compiling for OptiX/CUDA. Specifically, those that take a string as an argument, since strings are not fully supported on the GPU. Once string support has been extended, we will bring in the remaining noise functions.

## Tests

There is a new test (testoptix-noise) that renders a scene using several noise variants.

## Checklist:


lgritz:
Exciting!

I'll have more specific comments soon, there's much to go over.

Just out of curiosity, how big a job do you think it is to make this all work with generic Cuda? What I'm thinking is, if we can augment testshade with a fairly straightforward way to execute the shaders on Cuda in big arrays (akin to how testshade exercises the shading system on an "image" without a renderer -- only a handful of our tests need "testrender"), then instead of having to write entire new "scene tests" for testoptix as you add features, we can just use the existing testsuite unit tests for the Cuda side as well.

timgrant:
The idea of implementing a CUDA version of testshade has come up before. It would certainly be very handy for debugging, catching regressions, comparing with the CPU version, etc. Applying a little test-driven development would probably be a good thing.

My gut feeling is that an initial implementation would have to be constrained much like testoptix is currently, with the goal of increasing test coverage as the CUDA backend matures. We'll look into it and let you know what we think is feasible.

lgritz:
The running of the testsuite already is set up to run each test twice -- once unoptimized and once optimized. It can also have a third round to run through the tests that are expected to work for OptiX.

So primarily, I'm thinking that a little bit of work to make a simple Cuda test harness extension to testshade now will save you lots of trouble as you continue your work. Adding a new feature can simply "enable existing test XYZ for OptiX" instead of having to create an entirely new testoptix-oriented test for every individual feature as you implement it.

lgritz:
I see. Only do it if it makes sense to save you work. I was just seeing the trend of your making a new testoptix thingie for every little thing and hoping to save you some trouble.

CMakeLists.txt
```diff
@@ -96,6 +96,7 @@ option (USE_BOOST_WAVE "Use Boost Wave for C preprocessor (alternative is to use
 option (USE_QT "Include Qt support" ON)
 option (USE_OPTIX "Include OptiX support" OFF)
 set (OPTIX_EXTRA_LIBS CACHE STRING "Extra lib targets needed for OptiX")
+set (CUDA_TARGET_ARCH "sm_35" CACHE STRING "CUDA GPU architecture (e.g. sm_35)")
```
lgritz: Just out of curiosity: what is it that requires sm_35?

src/cmake/externalpackages.cmake
```diff
@@ -223,7 +223,7 @@ if (USE_CUDA OR USE_OPTIX)
         message (STATUS "CUDA_TOOLKIT_ROOT_DIR = ${CUDA_TOOLKIT_ROOT_DIR}")
     endif ()
 
-    find_package (CUDA 7.0 REQUIRED)
+    find_package (CUDA 8.0 REQUIRED)
```
lgritz: Curiosity: what is in this patch that required raising the Cuda floor? Do you think we'll end up sticking with 8.0? Or do you anticipate that it will likely creep up to 9?

src/cmake/externalpackages.cmake
```diff
@@ -238,6 +238,21 @@ if (USE_CUDA OR USE_OPTIX)
     if (NOT ${nvptx_index} GREATER -1)
         message (FATAL_ERROR "NVTPX target is not available in the provided LLVM build")
     endif()
+
+    if (${CUDA_VERSION_MAJOR} GREATER 8 AND ${LLVM_VERSION} LESS 6)
```
lgritz: FYI: we've had a little bit of trouble with llvm6 at our site, so we're hoping to stick with llvm 5.0.1 for now. We are currently on Cuda 8.0 and Optix 5.0. A hard requirement for either LLVM6 or Cuda 9 would force up to need some upgrades (including drivers across the facility, which is a bigger PITA, I think we are on a driver version that's slightly too old for Cuda 9).

src/liboslexec/CMakeLists.txt
```diff
@@ -123,45 +123,76 @@ MACRO ( LLVM_COMPILE llvm_src srclist )
...
-            -D__CUDACC__  -D__CUDA_ARCH__=350
-            --language=cuda --cuda-gpu-arch=sm_35 --cuda-device-only --cuda-path="${CUDA_TOOLKIT_ROOT_DIR}"
-            -Wno-deprecated-register
-            -O3 -Wno-format-security -emit-llvm
-            ${llvm_src} -c -o ${llvm_bc_cuda}
-        COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/serialize-bc.bash" ${llvm_bc_cuda} ${llvm_bc_cuda_cpp} "osl_llvm_compiled_ops_cuda"
-        MAIN_DEPENDENCY ${llvm_src}
-        DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/serialize-bc.bash"
-                ${exec_headers} ${PROJECT_PUBLIC_HEADERS}
+            -D__CUDACC__ -DOSL_COMPILING_TO_BITCODE=1
```
lgritz: I found that I also needed

        "-I${Boost_INCLUDE_DIRS}"

because <boost/thread/tss.hpp> is included transitively by opnoise.cpp (via oslexec_pvt.h).

How are you able to compile without that? Unless maybe your boost headers are in a standard install path.

src/liboslexec/CMakeLists.txt
```diff
@@ -123,45 +123,76 @@ MACRO ( LLVM_COMPILE llvm_src srclist )
...
+            -DNDEBUG -DOIIO_NO_SSE
+            --language=cuda --cuda-device-only --cuda-gpu-arch=${CUDA_TARGET_ARCH}
```
lgritz: I found that on my end, I need this line to also add

    --cuda-path=${CUDA_TOOLKIT_ROOT_DIR}

Or, maybe you think it should be ${CUDA_LIB_FLAGS}, which you set in externalpackages.cmake, but you only add the `--cuda-path` when using Cuda > 8 (I am using 8.0, so that wouldn't come into play).

Any ideas about whether that logic needs adjustment?

lgritz:
By the way, the other nice thing about having a Cuda mode for testshade is that it would allow a more direct performance comparison of executing shaders, divorced from the orthogonal issue of OptiX ray tracing speed versus testrender (which will be slaughtered, it's not designed for speed at all, no biggie, but in the process it will obfuscate the shading speed comparison). The other way that testshade is nice is that it removes all of the "renderer overhead" and allows performance testing (as well as correctness testing) of just the shader execution.

CMakeLists.txt
timgrant: The funnel shift intrinsic used to implement left rotate was added in SM35. It seems to me a like a decent middle ground to use as a default, and the CMake variable should hopefully allow users to choose another architecture if needed.

src/liboslexec/CMakeLists.txt
timgrant: Yes, I meant to add CUDA_LIB_FLAGS to the clang build command in liboslexec, as it's used in testoptix. I must have accidentally reverted the change somehow. Otherwise, I think the logic is okay. ``--cuda-path`` should be supplied only when the CUDA 8.0 is being used, since I found that specifying it with CUDA 9.0/9.1 resulted in mutliply-defined symbol errors.

src/liboslexec/CMakeLists.txt
timgrant: Hmm, boost must be in my path. I'll re-add the include.

src/cmake/externalpackages.cmake
timgrant: Nothing required it, but since I have only tested 8.0+ I decided to narrow the aperture somewhat.

timgrant:
The problem with the missing noise functions stems from how DECLs are guarded in builtindecl.h. I'm not sure what a good solution is, but I'm looking at it now.

timgrant:
The issue seems to be that when the noise ops are serialized to bitcode, it's necessary to avoid adding certain functions to the ``llvm_helper_function_map``, but only when targeting PTX.

I have tried augmenting the definition of ``DECL`` to take an extra argument that indicates whether or not to add a function to the map:

```C++
#define DECL_ALWAYS(name,signature) DECL(name,signature,true)
#define DECL_NODEVICE(name,signature) DECL(name,signature,false)
```

Then you can specify which functions should or shouldn't be added at run time:

```C++
DECL_ALWAYS (osl_add_closure_closure, "CXCC")
DECL_ALWAYS (osl_mul_closure_float, "CXCf")
...
DECL_NODEVICE (osl_hash_ii,  "ii")
DECL_NODEVICE (osl_hash_if,  "if")
```

```C++
static void
initialize_llvm_helper_function_map (bool device_compile)
{
    if (llvm_helper_function_map_initialized)
        return;  // already done
    spin_lock lock (llvm_helper_function_map_mutex);
    if (llvm_helper_function_map_initialized)
        return;
#define DECL(name,signature,always_add)                                     \
    if (! device_compile || always_add) {                                   \
        llvm_helper_function_map[#name] = HelperFuncRecord(signature,name); \
        external_function_names.push_back (#name);                          \
    }
#include "builtindecl.h"
#undef DECL
    llvm_helper_function_map_initialized = 1;
}
```

This approach is working for me, but maybe there's a better way to handle it?

lgritz:
I need to think about this. Trying to avoid clutter. Let me get back to you this afternoon.

lgritz:
What if we had just one DECL, it has a 3rd argument that is a 'flags' bitfield ('or'-ed pow2 bits, I mean), and one of those flags can be DECL_NO_CUDA for ones that should be excluded for the Cuda path? This seems  like it might leave a little more room for non-awkward expansion for other back ends or special cases when they are added. Can we use variadic macros to make it so that the ones that don't need special cases can stay as-is rather than have to pass a 0 or default token?

timgrant:
My initial take was similar to what you propose (sans the bitfield cleverness), so it should be easy to implement. I'll try it out and let you know how it goes.

timgrant:
After doing a little digging, it turns out that this might be a bit of a moot point. When targeting CUDA, LLVM throws a runtime error when compiling two of the hash functions:

```C++
OSL_SHADEOP OSL_HOSTDEVICE int
osl_hash_iv (float *x)
{
    return inthashf (x);
}

OSL_SHADEOP OSL_HOSTDEVICE int
osl_hash_ivf (float *x, float y)
{
    return inthashf (x, y);
}
```

```
llvm_util.cpp:724: failed assertion 'llvm::isa<llvm::Function>(c)': Declaration for osl_hash_ivf is wrong, LLVM had to make a cast
```
One working solution is to use ``void*`` instead of ``float*`` for the shadeops parameters, and cast back to ``float*`` when calling inthashf:

```C++
OSL_SHADEOP OSL_HOSTDEVICE int
osl_hash_ivf (void *x, float y)
{
    return inthashf (static_cast<float*>(x), y);
}
```

Any downsides to this?

lgritz:
Yes, by all means, declare as `void*` and cast internally (golly, I think we almost everywhere else do this anyway, not sure why this ever worked), and reap the wonderful benefits of not changing the DECL's.

timgrant:
It was a bit of a thrill peering over the edge into the abyss of variadic macros, but I'm glad I don't have jump into it.

lgritz:
A few extra `void*`'s and typecasts are among the lesser compromises one makes when trying to JIT a large portion of the rendering pipeline. :-)

lgritz:
LGTM, and builds and runs correctly on my end, yay!
Merging.

PR #878, opened 2018-03-28
Errors for writing to read-only function parameters or globals

lgritz:
It has been a longstanding problem that oslc did not properly flag
attempts to write to function paramseters that were not marked as
'output'.

I've rigged this so that in the current master (and any past releases,
assuming we backport this to release branches), it will just be a
warning. In future versions, it will be a full error. That should give
people some advanced warning (no pun intended) to fix any problems that
this catches that previously went unnoticed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,21 @@
-Release 1.10? -- ?? 2018? (compared to 1.9)
+Release 1.10? -- ?? 2018 (compared to 1.9)
 --------------------------------------------------
 Dependency and standards changes:
-* **LLVM 4.0 / 5.0 / 6.0**: Support has been removed for LLVM 3.x.
+* **LLVM 4.0 / 5.0 / 6.0**: Support has been removed for LLVM 3.x,
+  added for 6.0.
 * **OpenImageIO 1.8+**: This release of OSL should build properly against
   OIIO 1.8 or newer. Support has been dropped for OIIO 1.7.
+
+New back-end targets:
+* **OptiX** Work in progress: Experimental back end for NVIDIA OptiX GPU ray
+  tracing toolkit. #861
+    - Build with USE_OPTIX=1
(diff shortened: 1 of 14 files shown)


fpsunflower:
Can this be applied to globals as well? I don't think we enforce those cases either.

lgritz:
Yes, great idea. I will fix that and amend the PR.

lgritz:
Amended. Now will correctly warn (future: error) if you try to write to a read-only "global" such as `u`.

marsupial:
Is there a larger plan to allow local modification of arguments, or will an explicit copy always be necessary?

GlobalTable:
- Rename **writeable** to **readonly**? **writeable** seems less intuitive than **readonly** which is how the property is used/described elsewhere (i.e Symbol).
- Having constructor arguments that match field names will cause errors/warning in some compilers.


oslc-err-write-nonoutput:
- Might it make sense to run oslc with -Werror and have the test fail so it doesn't break when OSL_VERSION >= 11001.

lgritz:
Modification -- I wasn't planning on it. Function parameters in OSL are always pass by reference, the only issue was that they aren't supposed to be writeable unless marked as output. The original idea was to completely avoid any copies, stack frames, etc., by just treating param values as simple aliases of variables in the caller's scope. Changing to "pass by value" semantics is not without merit, but it's something that should be considered carefully and separately if anyone thinks it's important. In both the 8 years of operational OSL and the even longer history of RSL, dating to ~1990, I don't recall any significant user agitation about this issue, but if people feel strongly about it, widely through the OSL community, it's something we can revisit.

Global table -- you're right, it's less awkward to make it 'readonly', I will change that and amend.

oslc-err-write-nonoutput -- yes, quite right, fixing.

lgritz:
Chris oks via chat. Merging.

PR #879, opened 2018-03-28
Add forgotten -Werror to oslc help message

lgritz:


fpsunflower:
LGTM

Any reason we aren't using OIIO's ArgParse here?

lgritz:
I think it just predates ArgParse and I never went back to fix it. And the way `-I` and `-D` work needed a change to ArgParse that I had been too lazy to do. I'll come back to that, for now just want to fix the missing help line.

PR #880, opened 2018-04-09
SS attrib "error_repeats" controls suppression of repeat errors & warnings

lgritz:
Previously, exactly identical error and warning messages (including from shaders themselves) printed the first time and suppressed the duplicates. The rationale is that an error in a shader could easily turn into thousands of error messages if duplicates weren't suppressed. There wasn't a way to turn this behavior off.

Now it's controlled by ShadingSystem attribute "error_repeats", an int, which has default value 0 (do not issue repeated errors -- the current behavior). If changed to 1, repeated errors and warnings will be allowed to pass to the error handler, without suppression in the ShadingSystem.

Furthermore, the act of setting the attribute (to either 0 or 1) has the side effect of clearing the list of remembered errors. This means that even if you want error suppression, poking it in this way will reset the counters, which might be a handy thing to do at the start of each frame (if you are using a single ShadingSystem to render multiple frames) so that you start the counts fresh for each render, rather than for each ShadingSystem.

This patch is intended to be applied to 1.8, 1.9, and master.

### DIFF
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -643,6 +643,7 @@ ShadingSystemImpl::ShadingSystemImpl (RendererServices *renderer,
       m_lazy_userdata(false), m_userdata_isconnected(false),
       m_clearmemory (false), m_debugnan (false), m_debug_uninit(false),
       m_lockgeom_default (true), m_strict_messages(true),
+      m_error_repeats(false),
       m_range_checking(true),
       m_unknown_coordsys_error(true), m_connection_error(true),
       m_greedyjit(false), m_countlayerexecs(false),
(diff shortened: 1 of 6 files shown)


MasterZap:
Great, thanks. Otherwise the workaround is nuking the shading system?

lgritz:
Yeah, deleting the SS and making a new one would also effectively reset the counters. But that's an unfortunately heavy-weight way to do it.

If I merge this patch, do you have the opportunity to upgrade which OSL version you are using so that you pick up the fix? Or are you stuck for the rest of the year?

lgritz:
Merged into RB-1.8, RB-1.9, and master.

MasterZap:
No I'm grabbing this for sure. Thanks!

src/liboslexec/shadingsys.cpp
```diff
@@ -1278,6 +1281,14 @@ ShadingSystemImpl::getattribute (string_view name, TypeDesc type,
     ATTR_DECODE ("stat:mem_inst_connections_current", long long, m_stat_mem_inst_connections.current());
     ATTR_DECODE ("stat:mem_inst_connections_peak", long long, m_stat_mem_inst_connections.peak());
 
+    if (name == "error_repeats") {
```
MasterZap: Fun fact: Realized you are actually in getattribute, not setattribute here....

lgritz: Sorry, you are right. Fix is #883 and I will backport to all branches where I merged the error_repeats the first time.

PR #883, opened 2018-04-25
Fix error_repeats: clearing happened for getattribute instead of attribute

lgritz:
Thanks, Zap, for noticing this error.


MasterZap:
Much nicer. When can we get this integrated into 1.8 branch? 
Don't want to be pushy but I have some deadlines approaching :)

/Z

lgritz:
Running CI tests now, will merge as soon as they are done.

lgritz:
Merged into master, RB-1.9, and RB-1.8. And next Tuesday is when we will cut new official tagged releases.

PR #885, opened 2018-05-05
RendererServices: make non-abstract

lgritz:
Allowing the base class to be used as a trivial RendererServices, can
be helpful for certain applications where not much renderer functionality
is needed.


fpsunflower:
LGTM

Good idea, didn't realize it was this few methods.

lgritz:
I believe that originally I must have thought, "make all of the methods that every renderer will need be pure virtual, so nobody will forget to supply them and have unexpected bugs as a result."

But today I'm thinking, "not every use of OSL is a renderer, so make it as easy as possible to pass a trivial RendererServices without having to subclass and supply methods for things you won't use."

PR #887, opened 2018-05-07
OSL built-in displace() and bump() were documented but never implemented

lgritz:
Add them now. Also, an additional variety that takes a vector offset in
a non-common space.

There's no good reason why these were never implemented. I think that
people just used the `P += ...; N = calculatenormal(...)` idiom
directly. But they were always documented in the OSL Language Spec, so I
see no reason not to add them now.

### DIFF
--- a/src/shaders/stdosl.h
+++ b/src/shaders/stdosl.h
@@ -407,6 +407,35 @@ int hash (point p, float t) BUILTIN;
 
 
 // Displacement functions
+void displace (vector offset) {
+    P += offset;
+    N = normalize (calculatenormal (P));
+}
+void displace (float amp) {
+    displace (amp * N);
+}
+void displace (string space, float amp) {
+    float len = length (transform (space, N));
+    displace (amp * len * N);
+}
+void displace (string space, vector offset) {
+    displace (transform (space, "common", offset));
+}
+
+void bump (vector offset) {
+    N = normalize (calculatenormal (P+offset));
+}
+void bump (float amp) {
+    bump (amp * N);
+}
+void bump (string space, float amp) {
+    float len = length (transform (space, N));
+    bump (amp * len * N);
+}
+void bump (string space, vector offset) {
+    bump (transform (space, "common", offset));
+}
+
 
 
 // String functions
--- a/src/doc/languagespec.tex
+++ b/src/doc/languagespec.tex
@@ -3823,26 +3823,28 @@ computed (using {\cf filterwidth()}).
 \index{functions!displacement}
 
 \apiitem{void {\ce displace} (float amp) \\
+void {\ce displace} (vector offset) \\
 void {\ce displace} (string space, float amp) \\
-void {\ce displace} (vector offset)}
+void {\ce displace} (string space, vector offset)}
 \indexapi{displace()}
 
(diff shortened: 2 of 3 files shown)


src/shaders/stdosl.h
```diff
@@ -407,6 +407,35 @@ int hash (point p, float t) BUILTIN;
...
 // Displacement functions
+void displace (vector offset) {
+    P += offset;
+    N = normalize (calculatenormal (P));
+}
+void displace (float amp) {
+    displace (amp * N);
+}
+void displace (string space, float amp) {
+    float len = length (transform (space, N));
```
fpsunflower: Does this do the right thing if the scale is not uniform?

src/shaders/stdosl.h
```diff
@@ -407,6 +407,35 @@ int hash (point p, float t) BUILTIN;
...
+    displace (amp * len * N);
+}
+void displace (string space, vector offset) {
+    displace (transform (space, "common", offset));
+}
+
+void bump (vector offset) {
+    N = normalize (calculatenormal (P+offset));
```
fpsunflower: Isn't it discouraged to change `N` in a surface shader? Most closures take a normal as an argument.

fpsunflower: I think I would prefer for `bump` and `displace` to return a `normal` and `point` respectively, and let the shader writer manage the assignment to global variables.

src/shaders/stdosl.h
lgritz: I think. But it might depend on what you think is "right".

src/shaders/stdosl.h
lgritz: The idea is that you should call one of these in the "root" shader, which is passed in the accumulated offset or amount.

If you're randomly calling it in mid-graph nodes, perhaps multiple times, you're probably going to do weird things.

lgritz: Also, the idea was, some day (or in some renderers) we might want displace() to have additional side effects. For example, maybe it modifies dPdu, dPdv? Or do other renderer-specific things? I wanted a function that does the "assignment to global variables" part, which may someday grow to have per-renderer trickiness. The "accumulate and return total" part is already trivial.

src/shaders/stdosl.h
mprater: :) Yeah...  I've been using this idiom forever. It makes intuitive sense, and it does give the user some form of non-uniform control over the displacement magnitude. However, lately I've switched to using the cube-root of abs(determinant(matrix(space,"common"))) as the scale factor. This change of heart is due to my experience using non-uniformly scaled coordinate systems to control displacement. With the simple length(transform(N)) method, the results are never quite what I expect or desire, and so the presumed control it allows doesn't, in my recent opinion, live up to the expectation. As a result, I've concluded that preserving a volumetric sense of relative scale is more useful in the rare instance when a user must use a non-uniformly scaled space to control displacement magnitude. Since displacement can be viewed as a volumetric operation, I now see this as making more intuitive sense. But, all that said, the simple approach works perfectly in the case of uniformly scaled coordinates, and is cheaper to compute. And since that's by far the most likely scenario, it may be the right implementation choice.

PR #888, opened 2018-05-11
Fix llvm7

Lucretia:
## Description

Provides 2 patches to build on current master of LLVM.

## Tests

None

## Checklist:

### DIFF
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -1503,7 +1503,11 @@ LLVM_Util::write_bitcode_file (const char *filename, std::string *err)
     std::error_code local_error;
     llvm::raw_fd_ostream out (filename, local_error, llvm::sys::fs::F_None);
     if (! out.has_error()) {
+#if OSL_LLVM_VERSION >= 70
+        llvm::WriteBitcodeToFile (*module(), out);
+#else
         llvm::WriteBitcodeToFile (module(), out);
+#endif
         if (err && local_error)
             *err = local_error.message ();
     }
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM!

Any chance you can send me a signed CLA?

lgritz:
> "For something so trivial??"  

If I owned the project personally, certainly, I could make the call that short contributions like this don't need a CLA. But this project is sponsored by a company that is very legally conservative about this kind of thing. They give me remarkable freedom to administer this project as I see fit -- including developing the whole thing in public and giving away every last bit of it -- and almost the only restriction they stipulate is that I not do any merges into the repo without a CLA on file from the contributor. Given how open the whole project is (and that they pay my salary), it seems like a very small concession.

> I don't have a working printer right now.

It's the same bit pattern if you paste a signature image onto the PDF and then email it to us, as if you printed, signed, re-scanned, and then emailed. There's no need to do it the hard way.

I'll look into your suggestion for  https://cla-assistant.io/
That's an interesting long-term solution, but I would expect it to take weeks or months of back and forth with the lawyers to decide if that's an adequate substitute. The gears turn slowly in big companies.

Lucretia:
Do I just need to sign and have my name printed on it? I'm not sticking an address on it.

Lucretia:
P.S :  My signature's pretty crap and changes all the time, an X would just as good ;)

lgritz:
It doesn't go anywhere public. Nobody will see your address.

We just file it away and don't make any attempt to verify what your signature looks like or where you live. We trust you to fill it out as completely and seriously as you would for any other legal document.

lgritz:
Closing. This doesn't fully work and I'll do it myself.

PR #890, opened 2018-05-30
Device Strings

timgrant:
## Description

This pull request adds a new string representation, ``device_string``, which is intended to be a GPU-friendly alternative to ``OIIO::ustring``. A ``device_string`` is essentially a struct consisting of an integer ID and a pointer to a C-style string:

```C++
    struct device_string {
        uint64_t m_tag;
        const char* m_chars;
    };
```

The ID can be used to perform fast comparison between two ``device_string`` values, similar to how pointer comparisons are used with ``ustring``. The ``char*`` is available for cases where the string contents are needed (mainly ``printf``).

``device_string`` only supports a limited sub-set of string functionality. No dynamic operations are supported.

___IMPORTANT NOTE___
The minimum version of OptiX has been increased to 5.1, to accommodate ``device_string`` initializers in compiled C++/CUDA code.

#### Assigning IDs
For comparisons between compiled and user-supplied strings to be valid, all parties (especially ``liboslexec`` and the renderer) must agree on their IDs. This approach involves assigning fixed IDs to some subset of "important" strings (closure parameters, texture options, etc.), while allowing the user to assign additional tags via the OSL ShadingSystem. Shader compile-time strings that do not have a pre-defined tag fall back to the ``ustring`` hash.

The tags for important, common strings are defined in a shared header file (``OSL/device_string.h``):  
```c++
    enum StringTags: uint64_t {
        EMPTY_STRING = 0,
        PERLIN,
        UPERLIN,
        ...
        UNKNOWN_STRING = ~0u
    };
```

Additional string tags can be registered int the renderer setup code (a more detailed example is in ``testoptix/shading.cpp``):
```C++
    // this call will fail if "test string" has been registered with a different tag
    shadingsystem->register_string_tag ("test", 1234u);
    // the renderer can query the ShadingSystem for registered strings
    uint64_t tag = shadingsystem->lookup_string_tag ("perlin");
```

It should be noted that no attempt is made to guarantee the uniqueness of ``device_string`` IDs, as that implies a central run-time registry, which is a mechanism we are trying to avoid. This places some burden on the renderer to ensure that tag values are used consistently.

## Tests

We have added a ``printf`` test to the testsuite. More tests will be forthcoming in subsequent pull requests that exercise string functionality more thoroughly.

## Checklist:

### DIFF
--- a/src/liboslexec/llvm_gen.cpp
+++ b/src/liboslexec/llvm_gen.cpp
@@ -2847,15 +2847,7 @@ LLVMGEN (llvm_gen_noise)
     std::string funcname = "osl_" + name.string() + "_" + arg_typecode(&Result,derivs);
     std::vector<llvm::Value *> args;
     if (pass_name) {
-        if (! rop.use_optix()) {
-            args.push_back (rop.llvm_load_value(*Name));
-        }
-        else {
-            llvm::Value* name_arg = (Name->is_constant())
-                ? rop.getOrAllocateLLVMGlobal (*Name)
-                : rop.llvm_load_value (*Name, 0, nullptr, 0);
-            args.push_back (name_arg);
-        }
+        args.push_back (rop.llvm_load_value(*Name));
     }
     llvm::Value *tmpresult = NULL;
     // triple return, or float return with derivs, passes result pointer
--- a/src/liboslexec/llvm_gen.cpp
+++ b/src/liboslexec/llvm_gen.cpp
@@ -358,12 +358,17 @@ LLVMGEN (llvm_gen_printf)
                         s += " ";
                     s += ourformat;
 
-                    // In the OptiX case, we need to use the device-side string constant,
-                    // as added to the LLVM Module.
-                    llvm::Value* loaded = (simpletype.basetype == TypeDesc::STRING &&
-                                           sym.is_constant() && rop.use_optix())
-                        ? rop.getOrAllocateLLVMGlobal (sym)
-                        : rop.llvm_load_value (sym, 0, arrind, c);
+                    llvm::Value* loaded = nullptr;
+                    if (rop.use_optix() && simpletype.basetype == TypeDesc::STRING) {
+                        // In the OptiX case, we use the device_string.
+                        llvm::Value* device_string = (sym.is_constant())
+                            ? rop.getOrAllocateLLVMGlobal (sym)
+                            : rop.llvm_load_value (sym, 0, arrind, c);
+                        loaded = rop.llvm_load_device_string_char_ptr (device_string);
+                    }
+                    else {
+                        loaded = rop.llvm_load_value (sym, 0, arrind, c);
+                    }
 
                     if (simpletype.basetype == TypeDesc::FLOAT) {
                         // C varargs convention upconverts float->double.
@@ -386,6 +391,17 @@ LLVMGEN (llvm_gen_printf)
         }
     }
 
+    if (rop.use_optix() && arg > (format_arg + 2)) {
+        ASSERT (0 && "OptiX printf only supports 0 or 1 arguments at this time");
+    }
+
+    // In OptiX, printf currently supports 0 or 1 arguments, and the signature
+    // requires 1 argument, so push a null pointer onto the call args if there
+    // is no argument.
+    if (rop.use_optix() && arg == format_arg + 1) {
+        call_args.push_back(rop.ll.void_ptr_null());
+    }
+
     // Some ops prepend things
     if (op.opname() == op_error || op.opname() == op_warning) {
         std::string prefix = Strutil::format ("Shader %s [%s]: ",
@@ -395,11 +411,15 @@ LLVMGEN (llvm_gen_printf)
     }
 
     // Now go back and put the new format string in its place
-    call_args[new_format_slot] = (! rop.use_optix())
-        ? rop.ll.constant (s.c_str())
-        // In the OptiX case, we need to use the pointer to the constant format
-        // string added to the LLVM Module
-        : rop.llvm_get_pointer (format_sym);
+    if (! rop.use_optix()) {
+        call_args[new_format_slot] = rop.ll.constant (s.c_str());
+    }
+    else {
+        // In the OptiX case, we need to use the pointer to the device_string
+        // added to the LLVM Module.
+        llvm::Value* device_string = rop.getOrAllocateLLVMGlobal (format_sym);
+        call_args[new_format_slot] = rop.llvm_load_device_string_char_ptr (device_string);
+    }
 
     // Construct the function name and call it.
     std::string opname = std::string("osl_") + op.opname().string();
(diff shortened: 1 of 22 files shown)


src/liboslexec/llvm_gen.cpp
```diff
@@ -386,6 +391,17 @@ LLVMGEN (llvm_gen_printf)
         }
     }
 
+    if (rop.use_optix() && arg > (format_arg + 2)) {
+        ASSERT (0 && "OptiX printf only supports 0 or 1 arguments at this time");
```
lgritz: Can we make this into some kind of warning and turn the printf into something harmless, rather than an assertion that will bring down the whole app?

lgritz:
So a renderer is expected to pre-register a bunch of strings it thinks is important?
And any not in the table will just use the ustring hash?
Why not use the ustring hash for all of them? Is there advantage to having the pre-registered ones?

src/liboslexec/llvm_gen.cpp
timgrant: Definitely. If you find other instances, let me know and I will update them.

lgritz: Also, I just had the idea that I can do some constant folding of printf in the optimizer, which I bet in a wide array of cases can turn printfs into either constant strings or just one argument left.

lgritz:
OK, dumb question:

If we rigged ustring to have a (build time? run time?) option to allocate all of its unique string character memory using cudaMallocManaged(), could we then just pass plain old ustrings and call it a day, because it could use the pointer address for the assignments and comparisons, and when it needs the characters, they would simply be available through the magic of unified memory?

timgrant:
Using the hash in all cases would definitely be more convenient. The main problem that I ran into is statically computing the ustring hash for arbitrary strings in compiled code. There is probably some combination of macro/template/constexpr wizardy to make that work, but I was not successful in my attempts.

fpsunflower:
Using the hash alone would still run the risk of collisions right?

Remind us why the pointer itself is not good enough? Why do you need statically known integers for the strings given that you are making all the ptx at runtime anyway?

lgritz:
Chris: yes, the hash could in theory collide. The character addresses cannot, they are necessarily unique 64 bit values.

In normal ustring use, it's not the hash we use to compare strings, but the address of the unique representation of characters. This is runtime-dependent, but we don't care. 

The hashes are certainly deterministic, they ought to be able to be precomputed for strings we care about. But like Chris says, using the addresses instead of the hashes is preferred, and safer.

What's the circumstance in which the hash value needs to be known at compile time? Is it strictly for the named closure parameter names, which must be pre-compiled to PTX?

I'm sure you've probably been down these roads already, but of course I have to chase it down myself to be confident that this is the least-awkward design we can come up with.

Can the fact that all closures need to be registered (ShadingSystem::register_closure) in a way that communicates the names of the params, somehow be used to communicate the string addresses (though they be runtime determined) in time for the PTX compilation?

src/liboslexec/oslexec_pvt.h
```diff
@@ -913,6 +926,9 @@ class ShadingSystemImpl
     mutable std::map<ustring,long long> m_group_profile_times;
     // N.B. group_profile_times is protected by m_stat_mutex.
 
+    // A mapping from strings to integers, used for constructing tagged strings
+    std::map<ustring, uint64_t> m_tag_map;
+
```
marsupial: Is it worth specializing so that key comparison uses `ustring::c_str()` avoiding the `ustring::compare` call to `strcmp`, or is lexicographical sorting needed?

src/liboslexec/shadingsys.cpp
```diff
@@ -1013,6 +1040,18 @@ ShadingSystemImpl::query_closure(const char **name, int *id,
...
+{
+    if (m_tag_map.count(str) && m_tag_map[str] != tag)
+        return false;
+
+    m_tag_map[str] = tag;
+    return true;
+}
+
+
+
```
marsupial: Can reduce to one lookup
```
// Insert or validate previous string has the same tag
return m_tag_map.emplace(str, tag).first->second == tag;
```

src/liboslexec/shadingsys.cpp
```diff
@@ -1888,6 +1927,14 @@ ShadingSystemImpl::getstats (int level) const
...
 
 
+uint64_t
+ShadingSystemImpl::lookup_string_tag (ustring str)
+{
+    return m_tag_map.count(str) ? m_tag_map[str] : StringTags::UNKNOWN_STRING;
+}
+
+
+
```
marsupial: Can reduce to one lookup
```
auto itr = m_tag_map.find(str);
return itr != m_tag_map.end() ? itr->second : StringTags::UNKNOWN_STRING;
```

src/liboslexec/shadingsys.cpp
```diff
@@ -1013,6 +1040,18 @@ ShadingSystemImpl::query_closure(const char **name, int *id,
 
 
 
+bool
+ShadingSystemImpl::register_string_tag (ustring str, uint64_t tag)
+{
+    if (m_tag_map.count(str) && m_tag_map[str] != tag)
+        return false;
+
+    m_tag_map[str] = tag;
```
marsupial:
```
// Insert or validate previous string has the same tag
return m_tag_map.emplace(str, tag).first->second == tag;
```

src/liboslexec/shadingsys.cpp
```diff
@@ -1888,6 +1927,14 @@ ShadingSystemImpl::getstats (int level) const
 
 
 
+uint64_t
+ShadingSystemImpl::lookup_string_tag (ustring str)
+{
+    return m_tag_map.count(str) ? m_tag_map[str] : StringTags::UNKNOWN_STRING;
+}
```
marsupial:
```
auto itr = m_tag_map.find(str);
return itr != m_tag_map.end() ? itr->second : StringTags::UNKNOWN_STRING;

```

timgrant: Cool, I'll integrate your suggestions if the alternative representation does not pan out.

lgritz:
Yeah, we should have updated the ticket... Tim and I (and some others) have had some email and videoconf follow-ups on this PR and we're looking into variants of the strategy presented here. So take the code here with a grain of salt, it may not be the final stab at this.

marsupial:
Were just some simple notes while I was reading through, as this does relate to some issues I'm also hitting.  Good to know a bit of the larger context, and hope the following doesn't inject into an already complicated affair, but should be easy enough to discard if the need arrises.
##
Would it be advantageous for these methods to be a part of **RendererServices**?
One could then do the following without requiring a back pointer to the owning **ShadingSystem**.
```
RendererServices::TextureHandle* RendererServices::get_texture_handle(ustring file) {
  uint64_t texTag = this->lookup_string_tag(file);
  if (texTag == StringTags::UNKNOWN_STRING) {
    texTag = this->bindTexture(file);
    this->register_string_tag(file, texTag);
  }
  return (RendererServices::TextureHandle*) texTag;
}
```
##
Are other execution models that might have even less string manipulation available being considered?
Is having `device_string` as a simple an index into buffers the renderer provides a possibility?
For example the `const char*` inside `device_string` is not valid in OpenCL-1.x.
##
Furthermore, how much string manipulation is available through Optix? Do/will patterns like this work?
```
string CdTex;
getattribute("", CdTex);
texture(concat("/path/to/", CdTex, ".exr"), u, v);
```
##
I've done some work const-folding string formatting & concatenation, but it does so on the **llvm::Module**.  Is it a path worth going down to handle as an LLVM optimization-pass or preferable to do so in **OSL::RuntimeOptimizer**?

ee33:
As Larry said, Tim explored putting the ustring table in Unified Memory where it would be accessible to both CPU and GPU, using cudaMallocManaged().  This might work someday, but at present, there are enough subtle interactions with the CUDA memory manager that it doesn't seem prudent to use in a library like this.  (What if the CUDA context is destroyed, what if some other plugin is running a GPU kernel while the table is being updated, etc.)

Meanwhile, the discussion here inspired Tim to write a simplified DeviceString class.  We are in the process of withdrawing this PR, and posting a new PR, with "device strings Take 2".

PR #894, opened 2018-06-23
New shader types: camera, image

lgritz:
'camera' shaders can do lens or projection effects, and are allowed to
modify P and I.

'image' shaders can do image processing or post-render passes on pixels.

For image shaders, we're presuming that u,v are image coordinates, and
that texture lookups are capable of referencing the AOV buffers, and
have special semantics that when it's clear you are directly looking up
at u,v, it directly grabs corresponding pixels and avoids "double
filtering". (The exact implementation of these details are coming
later.)

If the docs are a little vague, that's because we're still exploring how
they can and should be used. For this review, we're just allowing those
new names and types when compiling shaders, which primarily boils down
to what global variables you're allowed to mess with inside of them.

In the process, I fixed a longstanding bug, which is that we were not
properly checking that shaders weren't modifying global variables that
were documented to be read-only in that shader type. Hopefully this
won't break people much... you really shouldn't have been depending on
breaking the rules.

Note that I've changed the internal ShaderType and ShaderUse to be fully
modern 'enum class' with proper namespacing.

### DIFF
--- a/src/doc/languagespec.tex
+++ b/src/doc/languagespec.tex
@@ -60,7 +60,7 @@ Language Specification
 Editor: Larry Gritz \\
 \emph{lg@imageworks.com}
 }
-\date{{\large Date: 31 Mar 2018 \\
+\date{{\large Date: 22 Jun 2018 \\
 % (with corrections, 31 Mar 2018)
 }
 \bigskip
@@ -583,17 +583,11 @@ The attribute state also includes shader assignments --- the shaders or
 shader groups for each of several \emph{shader uses}, such as surface
 shaders that designate how light reflects or emits from each point on a shape,
 displacement shaders that can add fine detail to the shape on a
-point-by-point basis, and volume shaders that describe how light is
-scattered within a region of space.  A particular renderer may have
-additional shader types that it supports.
-
-%\subsection*{Shader types}
-%
-%There are several types of shaders: surface, displacement, light,
-%volume.  The type of shader determines what it's allowed to do (for
-%example, only displacement shaders can alter the surface position).
-%There is also a generic kind of shader that can be used for any of the
-%shader uses.
+point-by-point basis, volume shaders that describe how light is
+scattered within a region of space, camera shaders that can specify
+projection or lens effects, and image shaders that can perform image
+processing or compositing operations on output images.
+A particular renderer may support only a subset of shader types.
 
 \subsection*{Shader execution state: parameter binding and global variables}
 
@@ -916,9 +910,10 @@ ordinary execution instructions (such as assignments, etc.).
 \index{shader types} \index{types!shader}
 
 Shader types include the following: {\cf surface}, {\cf displacement},
-{\cf light}, {\cf volume}, and generic {\cf shader}.  Some operations
+{\cf light}, {\cf volume}, {\cf camera}, {\cf image},
+and generic {\cf shader}.  Some operations
 may only be performed from within certain types of shaders (e.g., one
-may only call {\cf displace()} or alter \P in a displacement shader),
+may only call {\cf displace()} or alter \P in a displacement or camera shader),
 and some global variables may only be accessed from within certain types
 of shaders (e.g., {\cf dPdu} is not defined inside a volume shader).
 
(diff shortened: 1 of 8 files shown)


src/doc/languagespec.tex
```diff
@@ -2812,56 +2821,23 @@ \section{Global variables}
...
+\I            & R        &              & R      & RW     &       \\
+\N            & RW       & RW           &        &        &       \\
+{\cf Ng}      & R        & R            &        &        &       \\
+{\cf dPdu}    & R        & R            &        & R      &       \\
+{\cf dPdv}    & R        & R            &        & R      &       \\
+%{\cf Ps}      &          &              & R      &        &       \\
+{\cf u, v}    & R        & R            & R      & R      & R     \\
+{\cf time}    & R        & R            & R      & R      & R     \\
+{\cf dtime}   & R        & R            & R      & R      & R     \\
+{\cf dPdtime} & R        & R            & R      & R      & R     \\
```
fpsunflower: I don't think `time`, `dtime`  or `dPdtime` are relevant to the image shaders (for now at least).

src/include/osl_pvt.h
```diff
@@ -44,9 +44,8 @@ class StructSpec;
 /// Kinds of shaders
 ///
 enum ShaderType {
```
fpsunflower: Did you mean to make this one an `enum class` as well?

src/doc/languagespec.tex
lgritz: yeah, you're probably right

src/include/osl_pvt.h
lgritz: yes. why did  that work?

lgritz:
Self-doubt question: am I barking up the wrong tree here entirely, and we should just hold off on this while we just use fully generic "shader" shaders and rig them up to the renderer however, and then formalize it all only after we've gotten some more experience?

src/liboslcomp/oslcomp.cpp
```diff
@@ -1149,8 +1158,30 @@ OSLCompilerImpl::check_write_legality (const Opcode &op, int opnum,
...
+    // Check for writing to globals.
+    if (sym->symtype() == SymTypeGlobal) {
+        static ustring Ci("Ci"), P("P"), I("I"), N("N"), Ng("Ng");
+        static ustring dPdu("Pdu"), dPdv("Pdv"), u("u"), v("v");
+        static ustring time("time"), dtime("dtime"), dPdtime("dPdtime");
+        ShaderType stype = shadertype();
+        ustring symname = sym->name();
+        bool ok = true;
+        if (stype == ShaderType::Surface)
+            ok = (symname == Ci || symname == N);
```
fpsunflower: I don't think surface shaders are allowed to write to `N` ? (since most closure take in the bump normal)

src/liboslcomp/oslcomp.cpp
```diff
@@ -1149,8 +1158,30 @@ OSLCompilerImpl::check_write_legality (const Opcode &op, int opnum,
...
-    // FIXME -- check for writing to globals.  But it's tricky, depends on
-    // what kind of shader we are.
+    // Check for writing to globals.
+    if (sym->symtype() == SymTypeGlobal) {
+        static ustring Ci("Ci"), P("P"), I("I"), N("N"), Ng("Ng");
+        static ustring dPdu("Pdu"), dPdv("Pdv"), u("u"), v("v");
+        static ustring time("time"), dtime("dtime"), dPdtime("dPdtime");
+        ShaderType stype = shadertype();
+        ustring symname = sym->name();
+        bool ok = true;
```
fpsunflower: Generic shaders should be `ok=false` ?

src/liboslcomp/oslcomp.cpp
lgritz: That's what the docs said.

src/liboslcomp/oslcomp.cpp
```diff
@@ -1149,8 +1158,30 @@ OSLCompilerImpl::check_write_legality (const Opcode &op, int opnum,
...
+        if (stype == ShaderType::Surface)
```
fpsunflower: Might be more clear as a `switch` statement?

src/liboslcomp/oslcomp.cpp
lgritz: At least in testsuite, we seem to depend on generic shading being able to do anything.

lgritz:
I have a few typos and whatnot. Need to update

src/doc/languagespec.tex
lgritz: ok, dPdtime, certainly. But might time and dtime (for image shaders) be the frame time and duration? I can imagine wanting time-varying effects implemented in these shaders, for example.

aconty:
I think it makes sense to have different shader types, at least for the image one, whose globals will have nothing to do with surface.

src/liboslcomp/oslcomp.cpp
fpsunflower: Is that a bug or a feature? Might be beyond the scope of this review, but I think it would be worth clamping down on this.

lgritz: I don't know if it's a bug or feature. I'm going to open a separate discussion about this on osl-dev, there's an important question of future direction here and I want a diversity of opinions.

src/doc/languagespec.tex
fpsunflower: Makes sense. Maybe we should fix our renderer to provide `time` as `frame_time+shutter_sample` for the other shader types then.

fpsunflower:
LGTM

lgritz:
Abandoning this approach in favor of #899 instead.

PR #899, opened 2018-07-10
ShaderType/ShaderUse refactoring

lgritz:
Deprecate the use of shader "types" (surface, displacement, etc.)
designated in OSL source code. Those words will still be accepted for
backward-compatibility, but they are no longer required or encouraged.
Generic "shader" type is the only one we will talk about as we move
forward.

Excise the ShaderUse enum. Renderers (or non-renderers using OSL for
programmability) are allowed to determine and document whatever type of
shader "uses" they want to support. The only enforcement from the
liboslexec side is that all the shader nodes within one group must be
tagged with the same "use" string.

The "use" parameter to ShadingSystem::Shader() is intended to be removed
entirely, eventually. However, that's the first of three (optional)
parameters to Shader(), so it's a subtle change that can break renderers
in annoying and confusing ways if we do it all at once, so we'll do it
in stages: For this major release, we will make all three parameters to
Shader() be required (no optional params), which even if it generates a
compile-time error for a renderer, is a very obvious one to fix. Then in
a later major release (2.1?), after we know that all renderer code is
making 3-argument calls to Shader(), we will introduce a 2-argument
version that lacks a "use" parameter (can't break renderer code) and
deprecate the 3-argument call that has the useless "use" parameter, and
then in a later release, we can remove the 3-argument version once and
for all.

Add new queries for shader groups: "globals_read" and "globals_write",
which return a bit field of which shaderglobals fields will be read or
written, respectively, by the shader group after it's optimized.
The bitfield meanings are given by the new enum class SGBits, defined
in shaderglobals.h. This is easier to query (as an `int`) than to sort
through the list of strings you'd get with "globals_needed", and allows
renderers (if they so desire) to (a) verify that shaders don't use or
modify globals they aren't supposed to, for the type of shader use they
are being used for; (b) potentially to avoid any expensive computations
that would need to be performed to prepare certain shaderglobals fields,
for shader groups that won't end up using them.

Also make the `I` global variable be writeable and clarify that if a
global is writeable, it's writeable from any shader type, and we don't
intend to do differing checks based on shader type.

### DIFF
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -405,21 +405,35 @@ ShadingSystem::query_closure (const char **name, int *id,
 
 
 
+static OIIO::cspan< std::pair<ustring,SGBits> >
+sgbit_table ()
+{
+    static const std::pair<ustring,SGBits> table[] = {
+        { ustring("P"),       SGBits::P },
+        { ustring("I"),       SGBits::I },
+        { ustring("N"),       SGBits::N },
+        { ustring("Ng"),      SGBits::Ng },
+        { ustring("u"),       SGBits::u },
+        { ustring("v"),       SGBits::v },
+        { ustring("dPdu"),    SGBits::dPdu },
+        { ustring("dPdv"),    SGBits::dPdv },
+        { ustring("time"),    SGBits::time },
+        { ustring("dtime"),   SGBits::dtime },
+        { ustring("dPdtime"), SGBits::dPdtime },
+        { ustring("Ps"),      SGBits::Ps },
+        { ustring("Ci"),      SGBits::Ci }
+    };
+    return table;
+}
+
+
+
 SGBits
 ShadingSystem::globals_bit (ustring name)
 {
-    // **MUST** match the order of SGBits!!
-    static ustring globalnames[] = {
-        ustring("P"), ustring("I"), ustring("N"), ustring("Ng"),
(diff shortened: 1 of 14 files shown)


src/liboslexec/shadingsys.cpp
```diff
@@ -405,6 +405,48 @@ ShadingSystem::query_closure (const char **name, int *id,
 
 
 
+SGBits
+ShadingSystem::globals_bit (ustring name)
+{
+    // **MUST** match the order of SGBits!!
```
fpsunflower: I'm not a big fan of these cases where things need to stay in sync. Since its a plain linear search in the end, why not set it up as:

`pair<ustring, SGBit> globals[]`

So that you can list the name and corresponding `SGBit` together (both functions can share the same table)

src/liboslexec/shadingsys.cpp
```diff
@@ -405,6 +405,48 @@ ShadingSystem::query_closure (const char **name, int *id,
...
+    static ustring globalnames[] = {
+        ustring("P"), ustring("I"), ustring("N"), ustring("Ng"),
+        ustring("u"), ustring("v"), ustring("dPdu"), ustring("dPdv"),
+        ustring("time"), ustring("dtim"), ustring("dPdtime"), ustring("Ps"),
```
fpsunflower: typo: `"dtime"`

fpsunflower:
LGTM

PR #900, opened 2018-07-18
Fix incorrect array length on unbounded arrays specified via relaxed parameter type checking

fpsunflower:
If you tried to pass in something like:
`float[9] values 1 0 0 0 1 0 0 0 1`

To a shader that had `color values[]` declared, OSL would understand that as:
`color[9] values ...`

Instead of the expected:
`color[3] values ...`

This only comes up if you are using relaxed type checking of course.


lgritz:
LGTM!

Issue #723, opened 2017-03-15
OSLC performance problem when building shaders with large static arrays


luisbarrancos:
When implementing a value noise function in OSL for [appleseed](https://github.com/appleseedhq) with the use of a large table (64k) of RNG numbers generated via drand48 and a known seed in order to match values with a DCC app, i noticed the resulting shader would take a considerable amount of time to compile, on average 26 seconds.
Three runs:
```
real	0m26.780s
user	0m26.686s
sys	0m0.051s
```
```
real	0m27.068s
user	0m26.918s
sys	0m0.060s
```
```
real	0m27.343s
user	0m27.189s
sys	0m0.052s
```
If using large static arrays would be unavoidable, what would be the best way to address this in OSL, and is this a known case?
There are other shaders involving the use of this RNG table and the cumulative effect is somewhat considerable.

The shader was built with OSL 1.7.6, LLVM 3.5 on Fedora 23. Zipfile attached.
[value_noise_test.zip](https://github.com/imageworks/OpenShadingLanguage/files/844670/value_noise_test.zip)

fpsunflower:
The table is not used in `value_noise_2d` and removing it cuts the time roughly in half. This suggests the bottleneck is just in the parsing code somewhere.

I have a feeling there is just an `O(n^2)` type bug hiding somewhere, it should be possible to fix.

fpsunflower:
Yup the whole thing is stuck on `ASTNode::append` which does a linear search to add each element to the end of the array. I'm sure there will be an easy way to reorganize this, Larry is more familiar with this code to suggest the proper fix though.

dictoon:
> I have a feeling there is just an O(n^2) type bug hiding somewhere
Yup, that was my immediate suspicion when I saw the problem. Immediately reminded me of this blog: https://accidentallyquadratic.tumblr.com/
:)

dictoon:
Hello! This is still an everyday annoyance for us: The time it takes to compile appleseed is _largely_ dominated by the time it takes to compile a handful of OSL shaders. Any chance for a fix?

boberfly:
Hi all. Recently I tried building Appleseed's shaders against OSL 1.9.8 + LLVM 5.0.1 to match what Gaffer has defaulted to on Windows, and these shaders actually do generate but crash on exit which fails the build process. 

[...]

PR #901, opened 2018-07-19
Speed up parsing of long initializer lists.

lgritz:
If you have something like

    float table[65536] = { 0.1, 0.35, -0.72, ... }

This might take TENS of seconds for oslc to parse, because the mechanics
of how that long list of values is parsed -- it involves a singly-linked
list of ASTNodes, and with each new one, we are appending to the far end
of the linked list. This leads to an O(n^2) parsing time, where n is the
length of the list.

The straightforward (but wrong!) way to fix it is to change the Bison
grammar for init_expression_list from "left-recursive" to "right-recursive".
That makes it natural to prepend items (O(1)) to the accumulating list,
but it's very inefficient for Bison and can overrun Bison's fixed-size
parsing stack. Bison docs discurage use of right-recursion in grammars,
since left-recursion grammar can parse arbitrarily deep recursive
structures in a fixed memory size.

OK, so how do we fix this? Trickery! Keep left recursion, but accumulate
in the reverse order (i.e., prepend, not append), and then do a single
reverse of the backwards list at the very end. That makes the whole
parsing affair O(n) instead of O(n^2).

Sounds good, but one final wrinkle is that the list we need to reverse
is is a raw pointer to the head, followed by ref-counted pointers to
each subsequent element. In order to keep from accidentally freeing
things if the ref counts reach zero, we have our fancy reverse()
function rely on a brand spanking new feature of OIIO's intrusive_ptr<>
to add a release() method (https://github.com/OpenImageIO/oiio/pull/1986).
But since that doesn't exist in older OIIO, to avoid a compatibility
break, we only do our special left recursion + backwards list + reversal
trick if the OIIO version we build against has that new feature.
Otherwise, it's #ifdef'ed out and we just do it the old slow way.

This fixes OSL issue #723, and for the pathological test case presented
in that issue, it improves compile time (on my laptop) from ~26 seconds
to 0.4 seconds.

I also checked that this funny business with the reversal doesn't slow
down the simple cases for ordinary shader. Compiling the whole
collection of 500+ MaterialX shaders took about 20.5 seconds, and with
the new parsing code is exactly the same, or at least, too close to
accurately discern in the time trials.

### DIFF
--- a/src/liboslcomp/ast.cpp
+++ b/src/liboslcomp/ast.cpp
@@ -77,37 +77,17 @@ ScopeExit print_node_counts ([](){
 
 
 
-ASTNode *
-reverse (ASTNode *list)
+ASTNode::ref
+reverse (ASTNode::ref list)
 {
-    // trivial case: empty or just one item
-    if (!list || !list->nextptr())
-        return list;
-
-#ifndef OIIO_REFCNT_HAS_RELEASE
-    ASSERT (0 && "Do not call reverse() if you didn't build with an OIIO "
-                 "new enough to have intrustive_ptr::release().");
-    return nullptr;
-#else
-    // Turn the list (whose `next` fields are ref-counted pointers) into
-    // a vector of raw pointers, releasing the ref-counted pointers in the
-    // process (but don't free the nodes!).
-    std::vector<ASTNode *> nodes;
+    ASTNode::ref new_list;
     while (list) {
-        nodes.push_back (list);
-        list = list->m_next.release();
+        ASTNode::ref next = list->next();
+        list->m_next = new_list;
+        new_list = list;
+        list = next;
     }
-
-    // reverse the list
-    std::reverse (nodes.begin(), nodes.end());
-
-    // reassemble!
-    for (size_t i = 0, e = nodes.size(); i < e; ++i)
-        nodes[i]->m_next.reset ((i+1 < e) ? nodes[i+1] : nullptr);
-
-    // return the new head of list
-    return nodes[0];
-#endif
+    return new_list;
 }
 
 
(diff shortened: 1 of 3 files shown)


lgritz:
Fixes #723

src/liboslcomp/ast.cpp
```diff
@@ -76,6 +77,41 @@ ScopeExit print_node_counts ([](){
...
+    // trivial case: empty or just one item
+    if (!list || !list->nextptr())
+        return list;
+
+#ifndef OIIO_REFCNT_HAS_RELEASE
+    ASSERT (0 && "Do not call reverse() if you didn't build with an OIIO "
+                 "new enough to have intrustive_ptr::release().");
+    return nullptr;
+#else
+    // Turn the list (whose `next` fields are ref-counted pointers) into
```
fpsunflower: Why can't you do something like this?

```
ASTNode::ref reverse(ASTNode::ref root) {
  ASTNode::ref new_root = nullptr;
  while (root) {
    ASTNode::ref next = root->next;
    root->next = new_root;
    new_root = root;
    root = next;
  }
  return new_root;
}
```

Why do you need to allocate a whole new vector + do 3 passes?

lgritz: You're right, this is simpler.

lgritz:
Update pushed with simplified implementation of reverse(). Thanks for the suggestion.

But it's still the case that reverse returns a ref-counted pointer to the head, but at this point in the grammar we MUST have a raw pointer, so I still need to use the "dangerous" `intrusive_ptr<>::release()`.

fpsunflower:
Probably a much bigger change, but why does the grammar use raw pointers if the `ASTNode` objects are all meant to be reference counted?

lgritz:
Bison just works that way. The type returned by each grammatical construct is an undiscriminated union. You don't want to shove anything in their with a nontrivial ctr/dtr. The raw pointer is eventually converted to a managed pointer as I build the AST.

fpsunflower:
Ah I see. LGTM!

lgritz:
Thanks for all the tips. You always, always, make my code better.

PR #902, opened 2018-07-20
Device Strings

timgrant:
## Description

This is our second take on a GPU-friendly alternative to ``OIIO::ustring``. The main premise is that string constants reside in a block of CUDA device memory, with one entry for each canonical string. As with ``ustring``, two strings can be tested for equality by comparing their pointer values.

As a convenience, the length of the string and its hash value are stored in the 16 bytes preceding the characters in device memory, resulting in the following layout:

```
struct TableReb {
    size_t hash;
    size_t len;
    char  str[len+1];
};
```

The ``testoptix`` project has been updated to include a ``StringTable`` class that manages the device memory and creates OptiX variables for each of the strings.

As in take one, only static string operations are currently supported.

### Variables & STRDECLs
Creating variables is especially important for the set of "standard" strings (generally, members of the ``Strings`` namespace), to prevent raw pointer values from appearing in the compiled shaders, which helps make the generated PTX more friendly to OptiX's caching mechanisms.

The standard strings are enumerated in ``<OSL/strdecls.h>``. The STRDECLs are used to create the global variables for the strings in ``<OSL/device_string.h>``(which declares the variables) and ``StringTable::init`` (which creates the entries in the string table).

## Tests
The testsuite has been updated with some simple tests to test basic string operations (assignment, comparison, strlen, etc.).

## Checklist:


src/liboslexec/llvm_gen.cpp
```diff
@@ -386,6 +388,20 @@ LLVMGEN (llvm_gen_printf)
         }
     }
 
+    if (rop.use_optix() && arg > (format_arg + 2)) {
+        std::cerr << "WARNING: "
```
lgritz: Note to self: I never thought to do it before, but I bet I could "constant fold" printfs so that quite often they would collapse to 0 or 1 true runtime arguments.

src/testoptix/stringtable.cpp
```diff
@@ -0,0 +1,126 @@
+#include "stringtable.h"
```
lgritz: Let's put the usual license boilerplate here.

lgritz: This can be fixed later, but if we think StringTable is going to be useful to all renderers using OSL+OptiX, at some point we should probably move it from testoptix to liboslexec.

src/testoptix/stringtable.h
```diff
@@ -0,0 +1,100 @@
...
+#include <map>
+#include <utility>
+
+#include <OpenImageIO/ustring.h>
+#include <OSL/oslexec.h>
+
+#include <cuda_runtime_api.h>
+#include <optix_world.h>
+
+
```
lgritz: Should this class be in namespace OSL? (Less important if it's just for testoptix, more important if it's going to be "exported" later to be a utility useful for any renderer using OptiX+OSL.)

src/testoptix/testoptix.cpp
```diff
@@ -122,6 +122,8 @@ void getargs(int argc, const char *argv[])
                 "--debug", &debug1, "Lots of debugging info",
                 "--debug2", &debug2, "Even more debugging info",
                 "--runstats", &runstats, "Print run statistics",
+                "--saveptx", &saveptx, "Save the generated PTX",
+                "--warmup", &warmup, "Perform a warmup launch",
```
lgritz: Just for my edification: what's a warmup launch and under what circumstances is it helpful?

src/testoptix/stringtable.h
ee33: I think StringTable should go into OSL namespace if and when it goes into OSL proper (vs textoptix).  Which, as you say, we can do later.  For now it's in testoptix which is presumably "copy and modify" code for anyone using OptiX+OSL.

src/cmake/modules/FindOptiX.cmake
```diff
@@ -69,6 +68,13 @@ set(OPTIX_VERSION_STRING ${OPTIX_VERSION_MAJOR}.${OPTIX_VERSION_MINOR}.${OPTIX_V
 
 if (OPTIX_FOUND)
     message (STATUS "OptiX version = ${OPTIX_VERSION_STRING}")
+    if (NOT OptiX_FIND_QUIETLY)
+        message (STATUS "OptiX includes  = ${OPTIX_INCLUDE_DIR}")
+        message (STATUS "OptiX libraries = ${OPTIX_LIBRARIES}")
+    endif ()
+    if (${OPTIX_VERSION_MAJOR}.${OPTIX_VERSION_MINOR} STRLESS "5.1")
```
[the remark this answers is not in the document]
timgrant: That does seem a bit cleaner. I will update it.

src/liboslexec/backendllvm.cpp
```diff
@@ -416,9 +428,17 @@ BackendLLVM::createOptixVariable(const std::string& name, const std::string& typ
     // Refer to the OptiX API documentation and optix_defines.h in the OptiX SDK
     // for more information.
 
+    ASSERT (use_optix() && "This function is only supported when using OptiX!");
+
+    auto mangle_name = [](const std::string& name, const std::string& prefix) {
+        return "_ZN" + std::to_string (prefix.size() + 13) + "rti_internal_" +
```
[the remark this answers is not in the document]
timgrant: I always forget about those handy formatting functions. I will fix it.

src/liboslexec/llvm_gen.cpp
```diff
@@ -358,12 +358,14 @@ LLVMGEN (llvm_gen_printf)
...
-                    // In the OptiX case, we need to use the device-side string constant,
-                    // as added to the LLVM Module.
-                    llvm::Value* loaded = (simpletype.basetype == TypeDesc::STRING &&
-                                           sym.is_constant() && rop.use_optix())
-                        ? rop.getOrAllocateLLVMGlobal (sym)
-                        : rop.llvm_load_value (sym, 0, arrind, c);
+                    llvm::Value* loaded = nullptr;
+                    if (rop.use_optix() && simpletype.basetype == TypeDesc::STRING) {
+                        // In the OptiX case, we use the device-side string.
+                        loaded = rop.llvm_load_device_string (sym);
```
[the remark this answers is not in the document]
timgrant: I think I had a TODO in there at some point, which I will re-add.

src/liboslexec/backendllvm.cpp
```diff
@@ -430,35 +450,61 @@ BackendLLVM::createOptixVariable(const std::string& name, const std::string& typ
...
     addGlobalVariable (mangle_name (name, "semantic"  ), 1, 1, &empty, "");
     addGlobalVariable (mangle_name (name, "annotation"), 1, 1, &empty, "");
-
-    return g_var;
 }
 
 
 
 llvm::Value *
 BackendLLVM::getOrAllocateLLVMGlobal (const Symbol& sym)
```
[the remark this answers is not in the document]
timgrant: I will do a pass to make this more clear.

src/testoptix/testoptix.cpp
timgrant: There is overhead associated with the first CUDA launch that I wanted to isolate from the actual rendering time, to get a better sense of what kind of performance to expect in an interactive context.

src/testoptix/stringtable.h
timgrant: StringTable is kind of a minimal working example. It needs a bit of work to reach library-grade status.

src/testoptix/testoptix.cpp
lgritz: Total aside: is there someplace I can learn more about this... how big the "first batch" penalty is, and ways to mitigate it? Is it first batch per app? Per kernel?

src/testoptix/stringtable.h
lgritz: Gotcha. Fine, then. I was thinking that if it's substantially what any OptiX renderer would need, we would want them to use it, not duplicate it. But if it's a minimal example, that's fine. What else would a real renderer want from this class?

lgritz:
Pending the several minor suggestions I made, this looks good. Let me know when you've updated with whatever you wish to add to this PR, and I'll do the merge.

Also, if you can rebase to the current master, that would be helpful. Be careful, when I tried, there were some minor merge conflicts that had to be touched up by hand, but it wasn't too bad.

timgrant:
Thanks, Larry. I've followed your suggestions and will perform the rebase shortly.

timgrant:
I did the rebase a couple of days ago, is that showing up in the review?

lgritz:
Yes, sorry, I missed it!
I will do a quick re-test here and merge if nothing breaks on the work machine.

lgritz:
merged, thanks

PR #903, opened 2018-07-20
MaterialX support

lgritz:
Submitted on behalf of Derek Haase. Support for MaterialX 1.36 -- support for lots of new (or previously missing) functionality, especially extending a number of the math operators to certain vector types.

Also, "exponent" has been renamed "power", and "pack" has been renamed "combine".


lgritz:
LGTM, merging

PR #904, opened 2018-07-25
Add sign() and log() to vector struct types.

lgritz:
Ass missing `sign` node to MaterialX.
Change MaterialX `ln` implementation from log2 to correct log.


aconty:
LGTM!

PR #905, opened 2018-07-25
MaterialX 1.36 additions

lgritz:
New nodes for MaterialX 1.36:
* transformpoint, transformvector, transformnormal
* rgbtohsv, hsvtorgb

Also adds to our contributed shader headers:
* vector4 version of transform (by space name or by matrix)
* color4 version of transformc (converts the color portion, leaves alpha alone)


lgritz:
Amended PR with additional changes:

* Add new MX 1.36 node "rotate", mark old "rotate2d" as deprecated.

* Add new MX 1.36 node "hsvadjust", mark old "hueshift" as deprecated.

* stdosl.h: add a new variant of rotate() that takes just an axis and rotates around the origin using that axis (the other, previously existing, version or rotate() takes two arbitrary points and rotates about the line connecting them).

lgritz:
Amended again:  add new MX 1.36 "extract" and "separate" nodes.

PR #907, opened 2018-07-30
MaterialX cleanup: ramp4, texture mapping, matrix math

lgritz:
More work to conform the MaterialX shaders to the newly published 1.36 spec.

* Fix definition of ramp4 top vs bottom.   In recent times, it's been decided that "top" always refers to v=1 and  "bottom" refers to v=0.

* Extend add/sub/mul/div to include matrices

* Various clean ups to texture mapping nodes:

  - image node: wasn't honoring addressmode or filtertype params.
  - Fixes to triplanarprojection: wasn't properly normalizing by the      combined weight of the three views, wasn't honoring filtertype param,      wasn't forcing periodic wrap.
  -  Implement the new "tiledimage" node.


lgritz:
Amended: extend add/sub/mul/div to include matrices per 1.36 spec

PR #909, opened 2018-08-07
MaterialX build strategy overhaul

lgritz:
* Radically simplify build_materialX_osl.py -- It previously had a whole
lot of unused complexity by being able to build multiple template
expansions at once, and to do ALL template expansions without the cmake
system. But we never used it that way. So simplify it to handle a single
shader and type at a time, which is how it's called from our cmake
scripts. This makes it easier to maintain and removes a lot of redundant
logic that was duplicated (possibly incorrectly) between cmake and
python.

* Modify MaterialX build process to handle "OTHERTYPE" so we can get a
templated cross-product of two type lists (though by default, OTHERTYPE
is "none", so it's just a linear list of types we expand to). OTHERTYPE
is also allowed to be "same" which means it's the same as the first TYPE.
This makes it easy to make a templated node where for a bunch of types T,
it has a T,T version, and also a T,float version.

* Implement MX "convert" node, which utilizes the ability to have two
templated types.

* Clean-up of lots of nodes where we special cased the vecN/colN versus
itself, versus vecN/colN versus float. Now that we can template on two
types, these all simplify to one template for the "same" case and the
"float" case.

* File/node naming convention: for all nodes that can be templated on
a single type, we continue to call the OSL shader mx_thefunction_thetype.osl
(and .oso). But for all nodes that can be templated on two types,
(add, smoothstep, etc.), they will be named mx_thefunction_type1_type2.osl.


pleprince:
Hi Larry,
The README is out of sync with these changes. It would be nice to fix it.
Cheers

lgritz:
Yes, thanks for the reminder

lgritz:
Fixed and pushed to master

PR #910, opened 2018-08-17
LLVM 7.0 and clang 7 compiles

lgritz:
* Modernize the way we generate memset and memcpy ops -- we still had
  some artifacts of LLVM 3.x; since LLVM 4+, there has been a more
  direct way to generate the equivalent of memset/memcpy (and be
  careful, the API changes slightly for LLVM 7.0).

* Some header file changes with llvm 7.

* Minor ref-vs-ptr miscellany for WriteBitcodeToFile and CloneModule.

* Slight change to addPassesToEmitFile call signature.


src/liboslexec/llvm_util.cpp
```diff
@@ -1149,62 +1154,40 @@ LLVM_Util::op_return (llvm::Value *retval)
...
-    llvm::Value* fill_val = llvm::ConstantInt::get (context(),
-                                                    llvm::APInt(8, val));
-    // Non-volatile (allow optimizer to move it around as it wishes
-    // and even remove it if it can prove it's useless)
-    llvm::Value *args[5] = {
-        ptr, fill_val, len, constant(align), constant_bool(false)
-    };
-    builder().CreateCall (func, llvm::ArrayRef<llvm::Value*>(args, 5));
+    builder().CreateMemSet (ptr, builder().getInt8((unsigned char)val),
+                            uint64_t(len), (unsigned)align);
```
fpsunflower: Are you sure this is right? Casting `llvm::Value*` to `uint64_t` isn't what you want, right ?

fpsunflower: I would also add that it seems confusing to have both this version and the overload above.

src/liboslexec/llvm_util.cpp
```diff
@@ -1591,8 +1582,14 @@ LLVM_Util::ptx_compile_group (llvm::Module* lib_module, const std::string& name,
     llvm::raw_svector_ostream assembly_stream (assembly);
 
     // TODO: Make sure rounding modes, etc., are set correctly
+#if OSL_LLVM_VERSION >= 70
     target_machine->addPassesToEmitFile (mod_pm, assembly_stream,
+                                         nullptr,  // FIXME: Correct?
```
fpsunflower: I look at the code, this seems correct.

src/liboslexec/llvm_util.cpp
lgritz: Quite right. For that one, I just wanted to pass the Value* directly.

src/liboslexec/llvm_util.cpp
lgritz: Looks right to me, too, but since I can't test the combination of OptiX/PTX and LLVM 7.0, I left this in so that if something breaks, we know I wasn't super sure.

src/liboslexec/llvm_util.cpp
lgritz: updated

lgritz:
I don't expect any measurable speed gain. I think it's just a matter of better notation since real API calls were added instead of hoop jumping with intrinsics.

svenstaro:
Can you backport this to the 1.9 branch please?

lgritz:
backport proposed in #919

PR #915, opened 2018-09-14
Remove use of a fixed path to call bash

sambler:
## Description

Not every system has bash installed in /usr using env to call bash is a better cross platform method.

## Checklist:


lgritz:
Good catch. LGTM.

PR #916, opened 2018-09-14
Fix testrender bug with undefined order of operations

lgritz:
A suspicious line in testrender seems to depend on undefined behavior
regarding the specific order of evaluation of quantities in a math expression.
It has been there forever, but only seems to fail on certain builds of gcc 5
and possibly clang 7.

It's just a one-line fix to remove the order-of-evaluation ambiguity.

Thanks, Tim Grant, for pointing this out!


fpsunflower:
LGTM!

PR #917, opened 2018-09-19
Minor changes to build with clang 7.

lgritz:
The changes in oslcomp.cpp seem only to be necessary with particular
combinations of clang7 and older llvm. Which we don't recommend!  Better
to use the same generation of clang and llvm. But in any case, may as
well make set those options explicitly.

Also switch SPI to build with clang7/llvm7 by default.


svenstaro:
Can you backport this to the 1.9 branch please?

lgritz:
yes, will do

lgritz:
backport proposed in #919

PR #918, opened 2018-09-21
Add Pixel magnifier

marsupial:
Still a bit messy, but does the job.
Slider is in there to either provide a clamp/normalization or choose magnification level.
Does neither yet, so dealer's choice.

### DIFF
--- a/src/osltoy/osltoyapp.cpp
+++ b/src/osltoy/osltoyapp.cpp
@@ -103,9 +103,10 @@ public:
     }
 };
 
-static void setBackground(QWidget* widget, QColor color) {
+static void setBackground(QWidget* widget, QColor bg, QColor fg) {
     QPalette palette = widget->palette();
-    palette.setColor(widget->backgroundRole(), color);
+    palette.setColor(widget->backgroundRole(), bg);
+    palette.setColor(widget->foregroundRole(), fg);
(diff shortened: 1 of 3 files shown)


lgritz:
This is neat. But I'm also in the process of adding ways for users to interact with the image, including with the mouse, and maybe you don't want that other stuff in the way. I'm thinking that this "zoom/inspect" functionality ought to be something you can toggle. Maybe it is only there while you're holding down the Ctrl key, say? Or only if you click an "inspector" checkbox? I think it's important to be able to play Lee's pong shader game. :-) 

Not sure about the the magnification yet. If it's not functional, let's at least hide it from the UI for now.

lgritz:
On both OSX and Linux, when I build this, I can see the pixel zoom, but there's no visible text?

marsupial:
Some times one can just fall in love with Qt

lgritz:
OK, LGTM for now. I may tinker with the UI when I add some other changes on top of it that I am in the middle of. Have some more interesting features coming.

marsupial:
Tinker away.  If not obvious, as it is, stashing the Magnifier in a RenderView was easiest, but if not was envisioning possibly multiple RenderViews in which case it should probably have a home in the Main Window.

PR #919, opened 2018-09-23
Support clang7 and llvm7 in OSL 1.9

lgritz:
Backported changes from master so that further 1.9 releases are also compatible with llvm 7.0 and clang 7.


svenstaro:
Works for me!

PR #922, opened 2018-10-05
Fix oslc read/write error for regex_search/regex_match

lgritz:
For these functions, oslc failed to mark the results output param as write only.

Fix the bug in oslc, and also put in a provision in loadshader that will correct the old already-compiled incorrect files.

Thanks to Steena Monteiro from Intel for pointing this out!

### DIFF
--- a/src/liboslcomp/typecheck.cpp
+++ b/src/liboslcomp/typecheck.cpp
@@ -1261,6 +1261,7 @@ ASTfunction_call::typecheck_builtin_specialcase ()
     }
 
     if (func()->readwrite_special_case()) {
+        int nargs = (int)listlength(args());
         if (m_name == "sincos") {
             argwriteonly (1);
             argwriteonly (2);
@@ -1268,11 +1269,16 @@ ASTfunction_call::typecheck_builtin_specialcase ()
                    m_name == "gettextureinfo" || m_name == "getmatrix" ||
                    m_name == "dict_value") {
             // these all write to their last argument
-            argwriteonly ((int)listlength(args()));
+            argwriteonly (nargs);
         } else if (m_name == "pointcloud_get") {
             argwriteonly (5);
         } else if (m_name == "pointcloud_search") {
             mark_optional_output(5, pointcloud_out_args);
+        } else if ((m_name == "regex_search" || m_name == "regex_match")
+                   && nargs == 3) {
+            // the kind of regex_search and regex_match that contains a
+            // results argument should mark it writeable.
+            argwriteonly (2);
         } else if (m_name == "split") {
             argwriteonly (2);
         } else if (func()->texture_args()) {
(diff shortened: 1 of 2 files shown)


fpsunflower:
LGTM

AlexMWells:
I just ran into a situation where the 2 parameter versions of regex_search(pattern, subj) were being forced to mark subj as write only, even though it is read only.  This in turn cause ops that created subj to be elided (as the optimizer thought no one was reading them.  And regex_search executed with a nullptr for subj and segfaulted.

Here is a suggested fix for the part in loadshader.cpp:
```
        if ((opname == regex_search || opname == regex_search) && m_nargs > 3) {
            // Be sure to not mark the 2 parameter versions as write only
            //     int regex search (string subject, string regex)
            //     int regex match (string subject, string regex)
            // by making sure the number of args is > 3
             m_master->m_ops.back().argwriteonly (2);
        }
```

lgritz:
I don't think that's correct, Alex.

There are never > 3 args, because nargs doesn't count the return value, only the parameters.

When there are 3 args, arg 1 is the output array to write to. When there are two args, neither parameter is written to.

I looked at the oso resulting from testsuite/string, which tests both the 2 and 3 arg versions, and the results in the oso file look correct to me.

AlexMWells:
The oso file is fine.  This is in the loadshader.cpp which has some code to hammer down older oso that might have it wrong.  And that hammer didn't check for the number of arguments/parameters at all.
diff --git a/src/liboslexec/loadshader.cpp b/src/liboslexec/loadshader.cpp
index b55e45f..4c098d5 100644
--- a/src/liboslexec/loadshader.cpp
+++ b/src/liboslexec/loadshader.cpp
@@ -406,8 +406,13 @@ OSOReaderToMaster::hint (string_view hintstring)
         // This was a bug prior to 1.10.
         static ustring regex_search("regex_search");
         static ustring regex_match("regex_match");
-        if (opname == regex_search || opname == regex_search)
+        if ((opname == regex_search || opname == regex_search) && m_nargs > 3) {
+            // Be sure to not mark the 2 parameter versions as write only
+            //     int regex search (string subject, string regex)
+            //     int regex match (string subject, string regex)
+            // by making sure the number of args is > 3
             m_master->m_ops.back().argwriteonly (2);
+        }
         return;
     }
     if (Strutil::parse_prefix(h, "%argderivs{")) {

AlexMWells:
In this code, m_nargs I think includes the return type, so int, string, string would be 3.  int, string, int *, string would be 4.  All I can say for sure is that loading the 2 param version hand m_nargs == 3.

lgritz:
OH, yes I see. I was looking in the original code in typecheck.cpp.
Patch coming up.

AlexMWells:
For reference/testing here is the test shader that was segfaulting:
```
shader u_subj_v_pattern (output vector cout = 0)
{

  string u_subj = "electric blue";
  string v_pat1 = concat(format("%g", P[0]), "ham");
  string v_pat2 = concat(format("%g", P[0]), "ue");
  string v_pat3 = concat(format("%g", P[0]), "$el");
  string v_pat4 = concat(format("%g", P[0]), "el[a-b]");
  string v_pat5 = concat(format("%g", P[0]), ".*tric");
  string v_pat6 = concat(format("%g", P[0]), "blue");
  

 
   
   int a = -1;
   int b = -1;
   int c = -1;
   int d = -1;
   int e = -1;
   int f = -1;


  a =  regex_match (u_subj, v_pat1);         // should not match
  b =  regex_match (u_subj, v_pat2);  // should match
  c =  regex_match (u_subj, v_pat3);
  d =  regex_match (u_subj, v_pat4);
  e =  regex_match (u_subj, v_pat5);       // should match
  f =  regex_search (u_subj,v_pat6);

  if( (a >=0) && (b>=0) && (c>=0) && (d>=0) && (e>=0) && (f>=0))
{
  cout = vector(a+b/u, tan(c+d/v+u), tan(e*f/v*u));
}  

}
```

 v_pat6 is treated as never being read and the concat that writes to it is elided, the subsequent regex_search taking vpat6 then crashes on the nullptr.

lgritz:
Also, it seems obvious that one of those two should be regex_match!

lgritz:
Proposed solution in #1186

PR #923, opened 2018-10-08
Rewrite OSOReaderToMaster::hint

lgritz:
Zap is seeing sporadic (but frustraingly impossible to reproduce) crash
reports from inside this function, specifically in the part that parses
%argrw hints.

There was a longstanding comment:   // FIXME -- use string_view ops here

So, sure, rewrite the whole thing using string_view and Strutil::parse
functions, and remove the assertions, and put extra guards to make sure
we aren't acessing a symbol or op (via blah.back()) if there aren't any
in the vector yet, and pretty much just let it ignore or be robust to any
malformed hints.

I've never seen a repro case, so I don't know that this will fix it, but
it's programmed a lot more defensively now, so hopefully there will be
no further crashes (after I backport this to whatever version Zap is
using).  I even put in an error message for a certain way to malform
%argrw, in the hopes that it might give us a clue, but even that is just
a guess.

### DIFF
--- a/src/liboslexec/loadshader.cpp
+++ b/src/liboslexec/loadshader.cpp
@@ -331,108 +331,70 @@ OSOReaderToMaster::parameter_done ()
 
 
 
-inline bool
-starts_with (const std::string &source, const std::string &pattern)
-{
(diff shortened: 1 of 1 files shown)


aconty:
LGTM!

PR #925, opened 2018-10-12
Change to search path rules for .oso files

lgritz:
New rules:

* If NO "searchpath:shader" attribute is specified to the ShadingSystem,
  the current working directory (".") will be searched.

* If a searchpath is specified, it will look ONLY in those directories,
  in order. The current working directory will only be searched if
  it is listed among the paths.


fpsunflower:
LGTM

PR #926, opened 2018-10-15
Fixes to make recent osltoy changes not break against OIIO 1.8

lgritz:
Recent changes used a new ImageBufAlgo convention only present in current
OIIO master, not older 1.8 that we still promise compatibility with.

Didn't catch this before because the Travis tests don't install qt and
therefore don't build osltoy at all. Add qt to the Mac travis setup so
that it at least builds to prevent this from happening.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 
 language: cpp
 sudo: false
-osx_image: xcode7.3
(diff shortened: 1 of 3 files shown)


aconty:
LGTM

PR #927, opened 2018-10-25
Default to use vanilla boost

marsupial:
Default Makefiles build to use vanilla boost.
Remove filesytem and wave from link flags, which seem no longer necessary.


lgritz:
LGTM.  (Note: this is an SPI-only change, won't affect anybody else, so don't panic.)

PR #929, opened 2018-10-29
Wean ourselves off of 'operator int' of ustring, and bug fixes

lgritz:
OIIO's ustring has an "operator int()" that is possibly ill-advised, as
claimed by Thiago Ize and John Haddon from an oiio-dev thread from back
in 2015 (I'm doing some old thread necromancy). They claimed that an
operator int (or for that matter bool) could result in subtle bugs from
type conversions you don't expect to happen.  The fact that there's no
equivalent operator int/bool for std::string or std::string_view
bolsters their argument.

I started off just trying to eliminate all OSL uses that assume this
operator would work (the proper idiom is simply `!str.empty()`). Along the
way, I found *TWO* bugs related to this -- one in osltoy and one
when constant folding "wrap" arguments to texture functions! So Thiago
and John were surely right, and I am appropriately chastened and
belatedly trying to find a way to get rid of the troublesome operator.

### DIFF
--- a/src/liboslcomp/oslcomp.cpp
+++ b/src/liboslcomp/oslcomp.cpp
@@ -949,7 +949,7 @@ OSLCompilerImpl::write_oso_file (const std::string &outfilename,
             lastline = -1;
         }
 
-        if (/*m_debug &&*/ op.sourcefile()) {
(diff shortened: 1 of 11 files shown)


aconty:
LGTM!

lgritz:
Backporting to 1.9.

PR #931, opened 2018-11-06
oslc: Fix type constructor dealing with ambiguity in function return types

### LINKED ISSUE #930: constructor cast doesn't fully work with return value type deduction
Using the latest OIIO and OSL master built with llvm 5.0.2 on linux, passing the following to oslc:

```
shader cell_diffuse(output closure color result = 0)
{
   result = diffuse(N) * cellnoise(u, v);
}
```

results in
```

osl_test.osl:3: warning: Ambiguous call to 'cellnoise (float, float)'
  Chosen function is:
    float cellnoise (float, float)
  Other candidates are:
    color cellnoise (float, float)
    vector cellnoise (float, float)
```

I'm OK with this warning since it is ambiguous. The odd part is that when I tried to fix it like this:

`   result = diffuse(N) * float(cellnoise(u, v));`

it still produces the same warning. If instead I do

`   result = diffuse(N) * (float)cellnoise(u, v);`

Then I correctly get no warning. According to the OSL spec I get the impression that a C-style typecast is the same as a C++-style type constructor, so I would think both versions should behave the same.

lgritz:
If you have a function that is polymorphic based on return type --
like cellnoise, which could return a float or a color -- you will get
a warning about the ambiguity in certain situations which you can resolve
with an explicit type cast:

    (float) cellnoise (u, v)    // the cast disambiguates which you want

But Thiago found that the similar idiom using a type constructor (rather
than a cast) like this:

    float (cellnoise (u, v))    // fails to disambiguate cellnoise?

fails to disambiguate and gives a warning about ambiguous types (it will
choose the float-returning kind, but will warn about the ambiguity, which
will be an error if you compile with `-Wall`).

This patch fixes the problem by making ASTtype_constructor::typecheck()
aware that float(expr) ought to expect expr to be a float if possible,
and also triple(expr,expr,expr) should expect all three of the channel
expressions to disambiguate to float. It previously made no such inference
in those cases. Note, however, that triple(expr) did previously know that
the expression ought to prefer to disambiguate to the same triple type.

While investigating, I also saw that the oso generated by that same
constructor sometimes did an extra unnecessary temporary assignment, so
while I was ast it I also fixed that with the changes you see to
ASTtype_constructor::codegen (in fact, these changes altered more code
than the type checking part).

Fixes #930

### DIFF
--- a/testsuite/typecast/test.osl
+++ b/testsuite/typecast/test.osl
@@ -51,4 +51,13 @@ test ()
         // point q[2] = { ... };
         // color c = (color) q[1];
     }
+
+    // Test some tricky cases -- these should not warn! The constructors
+    // ought to be able to figure out the desired type of the ambiguous
+    // function arguments.
+    closure color result = diffuse(N) * float(cellnoise(u, v));
+    color c1 = float(cellnoise(u,v));
+    color c2 = color(cellnoise(u,v));
+    color c3 = color(cellnoise(u,v),0.0,0.0);
+    color c4 = c1 * cellnoise(u,v);
 }
--- a/src/liboslcomp/codegen.cpp
+++ b/src/liboslcomp/codegen.cpp
@@ -1697,11 +1697,19 @@ ASTtype_constructor::codegen (Symbol *dest)
         // Doesn't fit the pattern, drop to the usual case...
     }
 
+    // Special case: construct float(float_expr) -- just put it directly
+    // in the dest.
+    Symbol *argevaldest = nullptr;
+    if (dest && typespec().is_float() &&
+          nchildren() == 1 && child(0)->typespec().is_float()) {
+        argevaldest = dest;
+    }
+
     std::vector<Symbol *> argdest;
     argdest.push_back (dest);
     int nargs = 0;
     for (ref a = args();  a;  a = a->next(), ++nargs) {
-        Symbol *argval = a->codegen();
+        Symbol *argval = a->codegen(argevaldest);
         if (argval->typespec().is_int() && !typespec().is_int()) {
             // Coerce to float if it's an int
             if (a->nodetype() == literal_node) {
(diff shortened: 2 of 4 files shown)


testsuite/typecast/test.osl
```diff
@@ -51,4 +51,13 @@ test ()
...
     }
+
+    // Test some tricky cases -- these should not warn! The constructors
+    // ought to be able to figure out the desired type of the ambiguous
+    // function arguments.
+    closure color result = diffuse(N) * float(cellnoise(u, v));
+    color c1 = float(cellnoise(u,v));
+    color c2 = color(cellnoise(u,v));
+    color c3 = color(cellnoise(u,v),0.0,0.0);
+    color c4 = c1 * cellnoise(u,v);
```
fpsunflower: This last one seems ambiguous to me. `c1 * float` and `c1 * color` seem equally valid options.

lgritz: We've always disambiguated `triple OP ambig` to prefer the triple interpretation.

fpsunflower: So `closure color * ambig` stays `ambig` but `color * ambig` resolves to `color`?

lgritz: I should fix that, too. I will amend.

fpsunflower: I like the ambiguity in the `closure color` case though.

I agree that `color + ambig` should probably resolve to `color`,  but `*` and `/` seem inherently ambiguous to me, and I think it would be better to keep the warning.

lgritz: Just to be clear, this review doesn't change anything about the `ambig OP color` case.

fpsunflower: Sure - just brought it up since the two cases were right next to each other in the test.

As far as this particular patch, no objections at all.

PR #932, opened 2018-11-08
Fix texture3d lookups with derivatives

fpsunflower:
One side was trying to make use of dresultdz while the llvm codegen was always providing nullptr. Since the shading system only supports Dual2 at the moment for derivative lookups, it seems more consistent to forget about trying to come up with a meaningful Z derivative for the time being.


aconty:
LGTM

lgritz:
LGTM2

lgritz:
Currently, OIIO's texture3d just does a trilinear lookup from the volume, it does not actually "filter" (let alone mipmap), so the dz was just passed around as a formality in case we wanted to expand the functionality later. This is a fine fix.

lgritz:
Merged and will backport.

Issue #895, opened 2018-06-30
osltoy (master) failing to build with link errors with msvc14


exavi:
### Problem

osltoy seems to not build correctly with MSVC14.

**Expected behavior:** 

Full build of osl, including osltoy.

**Actual behavior:**  

If I don't specify the Qt5 location, osltoy won't try to build, and everything works as excpected, I get usable libraries, however, when building with osltoy, it fails to build with the following error: 
(I can send a gist of the whole log, but this is the only error)

`osltoyapp.obj : error LNK2019: unresolved external symbol "public: __cdecl OSL_v1_10::OSLQuery::Parameter::Parameter(struct OSL_v1_10::OSLQuery::Parameter const &)" (??0Parameter@OSLQuery@OSL_v1_10@@QEAA@AEBU012@@Z) referenced in function "class std::shared_ptr<class OSL_v1_10::ParamRec> __cdecl std::make_shared<class OSL_v1_10::ParamRec,struct OSL_v1_10::OSLQuery::Parameter const &>(struct OSL_v1_10::OSLQuery::Parameter const &)" (??$make_shared@VParamRec@OSL_v1_10@@AEBUParameter@OSLQuery@2@@std@@YA?AV?$shared_ptr@VParamRec@OSL_v1_10@@@0@AEBUParameter@OSLQuery@OSL_v1_10@@@Z) [C:\libs\__sources\OpenShadingLanguage-master\OpenShadingLanguage-master\build_rel\src\osltoy\osltoy.vcxproj]  C:\libs\__sources\OpenShadingLanguage-master\OpenShadingLanguage-master\build_rel\src\osltoy\Release\osltoy.exe : fatal error LNK1120: 1 unresolved externals [C:\libs\__sources\OpenShadingLanguage-master\OpenShadingLanguage-master\build_rel\src\osltoy\osltoy.vcxproj]`

### Steps to Reproduce

I ran cmake directly, rather than make, not sure if this would make any difference, but it's worth mentioning, here is the command I'm running it with, not very useful but shows the versions I'm using for everything:

```
[...]

PR #933, opened 2018-11-18
Export struct OSLQuery::Parameter for Windows

darkhorse64:
## Description

On Windows, osltoy fails to link because of a missing OSLQuery::Parameter constructor. Even when a class is properly exported, nested structs or classes within this class must be exported too.

## Tests

Just made myself sure it compiles and links :-^


lgritz:
LGTM, thanks.

lgritz:
Fixes #895

PR #936, opened 2018-11-21
SS::optimize_group: add optional ctx pointer

lgritz:
If supplied, it uses the passed context if needed, rather than allocating a temporary one.

Suggestion by Olivier Paquet.


fpsunflower:
LGTM

PR #937, opened 2018-11-22
Improve error reporting when compiling from memory buffer

lgritz:
The `OSLCompiler::compile_buffer()` method, which is used to compile
osl source code from a memory buffer, had unhelpful error messages
such as:

    <buffer>:11: error: Unknown struct name: x

because of course, there was no filename, just a memory buffer. This
is ugly and confusing to users.

This patch adds an optional parameter to compile_buffer() that lets you
supply a "filename" to use for error reporting of the mem-buffer-compiled
shader. This makes pretty error messages if you supply a good name:

    zap.osl:11: error: Unknown struct name: x

An application of this is shown with the changes in osltoy.

For my own debugging purposes, I also implemented a `-buffer` option to
oslc, which just forces a file read, compile-from-buffer, and then file
write, so that it's easier for me to exercise the buffer code path.

Also, while I was there, I made the unrelated change of altering the
return value of `OSLCompilerImpl::retrieve_source` from a std::string
to a string_view, which should reduce the number of pointless string
copies during oso output.


lgritz:
Amendment: I pushed an update to *remove* the retrieve_source change. I'm going to do that (and some other enhancement) as a separate review.

fpsunflower:
LGTM

PR #938, opened 2018-11-22
Improve oso output efficiency with less redundant work on retrieve_source

lgritz:
To make debugging easier and be more human-readable, oso files are littered
with comment lines reproducing the source code, so that you can see which
source corresponds to which "oso assembly".

The workhorse here is the retrieve_source(filename,line) function. But
it was unspeakably inefficient, reading the entire source file every
time an op needed a different source file than the previous op, and
scanning the file to find the n-th line *every* time. This seems like
the lookup is O(n^2) and also that we're rereading source files
potentially thousands of times.

Theory: If we cache all the files we read, so we never read each source
file more than once, and also remember which line we used last (we tend
to go in increasing order in a source file) and resume the search from
there, backtracking only when the line is less than the last one needed,
it ought to be blazingly fast!

Practice: It barely makes any difference.  Compiling all 600 MaterialX
source osl files takes 30 seconds on my laptop, and this fix saves maybe
a second or two at most.

Oh well. But I already made the change, so may as well keep the more
efficient algorithm. Perhaps there are other cases where it gives a bigger
advantage.

### DIFF
--- a/src/liboslcomp/oslcomp.cpp
+++ b/src/liboslcomp/oslcomp.cpp
@@ -534,6 +534,7 @@ OSLCompilerImpl::compile (string_view filename,
     std::vector<std::string> includepaths;
     m_cwd = OIIO::Filesystem::current_path();
     m_main_filename = filename;
+    clear_filecontents_cache();
 
     read_compile_options (options, defines, includepaths);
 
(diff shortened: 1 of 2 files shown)


fpsunflower:
LGTM

lgritz:
Note: I'm going to merge this into master but not immediately into RB-2.0 (this therefore marks the first official divergence). Seems too risk to do immediately before a RC freeze. If there are no reported problems in master, I'll backport it to 2.0 before the next monthly update.

lgritz:
Sheesh, of course I meant 1.10, not 2.0. Mixing up the version numbers of the OIIO and OSL releases which in both cases I hope to release RC's today.

PR #939, opened 2018-11-26
testshade: check that no leftover errors are in the TS or IC

lgritz:
Before testshade exits, check the ImageCache and TextureSystem and ask for any remaining error messages. In theory, all should have been properly captured and sent to the ErrorHandler and nothing should be left. If there is (that is, texture errors went unnoticed), consider it an error of the testshade execution. This is an aid to debugging interactions between OSL and the texture system.

### DIFF
--- a/src/testshade/testshade.cpp
+++ b/src/testshade/testshade.cpp
@@ -999,11 +999,16 @@ test_shade (int argc, const char *argv[])
 {
     OIIO::Timer timer;
 
+    // Request a TextureSystem (by default it will be the global shared
+    // one). This isn't strictly necessary, if you pass nullptr to
+    // ShadingSystem ctr, it will ask for the shared one internally.
+    TextureSystem *texturesys = TextureSystem::create();
+
     // Create a new shading system.  We pass it the RendererServices
     // object that services callbacks from the shading system, NULL for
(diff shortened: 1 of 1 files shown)


src/testshade/testshade.cpp
```diff
@@ -1239,7 +1239,24 @@ test_shade (int argc, const char *argv[])
 
     // We're done with the shading system now, destroy it
     shadergroup.reset ();  // Must release this before destroying shadingsys
+
+    TextureSystem *texturesys = shadingsys->texturesys();
```
fpsunflower: It seems a little sneaky to extract the texture system from the shadingsys, delete the latter and then print out info from the texture system.

I realize it works because this code is setup for the shared case, but is it maybe worth moving the `delete shadingsys` down a bit just in case? Or do you specifically need to catch errors that could have occurred during the destruction process?

lgritz: Exactly, I wanted to catch anything that happens during SS teardown.

But you are right. What I can do is ask for the TS up front, pass it to the SS ctr rather than implicitly relying on the shared one. Then I'll know that the TS outlives the SS.

lgritz: amended

fpsunflower:
LGTM!

PR #944, opened 2018-12-13
Demote writing to non-output function parameters to be a warning.

lgritz:
It was never supposed to be allowed. It used to be ignored in most
cases, error neither detected nor issued. PR #878 "fixed" it to detect
the error. But there were too many shaders that did it, users are
confused and it breaks things.

So this patch downgrades it from an error to a warning. If you ignore
the warning, you get what you deserve (which is probably not
bad... people lived with it for years). Anybody who wants it to be fully
illegal can always compile with -Werror to turn warnings in to full
errors.

I also "uniformized" two separate error messages on the same general
topic, so they look the same.

### DIFF
--- a/src/liboslcomp/ast.cpp
+++ b/src/liboslcomp/ast.cpp
@@ -943,10 +943,11 @@ ASTNode::check_symbol_writeability (ASTNode *var)
 
     if (dest) {
         if (dest->readonly()) {
-            if (OSL_VERSION >= 11001)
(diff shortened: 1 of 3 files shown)


lgritz:
Got ok offline from Chris and Zap says it works for him. Merging.

lgritz:
Zap, I'll backport to RB-1.10 as soon as the CI tests complete.

PR #945, opened 2018-12-13
Remove long-deprecated methods

lgritz:
These were marked as deprecated in 1.8. The latest release is 1.10. Let's remove it from master.

### DIFF
--- a/src/include/OSL/oslcomp.h
+++ b/src/include/OSL/oslcomp.h
@@ -41,9 +41,6 @@ namespace pvt {
 
 class OSLCOMPPUBLIC OSLCompiler {
 public:
-    OSL_DEPRECATED("Directly construct or new an OSLCompiler")
-    static OSLCompiler *create ();
-
     OSLCompiler (ErrorHandler *errhandler=NULL);
     ~OSLCompiler ();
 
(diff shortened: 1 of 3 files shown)


fpsunflower:
Any reason not to remove the other deprecated methods?

lgritz:
sure, may as well get the others

lgritz:
I only found one additional declaration (which was so deprecated that apparently I'd already removed the implementation).

Did you have something particular in mind?

fpsunflower:
I see 16 occurrences of `DEPRECATED(1.8)` in my local checkout ...

fpsunflower:
Ahem ignore me, I was looking at OIIO, not OSL :P

lgritz:
Aha, gotcha. Yeah, the OIIO 1.8 deprecations won't be removed for a while.

Usually on the version it mentions, it's just a comment "don't use this."
On the following major release, I use the OIIO_DEPRECATED macro ([[deprecated]] underneath, or the equivalent), which makes it a compile error (unless you turn it off)

lgritz:
sorry, wasn't finished...
Then on the release AFTER THAT, I actually remove it.

lgritz:
Is this ok, then, @fpsunflower ?

fpsunflower:
Yep - sounds good!

PR #946, opened 2018-12-13
Change all calls to Strutil::format to Strutil::sprintf

lgritz:
This is related to OIIO 2.0 adding a new "sprintf" function that currently is a synonym for "format", but we anticipate that format will one day be modified to conform to an upcoming C++20 std::format that uses a different notation for formatting commands (more like python string formatting, quite nice).

In oslconfig.h, if the version of OSL is too old to have sprintf, define it as a synonym for format.

### DIFF
--- a/src/liboslcomp/codegen.cpp
+++ b/src/liboslcomp/codegen.cpp
@@ -648,16 +648,16 @@ ASTNode::one_default_literal (const Symbol *sym, ASTNode *init,
         completed = false;
     } else if (type.is_int()) {
         if (islit && lit->typespec().is_int())
-            out += Strutil::format ("%d", lit->intval());
+            out += Strutil::sprintf ("%d", lit->intval());
         else {
             out += "0";  // FIXME?
             completed = false;
         }
     } else if (type.is_float()) {
         if (islit && lit->typespec().is_int())
-            out += Strutil::format ("%d", lit->intval());
+            out += Strutil::sprintf ("%d", lit->intval());
         else if (islit && lit->typespec().is_float())
-            out += Strutil::format ("%.9g", lit->floatval());
+            out += Strutil::sprintf ("%.9g", lit->floatval());
         else {
             out += "0";  // FIXME?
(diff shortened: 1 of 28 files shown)


src/liboslcomp/codegen.cpp
```diff
@@ -648,27 +648,27 @@ ASTNode::one_default_literal (const Symbol *sym, ASTNode *init,
...
             out += "0";  // FIXME?
             completed = false;
         }
     } else if (type.is_float()) {
         if (islit && lit->typespec().is_int())
-            out += Strutil::format ("%d", lit->intval());
+            out += Strutil::sprintf ("%d", lit->intval());
         else if (islit && lit->typespec().is_float())
-            out += Strutil::format ("%.8g", lit->floatval());
+            out += Strutil::sprintf ("%.8g", lit->floatval());
```
fpsunflower: Not related to this refactor obviously, but why are these `%.8g` and not `%.9g` ?

lgritz: I can only guess that at the time I thought 8 was plenty? I'll fix that while I'm at it.

lgritz: %.9g is the minimum that's guaranteed to be full precision for all fp32 code points?

lgritz: Actually, there are number of %.8g instances in the code base. I think I'll make that fix a separate PR to isolate it.

fpsunflower: Yes.

https://randomascii.wordpress.com/2013/02/07/float-precision-revisited-nine-digit-float-portability/

lgritz: I'm not going to fix the precision for this patch, that will be the next one.

lgritz:
Is this ok to merge? (The precision issue you pointed out was already merged separately.)

fpsunflower:
Yep. LGTM!

PR #947, opened 2018-12-17
Fix bug in oslc loop code generation

### LINKED ISSUE #776: while statement doesn't work correctly
### Problem

**Expected behavior:** 
"while" statement should support a conditional that includes only a variable.
Now "if " statement in OSL works.
**Actual behavior:**  
while statement doesn't work if I used C-style conditional.

### Steps to Reproduce

1. if I used only a variable as conditional,while statement didn't work correctly.
```
int hit=1;
while(hit){
    hit = 0;
}
```
2. To work it,I had to specify conditional using a comparison operator explicitly.
```
int hit=1;
while(hit==1){
   	hit = 0;
}
```

lgritz:
The problem was specific to using an integer variable directly as the
loop test, rather than a comparison. So for example, this would work
fine:

    i = 3;
    while (i > 0) {
        --i;
    }

But this would loop forever:

    i = 3;
    while (i) {
        --i;
    }

Because of a hilarious chain of events, this would eventually lead to
the optimizer thinking that since the value of i was nonzero at the
beginning of the loop, it must be equivalent to while(1). There was a
subtle assumption that the scope of the loop control variable would be
entirey inside the condition code. That was true of the the temporary
created to hold the result of "i > 0", but not true of the direct use of
i itself.  The solution is to force it to create a bool comparison
(which turns into the same machine code later anyway, so there is no
efficiency issue here).

Fixes #776

### DIFF
--- a/src/liboslcomp/codegen.cpp
+++ b/src/liboslcomp/codegen.cpp
@@ -1414,7 +1414,8 @@ ASTloop_statement::codegen (Symbol *)
     codegen_list (init());
 
     int condlabel = m_compiler->next_op_label ();
-    Symbol *condvar = cond()->codegen_int ();
+    Symbol *condvar = cond()->codegen_int (/*dest=*/    nullptr,
+                                           /*boolify=*/ true);
 
     // Retroactively add the argument
     size_t argstart = m_compiler->add_op_args (1, &condvar);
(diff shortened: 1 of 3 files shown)


aconty:
LGTM

PR #948, opened 2018-12-19
Fix array initialization counter

marsupial:
## Description
Fixes the the element count checking that occurs when initializing an array.

## Tests
Added oslc-err-initlist

### DIFF
--- a/src/liboslcomp/typecheck.cpp
+++ b/src/liboslcomp/typecheck.cpp
@@ -816,7 +816,9 @@ public:
         // Every element of the array is the same type
         TypeSpec elemtype = expected.elementtype();
 
-        int nelem = 0;
+        // Start at 1, as oslc would have already failed in either the case of
+        // an empty initializer list, or zero-length array.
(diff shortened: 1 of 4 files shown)


lgritz:
Oh, so the checks were there, but it was just an off-by-one error?

marsupial:
Yeah, curios if there's any obvious cases I'm missing to test further?

lgritz:
LGTM. I'll merge as soon as the CI tests are finished.

I don't suppose you already tried on your side to build the SPI shading library against this, just to see if the previous failure to find this error had let bugs creep into it?

marsupial:
Built the library yes, haven't run the test suite (was hoping to do it overnight).
As it caught no hidden errors, would be shocked if it changed any behavior.

lgritz:
Yeah, I don't expect any change to runtime. If it doesn't come up with new errors when compiling (i.e., errors missed before), I think you're home free. So I'll merge. If anything goes wrong, we can fix.

PR #951, opened 2018-12-23
oslc: Fix crash when encountering `for (;;)`

### LINKED ISSUE #688: Crash when compiling infinite for loop
A for loop without a condition statement will crash oslc:

```
for (;;)
{
    // stuff goes here
}
```

A possible fix is to ensure there's always a valid condition statement. An emtpy statement as above corresponds with a constant `true` value. For instance writing the `ASTloop_statement` as below fixes the crash.

```
ASTloop_statement::ASTloop_statement (OSLCompilerImpl *comp, LoopType looptype, ASTNode *init,
                    ASTNode *cond, ASTNode *iter, ASTNode *stmt)
    : ASTNode (loop_statement_node, comp, looptype, init, cond, iter, stmt)
{
    // replace empty condition with True value (in case it's left out in a for loop)
    if (!cond) { m_children[1] = new ASTliteral (comp, 1); }
}
```

lgritz:
An empty conditional clause of a `for` statement was leading to bad
things. It's understood that `for (;;)` means `for (;1;)`, so just
make sure that happens when parsing.

Fixes #688

### DIFF
--- a/src/liboslcomp/ast.cpp
+++ b/src/liboslcomp/ast.cpp
@@ -859,6 +859,18 @@ ASTconditional_statement::childname (size_t i) const
 
 
 
+ASTloop_statement::ASTloop_statement (OSLCompilerImpl *comp, LoopType looptype,
+                                      ASTNode *init, ASTNode *cond,
+                                      ASTNode *iter, ASTNode *stmt)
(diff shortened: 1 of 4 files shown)


aconty:
LGTM!

PR #954, opened 2018-12-31
Fix bug with inverse spline

### LINKED ISSUE #778: splineinverse( "bezier", ... ) fails for first and last segment
This expression:

```
testshade -g 256 256 -center -expr 'float cvs[4]={0,1.0/3,2.0/3,1}; result = v < 0.5 ? spline( "bezier", u, cvs ) : splineinverse( "bezier", u, cvs );' -o result spline.exr && display spline.exr
```

declares a set of uniformly increasing control vertices, and then uses them to shade a gradient.  The top of the gradient uses spline(), the bottom uses splineinverse(), both with a bezier basis.  I would expect top and bottom to be the same, because these values correspond to an identity curve between 0 and 1, but instead, the first and last third of the bottom are cut off.

This seems to be a natural consequence of this code:
https://github.com/imageworks/OpenShadingLanguage/blob/master/src/liboslexec/splineimpl.h#L191
which clamps the input range of the splineinverse to the values of the second and second last knot.  This seems clearly wrong for a bezier curve, where the second knot is only used for tangent information, and the first knot is a position.  It would be even weirder to do this to a hermite curve.

I'm not sure what the correct solution is in this context - for the general case of a cubic basis, it would probably be necessary to evaluate the curve at the minimum and maximum parameter value in order to get the correct range to clamp to.  But if the current approach is working well for everything except bezier and hermite, it would be good enough for my purposes if we just switched to using knots 0 and N instead of 1 and N - 1 if the basis is bezier or hermite.

lgritz:
We were clipping the range incorrectly for bezier, where it needs to step by 3 CVs for each span. Worked ok for the other spline types, which step by 1.

Add a test that verify that a spline designed to give a f(x)=x response does so, and also that its inverse is perfect.

Fixes #778


aconty:
LGTM

PR #956, opened 2019-01-03
Revise shader install path and stdosl.h search

### LINKED ISSUE #955: shaders are installed into non standard directory in linux build
I've come across this error while packaging for openSUSE.
The shaders are installed directly under $PREFIX /usr I've attached the patch I made to ensure that the installation adheres to the linux standard base. I use -DCMAKE_INSTALL_DATADIR:PATH=%{_datadir}/%{name} other distros can put in their own paths.
[OpenShadingLanguage-shaderinstalldir.patch.txt](https://github.com/imageworks/OpenShadingLanguage/files/2720561/OpenShadingLanguage-shaderinstalldir.patch.txt)

lgritz:
Issue #955 points out that we install the example shaders (and stdosl.h)
in PREFIX/shaders. That may be fine for a totally custom directory, but
it's nonstandard and rightly frowned upon if the prefix is a system-wide
place like /usr. Just like docs go in PREFIX/share/doc, probably the
shaders ought to go in PREFIX/share/shaders.

So this patch makes that change. Does anybody think that will be
extremely disruptive?

Modify oslc/liboslcomp to search in the new location (as well as
the old), and while we're at it, I was able to significantly clean up
and compress find_stdoslpath(). Also adjusted what in retrospect was
a logical error in how we search for stdosl: I now think that the
user-supplied include paths (for example, `oslc -Ifoo -Ibar`)
probably ought to be the first places searched, not the last (as it was
before).

Fixes #955


lgritz:
Opinions:

PREFIX/share/shaders ?
PREFIX/share/OSL/shaders ?
PREFIX/share/shaders/OSL ?

fpsunflower:
I think I'd vote for `PREFIX/share/shaders` since I assume `PREFIX` would already be somewhere specific to OSL ? or do we need to worry about the case where `PREFIX=/usr`?

lgritz:
Yeah, /usr is precisely what we're concerned about here.

StefanBruens:
I don't think the current patch is good enough:

- Files in the /usr/share/ hierarchy are typically placed below a project specific directory, in this case e.g. `/usr/share/osl/` or `/usr/share/OSL/`
- below this directory the hierarchy is up to the project, so e.g. `/usr/share/osl/shaders/` and `/usr/share/osl/shaders/MaterialX/`

CMAKE_INSTALL_DATADIR can be either an absolute or relative (to CMAKE_INSTALL_PREFIX) path. In either case CMAKE_FULL_INSTALL_DATADIR is the resulting absolute path, which can be used with cmakes configure_files to create a #define with a matching path:
```
// Keep this in sync with the install path in shaders/CMakeLists.txt and shaders/MaterialX/CMakeLists.txt
#define SHADER_INSTALL_DIR @CMAKE_FULL_INSTALL_DATADIR@/shaders/
```
This simplifies the determination of the stdosl.h path significantly.

StefanBruens:
Note, as GNUInstallDirs is included, `CMAKE_INSTALL_DATADIR` is typically set to "share", so the current patch already changes the shaders install path from `CMAKE_INSTALL_PREFIX/shaders` to  `CMAKE_INSTALL_PREFIX/share/shaders`.

lgritz:
Sorry for the delay.

Amended based on your comments.

Now defaults to .../shared/OSL/shaders
And use `CMAKE_INSTALL_FULL_DATADIR` as suggested, also define that location in oslversion.h.

lgritz:
In this new amendment, although the shaders are now in .../share/OSL/shaders, the docs are still in share/doc/OSL. Good or bad? On one hand, I like the idea of all the OSL stuff being together under share/OSL, but on the other hand, clearly share/doc/PACKAGE is the overall pattern I see on the systems I'm looking at. Suggestions here are appreciated, I have no strong preference.

StefanBruens:
The documentation path can be easily set with the standard CMAKE_INSTALL_DOCDIR, so deviations are not a big issue.

lgritz:
@StefanBruens Is that an approval of this PR as it currently stands?

StefanBruens:
Sorry, haven't had time to actually verify it, gimme time until the weekend.

lgritz:
No rush! Just let me know.

lgritz:
Hey, @StefanBruens did you ever get a chance to test this?

lgritz:
Anybody object to this? Last chance.

lgritz:
One more adjustment on my end: added a cmake config option that lets you override the shader install location.

lgritz:
I will merge this today if there are not objections.

PR #957, opened 2019-01-04
Fix group_name, group_init_name, group_entry_name

lgritz:
I didn't notice this in Tim's code before today, but he made the
retrieval of these group attributes more complex than they needed to be.
Instead of messing with pointers (which we avoid), you can just directly
getattribute(name,string&). So now we use that. Also eliminate group_name
entirely, since it's redundant with the existing groupname.

### DIFF
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -1424,6 +1424,7 @@ ShadingSystemImpl::getattribute (ShaderGroup *group, string_view name,
 {
     if (! group)
         return false;
+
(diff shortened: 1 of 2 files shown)


aconty:
LGTM, but wasn't "group_name" a more consistent keyword with the underscore?

lgritz:
"groupname" was first, there for a long time, we use it in a bunch of places and document it. I think Tim either didn't realize it was there, or didn't understand that there was already a simple way to retrieve a string. The "group_name" was added recently, undocumented, and used in only one place in the whole codebase.

PR #958, opened 2019-01-04
Deprecate the auto-allocation of contexts and per-thread-info

### LINKED ISSUE #691: __nptl_deallocate_tsd() crash when closing an app with OSL loaded as a library on Linux
This is pretty esoteric and may not be reasonably easy to fix for all uses, but I'm filing this so at least it's a known errata.

Due to [boost Ticket #3926](https://svn.boost.org/trac/boost/ticket/3926) and the use of boost::thread_specific_ptr to store thread-local data, when OSL is used as a shared library, you can run into occasional crashes in __nptl_deallocate_tsd(), which is the pthread thread local storage cleanup routine, when the application closes.

The sequence of events that leads to the problem is
1) load OSL as a shared library
2) start shading system and optimize a shader
3) this causes boost::thread_specific_ptr to register a pthread thread specific key and destructor pointing to code in the loaded library
4) clean up the shading system and unload the library
    --  at this point, the code for destructor that was registered with the pthread is unloaded because the LLVM_Util::PerThreadInfo are intentionally leaked, even having OSL clean up the shading system does not de-register the thread specific destructor.
5) closing the app causes the pthread library to do clean-up, calling the destructors for any per-thread
keys that are still registered.  Since the shared library is unloaded, this causes a segfault

We've worked around it by managing the thread specific data ourselves and passing it in for shading context PerThreadInfo, and adding the plumbing to get it all the way down into LLVM_Util::PerThreadInfo.  That may not be a good solution for users who don't manage their own thread local data.

lgritz:
There were a variety of places that took a Context* that defaulted
to null, where passing null means to let the called funtion allocate
a context.

This relies on a boost::thread_specific_pointer which could be easily
mishandled and lead to problems.

So with this patch, we move toward requiring the app to allocate and
own the contexts and PerThreadInfo objects.

* No more default null parameter value on contexts or threadinfo.

* Explicitly passing nullptr will still work (use with caution!) on
  2.0 with a warning, but will be a full error in 2.1 as that functionality
  is totally removed.

* execute() takes a ShadingContext& instead of ShadingConext*, to emphasize
  that it must not be null. (For 2.0, the pointer one will still exist
  and work, but is slated for deprecation.)

* Fixed comments that document "it's advised o pass a contexts, but simple
  apps can choose not to" to now imply that it's required.

* A bit of reorganization of SS::getattribute -- some group attrib
  queries required the group to already be optimized, and there were many
  calls to optimze_group that all had the implicit null context. These
  are all coalesced now, and it allocates a temporary context.

Fixes #691

### DIFF
--- a/src/liboslexec/context.cpp
+++ b/src/liboslexec/context.cpp
@@ -76,11 +76,13 @@ ShadingContext::execute_init (ShaderGroup &sgroup, ShaderGlobals &ssg, bool run)
     if (sgroup.nlayers()) {
         sgroup.start_running ();
         if (! sgroup.optimized()) {
-            shadingsys().optimize_group (sgroup);
+            auto ctx = shadingsys().get_context(thread_info());
+            shadingsys().optimize_group (sgroup, ctx);
             if (shadingsys().m_greedyjit && shadingsys().m_groups_to_compile_count) {
                 // If we are greedily JITing, optimize/JIT everything now
                 shadingsys().optimize_all_groups ();
             }
+            shadingsys().release_context(ctx);
         }
         if (sgroup.does_nothing())
             return false;
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -2557,6 +2557,7 @@ ShadingSystemImpl::get_context (PerThreadInfo *threadinfo,
 {
     if (! threadinfo) {
 #if OSL_VERSION < 20200
+        threadinfo = get_perthread_info ();
         warning ("ShadingSystem::get_context called without a PerThreadInfo");
 #else
         error ("ShadingSystem::get_context called without a PerThreadInfo");
@@ -2853,14 +2854,10 @@ ShadingSystemImpl::optimize_group (ShaderGroup &group, ShadingContext *ctx)
 
     double locking_time = timer();
 
-    bool perthread_allocated = false;
     bool ctx_allocated = false;
     PerThreadInfo *thread_info = nullptr;
     if (! ctx) {
-        if (! thread_info) {
-            thread_info = create_thread_info();
-            perthread_allocated = true;
-        }
+        thread_info = create_thread_info();
         ctx = get_context(thread_info);
         ctx_allocated = true;
     }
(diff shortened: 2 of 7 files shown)


src/liboslexec/shadingsys.cpp
```diff
@@ -2571,8 +2546,14 @@ ShadingContext *
 ShadingSystemImpl::get_context (PerThreadInfo *threadinfo,
                                 TextureSystem::Perthread *texture_threadinfo)
 {
-    if (! threadinfo)
-        threadinfo = get_perthread_info ();
+    if (! threadinfo) {
+#if OSL_VERSION < 20200
+        warning ("ShadingSystem::get_context called without a PerThreadInfo");
```
fpsunflower: Did you intend to auto-create a thread info here? otherwise it just crashes at the line below, no?

src/liboslexec/shadingsys.cpp
```diff
@@ -2849,9 +2844,15 @@ ShadingSystemImpl::optimize_group (ShaderGroup &group, ShadingContext *ctx)
 
     double locking_time = timer();
 
+    bool perthread_allocated = false;
```
fpsunflower: Isn't this the same as `ctx_allocated`?

lgritz:
I pushed an update to address review comments.

src/liboslexec/context.cpp
```diff
@@ -76,11 +76,13 @@ ShadingContext::execute_init (ShaderGroup &sgroup, ShaderGlobals &ssg, bool run)
     if (sgroup.nlayers()) {
         sgroup.start_running ();
         if (! sgroup.optimized()) {
-            shadingsys().optimize_group (sgroup);
+            auto ctx = shadingsys().get_context(thread_info());
+            shadingsys().optimize_group (sgroup, ctx);
```
fpsunflower: Could we use `*this` as the context?

lgritz: I don't think that's safe. Any setup that has happened before the execute_init could get clobbered. The "scratch context" that's needed for optimization can't be the same one that's already half set up to execute the shader.

lgritz: I did consider that. I just had trouble convincing myself that it was fully safe, and would not be super brittle (like, working fine now, but any number of future changes by unsuspecting developer/me would re-break it). Seemed safer to just request a fresh context. It's a pool, so it's very inexpensive to keep grabbing and then returning a simple scratch context.

fpsunflower: Makes sense.

LGTM!

lgritz:
I got the LGTM from Chris but I'm going to let this sit until next week before merging, just in case anybody externally has concerns to voice.

lgritz:
Oops, didn't mean to delete that. Re-opening.

PR #960, opened 2019-01-04
Fix transformc with derivs when using OCIO

lgritz:
I have no idea how this ever worked. In fact, it must never have, because
our checked-in reference output was wrong. But it kept matching the same
wrong until it didn't, and became a different kind of wrong on at least
one machine.

When doing transformc, where the space names necessitated using OCIO for
the transformation math, and only for the derivative case (the non-deriv
was fine all along), we were simply passing the wrong stride parameters
to the colorproc->apply() call.

Also, I noticed that we were trying to save a per-context "last
transformation" to avoid the lookup when doing repeats of needing the
same colorproc (basically, a 1-item LRU cached lookup of the
colorproc). But we never saved the space name, so the cache was not
operating.


fpsunflower:
LGTM

PR #961, opened 2019-01-08
[SPI] Update to latest partio version.

marsupial:
## Description
Older version uses an out of date Alembic.


lgritz:
LGTM, merging

Issue #964, opened 2019-01-13
Building of osltoy fails with "class-memaccess" error/warning with current GCC 8


StefanBruens:
### Problem

```
/home/stefan/Documents/Sources/OpenShadingLanguage/src/osltoy/osltoyrenderer.cpp: In constructor ‘OSL_v1_11::OSLToyRenderer::OSLToyRenderer()’:
/home/stefan/Documents/Sources/OpenShadingLanguage/src/osltoy/osltoyrenderer.cpp:82:42: error: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct OSL_v1_11::ShaderGlobals’ with no trivial copy-assignment; use assignment or value-initialization instead [-Werror=class-memaccess]
     memset (&sg, 0, sizeof(ShaderGlobals));
                                          ^
In file included from /home/stefan/Documents/Sources/OpenShadingLanguage/src/include/OSL/oslexec.h:34,
                 from /home/stefan/Documents/Sources/OpenShadingLanguage/src/osltoy/osltoyrenderer.cpp:33:
/home/stefan/Documents/Sources/OpenShadingLanguage/src/include/OSL/shaderglobals.h:61:8: note: ‘struct OSL_v1_11::ShaderGlobals’ declared here
 struct ShaderGlobals {
        ^~~~~~~~~~~~~
```
**Expected behavior:**  

Builds without errors/warnings

**Actual behavior:**  

### Steps to Reproduce

1. Build with -Werror=class-memaccess

### Versions

* OSL branch/version: master (2019-01-13)
* OS: openSUSE Tumbleweed
* C++ compiler: 8.2.1
* LLVM version: 
* OIIO version:

lgritz:
Long ago fixed by #965 but failed to auto-close this issue.

PR #965, opened 2019-01-13
Make initialization of ShaderGlobals C++ compliant

StefanBruens:
ShaderGlobals aggregates several non POD types, GCC warns when these
are initialized using memset:
error: ‘void* memset(void*, int, size_t)’ clearing an object of type
‘struct OSL_v1_11::ShaderGlobals’ with no trivial copy-assignment;
use assignment or value-initialization instead [-Werror=class-memaccess]
     memset (&sg, 0, sizeof(ShaderGlobals));

Use value-initilization instead.

Fix for #964.

## Description

See #964 

## Checklist:

### DIFF
--- a/src/osltoy/osltoyrenderer.cpp
+++ b/src/osltoy/osltoyrenderer.cpp
@@ -78,8 +78,8 @@ OSLToyRenderer::OSLToyRenderer ()
     m_attr_getters[ustring("camera:shutter_close")] = &OSLToyRenderer::get_camera_shutter_close;
 
     // Set up default shaderglobals
-    ShaderGlobals &sg (m_shaderglobals_template);
-    memset (&sg, 0, sizeof(ShaderGlobals));
+    ShaderGlobals &sg = m_shaderglobals_template;
+    sg = {};
     Matrix44 Mshad, Mobj;  // just let these be identity for now
     // Set "shader" space to be Mshad.  In a real renderer, this may be
     // different for each shader group.


src/osltoy/osltoyrenderer.cpp
```diff
@@ -78,8 +78,8 @@ OSLToyRenderer::OSLToyRenderer ()
     m_attr_getters[ustring("camera:shutter_close")] = &OSLToyRenderer::get_camera_shutter_close;
 
     // Set up default shaderglobals
-    ShaderGlobals &sg (m_shaderglobals_template);
-    memset (&sg, 0, sizeof(ShaderGlobals));
+    ShaderGlobals &sg = m_shaderglobals_template;
+    sg = {};
```
lgritz: I worry about the efficiency. At best, this will be the same as the memcpy, but it's possible that it will do something less efficient as it fills in fields one by one. I would instead suggest the following simpler fix to the warning:

```suggestion
    memset ((void*)&sg, 0, sizeof(ShaderGlobals));
```

lgritz: I meant memset, not memcpy, of course. My point, though, is that what you are literally asking for is to construct an SG and then copy it where we want it (possibly field by field). We are all hoping that the compiler figures out how to replace that with the equivalent memset to 0, but I feel better about just saying memset explicitly and not taking any chances. Call me old fashioned and superstitious?

lgritz:
As an aside, we really do need a CLA in order to accept and merge a PR.

src/osltoy/osltoyrenderer.cpp
StefanBruens: 1. Performance is totally irrelevant here.
2. The compiler is clever enough to avoid double initialization (checked the generated machine code with objdump).

lgritz: Oh yes, I see this is in the constructor. So you're right, there's no perf issue in this spot.

But are you quite sure this is correct? ShaderGlobals doesn't have a constructor defined, so the default constructor will just run the default constructor of the individual fields. But as an example, there are many Vec3 fields, and they are Imath::Vec3, whose default constructor leaves them uninitialized. 

For that matter, `m_shaderglobals_template` will just have had its default initializer run (which I think does nothing), so the `sg = {}` I think is effectively a no-op anyway, is it not? Because you're taking a just-been-default-constructed SG and then overwriting it with a default-constructed SG.

So I'm confused about how all the fields will come to be zeroed out in your proposal. Can you explain?

lgritz:
We haven't seen an update about this PR for many weeks. There's an outstanding question about whether it's correct (i.e. will zero out all fields versus leaving them uninitialized), and as well I'm unable to merge into this repository without receiving a CLA from you and/or your organization.

lgritz:
Months have gone by without update, response, or CLA. Closing. Please resubmit if this is important to anybody.

PR #967, opened 2019-01-14
Allow shader params (arrays/structs) empty brace initialization lists

lgritz:
Still gets length 1 (we don't allow 0-length arrays, that's a whole
other can of worms), but this is now treated as equivalent to
`= { default_value }` (0 for int or float, "" for string).

It's syntactic sugar, but it removes some clutter for people who want to
have shader parameters that are arrays of undetermined length (and thus are
probably expected to have a runtime-assigned instance value or be
conntected to an output of an earlier layer within the shader group),
removes the requirement for spelling out an initializing value that
won't be used anyway.

### DIFF
--- a/src/liboslcomp/codegen.cpp
+++ b/src/liboslcomp/codegen.cpp
@@ -968,7 +968,7 @@ ASTNode::codegen_initlist (ref init, TypeSpec type, Symbol *sym)
                 break;
             }
         }
-        if (all_const) {
+        if (all_const && init) {
             std::vector<char> arrayvals (type.simpletype().size());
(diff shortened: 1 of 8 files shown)


fpsunflower:
Maybe worth adding a test for the `struct` case as well?

lgritz:
Added test. It worked!
Also amended the commit message to clearly explain that it works for structs as well.

fpsunflower:
LGTM!

fpsunflower:
Did this fall through the cracks? Or was the an issue remaining?

lgritz:
Yes, I think it just fell through the cracks. Will rebase and re-test and if all is well, will merge tomorrow.

PR #973, opened 2019-02-19
Device String Update

timgrant:
## Description
This update brings some minor enhancements to how DeviceStrings are handled. The changes are intended to facilitate passing strings as userdata parameters, and passing strings out as closure parameters.

I have included a macro for casting a ``char*`` to a ``DeviceString``, in the spirit of the ``USTR`` macro. It's demonstrated in ``rend_get_userdata``. It will be useful when adapting the remaining "shadeops" functions that take string parameters.

## Tests
A case for supplying a userdata string has been added to the ``testoptix`` testsuite.

## Checklist:


lgritz:
Hi, Tim. I ran into two problems with this patch:

1. Though it builds for me when I use Cuda 8.0 and OptiX 5.1, I get some compilation and link flags when using Cuda 9 or 10 or OptiX 6. Have you tried any of those combinations on your end?

2. When using Cuda8/OptiX5.1 that builds, it passes testoptix, but testoptix-noise fails with some pattern change on the "hash".

timgrant:
Thanks, Larry. I haven't tested all of those configurations, so I'll do that now and see what snags I hit. I'm not sure why the hash noise would be different, but I'll take a look at that, too.

lgritz:
I see, in externalpackages.cmake, circa line 243, there's an "if" that sets --cuda-path for Cuda <= 8 but not >8. When I change this so that it sets CUDA_LIB_FLAGS regardless of version, then I can get it to build with cuda 9 and 9.1 (but not 10, which says "cannot find libdevice for sm_35).

So that's progress. Do you recall why you had this clause in externalpackages.cmake?

lgritz:
Also: apologies about the "optimize_group" change, that was my fault, I should have tested the optix build when I committed #958

lgritz:
When I compile against Cuda 9.1 and OptiX 6.0, I get various runtime failures.

terminate called after throwing an instance of 'optix::Exception'
  what():  Failed to load OptiX library

lgritz:
Aha, @fpsunflower informs me that to use OptiX 6, I need to upgrade the driver on my workstation. So that's probably the reason for that bit of the problem.

Still not sure why I can't use Cuda 10.

But this isn't critical, we can come back to it. The biggest holdup in merging your patch is the change in appearance of the hash on testoptix-noise.

timgrant:
The way OptiX SDK versions, CUDA versions, and LLVM versions are matched in the build is kludgy at best. Since OptiX is now a driver component, that makes the driver version another variable which needs to be accounted for. I'll start thinking about build improvements.

I don't think CUDA 10 is currently well supported in LLVM. I've tested up to CUDA 9.2 with a source build of LLVM 7.0.0 and it's working for me.

Regarding the reference image, for some reason the output for the noise test is different for OptiX 5.1 and OptiX 6.0. It will take a bit of digging to get to the bottom of it.

lgritz:
Gotcha, ok, for right now, I'll stick to Cuda 9.x.

timgrant:
In retrospect, using hash noise in this test was not a good idea, since it's sensitive to single bit differences. For some pixels, the texture coordinates being provided to the hash function are very slightly different between OptiX 5 and OptiX 6 (probably due optimization changing the order of some floating-point operations), so those pixels end up being completely different.

As eye-popping as the current reference image is, I should probably update the test to use a less sensitive noise variant.

lgritz:
Oh yes, that makes sense. Certainly `hash()` is ok to use as unit test where specific known values are used as inputs, but should never be used as a "rendering" test where the inputs are based on intersection positions, because of course we expect those to periodically (or from platform to platform) change ever so slightly.

Do you want to modify that on your end, or should we commit this PR and then I can immediately alter the testoptix-noise test myself and replace it with some other noise variety and its new reference image?

lgritz:
After I merge this, I will also check in a new reference image for that test. Note that the TravisCI has no GPUs and so does not build in USE_OPTIX=1 mode, so neither build breaks nor failed tests specific to the optix path are caught by our usual CI. (Which is how that other break I caused went unnoticed as well.)

lgritz:
LGTM, @timgrant. I'm testing now and if it seems ok I will do the merge.

One more question: did you have any thoughts on what I asked earlier about how in externalpackages.cmake, you had an `if` clause that adds the `--cuda-path=` argument only when the Cuda version <= 8? I found that in fact this FAILS for me with Cuda 9 and 9.1, but it works just fine if I always do the `--cuda-path=` (i.e., make the `if` less complicated). Opinions?

Also, apologies, I'm a little slow today, I do see now that you also updated the ref image when you changed the noise!

marsupial:
Hello. I'm finding that when setting/uploading to an optix::Buffer, Optix (or CUDA) is expecting the name to be fully mangled as it would be in C++. Which also affects the standard OSL::DeviceStrings.

This can be seen in #974, by chopping off the last commit (84075d6a353a20c572aed84fe38d996827d23c66)

timgrant:
What are the ``optix::Buffers`` you're trying to set?

The inclusion of device_string.h and the ``STRDECLs`` in rend_lib.h should create OptiX variable declarations for all of the standard strings, and ``StringTable::init()`` should provide the definitions (backing in the device memory table, etc.).

Then in your device code, you can do things like this:

```C++
case MICROFACET_ID: {
    const char* mem = ((ClosureComponent*) cur)->mem;
    const char* dist_str = *(const char**) &mem[0];
    // Comparisons between the closure variable and "standard"
    // strings are possible
    if (HDSTR(dist_str) == OSL::DeviceStrings::default_)
        printf("dist is default\n");
}
```

And in shadeops code (like opnoise.cpp, as excerpted from one of my branches):

```C++
#ifndef __CUDA_ARCH__
typedef OIIO::ustring StringParam;
#else
typedef DeviceString StringParam;
#endif

#ifdef __CUDA_ARCH__
namespace StringParams = OSL::DeviceStrings;
#else
namespace StringParams = OSL::Strings;
#endif

template<class R, class S, class T> OSL_HOSTDEVICE
inline void operator() (StringParam name, Dual2<R> &result,
                        const Dual2<S> &s, const Dual2<T> &t,
                        const S &sp, const T &tp,
                        ShaderGlobals *sg, const NoiseParams *opt) const {
    if (name == StringParams::uperlin || name == StringParams::noise) {
        PeriodicNoise noise;
        noise(result, s, t, sp, tp);
    }
    // etc...
}
```

I have not run into problems with C++ name mangling or namespaces when adapting the remaining noise functions (generic noise, Gabor, etc., for which I should probably open a PR sooner rather than later).

For the string parameters in opspline.cpp, the approach I've outlined above will hopefully work. If your issue is with ``getSplineBasis``, I'm not quite sure what the solution is.

marsupial:
I'm finding that all OSL::DeviceString (while readable variables in Cuda, are all empty)
https://github.com/marsupial/OpenShadingLanguage/commit/19ea151c6590fa0a2fb7d9ddd2ea3d0a191499a7 demonstrates the issue

I get `OSL::DeviceStrings (nil) (null)` unless the buffer is set with the mangled named (uncomment the changes in stringtable.cpp)

PR #974, opened 2019-02-23
Texture and Spline support for Optix

marsupial:
## Description
For Optix:
  Add basic texture lookup.
  Refactor spline evaluation code.
  Add 'texture-space' rendering to testoptix.

## Tests
Added a test for  texture lookup and spline.


lgritz:
Let's rebase this on top of the patches from Tim that I merged on Frieday.

timgrant:
I don't know if you saw my comment on [973](https://github.com/imageworks/OpenShadingLanguage/pull/973), but I took a closer look and the ``DeviceString`` variables in the ``OSL::DeviceStrings`` namespace weren't working as I had thought.

I brought my branch that integrates Gabor noise and generic noise (which make use of string parameters) up to date and have hopefully ironed out the issues with the standard strings. We'll try to get a PR open with these changes ASAP.

marsupial:
Came here after posting more in #973, so seems were on the same page, right?

timgrant:
Great minds think alike. I've opened #980 which shows one potential approach to string parameters.

marsupial:
@timgrant, I'm still getting the odd result of the spline functions coming in as an llvm::Constant vs. llvm::Function when opspline.cpp is compiled in bitcode during CUDA_SHADEOPS_COMPILE

Demonstrated by removing the comment at line 133 of `src/liboslexec/CMakeLists.txt`
and the include at line 278 of `src/testoptix/cuda/rend_lib.cu`

timgrant:
@marsupial, I'll build your branch tomorrow and see if anything pops out.

marsupial:
Thanks. Think I got it though, not sure where, but seemed to be a mis-alignment of a declaration/definition and LLVM was doing a bit-cast.

Hopefully da4e560 doesn't break the CPU spline code...

src/include/OSL/device_string.h
```diff
@@ -92,27 +92,27 @@ struct DeviceString {
...
 
+OSL_NAMESPACE_EXIT
+
+
 #ifdef __CUDA_ARCH__
-namespace StringParams = OSLDeviceStrings;
+namespace StringParams = OSL_NAMESPACE::DeviceStrings;
 #else
-namespace StringParams = OSL::Strings;
+namespace StringParams = OSL_NAMESPACE::Strings;
```
lgritz: Why this change?

src/include/OSL/dual.h
```diff
@@ -71,7 +72,7 @@ class Dual {
     OSL_HOSTDEVICE OIIO_CONSTEXPR14 Dual (const T &x) {
         m_data[0] = x;
         for (int i = 1; i <= PARTIALS; ++i)
-            m_data[i] = T(0.0);
+            m_data[i] = zero();
```
lgritz: why?

lgritz:
I like where you're going with this, but boy did I screw up in the last 48 hours. I've got a patch myself that does a huge refactor of this code (in fact, testoptix disappears and is subsumed by testrender) and was so excited about it that I didn't really think about the massive merge conflict that would ensue.

Um... I feel like that's my fault and it shouldn't be on your to fix it.

[ reply sits overnight ]

After sleeping on it, here's what I think the plan is:

1. I'm going to revise my (not yet submitted as PR) overhaul so that it doesn't remove testoptix; all the changes will be to testrender, and thus will not conflict.

2. I'll briefly test your patch, and if it works I will merge it. 

3. Submit my non-conflicting testrender overhaul.

4. At that point, we'll have a dead-man-walking testoptix that does texture and a testrender that does not (and, yuck, with testoptix bits that are significantly shuffled around and refactored). But it will all build and work.

5. Then I will do the work to backport your changes to my new scheme (don't do it yourself -- the last thing I want is us clobbering each other again), and submit that with a PR that has it all working in testrender and finally remove testoptix.

I'll try to do 1-3 over the weekend. Then I go on vacation, so we're not going to get things fully reconciled until I get back.

timgrant:
My mind is blown. I'm super excited to see your changes, Larry. I was actually going to bring my texture prototype up to speed this weekend (sorry for the delay!), so instead I'll look forward to your PR and see if I can make any helpful suggestions.

lgritz:
@timgrant I thought you had completely handed things off and weren't counting on getting anything more done for now?

I'd love for you to look over @marsupial (Roman)'s PR here and make suggestions about the approach.

The three of us are going to need to stop tripping over each other real soon.

Everybody please pre-declare what they are working on, if doing any major work on the OptiX support that is unlikely to be a smooth merge.

lgritz:
Also, @timgrant, feel free to post your stab at texture as a PR so we can compare and contrast.
(We don't have to merge all of these.)

Should I just hold off on merging Roman's texture PR

Also note to Roman: Please separate features into different PRs if possible, If this was a one PR for spline and another for texture, we could easily add the spline support right away and without any merge conlficts with what Tim or I had done. That said, this is an awesome and exciting stab at this and I in no way want to discourage you from working on the OptiX side.

lgritz:
Hmmm... I might simplify this operation by merging some of your individual commits separately and first.

lgritz:
Roman, what's up with the commit whose comment is "Fix Optix compilation problems for `Dual<float3>`"?

There's not much explanation. What were you trying to do there, what problem did it solve?

src/include/OSL/dual.h
marsupial: float3 isn't constructible via `T(0.0)`. At line 1087 is the Cuda specialization that returns `float3{0,0,0}`.

If I remember, tried moving it to `T(0.f)` first, but was still a compile-error.

src/include/OSL/device_string.h
marsupial: Was fighting custom-namespacing for a moment, that may be collateral damage and unnecessary.

lgritz: Makes sense, thanks.

src/include/OSL/dual.h
lgritz: I see, sure.

marsupial:
Hopefully good answers below, but if its just easier, feel free to close it.

At some point it was two branches (and this PR spline support only), but doing conflict resolution was just tremendously easier on one (and the spline and texture tests both needed the similar/additional test-harness). That said, having it only one test-runner is way better.

44389a912210a6253655cfd98bf710cec1b3a46e & 4cf8e399e75405a624a5b942f5bcface022c1b2c could probably be cherry-picked without much work. The only issue would be putting the change in _src/testoptix/optixrend.cpp_ in the new/right location during the context initialization then and moving/adapting the test itself.

```
    {
         optix::Buffer buffer = m_context->createBuffer(RT_BUFFER_INPUT, RT_FORMAT_USER);
         buffer->setElementSize(sizeof(pvt::Spline::SplineBasis));
         buffer->setSize(sizeof(pvt::Spline::gBasisSet)/sizeof(pvt::Spline::SplineBasis));

          pvt::Spline::SplineBasis* basis = (pvt::Spline::SplineBasis*) buffer->map();
         ::memcpy(basis, &pvt::Spline::gBasisSet[0], sizeof(pvt::Spline::gBasisSet));
         buffer->unmap();
         m_context[OSL_NAMESPACE_STRING "::pvt::Spline::gBasisSet"]->setBuffer(buffer);
     }
```

The texture part look's like more than it probably is if the lookup's can be tested elsewhere. Really just the addition of `RenderServices::good`, `OptixRenderer::get_texture_handle` and the four shader-ops in `rend_lib.cu` .

The `Dual<float3>` issue is that `float3(0)` is a compile-error.

lgritz:
I'm messing with it now to see how much of this PR can be merged without creating a problem with my other PR. Hopefully what's left will be small enough that I can reconcile it by hand without too much trouble.

lgritz:
Merged the float3 and spline refactor.

src/liboslexec/splineimpl.h
```diff
@@ -43,45 +47,100 @@ namespace Spline {
...
+    kBezier,
+    kBSpline,
+    kHermite,
+    kLinear,
+    kConstant,
+    kNumSplineTypes
 };
 
+#ifdef __CUDACC__
+    rtBuffer<SplineBasis> gBasisSet;
```
timgrant: You might consider making this a regular global variable rather than an OptiX Buffer, to avoid making OptiX a dependency of ``liboslexec``. It will bloat the size of the compiled code somewhat, but I think that's a worthwhile tradeoff.

If you look in ``gabornoise.cpp``, I added an ``OSL_DEVICE`` decorator to a couple of global float constants which is defined as ``__device__`` when compiling for a CUDA target. A similar pattern might work here.

src/liboslexec/splineimpl.h
```diff
@@ -43,45 +47,100 @@ namespace Spline {
 
 struct SplineBasis {
    int      basis_step;
-   Matrix44 basis;
+   float    basis[4][4];
```
timgrant: Regretfully, I was a little lazy when adapting ``ImathMatrix.h`` and didn't add the ``HOST_DEVICE`` decorators to the ``Matrix44`` templates. But it should be possible to use ``Matrix44`` for CUDA just like the lower-dimensional variants. It's probably not worth a PR in itself, but I can forward you a patch with a fully-dressed header.

lgritz:
Looks like the patch that synchronizes the CPU and GPU device strings can also be merged easily, and it passes my tests, so merging that as well.

src/liboslexec/splineimpl.h
lgritz: What we need to do is get a PR into OpenEXR itself that adds the decorators to all their matrix and vector classes at the source so that they are safe to compile for Cuda.

timgrant:
@lgritz , I'll look at Roman's texture stuff ASAP. But I might not have time to fully grok it today.

lgritz:
Tim's comments notwithstanding, Roman's patch works. I've pulled out a few independent pieces and merged them separately, then rebased rest (which contain, all mixed together: spline implementation, basic texture, and a log of refactor of testoptix). That rebased version nonetheless works.

Here's what I'm doing next:

* I refactored Roman's outstanding commits somewhat, basically turning them into into two remaining commits. The first contains most of the changes for spline (everything but a few odds & ends that require material from the texture patch) and almost works -- it makes all the remaining CPU side spine changes, most of the GPU changes, but fails the spline test (only for optix, passes CPU testsuite). The second patch contains the texture stuff, Roman's testoptix refactor, and the last bits needed to make spline work.

* I will therefore merge the remainder of Roman's work, and close this PR. (Everything in this PR will get merged, but I've refactored it into a different set of commits, so GH will think the PR was abandoned, but really it's approved and merged. There seems to be no way to tell GH that.)

* This split makes it easy for me to rebase my testrender overhaul right after the first commit, before the second. Thus absolutely minimizing the amount if icky merge stuff to work through.

* Then I will start modifying my patch to try to be fully rebased on all of Roman's work, and hopefully in the process preserving all of the important features he added, so that the final result will be totally working with spine and texture. It's going to be a hairy night of that, but if all goes well I'll have the edits and can try on OptiX at work tomorrow to verify, before heading out on vacation (not incommunicado, but definitely not someplace where I can directly test the optix path).

* Everybody keep Tim's suggestions in mind, we should come back when this is all done and reconsider if those small changes should be made. But let's wait until I get it all reconciled and tackle it when I get back from vacation, ok? I don't want another difficult merge. After Roman's patch is merged, everything in the testoptix directory should be considered stale, about to disappear and have the equivalent (but not exactly the same) pop up elsewhere.

timgrant:
SGTM 👍

lgritz:
OK, the last of this has been merged into master. I will close this PR.

Issue #963, opened 2019-01-12
Please add support for combined libLLVM


StefanBruens:
### Problem

libLLVM can either be built as component libraries or as combined libraries, see
https://wiki.gentoo.org/wiki/Project:LLVM/Shared_libraries

When LLVM is built as combined library, there is only libLLVM.so, and building
oslexec fails, as it uses LLVM_MCJIT_LIBRARY:

```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
LLVM_MCJIT_LIBRARY
    linked by target "oslexec" in directory /home/stefan/Documents/Sources/OpenShadingLanguage/src/liboslexec
```

**Expected behavior:**  

LLVM_MCJIT_LIBRARY should be set to a useful value. When using llvm-config, the correct library is returned:

```
$> llvm-config --libfiles mcjit
/usr/lib64/libLLVM.so
llvm-config --libs mcjit
-lLLVM
```

The issue can be worked around by setting "-DLLVM_MCJIT_LIBRARY='' " when running cmake.

### Versions

* OSL branch/version: 1.9.3, master
* OS: openSUSE TW 20190109
* C++ compiler: gcc 8.2
* LLVM version: 7.0.0
* OIIO version:

lgritz:
I'm not quite sure how to do this, and it's hard for me to test because all of my LLVM installs are the component library setup, I guess?

Since you seem to be in the situation of needing it, I don't suppose you can submit a PR that contains the necessary changes to our cmake scripts that make it work? (And, obviously, needs to continue working for those with the component library setup as well.)

StefanBruens:
It should be trivial:
- remove both find_library(LLVM ...) and find_library(LLVMMCJIT ...) from FindLLVM.cmake
- run the following when LLVM_LIBRARIES/LLVM_LIBRARY is not set (copied from the existing code):
```
if ((NOT LLVM_LIBRARY) AND (NOT LLVM_LIBRARIES))
    execute_process (COMMAND ${LLVM_CONFIG} --libfiles
                     OUTPUT_VARIABLE LLVM_LIBRARIES
                     OUTPUT_STRIP_TRAILING_WHITESPACE)
    string (REPLACE " " ";" LLVM_LIBRARIES "${LLVM_LIBRARIES}")
    set (LLVM_LIBRARY "")
endif()
```
[...]

PR #976, opened 2019-03-02
FindLLVM: Make more robust to library configuration of llvm

lgritz:
In particular, some system distributions bundle everything in one
libLLVM, without a separate libLLVMMCJIT.


lgritz:
@laerne, you want to take a look at this and see if it solves your "mcjit" library issue?

lgritz:
Fixes #963

laerne:
I've tried this one and its work fine as expected.

PR #977, opened 2019-03-04
Simplify flex/bison handling

lgritz:
Rely on CMake's own finding of flex/bison, simplifying our code to
just the helpful shortcut macro. Also make finding these REQUIRED,
which makes an easier-to-interpret error message than simply not
defining the macro.


lgritz:
objections?

lgritz:
chris says ok

PR #978, opened 2019-03-05
Improve warnings about ill-advised use of comma operator

lgritz:
We previously had warnings that attempted to catch mistakes in using
the "comma operator" when the user intended an "aggregate type
constructor", for example:

    color x = (a, b, c);    // comma operator! equivalent to: x = c;
    color x = color(a,b,c); // this is what they probably meant

However, the way we implemented the warning was specific to assignment:

    x = (a, b, c);

and variable declaration with initialization:

    color x = (a, b, c);

But it missed similar bugs in other constructs, for example

    color x = Cdiffuse * (a, b, c);

Here the user surely intends to multiply Cdiffuse channel-by-channel
with color(a,b,c), but what they will actually get is Cfiffuse scaled
by float c.

This patch changes the strategy for this warning, no longer tied to
variable declaration or assignment, but more generally any time the
comma operator is used inside expression parentheses, it will give a
warning.

### DIFF
--- a/src/liboslcomp/oslgram.y
+++ b/src/liboslcomp/oslgram.y
@@ -796,7 +796,21 @@ expression
                         $$->sourceline (@1.first_line);
                     }
                 }
-        | '(' compound_expression ')'           { $$ = $2; }
+        | '(' compound_expression ')'
+                {
+                    if ($2->nodetype() == ASTNode::comma_operator_node) {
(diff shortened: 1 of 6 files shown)


lgritz:
Does this seem ok to everybody?

lgritz:
Chris says ok

PR #979, opened 2019-03-06
OptiX: always seem to need --cuda-path flag, even for cuda 9

lgritz:
I needed this on my end to make Cuda 9 and 9.1 work.


lgritz:
@timgrant Do you have any opinions? Does this work for you on your end, or do you have any insight about why you had written it the old way in the first place?

Anybody else using the OptiX back end able to try this out and let me know if it works or doesn't? On my system, it seems to need this patch to work with Cuda 9.

timgrant:
As I recall, the intention was to avoid some issues I was having with doubly-defined symbols. But the errors may have been due to my build environment at the time. I tried your patch just now with LLVM 5 + CUDA 8 and LLVM 7 + CUDA 9 and didn't have any issues. So I say LGTM.

PR #980, opened 2019-03-06
Device noise update

timgrant:
## Description
This update incorporates the remaining noise variants (Gabor and generic) in the OptiX path. These variants were somewhat problematic since they have (u)string parameters, which made compiling them for the device impossible.

My solution was to add a conditionally-defined typedef for the string parameters based on the target:

```C++
#ifndef __CUDA_ARCH__
typedef OIIO::ustring StringParam;
#else
typedef DeviceString StringParam;
#endif

OSL_HOSTDEVICE
inline void operator() (StringParam noisename, Dual2<float> &result,
                        const Dual2<float> &x,
                        ShaderGlobals *sg, const NoiseParams *opt) const {
    result = gabor (x, opt);
}
```

And furthermore, to put the definitions for the standard strings into different namespaces:

```C++
#ifdef __CUDA_ARCH__
namespace StringParams = OSLDeviceStrings;
#else
namespace StringParams = OSL::Strings;
#endif

if (name == StringParams::uperlin || name == StringParams::noise) {
    Noise noise;
    noise(result, s);
}
```

It's perhaps a little awkward, so if there is a better way to handle this I am happy to hear it.

## Tests
I added a test scene to testoptix-noise that specifies ``gabor`` as the noise type, and sets ``lockgeom=0`` to exercise generic noise.

## Checklist:


lgritz:
Looks ok to me, and passes tests on my workstation.

@marsupial , does this look ok to you?

marsupial:
Also working, for me.
Likely gonna be some gnarly conflicts with #974 that'll take a moment.

lgritz:
OK if I merge, then?

marsupial:
Yeah, just gonna have to work through it.

timgrant:
I'm sorry if I caused any undue churn!

lgritz:
Not at all, @timgrant .

marsupial:
It's occuring to me that all of this Cuda namespacing might be the expectation the namespace is `OSL` versus the usage of `OSL_NAMESPACE`  and `OSL_NAMESPACE_STRING` macros..

It might be easier to just eat it for a minute and can then push a single PR addressing that vs another push, rebase, conflict cycle ?

lgritz:
Er, sorry, had already merged.

marsupial:
My poor words did mean might be easier to merge and deal with the namespacing after, so I'm happy with that.
Have a good weekend!

PR #982, opened 2019-03-13
testoptix enhancements to make more analogous to testrender

lgritz:
* Change -r to -res to match testrender (and for that matter, make -res
  work for testshade, not -g). (In both cases, the old oddballs are
  silent synonyms.)

* Respond to several additional OSL options, and set them the same
  was as testrender

* Allow --iters (much like it works in testshade) to make it easy to
  run all the rays many times to make accurate timing easier.

* Do some of synchronizing of near-identical code between
  testshade, testrender, and testoptix. And synchronize some of the
  analogous command line arguments (e.g., -res for everybody, not -res
  for testrender and -r for testoptix and -g for testsahde).

* Add empty `OPTIMIZEONLY` marker files in the optix test directories
  to signify that they don't need -O0 test cases run (currently, such tests
  will fail due to the OptiX path not yet supporting "init ops").

### DIFF
--- a/src/testoptix/testoptix.cpp
+++ b/src/testoptix/testoptix.cpp
@@ -66,28 +66,34 @@ using namespace OSL;
 
 namespace { // anonymous namespace
 
-ShadingSystem *shadingsys = NULL;
-bool debug1 = false;
-bool debug2 = false;
-bool verbose = false;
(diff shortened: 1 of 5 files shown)


fpsunflower:
LGTM

PR #983, opened 2019-03-20
Fix oslc assertion when passing init-lists to array parameters

lgritz:
It's easiest to explain the situation by example:

    void receive_array_3 (float array[3]) { }
    void receive_array_any (float array[]) { }

    receive_array_3 ({1.0, 2.0, 3.0});
    receive_array_any ({1.0, 2.0, 3.0});

Both of these would fail/crash by hitting an assertion in oslc.  The
general pattern is when passing an initializer-list (like just saying
`{ a, b, c }`) to a user function formal parameter that's an array.
They failed for somewhat different reasons if the formal parameter
was a concretely-sized array or if it was an unsized array.

A little bit of typecheck-fu handles this case.


src/liboslcomp/codegen.cpp
```diff
@@ -472,8 +472,16 @@ ASTcompound_initializer::codegen (Symbol *sym)
...
-	ASSERT(0 && "compound codegen");
-    return NULL;
+    if (m_typespec.is_sized_array()) {
+        if (!sym)
+            sym = m_compiler->make_temporary (m_typespec);
+        codegen_initlist (initlist(), m_typespec, sym);
+        return sym;
+    }
+
+    error("Possible bug: compound_initializer codegen does not know how "
```
fpsunflower: "Possible compiler bug" ?

(Just so that beginners don't think the compiler is blaming them ...)

fpsunflower:
Other than that minor wording suggestion, LGTM!

src/liboslcomp/codegen.cpp
lgritz: Yes, good point, will fix.

fpsunflower:
LGTM

PR #984, opened 2019-03-20
ShadingSystem: allow for thread-safe/reentrant shader specification

lgritz:
The ShaderGroupBegin / ShaderGroupEnd (with interspersed Parameter,
Shader, and ConnectShaders calls between them) relied on hidden state
indicating which shader group was the single one undergoing
specification.  This made the shader declaration process unable to
occur concurrenty for different shaders by different threads.

This patch introduces new versions of Parameter(), Shader(), and
ConnectShaders() that take an explicit `ShaderGroup&` parameter and
don't depend on any saved state. Remember that ShaderGroupBegin
returns a ShaderGroupRef, it's this dereferenced object that is passed
to the other functions that help specify the contents of the group.
This allows multithreaded apps to be building the scene, including
shader specification, from many threads concurrently.

The old stateful versions remain, and are fine to continue using for
any apps that are sure that all shader declaration will be done
serially (though we will perhaps mull over whether to eventually
deprecate it entirely).  Though internally, the amount of saved state
is now reduced to merely the one shared pointer that designates which
is the single active group being specified, the other former bits of
state either being entirely redundant or moving to be local to the
group itself.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -364,17 +364,20 @@ public:
     //    /* First layer - texture lookup shader: */
     //       /* Specify instance parameter values */
     //       const char *mapname = "colormap.exr";
-    //       ss->Parameter ("texturename", TypeDesc::TypeString, &mapname);
+    //       ss->Parameter (*group, "texturename", TypeDesc::TypeString,
+    //                      &mapname);
     //       float blur = 0.001;
-    //       ss->Parameter ("blur", TypeDesc::TypeFloat, &blur);
+    //       ss->Parameter (*group, "blur", TypeDesc::TypeFloat, &blur);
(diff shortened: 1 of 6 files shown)


fpsunflower:
LGTM!

Looks like exactly what I needed :+1:

lgritz:
Chris, this passes my usual testsuite (I even changed testshade/testrender to the new API calls), but I don't have a test that attempts to actually assemble shaders concurrently. Do you want me to merge, or would you like me to leave it here until you can pull it from my branch and try it out on your end for a sanity check for the thing you are working on?

fpsunflower:
Its easier to test if you just merge it. The changes look relatively safe - and it feels like a nice API addition.

lgritz:
Ok, then. Merging.

PR #985, opened 2019-03-21
Small fix for ConnectShaders

fpsunflower:
Recursive invocations need to use the same group. This bug only occurs when using the brand new APIs just added for threaded shader definition.


lgritz:
Oops. Of course. LGTM.

PR #988, opened 2019-03-26
Fix oslc assertion on type constructor for unknown struct

lgritz:
The assertion was not needed, errors issue before this code is reached.

### DIFF
--- a/src/liboslcomp/typecheck.cpp
+++ b/src/liboslcomp/typecheck.cpp
@@ -2272,8 +2272,12 @@ OSLCompilerImpl::code_from_type (TypeSpec type) const
             out = 's';
         else if (elem == TypeDesc::NONE)
             out = 'x';
-        else
-            ASSERT (0);
+        else {
+            out = 'x';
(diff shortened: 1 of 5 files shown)


src/liboslcomp/typecheck.cpp
```diff
@@ -2272,8 +2272,11 @@ OSLCompilerImpl::code_from_type (TypeSpec type) const
             out = 's';
         else if (elem == TypeDesc::NONE)
             out = 'x';
-        else
-            ASSERT (0);
+        else {
+            // error(ustring(), 0, "Possible compiler bug: %s(%s)",
```
fpsunflower: Remove this block of commented out code? (and maybe leave a comment at the end about how an empty string being returned signals an unknown type?)

lgritz: Updated.

lgritz: OK with this the latest update here?

fpsunflower:
LGTM!

PR #989, opened 2019-04-03
testoptix is dead. Long live testrender!

lgritz:
This extensive refactor fully merges all remaining testoptix
functionality into testrender, and deletes the remnants of testoptix.

There's still lots of refactoring I intend to do moving forward,
especially when I make an optix mode to 'testshade' as well. But this
is a good checkpoint to merge.


fpsunflower:
LGTM!

PR #991, opened 2019-04-05
Improve oslc type error detection for variable declaration with init

lgritz:
Turns out that some of the various type checking we do for simple
assignment,

    x = y

was not being checked at all for variable declaration with initialization,

    Type x = y

So this patch adds the necessary checks. And also goes back and improves
the error messages for regular variable assignment, changing from messages
like

    Cannot assign 'float' to 'int'

to the more clear (I think, because it names the variable, and looks
more like the grammatical construct being complained about)

    Cannot assign int i = float


fpsunflower:
LGTM

PR #993, opened 2019-04-09
oslc: slight adjustment to assignment typechecking

lgritz:
Consider int=float to be a warning, not an error.
This is partly because C allows it and we ran accross some
examples where it seems more strict than necessary.
(Of course it works without warning if you use a cast, showing
that any trucation is intentional.)


aconty:
LGTM2!

PR #996, opened 2019-04-16
testshade --optix

lgritz:
Works for simple cases. Use with caution, there is a LOT that doesn't
work in the OptiX path. Embarrassingly so. But this refactor makes it so
that the last of the old "testoptix" oriented tests now work using
"testshade --optix".

It also requires that environment variable TESTSHADE_OPTIX=1 must be
set.  This is for weird reasons that have to do with argument orders,
and will go away in the next refactor of testshade, so I don't even
want to bother explaining. It's about to get fixed.

There is a lot that doesn't work properly. Frankly, I would encourage
onlookers to wait for another couple of checkins from me before wasting
any time on this. The most embarrassing parts will be fixed soon.

To be honest, I don't recommend that people read this review too deeply.
This code is in flux and is about to get overhauled significantly yet again!
I'm just trying to checkpoint various stages of the work so that we keep
moving forward in steps that continue to work for all the ways we depend
on testshade and testrender without breaking them.


aconty:
LGTM

lgritz:
I'm going to merge this, but I'm getting spurious occasional errors on the Appveyor Windows build. I can't reproduce on Linux or OSX. On Appveyor/Windows, sometimes it succeeds, sometimes a few testshade tests fail. I haven't narrowed it down yet and don't have any good theories. I'm not even sure it's because of this PR or merely coincidental.

So I will merge anyway, then try to track down if it continues (and if it's not immediately fixed by the next batch of testshade changes).

Apologies if this makes it crash for anybody. If it does crash for you, though, could you please try to catch it in the debugger and tell me exactly where it's happening? Thanks.

PR #997, opened 2019-04-19
Improve FindOptiX.cmake logic

DeclanRussell:
## Description

Improves the logic for finding the OptiX libraries in two ways. Firstly, a fallback has been added to the OPTIX_find_api_library macro to set the library version based on the OPTIXHOME variable and our knowledge of the default directory structure of an OptiX install. Secondly, the logic now uses the version number that is already being fetched from the OptiX headers to correctly match the versioned OptiX libraries. 

## Checklist:

### DIFF
--- a/src/cmake/modules/FindOptiX.cmake
+++ b/src/cmake/modules/FindOptiX.cmake
@@ -39,13 +39,14 @@ macro(OPTIX_find_api_library name version)
     find_library(${name}_LIBRARY
         NAMES ${name}.${version} ${name}
         )
+    if (${name}_LIBRARY STREQUAL "${name}_LIBRARY-NOTFOUND")
+        if (WIN32)
+            set (${name}_LIBRARY "${OPTIXHOME}/lib64/${name}.${version}.lib")
+        else ()
+            set (${name}_LIBRARY "${OPTIXHOME}/lib64/lib${name}.so")
+        endif ()
+    endif()
 endmacro()
-
-OPTIX_find_api_library(optix 1)
-OPTIX_find_api_library(optixu 1)
-OPTIX_find_api_library(optix_prime 1)
-
-set (OPTIX_LIBRARIES ${optix_LIBRARY} ${optixu_LIBRARY} ${optix_prime_LIBRARY})
 
 # Pull out the API version from optix.h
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM, pending trying it out myself at work to verify that nothing breaks, and receiving the CLA from you guys (I know that's in progress).

@timgrant Do you have any opinions? Is there a more "official" NVIDIA-authored/maintained FindOptiX.cmake that is guaranteed to be bullet-proof and kept up to date with new OptiX releases?

lgritz:
(I'll try it at work on Monday; I don't have a good way to test it from home.)

lgritz:
Please ignore my slightly odd comment about the CLA... I received a "how do we send a CLA" email from the head of one company on the same day that you submitted your two PRs and assumed it was for your work, but I now see that it was coincidental and you are at an entirely different company. So I will need a CLA from your company as well. You can DM me to work out the details if you don't have this already in progress.

timgrant:
These changes look good to me.

The two most "official" FindOptiX.cmake versions are in the OptiX SDK and in the [OptiX Advanced Samples](https://github.com/nvpro-samples/optix_advanced_samples), but I can't see how they're substantially better.

lgritz:
@DeclanRussell Can you check out the FindOptiX.cmake in the OptiX SDK and make a judgment about whether it's a better idea to go with this PR or to just replace it entirely with the NVIDIA one? I think there's some utility in just using whatever the OptiX SDK includes, and absorbing any improvements they make going forward, unless diverging from that and maintaining our own brings some tangible advantage to us.

DeclanRussell:
Looking at the FindOptiX.cmake from the SDK it looks like the logic for finding the libraries functions the same as the the logic here prior to this PR. There is some additional logic in the SDK version regarding finding the DLLs which could be useful to add if you plan to run Optix tests a part of the testsuite but should probably be considered in its own ticket. Other than that the rest of the logic looks like legacy code for either finding 32 bit or mac OptiX libraries, neither of which are publicly available anyway. Therefore for now I would go with this pull request.

lgritz:
OK, thanks for looking into it. Let's stick to this PR, then.

lgritz:
LGTM and CLA was received. Merging.

PR #998, opened 2019-04-19
Allow OSL to build with USE_LLVM_BITCODE enabled on windows

DeclanRussell:
## Description
Enabling USE_LLVM_BITCODE is required to build the OptiX related features of OSL. Currently enabling this flag will fail to build on windows. There are a couple of reasons for this which are addressed by this pull request.

Firstly, the build uses a convenient bash script to turn LLVM-compiled bitcode into a C array that is compiled as C source code to be externally linked in to the build later. Unfortunately we cannot run bash files on windows. Therefore I have converted it into a python script which I found from [here](https://groups.google.com/forum/#!topic/osl-dev/FATRnv_WMuM).

Secondly, due to a type mismatch between the code compiled and the extern references i.e. the array in llvm_ops.bc.cpp being of type unsigned char[] and the extern type in llvm_instance.cpp being of type char[] this fails to link on windows. Therefore I have changed the types to allow these to link successfully. This does mean having to cast the array back to a char* to pass it as a parameter to the module_from_bitcode function which I am not super happy about but the alternative of changing the type to a char[] produces warnings on unix.

## Checklist:


src/cmake/cuda_macros.cmake
```diff
@@ -95,8 +95,8 @@ FUNCTION ( LLVM_COMPILE_CUDA llvm_src headers prefix llvm_bc_cpp_generated extra
     MAKE_CUDA_BITCODE (${llvm_src} "" llvm_bc "${extra_clang_args}")
 
     ADD_CUSTOM_COMMAND (OUTPUT ${llvm_bc_cpp}
-        COMMAND "${CMAKE_SOURCE_DIR}/src/liboslexec/serialize-bc.bash" ${llvm_bc} ${llvm_bc_cpp} ${prefix}
+        COMMAND "${CMAKE_SOURCE_DIR}/src/liboslexec/serialize-bc.py" ${llvm_bc} ${llvm_bc_cpp} ${prefix}
```
lgritz: If you chmod+x the .py file, that should be good enough on Unix, dunno about Windows, I'm always unsure if the `#!/usr/bin/env` trick will work there.

Is it better to just take no chances and make this line

    COMMAND python "${CMAKE_SOURCE_DIR}/src/liboslexec/serialize-bc.py" ${llvm_bc} ${llvm_bc_cpp} ${prefix}

??

lgritz:
The gist of it LGTM. Thanks for straightening this out!

Fails CI on Unix -- I think you need to `chmod +x` the serialize-bc.py python file, or to call python explicitly to run the script?

I will need the CLA (I know that's in progress) and I'll also need to wait until Monday to test this at work to make sure it doesn't break anything there.

lgritz:
Please ignore my slightly odd comment about the CLA... I received a "how do we send a CLA" email from the head of one company on the same day that you submitted your two PRs and assumed it was for your work, but I now see that it was coincidental and you are at an entirely different company. So I will need a CLA from your company as well. You can DM me to work out the details if you don't have this already in progress.

marsupial:
To avoid all the extra type/ casting changes, why not change line 12 of src/liboslexec/serialize-bc.py to:
f_out.write('char %s_block[] = {\n' % prefix)

src/cmake/cuda_macros.cmake
DeclanRussell: Pushed your second suggestion which appears to have worked great! Thanks

DeclanRussell:
@lgritz sure thing I'll get the CLA done once the office is back open tomorrow.

DeclanRussell:
@marsupial that is an option yes. Though this produced warnings for us which on our fork we promote to errors on unix. Is that the case here too?

marsupial:
Not worth holding anything up from my side, but I'm curious what the errors were when the type's actually align in declaration and definition ?

lgritz:
LGTM and CLA was received. Merging.

PR #1001, opened 2019-04-21
Pr/testsuite

marsupial:
## Description
Allows a build outside of the source tree to run and pass the testsuite.
Allows a build to test against it's possibly newer PTX versions rather than installing to PTX_PATH first.
Catch optix exceptions in testshade.

### DIFF
--- a/src/testrender/optixraytracer.cpp
+++ b/src/testrender/optixraytracer.cpp
@@ -56,6 +56,7 @@ OptixRaytracer::~OptixRaytracer ()
 }
 
 
+#ifdef OSL_USE_OPTIX
 
 std::string
 OptixRaytracer::load_ptx_file (string_view filename)
@@ -77,7 +78,7 @@ OptixRaytracer::load_ptx_file (string_view filename)
     return ptx_string;
 }
 
-
+#endif // OSL_USE_OPTIX
 
 bool
 OptixRaytracer::init_optix_context (int xres, int yres)
--- a/src/testrender/optixraytracer.cpp
+++ b/src/testrender/optixraytracer.cpp
@@ -28,10 +28,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include <OpenImageIO/filesystem.h>
+#include <OpenImageIO/sysutil.h>
 
 #include <OSL/oslconfig.h>
-
-#include <llvm/Support/FileSystem.h>
 
 #include "optixraytracer.h"
 #include "../liboslexec/splineimpl.h"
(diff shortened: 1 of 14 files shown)


marsupial:
All the comments are pretty related to me trying to get CUDA usage either a compile or link error when OSL was not built with Optix support.

I'm not terribly beholden to what I'm changing there, but I'm at least going to present a case before giving in:

Just for OSL, I get the intent... but this is a public header. So if it winds up being included it could change the meaning of **cudaMalloc** if someone had done a **using namespace OSL**, and it's not clear whether it'll either fail because of ambiguity or just silently change meaning.

And I'm not totally convinced that changing the meaning of **cudaMalloc/cudaFree** from "allocate this memory via Cuda" to "allocate some memory, no matter where" isn't going to cause larger headaches than CI-failure (i.e. if one translation unit winds up including _optix_compat.h_ to allocate, and another unit does the deallocation, which did not pick up the header)

Even the Exception (which is definitely more readable with the stubs) could fall victim.

```
void doInit(Things& thing) {
  using namespace OSL;
  try {
    thing.data = cudaMalloc();
    // What's was just called? Where will it be freed ?
  } catch (optix::Exception) {
    // Is this OSL::optix::Exception or ::optix::Exception ?
    // Might depend on include order or compiler and ambiguity rules
  } catch (::optix::Exception) {
    // Ok definitely an optix exception, but prior code will need to be changed.
  }
}
```

src/testrender/optixraytracer.cpp
```diff
@@ -54,22 +56,29 @@ OptixRaytracer::~OptixRaytracer ()
...
 OptixRaytracer::load_ptx_file (string_view filename)
 {
     std::string ptx_string;
-    std::string filepath = filename;
-    if (! OIIO::Filesystem::exists(filepath))
-        filepath = OIIO::Strutil::sprintf ("%s/%s", PTX_PATH, filename);
-    if (! OIIO::Filesystem::read_text_file (filepath, ptx_string)) {
+    std::vector<std::string> paths = {
+        OIIO::Filesystem::parent_path(
+            llvm::sys::fs::getMainExecutable("testrender",
```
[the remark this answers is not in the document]
marsupial: Ug, looked for that in Filesystem, not Sysutil

src/testrender/optixraytracer.cpp
```diff
@@ -58,15 +59,21 @@ OptixRaytracer::~OptixRaytracer ()
...
 OptixRaytracer::load_ptx_file (string_view filename)
 {
-    std::string ptx_string;
-    std::string filepath = filename;
-    if (! OIIO::Filesystem::exists(filepath))
-        filepath = OIIO::Strutil::sprintf ("%s/%s", PTX_PATH, filename);
-    if (! OIIO::Filesystem::read_text_file (filepath, ptx_string)) {
-        errhandler().severe ("Unable to load %s", filename);
-        ptx_string.clear();
+#ifdef OSL_USE_OPTIX
```
lgritz: Do we need this guard here? There is no code enclosed that makes OptiX API calls.

src/testshade/optixgridrender.cpp
```diff
@@ -62,6 +63,28 @@ OptixGridRenderer::OptixGridRenderer ()
 
 
 
+std::string
+OptixGridRenderer::load_ptx_file (string_view filename)
+{
+#ifdef OSL_USE_OPTIX
```
lgritz: Guard not necessary.

lgritz:
OK, good enough, and works for me, so I will merge.

PR #1002, opened 2019-04-21
Optix opcolor (start)

marsupial:
## Description
A bunch of simple commits before a larger one to get opcolor.cpp working with Optix.

7c16e9f might be me misreading the original implementation, but worth a good-look.

bba1477 adds come more custom Imath includes...something I think no-one wants, but tries to consolidate the includes and have oslconfig.h use Imathx.h as well.

I've had luck using #pragma clang force_cuda_host_device begin and using stock EXR includes, but want to check off my laptop if that's viable.


lgritz:
Oof, I really hate that we're taking on more and more *copies* of Imath files.
But I guess it makes sense, I don't have a better alternative at the moment.

My plan is to push these back to the upstream OpenEXR project ASAP, but even that is not a solution that eliminates this need, since even if the `__host__ __device__` nonsense went into OpenEXR today, it would have to propagate not only to the next release but to all the downstream clients to pick it up who might try to build OSL. So I guess there's no fix other than doing this ourselves on our copies, pushing back, and eventually eliminating our copies when our minimum required OpenEXR version has caught up to having them.

I guess this is a LGTM?

lgritz:
LGTM, merging.

PR #1003, opened 2019-04-22
Refactor testshade argument parsing

lgritz:
We had a weird chicken-and-egg problem after a recent testshade
refactor, where getargs parsing made the individual shadingsystem
calls to set up the group, but also set other options we wished had
been established before this parsing starts.

The solution is change the main getargs to NOT fully set up the shader
group, but instead to just stash away the group setup args for later
processing. Then after the renderer and SS are fully set up
considering all the other options, we re-parse the stashed options and
set up the shader group.

This fixes some embarrassing reliance on the TESTSHADE_OPTIX environment
variable during tests.


marsupial:
LGTM

PR #1004, opened 2019-04-24
Modify testsuite to run optix tests

lgritz:
We previously had each test run twice: once with -O0 and again with
-O2. Because one tests the actual literal code specified in the shader
(without the possibility of being optimized away) and the other tests
the optimization to make sure it still matches.

This patch extends the logic to allow the possibility of every test
running in optix mode as well, if the code base is built with
USE_OPTIX=1.

For now, it only runs optix mode if the individual test has "optix"
in its name, or if the test directory has a file called "OPTIX" present.
Also, if environment variable TESTSUITE_OPTIX=1, it will try to run
*all* tests with optix mode (that's just for me... most of them fail
at the moment).

The eventual goal is for these conditions to change, so that in a
USE_OPTIX=1 build, *all* tests run optix flavors unless not marked
with a NOOPTIX file in their directory.

At the moment, *6* tests pass for me in OptiX mode. Now we can start
attacking the features to get more and more tests to pass identically
on CPU and GPU.

I realized that there were many tests that are only trying to test
that oslc issues the right error or warning for various malformed
code, or that are only compiling the shader to test oslinfo, or a few
other such cases that run oslc only and never try to execute their
shaders. So I also liberally sprinkled NOOPTIX as well as NOOPTIMIZE
marker files in those test directories, to skip optix tests, and also
skip a second "optimized" run of the same test (which was pointless,
because it wasn't executing at all).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,6 +224,7 @@ macro ( TESTSUITE )
     if (_ats_FOUNDVAR AND NOT ${_ats_FOUNDVAR})
         set (_ats_LABEL "broken")
     endif ()
+    set (test_all_optix $ENV{TESTSUITE_OPTIX})
     # Add the tests if all is well.
     set (ALL_TEST_LIST "")
     foreach (_testname ${_ats_UNPARSED_ARGUMENTS})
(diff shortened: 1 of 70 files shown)


lgritz:
LGTM received on internal chat

PR #1006, opened 2019-04-25
Fix a subtle issue with parameter arrays

fpsunflower:
When specifying a parameter value for an unbounded array, supplying a scalar parameter would lead to a failed assert.

For example, shader declares: `string[] foo`, but the user provides the parameter type: `string` instead of `string[1]`.

It seems ok to me to allow this (at the very least, it shouldn't be an assert).


aconty:
LGTM

lgritz:
Certainly shouldn't be an assertion.

I'm on the fence about whether it should be ok to substitute a scalar when an array is needed. But I'm not sure I have a strong argument about it. Maybe the fact that it asserted before and generally nobody brought that to our attention means that in practice it doesn't come up enough to worry about whether we are creating some kind of moral hazard here.

LGTM, and it if causes problems by allowing this substitution, we can always revisit later.

PR #1009, opened 2019-04-30
testsuite runtest.py cleanup

lgritz:
* Fix bug from PR 1001, default value for missing OSL_SOURCE_DIR env
  variable should not have included the testsuite directory itself.

* Be sure to correctly redirect all stderr to the same saved file as
  stdout. This makes it easier to have testsuite entries that test
  warning or error output.


fpsunflower:
LGTM

PR #1010, opened 2019-05-01
Warn about missed optimizations

lgritz:
* Properly track gettextureinfo handle use in the stats.

* New ShadingSystem option: "opt_warnings" enables warnings about
  things that couldn't be optimized and may be performance issues.

* New ShadingSystem option: "gpu_opt_error" enables full error status
  of the subset of those warnings that are also hard no-go's when
  executing on GPUs.

As an example, the first such warning/error is for texture calls where
the name of the texture cannot be fully resolved to a known string
constant during the course of runtime optimization. We will add other
warnings/errors over time.

The warnings/errors are sent back to the renderer's ErrorHandler,
and as an example they look like this:

    Optimization warning:
      group: bumpy_material_14
      layer: disp_texture_2
      shaders/tex.osl:58: texture(): texture name cannot be reduced to a constant

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -335,6 +335,7 @@ TESTSUITE ( aastep allowconnect-err and-or-not-synonyms arithmetic
             noise-perlin noise-simplex
             pnoise pnoise-cell pnoise-gabor pnoise-perlin
             operator-overloading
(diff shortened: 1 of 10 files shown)


fpsunflower:
LGTM

PR #1011, opened 2019-05-01
Optix simplification / cleanup

marsupial:
## Description
Trying to simplify / cleanup portions of the current implementation.

Make OSL::pvt::gSplineData constant which can avoid avoid CPU->GPU upload.
Introduce BackendLLVM::llvm_load_string to localize the branching done when uploading strings.
Send actual bitcode bytes vs a point to std::vector (which is currently incurring a temporary allocation penalty)

### DIFF
--- a/src/liboslexec/backendllvm.h
+++ b/src/liboslexec/backendllvm.h
@@ -150,6 +150,15 @@ public:
     /// reside in a global variable, the groupdata struct, or a local value.
     llvm::Value *llvm_load_device_string (const Symbol& sym);
 
+    /// Convenience function to load a string for CPU or GPU device
+    llvm::Value *llvm_load_string (const Symbol& sym) {
+        if (!use_optix())
+            return llvm_load_value (sym);
+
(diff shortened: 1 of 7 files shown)


src/liboslexec/backendllvm.h
```diff
@@ -150,6 +150,15 @@ class BackendLLVM : public OSOProcessorBase {
     /// reside in a global variable, the groupdata struct, or a local value.
     llvm::Value *llvm_load_device_string (const Symbol& sym);
 
+    /// Convenience function to load a string for CPU or GPU device
+    llvm::Value *llvm_load_string (const Symbol& sym) {
```
lgritz: I feel like this would be better as:

    DASSERT(sym.typespec().is_string());
    return use_optix() ? llvm_load_device_string(sym) : llvm_load_value(sym);

lgritz:
Pending that change, LGTM.

lgritz:
LGTM, and my tests pass. Merging.

PR #1012, opened 2019-05-03
Optix compilation of opcolor.cpp

marsupial:
## Description
Fairly large and the Optix path will require new test-data for **testsuite/transformc** dues to the printing issues.  535a3f8 is a late addition to fix an issue that popped up yesterday running the testsuite locally so probably good to make sure it passes CI before changing any tests.

I think the `OptixRaytracer::synch_attributes` method in **testshade** will need to be made available to **testrender** as well (either through OSL itself or just a copy-paste), but left that out to get thoughts on which and make review a bit easier on the eyes.

## Tests
As above, **testsuite/transformc** will need to be updated.


src/liboslexec/opcolor.cpp
```diff
@@ -156,31 +248,29 @@ rgb_to_hsl (const COLOR3& rgb)
...
-template <typename COLOR3>
+template <typename COLOR3> OSL_HOSTDEVICE
 static COLOR3
 YIQ_to_rgb (const COLOR3& YIQ)
 {
-    static const Matrix33 M (1.0000,  1.0000,  1.0000,
-                             0.9557, -0.2716, -1.1082,
-                             0.6199, -0.6469,  1.7051);
-    return YIQ * M;
+    return YIQ * Matrix33(1.0000,  1.0000,  1.0000,
```
lgritz: Why this change? Are you confident that it won't run the Matrix33 constructor every time?

marsupial: The issue is that Cuda doesn't like statics (I guess because the parallel execution and/or the locking semantics which it can't ignore for C++11 ?).

I think Cuda was fine with a constexpr variable, but that is c++14.

I'll see what god-bolt says about the const fold and maybe test whether constructing vs. the lock that was there has any performance impact.

marsupial: My read of https://godbolt.org/z/vPjyUJ is that pretty much the same code is generated to do the math, just the locking goes away with this change ?

lgritz: OK

lgritz:
LGTM, mergine

PR #1013, opened 2019-05-07
CPU & Optix "rgb" <=> "sRGB" color transformation

marsupial:
## Description
Implement "rgb" <=> "sRGB" color transformation.

## Tests
Updated transformc to render 1 x 1 and enabled Optix test.


lgritz:
LGTM. Will merge after one more test on my end.

PR #1017, opened 2019-05-10
__attribute__((aligned())) does not work on Windows (no error is give…

ThiagoIze:
…n even though alignment never happens). Use C++11 alignas for cross-platform compatibility.

As we know from ```OIIO_ALIGN```, windows and linux use different mechanisms for specifying alignment before c++11. The cuda code is using the linux version and works on linux but on Windows it would compile without warning and then fail at runtime with alignment errors. The simplest fix is to just switch over to the C++11 ```alignas```, which is what we already do with ```OSL_ALIGNAS```. Note that this is cuda code so no need for the gcc fix in ```OSL_ALIGNAS```, but I suppose that could be safely substituted in if consistency is desired? 

## Checklist:


lgritz:
LGTM!

PR #1018, opened 2019-05-11
Enable OptiX on many tests

lgritz:
Add 'OPTIX' label files to a whole bunch of testsuite tests to force
them to run in OptiX mode also.

Make sure optix tests run both optimized and unoptimized.

Moany tests still ran 'testsuite -g 2 2' to run at 4 points as if on a
grid. This was an artifact from the early days when we had a "batched
shader bytecode interpreter" that handled "varying" specially and
needed to run on multiple points at once to ensure that they were
shading entire grids properly. But once we switched to LLVM JIT and a
"single point shading model", that was no longer necessary. But never
had a good reason to change them before.

Lots of tests "failed" OptiX mode just because printf wasn't working
correctly for multiple points. That's silly, we can test printf interleaving
separately and with one test. No need to make passing dozens or hundreds
of others contingent on that one problem, rather than hinging on whether
OptiX path can really implement the feature that the test was designed
to demonstrate.

So for a bunch of tests, I changed them from running 'testsuite -g 2 2' to
just running on one point, and enabled OptiX mode to get a pass.

All in all now, I believe we pass 37 tests in OptiX mode (both optimized
and not optimized). We will try to grow this week by week.

The review of this PR is not very exciting -- primarily it is just changing test
reference output that previously printed 4 points, into printing 1 point, and then
adding an 'OPTIX' file if the test passed in OptiX mode for both -O0 and -O2.


lgritz:
Got an off-list LGTM from Roman.

PR #1019, opened 2019-05-13
Install crash handlers with stack trace dump

lgritz:
Recently, OIIO sysutil.h added a way to set up a stack trace dump in
case of crash. Use those in OSL binary commands (it does not
automatically happen from the library). This only work with a new
enough version of OIIO, and only if that OIIO was compiled with a new
enough boost to contain boost's stack trace library (>= 1.65).

This is primarily to help developers figure out what's going on with
CI tests fail on remote machines -- you can't catch it in the
debugger, but maybe the stack trace will give you the clue you need as
to what's going on. I definitely don't expect it to be very helpful if
the crash happens in the JITed shader. But if the crash is in our code,
maybe it will help the detective work.

To that end, we also change the testsuite runtest.py so that when a
command fails entirely, print its console output.

And bump appveyor boost to 1.69, so that it's new enough to support
the stack trace library.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -49,7 +49,7 @@ before_build:
 
     $DWN_DIR = "C:\projects\OSLdownloads"
     $DEP_DIR = "C:\projects\OSLdeps"
-    $BOOST_ROOT = "C:\Libraries\boost_1_63_0"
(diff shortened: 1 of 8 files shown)


aconty:
LGTM!

PR #1026, opened 2019-05-24
Improve errors for badly formed serialized shaders

lgritz:
When using the variety of ShaderGroupBegin that takes a serialized
description of the group, errors in the individual 'shader', 'param',
or 'connect' commands were not bubbled up into a proper error message
for the ShaderGroupBegin call itself, nor caused it to return null
as the group.


fpsunflower:
I see now that each of the `Parameter`, `Shader` and `ConnectShaders` calls return a boolean. This seems sufficient for applications to handle errors on their end in the non-serialized case.

LGTM!

PR #1028, opened 2019-06-05
Add support for native OSL closures in OptiX

DeclanRussell:
## Description

The current implementation of the OptiX run time defines its own closure structures that replicate that of the native OSL closure structures. This is likely just legacy code as it was easier to use OptiX types like float3 than add support the native OSL types. Removing this duplication and using the native OSL closures improves the maintainability of the code as you will no longer need duplicate logic if the structures ever change.

Some notes regarding this PR that are worth thinking about:
* In OSL/oslclosure.h I have to align the ClosureComponent struct to 8 bytes instead of 16. This is because the memory pools we use to store these structs are aligned to 8 bytes and we need to align them in a compatible way to avoid misalignment issues.
* ~~There is an order dependency when including headers like OSL/oslclosure.h in device code. If OSL/oslconfig.h is not included prior, such defines as OSL_HOSTDEVICE will not be defined correctly and can causes compiler errors. You can avoid this by defining these defines in the build instead which I have found makes things a little nicer in other projects. Whether this is an issue or not is up for discussion, but I thought it was worth mentioning. This is likely outside of the scope of this PR anyway.~~

## Checklist:

### DIFF
--- a/src/include/OSL/oslclosure.h
+++ b/src/include/OSL/oslclosure.h
@@ -29,8 +29,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #pragma once
 
 #include <cstring>
-#include <OpenImageIO/ustring.h>
 #include <OSL/oslconfig.h>
+#include <OpenImageIO/ustring.h>
 
 OSL_NAMESPACE_ENTER
 
--- a/src/include/OSL/oslclosure.h
+++ b/src/include/OSL/oslclosure.h
@@ -94,17 +94,17 @@ struct OSLEXECPUBLIC ClosureColor {
 
     int id;
 
-    const ClosureComponent* as_comp() const {
+    OSL_HOSTDEVICE const ClosureComponent* as_comp() const {
         DASSERT(id >= COMPONENT_BASE_ID);
         return reinterpret_cast<const ClosureComponent*>(this);
     }
 
-    const ClosureMul* as_mul() const {
+    OSL_HOSTDEVICE const ClosureMul* as_mul() const {
         DASSERT(id == MUL);
         return reinterpret_cast<const ClosureMul*>(this);
     }
 
-    const ClosureAdd* as_add() const {
+    OSL_HOSTDEVICE const ClosureAdd* as_add() const {
         DASSERT(id == ADD);
         return reinterpret_cast<const ClosureAdd*>(this);
     }
@@ -120,20 +120,26 @@ struct OSLEXECPUBLIC ClosureColor {
 /// scaled to add parameters after the end of the struct. Alignment is
 /// set to 16 bytes so that 64 bit pointers and 128 bit SSE types in user
 /// structs have the required alignment.
+#ifdef __CUDACC__
+/// Notice in the OptiX implementation we align this to 8 bytes
+/// so that it matches the alignment of the memory pools.
+struct OSLEXECPUBLIC OSL_ALIGNAS(8) ClosureComponent : public ClosureColor
+#else
 struct OSLEXECPUBLIC OSL_ALIGNAS(16) ClosureComponent : public ClosureColor
+#endif
 {
     Vec3 w;                     ///< Weight of this component
 
     /// Handy method for getting the parameter memory as a void*.
-    void *data () { return (char*)(this + 1); }
-    const void *data () const { return (const char*)(this + 1); }
+    OSL_HOSTDEVICE void *data () { return (char*)(this + 1); }
+    OSL_HOSTDEVICE const void *data () const { return (const char*)(this + 1); }
 
     /// Handy methods for extracting the underlying parameters as a struct
     template <typename T>
-    const T* as() const { return reinterpret_cast<const T*>(data()); }
+    OSL_HOSTDEVICE const T* as() const { return reinterpret_cast<const T*>(data()); }
 
     template <typename T>
-    T* as() { return reinterpret_cast<const T*>(data()); }
+    OSL_HOSTDEVICE T* as() { return reinterpret_cast<const T*>(data()); }
 };
 
 
--- a/src/include/OSL/oslclosure.h
+++ b/src/include/OSL/oslclosure.h
@@ -30,7 +30,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <cstring>
 #include <OSL/oslconfig.h>
-#include <OpenImageIO/ustring.h>
 
 OSL_NAMESPACE_ENTER
 
--- a/src/testrender/cuda/rend_lib.cu
+++ b/src/testrender/cuda/rend_lib.cu
@@ -1,9 +1,9 @@
 #include <optix.h>
 #include <optix_device.h>
 #include <optix_math.h>
+#include <OSL/oslclosure.h>
 
 #include "rend_lib.h"
-#include <OSL/oslclosure.h>
 
 
 rtDeclareVariable (uint2, launch_index, rtLaunchIndex, );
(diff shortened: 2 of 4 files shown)


lgritz:
Nice, LGTM. I need to build and test this patch for OptiX when I get into the office and it it all looks fine I will merge.

Aside: The CI tests on Appveyor have been having sporadic failures for a while that I haven't been able to track down yet. Have you seen any unstable behavior on Windows on your side? I haven't heard anybody report it in real world use, and have never seen the failures on Linux or OSX.

DeclanRussell:
Great!

I haven't noticed anything personally, though most of my testing is on the optix tests. I tried running the test crashed in the CI on this pull request several hundred times but never managed to reproduce the crash. Interestingly however, the odd test did appear to hang for a split second, but did ultimately finish. Perhaps whatever is causing this hang on my machine is what is crashing the crash on the CI here?

src/include/OSL/oslclosure.h
```diff
@@ -120,20 +120,26 @@ struct OSLEXECPUBLIC ClosureColor {
 /// scaled to add parameters after the end of the struct. Alignment is
 /// set to 16 bytes so that 64 bit pointers and 128 bit SSE types in user
 /// structs have the required alignment.
+#ifdef __CUDACC__
+/// Notice in the OptiX implementation we align this to 8 bytes
```
fpsunflower: Would another solution be to force the GPU to align addresses to 16 bytes?

I realize that these types probably will not need to travel back and forth between CPU and GPU, but it seems preferable to keep the alignment the same, no ?

src/testrender/cuda/rend_lib.cu
```diff
@@ -20,56 +21,56 @@ OSL_NAMESPACE_EXIT
...
-    void* closure_component_allot (void* pool, int id, size_t prim_size, const float3& w)
+    void* closure_component_allot (void* pool, int id, size_t prim_size, const OSL::Color3& w)
     {
-        ((ClosureComponent*) pool)->id = id;
-        ((ClosureComponent*) pool)->w  = w;
+        ((OSL::ClosureComponent*) pool)->id = id;
+        ((OSL::ClosureComponent*) pool)->w  = w;
 
-        size_t needed   = (sizeof(ClosureComponent) - sizeof(void*) + prim_size + 0x7) & ~0x7;
+        size_t needed   = (sizeof(OSL::ClosureComponent) - sizeof(void*) + prim_size + 0x7) & ~0x7;
```
fpsunflower: Could use `(alignof(OSL::ClosureComponent) - 1)` instead of `0x7` to be a bit more explicit?

Also, while I realize that all the users of the pool are probably the same alignment right now, this isn't necessarily true in general. I believe the CPU side aligns first and then figures out the size (see code in the `SimplePool`  class).

src/testrender/cuda/rend_lib.cu
```diff
@@ -20,56 +21,56 @@ OSL_NAMESPACE_EXIT
...
         char*  char_ptr = (char*) pool;
 
         return (void*) &char_ptr[needed];
```
fpsunflower: Not really related to this review, but does anyone have any thoughts on how to handle overflow conditions? If I understand correctly, we are just hoping the pool is big enough here ...

src/include/OSL/oslclosure.h
lgritz: Remind me why the CPU side is 16? Are we really using SSE on these data and does the occasional one that's not 8-byte aligned instead of 16 really impact performance?

fpsunflower:
> There is an order dependency when including headers like OSL/oslclosure.h in device code. If OSL/oslconfig.h is not included prior, such defines as OSL_HOSTDEVICE will not be defined correctly and can causes compiler errors.

Can you explain this a bit more? `OSL/oslclosure.h` already includes `OSL/oslconfig.h`, so why does the order matter?

src/testrender/cuda/rend_lib.cu
timgrant: Sizing the closure pool is definitely a can of worms. Is it possible to statically bound the pool?

fpsunflower: I might be possible by just counting the closure creation/add/mul instructions in the post-optimized group -- but that is likely to be a very conservative value (ie: much bigger than necessary).

Its not an urgent issue -- just wondering what kind of error handling we might have available to us to at least diagnose when this happens.

DeclanRussell:
> > There is an order dependency when including headers like OSL/oslclosure.h in device code. If OSL/oslconfig.h is not included prior, such defines as OSL_HOSTDEVICE will not be defined correctly and can causes compiler errors.
> 
> Can you explain this a bit more? `OSL/oslclosure.h` already includes `OSL/oslconfig.h`, so why does the order matter?

Apologies, after looking at this again I realized this is actually incorrect. The issue was actually because Imathx.h is required to be included prior to the stock Imath headers in device code (see oslconfig.h). I've pushed e5057d0 that cleans up the headers a little for this.

src/include/OSL/oslclosure.h
DeclanRussell: Presumably 8 bytes is just the lowest common alignment needed for some data structures on device code, I don't know for sure though. Perhaps @timgrant can share some insight if this is the case and if aligning to 16 bytes would be ok?

timgrant: I think the main concern with increasing the alignment is that it might not make the best use of the closure pool, which is likely to be a limited resource. But that might not matter too much in practice. Would it be problematic to define the alignment differently for CPU and GPU?

fpsunflower: @lgritz The blender folks use SSE types in their closures and submitted the change to bump the alignment to 16.

@timgrant I think the expedient fix is to keep the code as is (different on cpu/gpu). Though I would still tweak the allocation code to use `alignof` so we only need to change it in one place if we ever change our mind.

lgritz:
So where are we with this PR? Are @DeclanRussell, @timgrant , and @fpsunflower all ok with it as it stands?

fpsunflower:
My comments about changing the `0x7` into `alignof(...)` still stands (and adjust the logic to align the current pointer first, rather than assuming everything allocated in the pool will have a matching alignment).

timgrant:
LGTM, pending @fpsunflower's suggestions.

src/testrender/cuda/rend_lib.cu
DeclanRussell: Sounds good to me!

Aligning the pointer sounds interesting. I'll check out the SimplePool class and see what I can do.

DeclanRussell:
Apologies for the lateness of these final changes, I got caught up attempting to build with CUDA 9.x on windows which turns out to be quite the challenge.

I've committed ee3988b to address @fpsunflower's suggestions. Ideally, we would just have a device side pool implementation similar to SimplePool, but that is a little outside the scope of this PR so I hope this logic will suffice for the time being.

lgritz:
LGTM and passes all my tests, with and without OptiX enabled. Merging!

lgritz:
Can I ask you to fix one more thing before I merge?

Can you put a comment near the `#include` that says *why* Imathx.h needs to be included before any OIIO files? I just fear that a year from now, somebody who doesn't remember this PR's details will move it back to the usual order (of "this-project" headers being last) and not know why things broke.

fpsunflower:
LGTM2!

DeclanRussell:
> Can you put a comment near the #include that says why Imathx.h needs to be included before any OIIO files? I just fear that a year from now, somebody who doesn't remember this PR's details will move it back to the usual order (of "this-project" headers being last) and not know why things broke.

Even better, I noticed that oslconfig.h already includes the OIIO headers required. So we can just remove the ustring include that was causing the issues as I have done in df4e666.

lgritz:
As an aside, the OpenEXR team is hoping to gear up for a 2.4 release in the coming months and I'm hopeful about getting the Cuda decorations merged upstream it in time, so that we can start to phase out maintaining separate copies on our side.

Issue #1025, opened 2019-05-23
Output closures appear to not be supported in OptiX


DeclanRussell:
### Problem

It appears that output closures are not supported when compiling OSL for OptiX.

**Expected behavior:**  
The shader should compile nicely into PTX ready for OptiX consumption.
**Actual behavior:**  
The code triggers and assert during the optimize_group method.

`C:\Some\Path\include\osl_pvt.h:275: failed assertion '! is_closure() && "Don't call this if it could be a closure"'`

### Steps to Reproduce

1. Attached is a modified version of testoptix that demonstrates the issue. You can run this with the following command line: testrender.exe -optix --debug2 -res 320 240 scene.xml out.exr
[OutputClosuresRepo.zip](https://github.com/imageworks/OpenShadingLanguage/files/3212065/OutputClosuresRepo.zip)

### Versions

* OSL branch/version: 1.11.0 -> master
* OS: windows
* C++ compiler: MSVC++ 14.0
* LLVM version: 8
* OIIO version: 2.1.0

timgrant:
Declan, I have reproduced your issue and understand what's going wrong.

During ShaderGroup optimization, shader parameters are initialized differently depending on whether nor not OptiX is enabled. In the OptiX-specific code path, there are several places where the type of the parameter is queried. Performing certain queries on closure parameters is not permitted. In this case, checking if a closure parameter is a string triggers the assert.

The patch to avoid the assert is pretty small:

```diff --git a/src/liboslexec/llvm_instance.cpp b/src/liboslexec/llvm_instance.cpp
index d4a947c..af49828 100644
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -497,7 +497,7 @@ BackendLLVM::llvm_assign_initial_value (const Symbol& sym, bool force)
         ll.op_branch (cond_val, no_userdata_block, after_userdata_block);
     }
 
-    if (use_optix() && ! sym.typespec().is_string()) {
+    if (use_optix() && ! sym.typespec().is_closure() && ! sym.typespec().is_string()) {
         ASSERT (! sym.has_init_ops() && "Init ops are not currently supported in OptiX");
 
         // If the call to osl_bind_interpolated_param returns 0, the default
@@ -510,7 +510,7 @@ BackendLLVM::llvm_assign_initial_value (const Symbol& sym, bool force)
 
         TypeDesc t = sym.typespec().simpletype();
[...]

PR #1029, opened 2019-06-06
OptiX closure param fix

timgrant:
## Description

As reported in #1025, ShaderGroup compilation fails when a shader uses closure parameters and OptiX is enabled. This is due to an improper type check that is only present in the OptiX path. The fix here is to check whether a parameter is of type closure before checking if it's a string.

## Tests

No new tests.

## Checklist:

### DIFF
--- a/src/testrender/testrender.cpp
+++ b/src/testrender/testrender.cpp
@@ -279,7 +279,7 @@ main (int argc, const char *argv[])
         rend->clear();
         delete shadingsys;
         delete rend;
-    } catch (const optix::Exception& e) {
+    } catch (const OSL::optix::Exception& e) {
         printf("Optix Error: %s\n", e.what());
     } catch (const std::exception& e) {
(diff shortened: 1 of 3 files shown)


src/testrender/testrender.cpp
```diff
@@ -279,7 +279,7 @@ main (int argc, const char *argv[])
         rend->clear();
         delete shadingsys;
         delete rend;
-    } catch (const optix::Exception& e) {
+    } catch (const OSL::optix::Exception& e) {
```
timgrant: Without the `OSL` namespace, both gcc and clang fail for me in Linux with this error: `ISO C++ forbids declaration of ‘optix’ with no type [-Werror=permissive]`. Since this change was necessary for me to compile and test, I decided to tack it onto this review.

lgritz:
LGTM, merging

PR #1030, opened 2019-06-18
Fix OSLQuery::init return value on success

marsupial:
## Description
Slight issue with OSLQuery::init always returning false.

## Tests
Testing against CI


lgritz:
Oops. LGTM

PR #1031, opened 2019-06-19
ASTNode: Make long chains of nodes (though m_next) free serially, not recursively

lgritz:
Avoids stack overflow for hideously complicated parse trees.


fpsunflower:
LGTM!

PR #1032, opened 2019-06-28
Respect lockgeom default for 'GroupSpecs' and remove ASSERT from ReParameter

marsupial:
Respect 'lockgeom' default when building from commands.

ReParameter was asserting after parameter lookup, if the symbol was not on the instance.
Since the call already has a return value to signal success, use that instead.


lgritz:
LGTM, I will merge when the CI tests return successfully.

PR #1033, opened 2019-07-01
Take advantage of proposed OIIO APIs to pass thread_info to texture related calls

fpsunflower:
This goes together with https://github.com/OpenImageIO/oiio/pull/2277

I've also attempted to constant fold get_texture_info even if the call would fail (and generate an error message to maintain runtime behavior).

Remove unused function get_texture_perthread from RendererServices to reduce confusion. It was never getting called internally.

### DIFF
--- a/src/include/OSL/rendererservices.h
+++ b/src/include/OSL/rendererservices.h
@@ -200,16 +200,13 @@ public:
 
     /// Given the name of a texture, return an opaque handle that can be
     /// used with texture calls to avoid the name lookups.
-    virtual TextureHandle * get_texture_handle (ustring filename);
+    virtual TextureHandle * get_texture_handle (ustring filename,
+                                                ShadingContext *context);
 
     /// Return true if the texture handle (previously returned by
     /// get_texture_handle()) is a valid texture that can be subsequently
     /// read or sampled.
     virtual bool good (TextureHandle *texture_handle);
-
-    /// Return a thread-specific opaque pointer to the texture system.
-    /// Knowing the context may help this go faster.
-    virtual TexturePerthread * get_texture_perthread (ShadingContext *context=NULL);
 
     /// Filtered 2D texture lookup for a single point.
     ///
@@ -322,14 +319,20 @@ public:
     /// can) use that extra information to perform a less expensive texture
     /// lookup.
     ///
-    /// Note to renderers: if sg is NULL, that means get_texture_info is
-    /// being called speculatively by the runtime optimizer, and it doesn't
-    /// know which object the shader will be run on.
-    virtual bool get_texture_info (ShaderGlobals *sg, ustring filename,
+    /// If the errormessage parameter is NULL, this method is expected to
+    /// handle the errors fully, including forwarding them to the renderer
+    /// or shading system. If errormessage is non-NULL, any resulting error
+    /// messages (in case of failure, when the function returns false) will
+    /// be stored there, leaving it up to the caller/shader to handle the
+    /// error.
+    virtual bool get_texture_info (ustring filename,
                                    TextureHandle *texture_handle,
+                                   TexturePerthread *texture_thread_info,
+                                   ShadingContext *shading_context,
                                    int subimage,
                                    ustring dataname, TypeDesc datatype,
-                                   void *data);
+                                   void *data,
+                                   ustring *errormessage);
 
 
     /// Lookup nearest points in a point cloud. It will search for
--- a/src/liboslexec/constfold.cpp
+++ b/src/liboslexec/constfold.cpp
@@ -55,7 +55,10 @@ static ustring u_nop    ("nop"),
                u_inversesqrt ("inversesqrt"),
                u_if     ("if"),
                u_eq     ("eq"),
-               u_return ("return");
+               u_return ("return"),
+               u_error  ("error"),
+               u_fmterror("%s");
+
 static ustring u_cell ("cell"), u_cellnoise ("cellnoise");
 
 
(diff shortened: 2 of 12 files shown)


fpsunflower:
Updated to not require any changes on the OIIO side.

src/liboslexec/constfold.cpp
```diff
@@ -2444,11 +2452,16 @@ DECLFOLDER(constfold_gettextureinfo)
             newop.argwrite (1, false);
             return 1;
         } else {
-            // Return without constant folding gettextureinfo -- because
-            // we WANT the shader to fail and issue error messages at
-            // the appropriate time.
-            (void) rop.texturesys()->geterror (); // eat the error
-            return 0;
+            // Constant fold to 0
```
fpsunflower: @lgritz This case seems to be broken, but I am not sure why. Can you spot any obvious mistakes in how I'm inserting the call to "error"? The output of the unit test suggests it isn't getting called at all.

lgritz: I'm not sure yet, but Ole just got me a test scene on Friday, that I'm looking into today, as a test case for deep dive into exactly why some of those don't get optimized away. Probably going to kick myself when I figure it out.

src/liboslexec/llvm_gen.cpp
```diff
@@ -2483,9 +2481,7 @@ LLVMGEN (llvm_gen_texture3d)
     args.push_back (rop.sg_void_ptr());
     RendererServices::TextureHandle *texture_handle = NULL;
     if (Filename.is_constant() && rop.shadingsys().opt_texture_handle()) {
-        texture_handle = rop.renderer()->get_texture_handle (*(ustring *)Filename.data());
-        if (! rop.renderer()->good (texture_handle))
-            texture_handle = NULL;
+        texture_handle = rop.renderer()->get_texture_handle (*(ustring *)Filename.data(), rop.shadingcontext());
```
lgritz: Does the logic change here slightly? It no longer tests for "good()", which encompasses both if the handle is null, but also if the file is broken somehow. So handles of broken files previously got changed to null here, but not now? Is that bad?

src/liboslexec/llvm_gen.cpp
```diff
@@ -2543,9 +2539,7 @@ LLVMGEN (llvm_gen_environment)
...
```
lgritz: No call to good()? Is that bad?

src/include/OSL/rendererservices.h
```diff
@@ -200,17 +200,14 @@ class OSLEXECPUBLIC RendererServices {
 
     /// Given the name of a texture, return an opaque handle that can be
     /// used with texture calls to avoid the name lookups.
-    virtual TextureHandle * get_texture_handle (ustring filename);
+    virtual TextureHandle * get_texture_handle (ustring filename,
```
lgritz: We need to make sure that the commit message is super clear that we are breaking compatibility with RendererServices. Renderers will need to update on their side. As soon as it's merged, we might want to also post to osl-dev just to be sure everybody knows.

src/liboslexec/constfold.cpp
fpsunflower: The shaders are using the idiom:

`gettextureinfo(filename, "channels", nchannel) == 0` as a way to detect if the file is valid.

Folding away this case seems worthwhile (although it isn't clear how often that case is actually triggered - since we never see an error message from it).



I also fixed the array case (did not require extra code), which should catch queries to `resolution` and `datawindow`, `displaywindow`. I'll push that in a second.

src/liboslexec/llvm_gen.cpp
fpsunflower: As you said -- the only thing the `filename` variant of the code on the OIIO side does is translate a filename to a handle before calling the handle version.

In other words -- if you already have a handle to the broken file, OIIO knows how to deal with that correctly and would not do anything differently if you called the filename version instead. So it seems like a pessimization to reset the handle to `NULL` in that case.

lgritz: OK, sounds reasonable.

lgritz:
Failing CI tests, looks like an error message that was previously issued is now MIA.

src/liboslexec/constfold.cpp
fpsunflower: Got it all working now.

The string in the testsuite output has to change, but that seems acceptable.

There are several cases where `gettextureinfo` fails but does not generate an error message and those cases can be folded without having to add a message into the opcode stream.

lgritz: Ah, that makes sense.

fpsunflower:
Updated the review. Error message is now working (modulo a tweak on the OIIO side to avoid doubled up messages).

The formatting of the error changes depending on the optimize flag since in one case the `get_texture_info` method of `RendererServices` is used, while in the other it looks like a plain shader error message. But the meat of the error is the same either way.

testsuite/gettextureinfo/ref/out-opt.txt
```diff
@@ -0,0 +1,56 @@
...
+green.tx averagecolor: 0.101961 0.501961 0.101961
+green.tx averagealpha: 1
+green.tx constantcolor: 0.101961 0.501961 0.101961
+green.tx constantalpha: 1
+green.tx foobar: not found (0)
+ERROR: Shader error [test]: Invalid image file "badfile": Image "badfile" does not exist. Also, it is not the name of an image format that OpenImageIO recognizes.
+
+badfile data: not found (0)
+green.tx exists? 1 (return 1)
+badfile3 exists? 0 (return 1)
```
lgritz: Is this problematic?

fpsunflower: Which part? This section of the output is the same as the old reference.

lgritz: Maybe just the part in parens? Maybe my fault?

lgritz: NO, never mind!

lgritz: "exists" succeeds for bad files, I forgot that.

lgritz:
LGTM!

Pending passing of at least the Linux CI tests. Travis is going slow today.

fpsunflower:
I should add this depends on https://github.com/OpenImageIO/oiio/pull/2278/ to pass correctly (otherwise a message gets doubled-up).

I generated the reference from the patched OIIO.

lgritz:
I'll straighten that all out with a back-port of the oiio part.

lgritz:
So don't worry if it fails, but only because of the test against the release branch of oiio, on this one test.

PR #1034, opened 2019-07-02
Avoid creating tiny std::vectors during constant folding where possible

fpsunflower:
Noticed we were making a bunch of tiny `std::vector`s while working on the `gettextureinfo` patch.

Probably a drop in the bucket compared to everything else that goes on during runtime optimization, but every little bit helps.

Also `rop.insert_code` also makes a fresh `Opcode` struct, so there's no need to tweak the read/write flags if they are the same as the default.

### DIFF
--- a/src/liboslexec/constfold.cpp
+++ b/src/liboslexec/constfold.cpp
@@ -1298,12 +1298,10 @@ DECLFOLDER(constfold_split)
         for (int i = 0;  i < n;  ++i)
             usplits[i] = ustring(splits[i]);
         int cind = rop.add_constant (TypeDesc(TypeDesc::STRING,resultslen),
-                                     &usplits[0]);
+                                     usplits.data());
         // And insert an instruction copying our constant array to the
         // user's results array.
-        std::vector<int> args;
-        args.push_back (resultsarg);
-        args.push_back (cind);
+        const int args[] = { resultsarg, cind };
         rop.insert_code (opnum, u_assign, args,
                          RuntimeOptimizer::RecomputeRWRanges,
                          RuntimeOptimizer::GroupWithNext);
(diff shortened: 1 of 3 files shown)


lgritz:
LGTM, although I have a suggest for how to make this even better:

Change

```
RuntimeOptimizer::insert_code (int opnum, ustring opname,
                               const std::vector<int> &args_to_add,
                               RecomputeRWRangesOption recompute_rw_ranges,
                               InsertRelation relation)
```

from vector to span, as follows:

```
void
RuntimeOptimizer::insert_code (int opnum, ustring opname,
                               cspan<int> args_to_add,
                               RecomputeRWRangesOption recompute_rw_ranges,
                               InsertRelation relation)
{
    insert_code (opnum, opname, args.begin(), args.end(),
                 recompute_rw_ranges, relation);
}
```

And then in a place where you did

        const int args_to_add[2] = { cosarg, rop.add_constant (c) };
        rop.insert_code (opnum, u_assign, args_to_add, args_to_add + 2,
                         RuntimeOptimizer::RecomputeRWRanges,
                         RuntimeOptimizer::GroupWithNext);

you can have the simpler (and original):

        const int args_to_add[2] = { cosarg, rop.add_constant (c) };
        rop.insert_code (opnum, u_assign, args_to_add,
                         RuntimeOptimizer::RecomputeRWRanges,
                         RuntimeOptimizer::GroupWithNext);

It will infer the right size of the array. It's less prone to bugs (especially cut-and-paste errors) if you don't need the caller to pass the length. And the same cspan signature will work with fixed length arrays like this, as well as for std::vector.

fpsunflower:
Updated. I do like this even better.

src/liboslexec/runtimeoptimize.cpp
```diff
@@ -495,17 +495,17 @@ RuntimeOptimizer::turn_into_nop (int begin, int end, string_view why)
 
 void
 RuntimeOptimizer::insert_code (int opnum, ustring opname,
-                               const int *argsbegin, const int *argsend,
+                               const OIIO::cspan<int> args_to_add,
```
lgritz: Just FYI, you shouldn't need the namespace here, in oslconfig.h, it aliases span and cspan inside the OSL namespace to be OIIO::span/cspan.

src/liboslexec/constfold.cpp
```diff
@@ -2727,19 +2708,15 @@ DECLFOLDER(constfold_pointcloud_search)
         int const_array_sym = rop.add_constant (const_valtype, const_data);
         // ... and add an instruction to copy the constant into the
         // original destination for the query.
-        std::vector<int> args_to_add;
-        args_to_add.push_back (value_args[i]);
-        args_to_add.push_back (const_array_sym);
+        const int args_to_add[2] = { value_args[i], const_array_sym };
```
lgritz: Not critical, but would you like this idiom more, or less, if this were

    cost int args_to_add[] = { ... }

(no magic numeral at all)???

fpsunflower: Yeah I guess that `2` is a bit redundant. I can update it.

lgritz: I'm just thinking, it's one less thing that can go wrong if at a later point args are added or deleted, or things are cut and pasted around.

fpsunflower:
Updated without magic constants.

lgritz:
LGTM

PR #1035, opened 2019-07-03
Constant fold array access even if they are out-of-bounds

fpsunflower:
This extends constant folding for `aref` even in the cases where the array index is out of bounds. If the user asked for range checking, we will generate an error instruction in the event that control flow reaches there, but the array index will be clamped regardless.

### DIFF
--- a/src/liboslexec/constfold.cpp
+++ b/src/liboslexec/constfold.cpp
@@ -57,7 +57,8 @@ static ustring u_nop    ("nop"),
                u_eq     ("eq"),
                u_return ("return"),
                u_error  ("error"),
-               u_fmterror("%s");
+               u_fmterror("%s"),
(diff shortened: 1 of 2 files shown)


src/liboslexec/constfold.cpp
```diff
@@ -778,19 +779,34 @@ DECLFOLDER(constfold_aref)
...
-            // isn't necessarily a reportable error, because it may be a
-            // code path that will never be taken.  Punt -- don't
-            // optimize this op, leave it to the execute-time range
-            // check to catch, if indeed it is a problem.
-            return 0;
-        }
-        ASSERT (index < A.typespec().arraylength());
+        int length = A.typespec().arraylength();
+        int index = *(int *)Index.data(), old_index = index;
+        if (index < 0) index = 0;
```
lgritz: Instead of these three lines, maybe:

    index = clamp(index, 0, length-1);

???

fpsunflower:
Addressing comments.

PR #1036, opened 2019-07-04
Avoid creating tiny std::vectors during LLVM code generation

fpsunflower:
This is inspired by the recent refactors in the constant folding code. There was an opportunity to apply the same idiom here as well as remove a bunch of un-necessary `std::vector` usage.

I made a small assumption about the number of arguments in one place (verified via an assert of course). Testsuite for OSL and our renderer suggest the limit is large enough.

### DIFF
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -30,6 +30,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <OSL/export.h>
 #include <OSL/oslversion.h>
+#include <OSL/oslconfig.h>
 
 #include <vector>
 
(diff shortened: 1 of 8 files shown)


src/include/OSL/llvm_util.h
```diff
@@ -354,36 +354,30 @@ class OSLEXECPUBLIC LLVM_Util {
     /// Generate code for a call to the function pointer, with the given
     /// arg list.  Return an llvm::Value* corresponding to the return
     /// value of the function, if any.
-    llvm::Value *call_function (llvm::Value *func,
-                                llvm::Value **args, int nargs);
+    llvm::Value *call_function (llvm::Value *func, OIIO::cspan<llvm::Value *> args);
```
lgritz: Let's change these to `cspan` (since we're in the OSL namespace already -- oslconfig.h already creates the right `using` type alias).

src/liboslexec/backendllvm.cpp
```diff
@@ -949,13 +949,13 @@ BackendLLVM::userdata_initialized_ref (int userdata_index)
...
-                                      bool deriv_ptrs)
+                                 OIIO::cspan<const Symbol *> args,
+                                 bool deriv_ptrs)
 {
-    std::vector<llvm::Value *> valargs;
-    valargs.resize ((size_t)nargs);
-    for (int i = 0;  i < nargs;  ++i) {
-        const Symbol &s = *(symargs[i]);
+    llvm::Value *valargs[16];
+    DASSERT(args.size() < int(sizeof(valargs) / sizeof(valargs[0])));
```
lgritz: I'm not crazy about this. Looks like  bug lurking if there's ever a function with > 16 args.
Let's leave the vector in this one case. Good candidate someday for a short_vector.

src/liboslexec/backendllvm.h
```diff
@@ -376,15 +376,24 @@ class BackendLLVM : public OSOProcessorBase {
...
     llvm::Value *llvm_call_function (const char *name, const Symbol &A,
-                                     const Symbol &B, bool deriv_ptrs=false);
+                                     bool deriv_ptrs=false) {
+        const Symbol *args[] = { &A };
+        return llvm_call_function (name, args, deriv_ptrs);
+    }
+    llvm::Value *llvm_call_function (const char *name, const Symbol &A,
+                                     const Symbol &B, bool deriv_ptrs=false) {
+        const Symbol *args[] = { &A, &B };
+        return llvm_call_function (name, args, deriv_ptrs);
```
lgritz: Just for fun, I think that this should also work:

    return llvm_call_function (name, { &A, &B }, deriv_ptrs);

yes?

src/include/OSL/llvm_util.h
fpsunflower: Sure, didn't notice the alias.

src/liboslexec/backendllvm.cpp
fpsunflower: Just to be clear - this would only trigger for a function handled by `gen_generic`. It seems a bit unlikely to have more than 16 arguments for a function that didn't need special handling at all.

But I'm happy to put in an overflow case that goes through `std::vector`.

src/liboslexec/backendllvm.h
fpsunflower: It does! And at that point, there isn't much need to add these overloads at all.

fpsunflower: I'll leave the overloads for the time being since they are just 1 line and save  a bunch of code.

For the `cspan<Symbol*>` case, the fact that the overloads take a `Symbol&` make the client code a smidge cleaner.

fpsunflower:
Addressing comments.

PR #1037, opened 2019-07-04
Fix sign/unsigned comparsison warning/error

marsupial:
## Description
With -Werror this is promoted to an error (on gcc-4.8)


lgritz:
LGTM!

Weird choice to begin with.

PR #1038, opened 2019-07-04
Bump minimum OIIO version to 2.0

lgritz:
We've been talking about this for a long time.
This is only in master. The 1.10 family of releases will always retain compatibility with OIIO 1.8.


fpsunflower:
LGTM!

Issue #1041, opened 2019-07-19
Crash in optimize_group when calling rotate in OptiX mode


DeclanRussell:
### Problem

I'm experiencing a crash when using the rotate function in OptiX mode. As far as I can tell it seem to stem from calling the `transform` function. Perhaps there are some ops that are not implemented yet. I get the following stack trace when running optimize_group a shader with it included.

```
>> OSL_v1_11::pvt::LLVM_Util::void_ptr
* OSL_v1_11::pvt::BackendLLVM::llvm_call_function
* OSL_v1_11::pvt::llvm_gen_generic
* OSL_v1_11::pvt::llvm_gen_transform
* OSL_v1_11::pvt::BackendLLVM::build_llvm_code
* OSL_v1_11::pvt::BackendLLVM::build_llvm_instance
* OSL_v1_11::pvt::BackendLLVM::run
* OSL_v1_11::pvt::ShadingSystemImpl::optimize_group
```

**Expected behavior:**  
Don't crash.
**Actual behavior:**  
Crash.

### Steps to Reproduce

Here is a patch for the testoptix test to reproduce the issue. Just simply run the test to reproduce the crash.
`testrender.exe -optix scene.xml -r 320 320 out.exr`
```
diff --git a/testsuite/testoptix/matte.osl b/testsuite/testoptix/matte.osl
index 77284cc..ca31f72 100644
--- a/testsuite/testoptix/matte.osl
+++ b/testsuite/testoptix/matte.osl
@@ -39,5 +39,7 @@ matte
             float UImin = 0, float UImax = 1 ]]
   )
 {
-    Ci = Kd * Cs * diffuse (N);
+    point Y = rotate(point(1, 0 , 0), 0.5, point(1, 0 , 0), point(0, 1, 0));
+    color X = color(Y);
+    Ci = diffuse (N) * X;
 }
```

### Versions

* OSL branch/version: master
* OS: Windows/Centos 6
* C++ compiler: clang
* LLVM version: 5.0.2
* OIIO version: master

PR #1042, opened 2019-07-21
Fix crash in optimize_group when compiling transform command for OptiX

DeclanRussell:
## Description

Here is a fix for #1041. @timgrant added a fix for the crash. Essentially matrices were not being dealt with in `BackendLLVM::getOrAllocateCUDAVariable`, hence the crash. The rest of these changes are to port the correct matrix ops to OptiX that are needed to use the transform functions. This meant adding the device decorations to the `Matrix44` class in `ImathMatrix_cuda.h` and to the matrix ops in `opmatrix.cpp`. I've only done the simple ops as the others will need some more thinking to port and also don't help fix the problem at hand.

## Tests

I've tested this against the repro in #1041 and can confirm it fixes the issue.

## Checklist:

### DIFF
--- a/src/include/OSL/ImathMatrix_cuda.h
+++ b/src/include/OSL/ImathMatrix_cuda.h
@@ -443,35 +443,35 @@ template <class T> class Matrix44
 
     T           x[4][4];
 
-    T *         operator [] (int i);
-    const T *   operator [] (int i) const;
+    IMATH_HOSTDEVICE T *         operator [] (int i);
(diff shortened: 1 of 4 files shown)


marsupial:
Passing on my MacBook, I can try on a workstation tomorrow.

lgritz:
Can you build on your MacBook with USE_OPTIX=1?

marsupial:
Yes, requires a bit of CMake fun (and dropping to shader model 3.0)

marsupial:
Passing on Linux too!

PR #1046, opened 2019-07-24
Raise minimum LLVM to 6.0, drop LLVM 5

lgritz:
LLVM 9 is about to branch... supporting 6,7,8 and soon 9 is more than enough, we don't
need to support 5 versions back.

Also, LLVM 6 is the minimum that can be used with the new OptiX work, so 5 will soon be very undesirable if not already.


lgritz:
Brought this up to date. LLVM 10 release is February, dropping 5 seems very reasonable.

fpsunflower:
LGTM

PR #1047, opened 2019-08-01
Some fixes for mingw compiler

Alexpux:
1. Allow to use pugixml-1.9
2. MIngw compiler use ````__declspec```` macroses too for export symbols
3. Fix export symbols by liboslexec to allow properly link osltoy

### DIFF
--- a/src/cmake/modules/FindPugiXML.cmake
+++ b/src/cmake/modules/FindPugiXML.cmake
@@ -12,12 +12,12 @@ find_path (PUGIXML_INCLUDE_DIR
            NAMES pugixml.hpp
            PATHS ${PUGIXML_HOME}/include
            /usr/local/include
-           /usr/local/include/pugixml-1.8)
+           PATH_SUFFIXES pugixml-1.8 pugixml-1.9)
 find_library (PUGIXML_LIBRARY
               NAMES pugixml
               PATHS ${PUGIXML_HOME}/lib
               /usr/local/lib
-              /usr/local/lib/pugixml-1.8)
+              PATH_SUFFIXES pugixml-1.8 pugixml-1.9)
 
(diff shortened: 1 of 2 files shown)


lgritz:
LGTM, thanks!

We really need a signed CLA for any contributions to this project. Are you able to do that?

Alexpux:
@Igritz what I need to do?

lgritz:
Instructions here: https://github.com/imageworks/OpenShadingLanguage/blob/master/CONTRIBUTING.md

Alexpux:
@lgritz I understand. Can we apply this patch without CLA as it doesnt contain any authors code - just changing some ifdefs. I wil try to send CLA later. I'm maintainer of MSYS2 project and support hundreds of packages with thousands patches so upstreaming this patches always take too many time even if patch is very simple.
Regards,
Alexey.

Alexpux:
@lgritz I have more patches but I think that not all can be accepted:
https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-openshadinglanguage
You can look into them if interesting

lgritz:
@Alexpux You see that I accepted your PR to OIIO without CLA for this reason -- it's clear that there is no real code to protect. But I fully control that project and therefore can make such judgment calls. For OSL, it's owned by Sony and while they give us amazing latitude to develop this project in public, one of the very few conditions they have asked for is that we have a strict CLA policy without exceptions.

If you can't get a CLA signed, I (or somebody else who can sign a CLA) will have to somehow reconstruct all of these changes independently.

Alexpux:
@lgritz I'm send signed CLA to email opensource@imageworks.com

lgritz:
We received the CLA, thanks. I will merge as soon as I run one more test on my end.

PR #1049, opened 2019-08-06
Allow aggregate component access with .[xyzrgb]

lgritz:
Spatial geometry triples (point, vector, normal) now support A.x, A.y,
A.z as synonyms for A[0], A[1], A[2], respectively. It's not the
notation I prefer (obviously, or I would have done it 25 years ago),
but other people really dig it, so fine by me.

Colors now support A.r, A.g, A.b as synonyms for A[0], A[1], A[2],
respectively.  To be honest, this worries me a little because I fear
we will regret it if we ever shift to supporting spectral rendering
(and the components will not necessarily represent red, green, and
blue, and may have arbitrary number). But again, people really seem to
like this, so fine, but don't come complaining to me in a few years
when your shaders all need rewriting for your new spectral renderer.

In both cases, it's for direct reference to lvalues (variables, including
references passed as arguments to functions):

    val = P.x;   // yes
    P.y = val;   // yes
    function_with_float_param (P.z);   // yes

But you can't use this notations on arbitrary expressions:

    val = function_returning_color().r;   // no
    val = (P + N).y;                      // no, double ick

We are also NOT supporting "swizzling."

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@ Dependency and standards changes:
   we may drop support for OIIO 1.x.
 
 OSL Language and oslc compiler:
+* New syntax to reference color and point/vector/normal components as
+  named struct components, such as C.r, C.g, C.b, or P.x, P.y, P.z.
+  #1049 (1.11.0)
 * oslc compilation speed-ups with faster retrieval of source lines when
   pasted into oso output. #938 (1.11.0)
 * Writing to function parameters not marked as `output` was only
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -330,6 +330,7 @@ TESTSUITE ( aastep allowconnect-err and-or-not-synonyms arithmetic
             logic loop matrix message
             mergeinstances-nouserdata mergeinstances-vararray
             metadata-braces miscmath missing-shader
+            named-components
             noise noise-cell
             noise-gabor noise-gabor2d-filter noise-gabor3d-filter
             noise-perlin noise-simplex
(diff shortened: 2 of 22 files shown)


fpsunflower:
LGTM!

lgritz:
I'm going to leave this up for a few days before merging, to give people a chance to comment, since it's rare that we make syntactic changes.

MasterZap:
Thank you!!!

Question: Do you specifically care about color vs vector/point/normal type, or will color.x and vector.r also work (not that I particulary care, just curious)

Also, yeah, the spectre of spectral rendering... well.. it's not like every existing shader accessing the individual components wouldn't need to be rewritten anyway for that case (i.e. the problem is component access itself, not the syntax to acheive it), so that's kinda a red herring anyway, isn't it?

Especially since we have no color.numcomponents or similar in OSL (or do we?)

/Z

lgritz:
As I implemented it, .[rgb] for colors, .[xyz] for point, vector, normal.

We don't have a numcomponents, but it could be added.

lgritz:
Additional detail: this is 100% forward and back compatible. It's strictly an oslc-side change, so a shader that says "C.r" and is compiled with a new enough oslc to support that syntax can be executed just fine with an older ShadingSystem. Underneath, C.r is just a synonym for C[0].

MasterZap:
Well, calling it compatible is overstating it, since it's not source compatible to older versions... and I deal w. shaders in source form, not OSO form....   so this will give me some headaches. But I like the change enough to be prepeared to somehow endure those headaches. Not sure how.... but..... I will endure :)

PR #1053, opened 2019-08-31
Fix private header inclusion of opcolor.h

marsupial:
## Description
Fixes reliance on liboslexec/opcolor.h in client code.
There isn't a super-great simple-solution here while still keeping liboptix out of liboslexec.

Final version seem like it should be extensible enough to at least handle other data upload that needs to convert device strings...but uploading OCIO transforms will probably change all of this anyway?


lgritz:
LGTM,  I think. I'll test it tomorrow and merge if all seems ok on my workstation with optix.

lgritz:
Tests seem fine on my workstation, merging.

PR #1054, opened 2019-09-04
Add remaining matrix ops to optix

DeclanRussell:
## Description

In this pull request I have ported the remaining matrix ops to OptiX. This therefore allows renderers to add support for transforms (camera, raster, screen space etc..) in the OptiX runtime. The majority of this PR was just adding the device decorations for the correct functions and using `llvm_load_string` in stead of `llvm_load_value` in the correct places so that device strings are dealt with correctly. Notice that I have left the `osl_get_matrix` and `osl_get_inverse_matrix` functions as extern functions. The design of this is because they need some kind of renderer callback. As extern this allows the renderer to directly implement these ops in the same way we do with such ops as `osl_allocate_closure_component`. I have created the basic functions for these in the rend_lib.cu file but have left the actual implementation of the OSL transforms for another enthusiastic contributor as to not take all the fun.

## Checklist:

### DIFF
--- a/src/liboslexec/llvm_gen.cpp
+++ b/src/liboslexec/llvm_gen.cpp
@@ -1516,7 +1516,7 @@ LLVMGEN (llvm_gen_construct_triple)
         llvm::Value *args[] = { rop.sg_void_ptr(),
             rop.llvm_void_ptr(Result), rop.ll.constant(Result.has_derivs()),
             rop.llvm_void_ptr(Result), rop.ll.constant(Result.has_derivs()),
-            rop.llvm_load_value(Space), rop.ll.constant(Strings::common),
+            rop.llvm_load_string(Space), rop.ll.constant(Strings::common),,
             rop.ll.constant((int)vectype) };
         RendererServices *rend (rop.shadingsys().renderer());
         if (rend->transform_points (NULL, from, to, 0.0f, NULL, NULL, 0, vectype)) {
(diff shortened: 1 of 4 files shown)


src/liboslexec/llvm_gen.cpp
```diff
@@ -1516,7 +1516,7 @@ LLVMGEN (llvm_gen_construct_triple)
         llvm::Value *args[] = { rop.sg_void_ptr(),
             rop.llvm_void_ptr(Result), rop.ll.constant(Result.has_derivs()),
             rop.llvm_void_ptr(Result), rop.ll.constant(Result.has_derivs()),
-            rop.llvm_load_value(Space), rop.ll.constant(Strings::common),
+            rop.llvm_load_string(Space), rop.ll.constant(Strings::common),,
```
lgritz: Maybe in the long run, it's smart to make llvm_load_value "string-aware" so that we don't have to change all these (including in future PRs covering other parts of the code)? Also, do you have an extra comma on this line?

src/liboslexec/opmatrix.cpp
```diff
@@ -278,7 +288,12 @@ osl_transform_triple (void *sg_, void *Pin, int Pin_derivs,
             else
                 osl_transformn_vmv(Pout, &M, Pin);
         }
+#ifndef __CUDACC__
         else ASSERT(0);
```
lgritz: Is there merit to changing OIIO's ASSERT so that it's it's CUDA-safe? (Is that possible?)

src/liboslexec/llvm_gen.cpp
DeclanRussell: I like that idea, I'll look into it. Yup, you're correct that comma got away from me. Fixed in dc976f7.

src/liboslexec/opmatrix.cpp
DeclanRussell: Sure, supposedly there is support for asserts in CUDA. I've never really used them though as it requires you to compile your CUDA code in debug mode. Thats probably just bad practice on my part though.

src/liboslexec/llvm_gen.cpp
DeclanRussell: I've made `llvm_load_string` "string-aware" in 5fc8b1a and reverted the modifications in `llvm_gen.cpp`. I'm not sure how robust the ternary condition is but it seems to work correctly in my testing.

lgritz:
LGTM.

Let me run some tests on my end and I'll merge afterwards.

lgritz:
Tests fine on my end, merging.

Issue #1040, opened 2019-07-16
Failed assertion: Init ops are not currently supported in OptiX


DeclanRussell:
### Problem

Init opts are currently not supported in the OptiX run time. This means that using such notation as initializing vector parameters from global variables (e.g. `vector v = vector(u, v, 0)` ) is not supported and you will hit the assert in the tittle of this ticket. This notation is used quite frequently in the publicly available OSL shaders that the 3DSMax guys have written [here](https://github.com/ADN-DevTech/3dsMax-OSL-Shaders) and its a shame not to be able to use them. Is there any plans on adding support for these anytime soon? At least having a graceful fallback so that the run time doesn't crash with these shaders would be great!

### Steps to Reproduce
The following patch will modify the testoptix test to reproduce the issue. Simply run,
`testrender.exe -optix scene.xml -r 320 320 out.exr`
This will give the following assert,
`llvm_instance.cpp:495: failed assertion '! sym.has_init_ops() && "Init ops are not currently supported in OptiX"'`
```diff
--- a/testsuite/testoptix/checkerboard.osl
+++ b/testsuite/testoptix/checkerboard.osl
@@ -41,7 +41,7 @@ checkerboard
         [[  string description = "scale factor for s coordinate" ]],
     float scale_t = 4
         [[  string description = "scale factor for t coordinate" ]],
-    color Ca = color(1, 1, 1)
+    color Ca = color(u, v, 1)
         [[  string description = "color of even squares" ]],
     color Cb = color(0, 0, 0)
         [[  string description = "color of odd squares" ]],
diff --git a/testsuite/testoptix/scene.xml b/testsuite/testoptix/scene.xml
index 1a2956e..e33f646 100644
--- a/testsuite/testoptix/scene.xml
+++ b/testsuite/testoptix/scene.xml
@@ -4,7 +4,6 @@
    <ShaderGroup>
       param float scale_s 20;
       param float scale_t 20;
-      param color Ca 0.1 0.1 0.1;
       param color Cb 0.5 0.5 0.5;
       shader checkerboard tex;
       shader matte layer1;

```

### Versions

* OSL branch/version: master
* OS: Windows 10 / Centos 6.0
* C++ compiler: clang 5
* LLVM version: 5.0.2
* OIIO version: master

lgritz:
Fixed by #1057

Issue #1027, opened 2019-05-31
Writing color user data doesn't work in OptiX mode


DeclanRussell:
### Problem

The way the OptiX run time exposes user data is by directly implementing the `osl_bind_interpolated_param()` function. There is already an example implementation of the callback in testrender for strings. However, it seems that if you try to use the callback for retrieving colors (and maybe other data types) the result does not get correctly passed along.

### Steps to Reproduce

1. [Here](https://github.com/imageworks/OpenShadingLanguage/files/3241176/WriteUserDataColorOptix.txt) is a patch that hard codes the user data callback so that if color user data is requested, it returns white. The patch also modifies the testoptix test so that you can simply reproduce the results with the command: testrender.exe -optix -res 320 240 scene.xml out.exr. Notice that the checkerboard.osl shader which is used for the background of the test now should simply return a color provided by user data.
2. Notice that by running the test the background is black, when the expected result should be white.
![optix_out](https://user-images.githubusercontent.com/9109767/58704862-55923b00-83a5-11e9-8581-a91f11f4e132.png)

3. Interestingly, if the user data callback returns zero, the default value of the parameter is correctly retrieved.
![ret_zero](https://user-images.githubusercontent.com/9109767/58704928-870b0680-83a5-11e9-8c92-47defc784ab9.png)

### Versions

* OSL branch/version: 1.11.0 -> master
* OS: windows
* C++ compiler: MSVC++ 14.0
* LLVM version: 8
* OIIO version: 2.1.0

timgrant:
I'd have to dig a bit more to figure out exactly why the userdata lookup is failing, but in the meantime here is a patch that should unblock you: [userdata-fix.txt](https://github.com/imageworks/OpenShadingLanguage/files/3287412/userdata-fix.txt). It makes the userdata lookup behave more like the CPU version. Lookup by name, type, etc., is possible, but requires a bit of setup.

I could maybe roll this change into #1029, or if we encounter more userdata wrinkles we can do a new PR.

DeclanRussell:
Thanks Tim, I just got around to trying this patch out. Looks like its solved the issue as far as I can tell, nice one!

lgritz:
Long since solved, but this issue didn't seem to auto-close as intended.

PR #1056, opened 2019-09-10
OptiX Fixes - or - How I Spent My Summer Vacation

timgrant:
## Description
This is a handful of small fixes that have built up over the last several months. Apologies for their somewhat wide-ranging nature, we can break them into several PRs if that would be helpful.

Issue #1040 is addressed by allowing init ops in the OptiX path. This had been conservatively disallowed, but there aren't any obvious reasons that init ops won't Just Work™.

Issue #1027 highlighted some shortcomings in userdata initialization. I've made `rend_get_userdata` look and behave more like the host version. This issue also flushed out a bug in a little-used corner of rtVariable initialization which has been patched in recent drivers.

There was [an issue](https://groups.google.com/forum/#!topic/osl-dev/qwymeCvdFOw) raised on the osl-dev group about variable updates no longer working. This was due to metadata for OptiX variables not being created. I fixed that issue and squished out a bit of duplication in CUDA/OptiX variable creation. I also relaxed the constraints on the types and the alignment of CUDA variables. This is to allow a wider range of parameter types to be initialized.

Lastly, it was discovered that closure capture was not working properly for strings. DeviceStrings are essentially a `char**`, and sometimes it is convenient to copy the `char*` (like when writing a closure param) and sometimes it's better to handle the `char**` (like when assigning a temporary variable). I added a bool to `llvm_load_device_string` to determine the semantics.

## Tests
I have not added any new tests. If the existing testsuite does not cover init ops and userdata initialization in OptiX, can it be easily extended?

## Checklist:

### DIFF
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -509,7 +509,7 @@ BackendLLVM::getOrAllocateCUDAVariable (const Symbol& sym, bool addMetadata)
 
     // TODO: Figure out the actual CUDA alignment requirements for the various
     //       OSL types. For now, be somewhat conservative and assume 8 for
-    //       non-scalar types. See: createOptixVariable()
+    //       non-scalar types.
     int alignment = (sym.typespec().is_scalarnum()) ? 4 : 8;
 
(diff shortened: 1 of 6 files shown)


timgrant:
Sounds good, Larry. I should have done that from the get-go.

PR #1057, opened 2019-09-12
Allow init ops in OptiX

timgrant:
## Description
This is a fix for issue #1040. Init ops had been conservatively disallowed, but there is no obvious reason that they won't work. I'm also relaxing the type and alignment checks when creating OptiX/CUDA variables, which again had been somewhat conservatively restricted.

I'm leaving as a TODO finding the optimal alignment for each of the types. The current values should work well for strings, scalars, and point-like types, but might not be optimal for matrices.

## Tests
No new tests

## Checklist:


lgritz:
Fixes #1040

PR #1060, opened 2019-09-12
Make rend_get_userdata behave more like the host version

timgrant:
## Description
Issue #1027 highlighted some shortcomings in userdata initialization. This patch makes `rend_get_userdata` match the host version more closely, especially with regard to the `userdata_initialized` flags.

This issue also flushed out a bug in rtVariable initialization that has since been patched. In some cases, only the first member of an aggregate type was correctly initialized with a default value.

## Tests
No new tests.

## Checklist:

- [x] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [x] I have previously submitted a [Contributor License Agreement](http://opensource.imageworks.com/cla/).
- [x] I have updated the documentation, if applicable.
- [x] I have ensured that the change is tested somewhere in the testsuite (adding new test cases if necessary).
- [x] My code follows the prevailing code style of this project.

PR #1061, opened 2019-09-13
Overhaul OptiX variable creation

timgrant:
## Description

There was [an issue](https://groups.google.com/forum/#!topic/osl-dev/qwymeCvdFOw) raised on the osl-dev group about variable updates no longer working in optix. This was due to a missing call to `createOptixMetadata`, which I had inadvertently removed in #973. I fixed that issue and squished out a bit of duplication in CUDA/OptiX variable creation.

I followed @fpsunflower's suggestion from #1057 and now use `TypeDesc` instead of `std::string` to communicate the type when creating a CUDA variable.

## Tests
No new tests added, but existing OptiX tests pass.

## Checklist:

### DIFF
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -325,30 +325,26 @@ BackendLLVM::getOrAllocateLLVMSymbol (const Symbol& sym)
 
 
 llvm::Value*
-BackendLLVM::addCUDAVariable(const std::string& name, const Symbol& sym,
-                             int size, int alignment, const void* data,
-                             const std::string& type)
+BackendLLVM::addCUDAVariable(const std::string& name, int size, int alignment,
+                             const void* data, TypeDesc type)
 {
     ASSERT (use_optix() && "This function is only supposed to be used with OptiX!");
 
     llvm::Constant* constant = nullptr;
 
-    if (type == "float") {
+    if (type == TypeDesc::TypeFloat) {
         constant = llvm::ConstantFP::get (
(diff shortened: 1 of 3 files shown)


lgritz:
Tim, the first commit causes two unit tests to fail when built in optix mode.

timgrant:
Thanks for letting me know. I'll look into it ASAP.

timgrant:
What tests are failing for you? It's possible that the failures are a manifestation of the rtVariable initialization bug I mentioned in #1060. I believe that bug affected `paramval-floatpromotion` and `vararray-default`, or something in those genres.

If those are the tests, it's possible that you need a newer driver. I have tested the two latest [Linux drivers](https://www.nvidia.com/en-us/drivers/unix/) and can confirm that those tests are passing, along with the rest of the testsuite*.

Since it looks like you're using the OptiX 6.0 SDK (based on your `Makefile-bits`), you should try the long-lived branch (430.50).

lgritz:
The two that fail for me with your patch are mergeinstances-vararray.optix and vararray-default.optix.

You think that means I need a newer *driver*?

timgrant:
Yes, possibly. There was a bug in rtVariable initialization that caused array-like types to be improperly initialized in some cases. This PR restores the creation of the metadata that makes CUDA variables visible to the OptiX compiler, and thus exposes the bug. Without this PR, variables remain plain old global CUDA variables which are statically initialized.

lgritz:
Yep, driver upgrade seems to fix. Thanks. I'll  run one more full testsuite and then merge.

PR #1064, opened 2019-09-18
OptiX string closure params fix

timgrant:
## Description
No issue has been created for it, but it has been found that attempting to access string closure params in the CUDA wrapper produces illegal access errors. I made following the `char**` optional in `llvm_load_device_string` to handle cases where a `DeviceString` value is the source for a `memcpy`.

## Tests
I modified the `MICROFACET_ID` in `process_closure` to return one color if the distribution is `default` and another color in the non-default case, and added a corresponding test to the testsuite. It handles default and non-default values, as well as both states of `lockgeom` for the `dist` parameter.

## Checklist:


lgritz:
LGTM, and it passes all tests. Merging!

PR #1066, opened 2019-10-02
Fix a discrepancy in ordinary vs dual math

fpsunflower:
Division for regular numbers is just `a / b`. But for values going through the dual class it was computed as: `a * (1 / b)`. The more verbose version was chosen because the intermediate inverse value then gets re-used to update the derivative components.

But this has the unfortunate side-effect of making the Dual codepath make ever so slightly different decisions than the regular one. This can lead to some really head scratching debugging sessions were seemingly innocuous changes in the code would change the bits you got out of the code based on if the symbol was tagged as needing derivatives or not.

Even if the new logic is (potentially) slightly slower, the more robust behavior is worth it IMHO.


lgritz:
Chris, can you rebase on top of the new master that includes the fix I merged that fixes the CI that broke after I changed the OIIO cmake minimum version?

fpsunflower:
Sure. Just rebased & pushed.

lgritz:
CI failing. Seems like a few tests have changed their outputs as a result of this change?

fpsunflower:
Yes, a number of rendering tests need updating. I can update those that fail on my machine, but different CI setups have different sets of failures.

lgritz:
Yeah, just update so it passes on your workstation, and then we'll see how many failures are still left.

fpsunflower:
Updated the `render-*` tests. I see odd crashes on the windows side, but I'm not sure if that is related.

lgritz:
We've had spurious crashes on the Windows CI for a while, I haven't narrowed it down and nobody has reported it in the real world. 

Looks like your new ref images pass all the CI tests!

LGTM, merge at will.

lgritz:
Hang on, don't merge yet.

lgritz:
Alex Wells points out that the code we JIT for division (see llvm_gen_div in llvm_gen.cpp), we do a true divide for the "value" and the inverse-mul trick for the derivs. Perhaps that's what we should do here as well.

fpsunflower:
All good points. Let me take another stab at this.

fpsunflower:
Updated. The dual division operator is now consistent with what `llvm_gen_div` does (modulo the check for division by 0), and the normalize call is made consistent.

lgritz:
LGTM

PR #1073, opened 2019-11-06
Handle escaped string literals (like "\tfoo\n") more correctly.

lgritz:
"Unescape" all string literals in .osl and .oso as they are parsed,
then "escape" them properly for output.  We weren't being careful and
consistent before. This led to at least two subtle behavioral errors:

1. oslinfo would appear to add an extra set of escape characters to
   metadata values that contained escape codes.

2. If s = "\n", then strlen(s) was 1 if s was a local variable, but was
   incorrectly 2 (wowza!) if s was a parameter!

### DIFF
--- a/src/include/osl_pvt.h
+++ b/src/include/osl_pvt.h
@@ -710,7 +710,10 @@ public:
         return ((const ustring *)data())[index];
     }
 
-    /// Stream output
+    // Stream output. Note that print/print_vals assume that any string
+    // values are "raw" and they will be converted to C source code "escaped
+    // string" notation for printing. For example, a newline characer will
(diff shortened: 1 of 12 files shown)


fpsunflower:
LGTM!

lgritz:
Hold on, I discovered one more edge case I don't get quite right. Will fix today and post an update...

lgritz:
Slight update pushed.

Also note that a semi-related bug (at least inasmuch as it also concerned certain escaped strings) was found on the OIIO Strutil side: https://github.com/OpenImageIO/oiio/pull/2386

fpsunflower:
LGTM

PR #1074, opened 2019-11-12
Uncomment old lines to fix rpath on macosx

aconty:
Build stopped working with rpath after latest overhaul. This seems to fix it.


lgritz:
As a further clue, the analogous lines were not commented out on the OIIO side. So I suspect that what happened is that somewhere in the middle of the build system overhaul, I commented the lines out just to see if they were still needed, nothing broke so I never restored it, not realizing that it worked only because my environment DYLD_LIBRARY_PATH explicitly contained the right path.

PR #1075, opened 2019-11-17
Ensure locale independence

lgritz:
I found several additional places where we were not being locale
independent, so if run in a locale with "," as the decimal separators,
output would change or bad things would happen.


fpsunflower:
LGTM

PR #1076, opened 2019-11-18
Be more explicit about which string formatting convention we use

lgritz:
With OIIO 2.0 as our minimum, change all the internal error reporting
functions taking printf-style arguments to use the errorf() convention,
thus reserving the name error() for an eventual future where we use
C++20 format conventions.

Also make sure that we are consistently using the right OIIO functions
themselves -- sprintf, not format, for example.

This doesn't affect any of OSL's public APIs, this is all strictly
internal and does not break public API compatibility.


fpsunflower:
LGTM!

PR #1079, opened 2019-11-21
Remove pointless assertions in TypeSpec class

lgritz:


lgritz:
review oked by @fpsunflower offline

PR #1080, opened 2019-11-23
Assertion overhaul

lgritz:
New macros:

* OSL_ASSERT / OSL_ASSERT_MSG -- always test, always print error if the
  test fails, abort only in debug build.

* OSL_DASSERT / OSL_DASSERT_MSG -- debug mode only test, print, abort.
  Does nothing for release build.

Eschew the old OIIO macros ASSERT/DASSERT, since it is perferable to not
purposely abort when in release mode.

Ported over from OIIO copies of a few other related macros.

This patch just adds definitions, we will change the existing macros
in subsequent PRs.

### DIFF
--- a/src/include/OSL/oslconfig.h
+++ b/src/include/OSL/oslconfig.h
@@ -187,4 +187,63 @@ using OIIO::cspan;
 #  define OSL_ALIGNAS(size) alignas(size)
 #endif
 
+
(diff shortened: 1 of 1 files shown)


lgritz:
Merging because this only adds things, can't break existing code.

PR #1081, opened 2019-11-28
oslc --embed-source : embed preprocessed source at end of oso file

lgritz:
This can be useful in debugging situations, since the oso file would
completely contain a readable (though preprocessed) and recompilable
version of the source.

In the oso file, this is signified by the hint `%preprocessed_source`,
and following that the remainder of the file is assumed to be source
code or comments and will be ignored by the oso parser.

### DIFF
--- a/src/liboslcomp/oslcomp.cpp
+++ b/src/liboslcomp/oslcomp.cpp
@@ -418,6 +418,8 @@ OSLCompilerImpl::read_compile_options (const std::vector<std::string> &options,
             m_optimizelevel = 2;
         } else if (options[i] == "-Werror") {
             m_err_on_warning = true;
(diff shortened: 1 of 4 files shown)


aghiles:
Very nice to see "--embed-source" back from the dead! :)

fpsunflower:
LGTM!

I assume you are planning to add a shading system option that would optionally recompile the pre-processed source on `.oso` load?

lgritz:
Maybe in the future? I wasn't going to do it right away.

fpsunflower:
Another idea that might have advantages from a debugging standpoint is to let `oslc` compile `.oso` files that have this metadata present.

Then you could directly diff the old and new and see what the new compiler did differently (if anything).

PR #1082, opened 2019-11-28
osl internals: be clear about printf style formatting

lgritz:
Found a couple more internals to change the names to be extra sure we
keep straight which ones are "printf-style" formatting ("%6.2f"), versus
future C++ converstion of std::format (Python-like: "{:6.2}"). Currently
we use exclusively printf-style, but some day there may be a mix, so I'm
removing ambiguity in the names. This is strictly internal, does not
change any public APIs.


lgritz:
Any objections ?

aconty:
LGTM!

PR #1083, opened 2019-12-04
Work around Cuda problem with the {fmt} lib that OIIO bundles

lgritz:
Just disable it when compiling Cuda code -- which doesn't need this
functionality anyway. That skips the headers that Cuda is complaining
about.


lgritz:
Merging without review -- fixes critical bug I need right now, and I don't think it can possibly break anybody who isn't building OSL for OptiX (which is already broken if they don't have this fix).

PR #1085, opened 2019-12-13
oslc can generate dependency files

lgritz:
This patch adds support for `-M`, `-MM`, `-MD`, `-MMD`, `-MF` flags,
akin to how they are used in gcc or clang, to generate dependency
files that can be used by make, ninja, etc., to express header
dependencies of osl files.  For example, output might look like:

    $ oslc -MMD test.osl
    Compiled test.osl -> test.oso
    $ cat test.d
    test.oso: test.osl \
      dist/macosx/share/OSL/shaders/color2.h \
      dist/macosx/share/OSL/shaders/stdosl.h

`-MD` and `-MMD` output a dependency file (all headers for `-MD`, just
user headers for `-MMD`), by default to `<inputbasename>.d`, but
overrideable by `-MF<filename>`.

`-M` and `-MM` are versions of `-MD` and `-MMD` but that print the
resulting dependency list to stdout.

We also adjust the printing of "Compiled foo.osl -> foo.oso" so that
it does not happen in the case of `-E`, `-M`, and `-MM` when other
specific output is expected to comprise the whole of stdout.

Also, I *think* that we now no longer use Boost Wave at all on any
platforms, so I put in an `#error` compiler directive to verify. If
nobody complains, I'll soon excise all of the Boost Wave specific
code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -325,7 +325,7 @@ TESTSUITE ( aastep allowconnect-err and-or-not-synonyms arithmetic
             pnoise pnoise-cell pnoise-gabor pnoise-perlin
             operator-overloading
             opt-warnings
-            oslc-comma oslc-D
+            oslc-comma oslc-D oslc-M
             oslc-err-arrayindex oslc-err-assignmenttypes
             oslc-err-closuremul oslc-err-field
             oslc-err-format oslc-err-funcoverload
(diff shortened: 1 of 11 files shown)


fpsunflower:
LGTM

lgritz:
I updated the PR to add a test.

And in adding a test, I realized that there is an important use for distinguishing between -M/-MM and -MD/-MMD: I use the "user header only" versions to exclude the listing of dependency stdosl.h, since its absolute path is murder to create a proper comparison file for the testsuite.

PR #1087, opened 2019-12-14
Replace ASSERT -> OSL_ASSERT and DASSERT -> OSL_DASSERT.

lgritz:
The old macros were from OIIO, namespace polluting, and ASSERT always
aborted the process. The new macros won't conflict with other apps,
and OSL_ASSERT prints error always but only aborts for debug mode,
which discussion concluded was the preferred behavior.

I also policed up the whole set of assertions, and there were many
instances of:

* ASSERT that I deemed downgradable to DASSERT.

* ASSERT/DASSERT that I deemed entirely unnecessary and deleted.

* A few asserts that should always have been OIIO_CHECK_* unit test
  macros.

### DIFF
--- a/src/include/OSL/oslclosure.h
+++ b/src/include/OSL/oslclosure.h
@@ -94,17 +94,17 @@ struct OSLEXECPUBLIC ClosureColor {
     int id;
 
     OSL_HOSTDEVICE const ClosureComponent* as_comp() const {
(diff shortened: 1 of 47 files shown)


lgritz:
Any objections?

aconty:
LGTM!

PR #1088, opened 2019-12-16
Refactor implementation of template<class T, int PARTIALS> Dual

AlexMWells:
## Description

Change Dual<T,PARTIALS> to not use array based data members and avoid any control flow loops for operations on it.  The change better enables compilers to keep values in registers via SROA (Scalar Replacement of Aggregates) optimizations as well as enable more efficient SIMD vectorization of loops using dual types.

The generic implementation of Dual defers storage of base class DualStorage which can be specialized for 1, 2, or 3 PARTIALS.  Specialized DualStorage have individual data members: m_val, m_dx, m_dy, m_dz instead of an array m_elem[PARTIALS].  To provide index based access to the data members, a ConstIndex<int> is introduced.  The DualStorage provides element access to applicable ConstIndex<#>:
    constexpr const T& elem (ConstIndex<0>) const { return m_val; }
    constexpr const T& elem (ConstIndex<1>) const { return m_dx; }
    constexpr const T& elem (ConstIndex<2>) const { return m_dy; }
    T& elem (ConstIndex<0>)  { return m_val; }
    T& elem (ConstIndex<1>)  { return m_dx; }
    T& elem (ConstIndex<2>)  { return m_dy; }
And support for non-constant index based access is removed.  So generic Dual implementation that loops over indices has been updated to use a mechanism to statically unroll a loop at compile time emitting the loop body with the correctly typed ConstIndex<#>.  To make this simple, a macro:
OSL_INDEX_LOOP(__INDEX_NAME, __COUNT, ...).  In practice, this means changing
     for(int i=0; i < elements;++i) {
          elem(i) += value;
     }
to
     OSL_INDEX_LOOP(i, elements, {
          elem(i) += value;
     });

Replaced index based component helper with explicit named helpers:
comp(xkit,0), comp(xkit,1)
changes to
comp_x(xkit), comp_y(xkit)
As it turned out all accesses where with hard coded 0, 1, 2 so no loss in functionality and improved readability.

Update opcolor implementation to avoid subscript operators [0],[1],[2] and use named data members .x, .y. z to avoid pointer aliasing issues.
Update opcolor implementation to avoid mat[?][?] operators that loose multidimensional C array data type  and access the underlying c array directly mat.x[?][?]

## Tests

Dual2 is used in existing code and existing tests should exercise it.
## Checklist:

- [X ] I have updated the documentation, if applicable.
- [X ] I have ensured that the change is tested somewhere in the testsuite (adding new test cases if necessary).
- [X ] My code follows the prevailing code style of this project.

### DIFF
--- a/src/include/OSL/dual.h
+++ b/src/include/OSL/dual.h
@@ -90,9 +90,10 @@ OSL_NAMESPACE_ENTER
 #else
     namespace pvt {
         template<typename T>
-        static constexpr T static_min(T a, T b) { return (b < a) ? b : a; }
-    } // namespace pvt
-    // explanation of passing code block as macro argument to handle
+        static OSL_HOSTDEVICE constexpr T static_min(T a, T b) {
+            return (b < a) ? b : a;
+        }
+    }    // explanation of passing code block as macro argument to handle
     // nested comma operators that might break up the code block into
     // multiple macro arguments
     // https://mort.coffee/home/obscure-c-features/
@@ -153,12 +154,12 @@ public:
     // To better enable Scalar Replacement of Aggregates and other
     // transformations, CLANG has easier time if the per element
     // constructors declared.
-    OIIO_CONSTEXPR14 DualStorage() {}
-    OIIO_CONSTEXPR14 DualStorage(const T & val, const T & dx)
+    OSL_HOSTDEVICE OIIO_CONSTEXPR14 DualStorage() {}
+    OSL_HOSTDEVICE OIIO_CONSTEXPR14 DualStorage(const T & val, const T & dx)
     : m_val(val)
     , m_dx(dx)
     {}
-    OIIO_CONSTEXPR14 DualStorage(const DualStorage &other)
+    OSL_HOSTDEVICE OIIO_CONSTEXPR14 DualStorage(const DualStorage &other)
     : m_val(other.m_val)
     , m_dx(other.m_dx)
     {}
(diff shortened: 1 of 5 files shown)


ee33:
I haven't looked at the code, but just wanted to remark, avoiding array members is a good GPU optimization as well.  A struct with an array often can't live in registers and instead has to go to local memory (meaning, per-thread DRAM).

lgritz:
Interesting!

lgritz:
I tried to build this on my end and found that when compiled with USE_OPTIX=1, there are some missing OSL_HOSTDEVICE decorations on the constructors for the Dual specializations. Just minor typo kind of problem. I have fixed it on my end and will push an update to your branch to amend the PR.

lgritz:
Well, it's not seeming to allow me to push to your branch, @AlexMWells. I think you checked the box, it says "Add more commits by pushing to the PR/Dual branch on AlexMWells/OpenShadingLanguage", but it won't let me. Might be a mixup about whether I'm "lgritz" versus "imageworks", I dunno.

In any case, you can see the changes I made here: https://github.com/lgritz/OpenShadingLanguage/commit/d436aa7cd09f3a12ade8e8b0d295ab64bc2df0cf
Can you please make those changes on your end, Alex, and push an update?

AlexMWells:
Yes, I will do that right now.

AlexMWells:
I promoted the change to PR/Dual.

And yes I see on the pull request the box “Allow edits from maintainers.” Is checked.  Let me know if there is a different permission I need to go change on my fork to enable maintainers access to edit.

lgritz:
Thanks, Alex. I don't think you did anything wrong, I think the mixup may be on my end, where it's failing to understand that my ordinary account is the same as my role as the maintainer of all the imageworks projects. That's my theory anyway.

lgritz:
All LGTM, Alex. I will merge this. I might do a little stylistic touch-up later (nothing you did wrong, just a couple spots where maybe you introduced a macro without realizing that we already had one elsewhere that did the same thing, that kind of thing).

lgritz:
I did some extensive production benchmarking here, and just for completeness I will point out that this change does not seem to have any performance degradation on our production scenes.

PR #1089, opened 2019-12-18
Excise support for Boost Wave

lgritz:
We don't need it anymore, we use clang library preprocessor on all
configurations now.


fpsunflower:
LGTM!

lgritz:
This is your kind of patch, @fpsunflower !

PR #1091, opened 2020-01-06
Fix up some build issues.

marsupial:
## Description
Fixes some minor build issues:

CMake caches **PUGIXML_INCLUDE_DIR** on first invocation of **find_path** making the conditional never taken on re-configure.

nvcc complains about the usage of **stderr** in OSL_ASSERT_XXX, so make it use **printf**

nvcc warns about _break_ never being reached


lgritz:
LGTM!

I will merge after the CI tests come back.

PR #1092, opened 2020-01-07
Add generation of cmake config files for export

lgritz:
This lets projects directly use OSL as imported targets without any
"FindOSL.cmake" file.


lgritz:
Does anybody object to this?
I've tried it (using a personal project that depends on OSL) and it works great.

PR #1093, opened 2020-01-07
Some stylistic touch-ups on compiler related macros

lgritz:
* Port a bunch of compiler identification and manipulation macros from
  OIIO's platform.h to OSL's oslconfig.h. (Renaming OIIO_blah -> OSL_blah
  obviously.)

* Rename the recently-added OSL_INLINE -> OSL_FORCEINLINE to match the
  similar macro from OIIO and use OIIO's definition which is is more
  robust to a wider range of compilers. Rename many places where we
  previously used OIIO_FORCEINLINE to the OSL name.

* Rename the recently-added OSL_EXPECT_TRUE and OSL_EXPECT_FALSE ->
  OSL_LIKELY and OSL_UNLIKELY, porting and matching the similar macro
  that we already used extensively from OIIO. Rename all the places
  where we used it with the OIIO name.

* Change bare references to `__cplusplus >= 201402L` into
  `OSL_CPLUSPUS_VERSION >= 14`.

* Change stray OIIO_CONSTEXPR to OSL_CONSTEXPR which we have had for
  a while now.

### DIFF
--- a/src/include/OSL/oslconfig.h
+++ b/src/include/OSL/oslconfig.h
@@ -63,6 +63,83 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #    error "This version of OSL requires C++11"
 #endif
 
+
+
+// In C++20 (and some compilers before that), __has_cpp_attribute can
+// test for understand of [[attr]] tests.
+#ifndef __has_cpp_attribute
+#    define __has_cpp_attribute(x) 0
+#endif
+
+// On gcc & clang, __has_attribute can test for __attribute__((attr))
+#ifndef __has_attribute
+#    define __has_attribute(x) 0
+#endif
+
+// In C++17 (and some compilers before that), __has_include("blah.h") or
+// __has_include(<blah.h>) can test for presence of an include file.
+#ifndef __has_include
+#    define __has_include(x) 0
+#endif
+
+
+
+// Define OSL_GNUC_VERSION to hold an encoded gcc version (e.g. 40802 for
+// 4.8.2), or 0 if not a GCC release. N.B.: This will be 0 for clang.
+#if defined(__GNUC__) && !defined(__clang__)
+#  define OSL_GNUC_VERSION (10000*__GNUC__ + 100*__GNUC_MINOR__ + __GNUC_PATCHLEVEL__)
+#else
+#  define OSL_GNUC_VERSION 0
+#endif
+
+// Define OSL_CLANG_VERSION to hold an encoded generic Clang version (e.g.
+// 30402 for clang 3.4.2), or 0 if not a generic Clang release.
+// N.B. This will be 0 for the clang Apple distributes (which has different
+// version numbers entirely).
+#if defined(__clang__) && !defined(__apple_build_version__)
+#  define OSL_CLANG_VERSION (10000*__clang_major__ + 100*__clang_minor__ + __clang_patchlevel__)
+#else
+#  define OSL_CLANG_VERSION 0
+#endif
+
+// Define OSL_APPLE_CLANG_VERSION to hold an encoded Apple Clang version
+// (e.g. 70002 for clang 7.0.2), or 0 if not an Apple Clang release.
+#if defined(__clang__) && defined(__apple_build_version__)
+#  define OSL_APPLE_CLANG_VERSION (10000*__clang_major__ + 100*__clang_minor__ + __clang_patchlevel__)
+#else
+#  define OSL_APPLE_CLANG_VERSION 0
+#endif
+
+// Define OSL_INTEL_COMPILER to hold an encoded Intel compiler version
+// (e.g. 1900), or 0 if not an Intel compiler.
+#if defined(__INTEL_COMPILER)
+#  define OSL_INTEL_COMPILER __INTEL_COMPILER
+#else
+#  define OSL_INTEL_COMPILER 0
+#endif
+
+// Tests for MSVS versions, always 0 if not MSVS at all.
+#if defined(_MSC_VER)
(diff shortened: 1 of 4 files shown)


lgritz:
@AlexMWells can you take a look and approve? It renames some of the macros you set up in your dual overhaul, to match names we use elsewhere.

src/include/OSL/oslconfig.h
```diff
@@ -75,34 +152,55 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-        #define OSL_NOINLINE __declspec(noinline)
-    #else
-        #define OSL_NOINLINE __attribute__((noinline))
-    #endif
+
+// OSL_NOINLINE hints to the compiler that the functions should never be
+// inlined.
+#if defined(__GNUC__) || defined(__clang__) || __has_attribute(noinline)
+#    define OSL_NOINLINE __attribute__((noinline))
+#elif defined(_MSC_VER) || defined(__INTEL_COMPILER)
```
AlexMWells: Don't think the defined(__INTEL_COMPILER) is necessary, its more of a linuyx vs. windows thing.

src/include/OSL/oslconfig.h
```diff
@@ -75,34 +152,55 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+// Some compilers define a special intrinsic to use in conditionals that can
+// speed up extremely performance-critical spots if the conditional is
+// usually (or rarely) is true.  You use it by replacing
+//     if (x) ...
+// with
+//     if (OSL_LIKELY(x)) ...     // if you think x will usually be true
+// or
+//     if (OSL_UNLIKELY(x)) ...   // if you think x will rarely be true
+// Caveat: Programmers are notoriously bad at guessing this, so it
+// should be used only with thorough benchmarking.
```
AlexMWells: Perhaps add comment about use in SIMD loops.
// When used in SIMD loops, can influence generation of a test to see if 
// all data lanes are false with a branch to skip the following code block vs. always executing 
// the code block with mask.

AlexMWells:
LGTM

src/include/OSL/oslconfig.h
lgritz: Can you elaborate on what "can influence" means?

lgritz: @AlexMWells I'm going to merge this as-is, but I can separately modify the comment as you suggested. But just so I'm sure myself and also write the correct comment, can you say more about what you mean? Does using this in SIMD loops make things better or worse, and why?

AlexMWells: Sure.  when vectorizing and a conditional branch is reached, a compiler will 
1.  Always execute the block with masking (where the mask includes the result of the condition).
2.  Test the conditional on all data lanes to see if they are all false, then jump over the basic block.

For small basic blocks that just do simple math or assignments, it can be better to use approach 1 and just always execute it masked vs. extra instructions to test the condition mask and jmp.  jmp introduces possibility of branch misprediction and may inhibit out of order execution.

For larger or expensive blocks, the cost of the test/jmp is easily amortized and option #2 would be better.  

A compiler will use internal heuristics to make the best decision between #1 and #2 it can.  But heuristics don't cover all cases and may not cover corner cases.

By "influence", I mean that it has been observed that if the compiler is hinted that a branch is unlikely, that it will choose take take option #2.  We make use of this to hint to the compiler that an expensive branch should avoided execution.

Most common use case is for expensive error handling code.  An example would be a check for a result of 0 or NAN, and that if true repeats an algorithm or performs a more expensive version of the algorithm.  The 99% case is that the error won't be happening and we really want to truly skip that block of code.  Usually the heuristic get this, but sometimes the block maybe small but contain an expensive instruction like divide or sqrt.  So even though masked off, it will still take clock cycles.

The inverse is also true, if hinted that a block is "likely", a compiler may choose to test if all lanes are true and execute an "unmasked" version of the block, else jmp to masked version.

I think I am trying to say that  "OSL_LIKELY and OSL_UNLIKELY can influence compiler heuristics on always executing a code block masked vs. emitting a test and jmp to skip a code block entirely."

lgritz: Thanks, that is a very helpful explanation.

lgritz: Would it be correct to say that in loops you hope get auto-vectorized, you should probably be more motivated to use LIKELY/UNLIKELY hints whenever you know it to be true, than you perhaps would in ordinary scalar code where the benefit of this hint is more subtle and hit-or-miss whether it actually helps?

fpsunflower: In ordinary code, the compiler uses these hints to layout the code favorably to the branch-predictor. But the benefit is definitely more subtle outside of really hot inner loops.

AlexMWells: I wouldn't go adding them to every conditional statement.  Treat them a fine tuning to override or prevent a compiler heuristic from making the wrong decision.  And yes they can help scalar code as well, however the heuristic being wrong for scalar code is not as detrimental as for vectorized code.

My most common use cases are when I notice unexpected branches always executing, I can use OSL_UNLIKELY to alter that.  Or in cases where I want to make sure the compiler doesn't get it wrong (even if it didn't), maybe a series of if statements that as a developer I know 1 and only 1 will ever executed.    Sometimes it might be self documenting to a reader.

lgritz: Yeah, my experience is that for ordinary code, you need to benchmark, it may or not be helpful in any measurable way. I think Alex seems to be saying that it's much more broadly helpful in vectorized code.

PR #1094, opened 2020-01-09
Add missing pugi libs to testrender linkage

lgritz:


lgritz:
@sambler Can you give this a try?

lgritz:
This has been reported (on the oiio-dev mail list) to correctly fix the problem. Merging.

PR #1096, opened 2020-01-15
Math and aliasing improvements

AlexMWells:
## Description

Move inclusion of Imath*cuda.h from Imathx.h to oslconfig.h
Avoid aliasing in helper math functions by not using Vec3::operator [] and instead accessing .x, .y, .z data members directly.  Likewise for Vec2.
Avoid aliasing in helper math functions by not calling Matrix44::operator [] which returns a float *.  Instead use the x[4][4] datamember directly which keeps it type safe and avoids casting down to a float *.
Made math helper functions static OSL_FORCELINE where possible to avoid any collisions with other translation units who might be compiled with different compiler flags (like target ISA).
To make some of the aliasing avoidance changes, needed to introduce some free functions to be called instead of existing base class methods (which exist in a different open source project).
Partitioned general purpose inverse of Matrix44 into:
  test_if_affine
  affineInverse - fast path that is SIMD friendly
   nonAffineInverse - slow path to be used outside SIMD loop to
                      handle any non-affine matrices
Changed math helpers declared in OSL namespace to work on OSL::Matrix44, OSL::Vec3, etc. instead of Imath::Matrix44<T> as the helpers would not have function should OSL::Matrix44 be defined as anything other than Imath::Matrix44<T>.
Updated Matrix22 to avoid taking addresses with memset and memcpy that could prevent compiler Scalar Replacement of Aggregates optimizations.
Avoid aliasing issues in Matrix22 implementation by removing Matrix22::operator[] and using .x[2][2] directly and avoiding Vec2::operator[].
Removed default parameter to Matrix22::inverse(bool singExc) and added Matrix22::inverse() const noexcept, to allow for improved execution within SIMD loops.
Unrolled small loops in math helpers to avoid introducing control flow and dependencies during execution.
Updated opmatrix.cpp to call free function versions of transpose and multDirMatrix to utilized the versions added to Imathx.h
Added OSL_PRAGMA(unquoted) as a cross platform means of declaring a pragma
Added OSL_INTEL_PRAGMA to emit pragmas that are specific to ICC
Added OSL_CLANG_PRAGMA to emit pragmas that are specific to Clang
Added OSL_CLANG_ATTRIBUTE to emit attributes that are specific to Clang

## Checklist:

### DIFF
--- a/src/include/OSL/oslconfig.h
+++ b/src/include/OSL/oslconfig.h
@@ -152,6 +152,25 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     #endif
 #endif // OSL_DEBUG
 
+#if defined(_MSC_VER)
+    #define OSL_PRAGMA(aUnQuotedPragma) __pragma(aUnQuotedPragma)
+#else
+    #define OSL_PRAGMA(aUnQuotedPragma) _Pragma(#aUnQuotedPragma)
+#endif
+
+#if __INTEL_COMPILER >= 1100
+    #define OSL_INTEL_PRAGMA(aUnQuotedPragma) OSL_PRAGMA(aUnQuotedPragma)
+#else
+    #define OSL_INTEL_PRAGMA(aUnQuotedPragma)
+#endif
+
+#ifdef __clang__
+    #define OSL_CLANG_PRAGMA(aUnQuotedPragma) OSL_PRAGMA(aUnQuotedPragma)
+    #define OSL_CLANG_ATTRIBUTE(value) __attribute__((value))
+#else
+    #define OSL_CLANG_PRAGMA(aUnQuotedPragma)
+    #define OSL_CLANG_ATTRIBUTE(value)
+#endif
 
 // OSL_FORCEINLINE is a function attribute that attempts to make the
 // function always inline. On many compilers regular 'inline' is only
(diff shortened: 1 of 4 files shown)


lgritz:
LGTM!

I will merge on Monday when I'm back at work, after doing a double check that nothing fails in OptiX mode.

src/include/OSL/oslconfig.h
```diff
@@ -152,6 +152,25 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     #endif
 #endif // OSL_DEBUG
 
+#if defined(_MSC_VER)
+    #define OSL_PRAGMA(aUnQuotedPragma) __pragma(aUnQuotedPragma)
+#else
+    #define OSL_PRAGMA(aUnQuotedPragma) _Pragma(#aUnQuotedPragma)
```
lgritz: Why does this work? Oh, `_Pragma` is C99. Does that work for all non-MSVS C++ compilers we might deal with? (I guess it does if it passed the CI.)

lgritz:
Passes unit tests at work when running in OptiX mode. Merging.

src/include/OSL/oslconfig.h
AlexMWells: https://en.cppreference.com/w/cpp/preprocessor/impl
_Pragma is part of c++11 standard.  
It was not implemented in the versions of MSVC we were using.
https://developercommunity.visualstudio.com/content/problem/325605/vc-does-not-support-the-standard-pragma.html
So we have this conditional compilation to use the existing MSVC __pragma which doesn't accept quotes.  So we standardized OSL_PRAGMA to be unquoted and supply the required quotes for _Pragma itself.

Issue #866, opened 2018-02-20
oslc un-escapes the path to stdosl.h


LazyDodo:
### Problem

When installing oslc in a path that has \n in it ( such as c:\osl_test\new\ ) oslc will
fail trying to compile any shader. due to the path being un-escaped somewhere. 

### Steps to Reproduce

```
c:\osl_test>new\osl\bin\oslc.exe test.osl
c:\osl_test\test.osl:1: error: error: could not find include file: c:\osl_test
ew\osl/shaders/stdosl.h

FAILED test.osl

c:\osl_test>rename new old

c:\osl_test>old\osl\bin\oslc.exe test.osl
Compiled test.osl -> mytest.oso

c:\osl_test>
```
I validated with procmon that the path is indeed unescaped

### Versions

* OSL branch/version: tested 1.7.5 and 1.9.3
* OS: Windows
* C++ compiler: msvc 2015

lgritz:
Hi, sorry for this to slip through the cracks. Can somebody verify whether or not this is still a problem in the current master and/or release branch?

lgritz:
#1073 seems to be on a related topic, but I'm not sure if it addresses this particular bug.

LazyDodo:
Not a problem! Between the time of reporting and now I'm only aware of a single blender user bumping into this one. 

I can still reproduce the issue 
```
c:\osl>old\bin\oslc.exe test.osl
Compiled test.osl -> test.oso

c:\osl>rename old new

c:\osl>new\bin\oslc.exe test.osl
c:\osl\test.osl:1: error: error: could not find include file: c:\osl
ew/shaders/stdosl.h
FAILED test.osl

c:\osl>new\bin\oslc.exe --help
oslc -- Open Shading Language compiler 1.10.8
(c) Copyright 2009-2019 Sony Pictures Imageworks, et al. All rights reserved.
```

lgritz:
So the problem is specifically when the path to oslc contains a directory starting with n?

Or is it also a problem when the osl file is in a directory starting with n?

um, I mean,

    > c:\path\to\new\oslc test.osl

    > oslc c:\path\to\new\test.osl

Both problems, just first, or just second?

lgritz:
Proposed fix in #1101.

Would you by any chance be able to test this on Windows?
If it works, I can backport to release branch as well.

LazyDodo:
[...]

PR #1101, opened 2020-01-24
Windows fix to escape stdosl path

lgritz:
You see, we are taking a file path, which on Windows might look like
"c:\path\to\new\osl" (the Windows design mistake we will all live with
forever), and then injecting that into the osl source parse string as
a double quoted string, it will be unescaped and turn into
"c:\path<tab>o<newline>ew\osl", which of course is wrong.

This leads to hilarious bugs on Windows if OSL is installed someplace
where one of the directories in the sequence starts with an "n", as
above.


lgritz:
Closes #866

lgritz:
Does anybody with Windows knowledge think this is a bad idea?

PR #1103, opened 2020-01-26
Don't merge a layer if it's an entry layer

derekhaase:
## Description
Assertion in merge_instances() with a shader group containing multiple identical entry layers.

This was initially encountered in the 1.9 branch, the issue is still present in the master branch; however only when built with the debug flag, due to the OSL_DASSERT change.

The error in master is:
src/liboslexec/shadingsys.cpp:3159: merge_instances: Assertion 'B->unused()' failed.

To fix this I've added a check into merge_instances() to ensure that a group is not an entry layer.

## Tests
1. Yes I've added a new test case.

2. For the record I am unable to get my test case to fail when running make test.
It seems to only run (for me) using the non-debug build, user error on my part?
The command I was running was 
  make debug test TEST=mergeinstances-duplicate-entrylayers

I did however verify that I receive the assert when running testshade manually with the debug binaries.

## Checklist:

- [ X] I have previously submitted a [Contributor License Agreement](http://opensource.imageworks.com/cla/).
- [ X] I have updated the documentation, if applicable.
- [ X] I have ensured that the change is tested somewhere in the testsuite (adding new test cases if necessary).
- [ X] My code follows the prevailing code style of this project.

### DIFF
--- a/testsuite/mergeinstances-duplicate-entrylayers/a.osl
+++ b/testsuite/mergeinstances-duplicate-entrylayers/a.osl
@@ -1,17 +1,12 @@
 shader a
 (
     float frequency = 1,
-
     string noiseType = "perlin",
-
     color amplitude = 1,
-
     int seed = 0,
-
     output color out = 0
 )
 {
     point domain = point(u + seed, v + seed, 0) * frequency;
-
     out = noise(noiseType, domain[0], domain[1]) * amplitude;
 }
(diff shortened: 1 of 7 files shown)


testsuite/mergeinstances-duplicate-entrylayers/a.osl
```diff
@@ -0,0 +1,17 @@
+shader a
+(
+    float frequency = 1,
+
```
lgritz: Totally not critical, but is there a reason why every second line is empty?

lgritz:
LGTM. Let me try it out at work tomorrow to confirm, then I will merge.

testsuite/mergeinstances-duplicate-entrylayers/a.osl
derekhaase: Fixed

PR #1104, opened 2020-01-27
Speedup string concat op

fpsunflower:
## Description

The string concatenation op was going through `sprintf` which forced at least one
memory allocation to happen, in addition to the expense of going through the parsing logic, etc ... This was roughly doubling the cost of this (already expensive) operation. This commit implements a more direct concatenation via `memcpy`, and uses a small stack buffer in most cases.

## Tests

Production shader which was abusing string concat renders 10% faster (now only `ustring` creation shows up in vtune). Obviously the offending shader will be rewritten, but this seemed like a low hanging fruit.


lgritz:
Tests pass. Merging.

PR #1106, opened 2020-01-28
Address strict warnings about unused parameters to functions

lgritz:
Fool around with -Wextra (optional, off by default for now), see what
turns up.


lgritz:
The punchline is: unlike OIIO (which turned up about a dozen legit bugs after enabling these warnings), no actual bugs were found in OSL from enabling these warnings. But no reason not to commit the warning fixes.

fpsunflower:
LGTM

PR #1107, opened 2020-01-29
oslc: add -MT option

lgritz:
Completing the recent dependency generation feature, -MT lets you
customize the name of the dependency target.

Also, to match the C/C++ compilers, we now accept both

    -MF<filename>       # all as one arg
    -MT<target>

    -MF <filename>      # two args
    -MT <target>


leekerley:
LGTM

PR #1108, opened 2020-01-31
SIMD Friendly Math and Noise refinements to work inside SIMD loops

AlexMWells:
Introduce sfm (SIMD Friendly Math) namespace with versions of math functions friendly to use inside SIMD loops.

Introduce IntHashBasedNoise to serve as base for HashNoise & CellNoise.
Reimplemented PeriodicHashNoise and PeriodicCellNose to use an adapter to wrap their input values and reuse existing HashNoise & CellNoise implementations.
Introduce code generation policies for Perlin based based functions and Noise objects in order to resuse the same perlin noise functions with or without SIMD intrinsics.  Using a "scalar"(non-intrinsic) version of the algorithm is required when calling from inside a SIMD loop.
Defined Scalar versions of any Noise objects that used SIMD intrinsics internally.
Provide optimized inthash, sfm_simplexnoise, sfm_gabornoise implementations suitable for use inside SIMD loops.
Introduce wide.h with Block<T, int Width> to store data as structure of arrays internally, more Wide and Masked wrappers to Block<T> be added in future.
Added osl::conjunction and osl::enable_if_type for compatibility with older C++ versions that do not have their equivalents defined in std::
Added OSL_NON_INTEL_CLANG = [0 | 1] to identify when compiler is a truely clang, as OSX still has __clang__ defined when using Intel compiler.
Added OSL_NODISCARD
Added OSL_OMP_PRAGMA(unquoted_text) that will only be emitted when -DOSL_OPENMP_SIMD and -DNDEBUG (disabled in debug builds)
Added OSL_FORCEINLINE_BLOCK on supported compilers will recursively inline the following statement
Split common portions of gabornoise.cpp into gabornoise.h that can be shared with sfm_gabornoise.h (the SIMD friendly version of gabornoise).

### DIFF
--- a/src/include/OSL/oslnoise.h
+++ b/src/include/OSL/oslnoise.h
@@ -198,8 +198,8 @@ bjfinal (const int4& a_, const int4& b_, const int4& c_)
 /// hash an array of N 32 bit values into a pseudo-random value
 /// based on my favorite hash: http://burtleburtle.net/bob/c/lookup3.c
 /// templated so that the compiler can unroll the loops for us
-template <int N> OSL_HOSTDEVICE
-OSL_FORCEINLINE unsigned int
+template <int N>
+OSL_FORCEINLINE OSL_HOSTDEVICE unsigned int
 reference_inthash (const unsigned int k[N]) {
     // now hash the data!
     unsigned int a, b, c, len = N;
@@ -232,7 +232,7 @@ using RequireAllSame = typename std::enable_if<conjunction<std::is_same<ListT, H
 // NOTE: only accept unsigned int's so that overloads with unsigned int's will be chosen.
 // Callers with "int" parameters will need to static_cast<unsigned int>(param)
 template<typename ...ListT, typename = RequireAllSame<unsigned int, ListT...> >
-OSL_HOSTDEVICE OSL_FORCEINLINE unsigned int
+OSL_FORCEINLINE OSL_HOSTDEVICE unsigned int
 inthash (ListT... uintList)  {
     const int count = sizeof...(uintList);
 	const unsigned int iv[count] = {uintList...};
@@ -251,7 +251,7 @@ inthash (ListT... uintList)  {
 	// The upshot is no need to create a k[N] on the stack and hopefully
 	// simpler time for optimizer as it has no need to deal with while
 	// loop and switch statement.
-	OSL_HOSTDEVICE OSL_FORCEINLINE unsigned int
+	OSL_FORCEINLINE OSL_HOSTDEVICE unsigned int
 	inthash (const unsigned int k0) {
 		// now hash the data!
 		unsigned int start_val = 0xdeadbeef + (1 << 2) + 13;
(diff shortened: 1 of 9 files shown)


lgritz:
I have not forgotten about this, @AlexMWells , it's just a lot to digest.

The gist of this work LGTM, but I want to do some benchmarking on my end before merging, and also double check that nothing breaks in OptiX mode, which the CI tests do not check. I feel like I need to read through it deeply one more time, too.

After I merge, I'll take a pass at porting a lot of the SIMD-friendly versions of things in OIIO, back to OIIO, and perhaps eventually removing them from here if we are willing to live with the assumption that consumers of the SIMD OSL might be expected to have a higher version floor of OIIO than those who don't need those features.

lgritz:
A lot of this breaks the build when USE_OPTIX=1
I'm sure it's not rocket science to fix, but will require a pass over it. I might be able to do it, but probably not for a couple days.

src/include/OSL/oslnoise.h
```diff
@@ -253,437 +355,370 @@ inthash_simd (const int4& key_x, const int4& key_y, const int4& key_z, const int
...
+// combinations be specified using polymorphism.  Main reason is so that
+// different versions can pass parameters by value vs. reference.
+//template <typename B, typename F>
+//OSL_FORCEINLINE OSL_HOSTDEVICE F select (B b, const F& t, const F& f) { return b ? t : f; }
+
+OSL_FORCEINLINE OSL_HOSTDEVICE float select(const bool b, float t, float f) {
+    // NOTE:  parameters must NOT be references, to avoid inlining caller's creation of
+    // these values down inside the conditional assignments which can create to complex
+    // of a code flow for Clang's vectorizer to handle
+    static_assert(!std::is_reference<decltype(b)>::value, "parameters to select cannot be references");
```
lgritz: Aside: I think these static_assert calls are probably a holdover from when this function was a template. But now fully specialized to float, they don't serve a purpose. Some time later, we should just remove them.

lgritz:
Yes, the fixes build cleanly with `USE_OPTIX=1` and in passes the OptiX-enabled OSL testsuite.

I will merge.

lgritz:
Note: as I explained before, I intend to migrate a good portion of the "sfm" functions back into OIIO fmath.h and OpenEXR's headers, and then undo some of these changes to use those versions rather than replicating that code here. It will require that people wanting full performance of the SIMD batch shading will need sufficiently new OIIO+OpenEXR, but I'm ok with that restriction.

src/include/OSL/oslnoise.h
AlexMWells: I think I left the static assert in there as strong teeth vs. just a comment, should someone come along later and change the parameters to references the static assert would fire.  In fact I might have done that myself and not wanted to repeat the same mistake.  It took a while to track it down, as inlining an entire AST tree inside a select because it is a reference is not obvious

lgritz: I'm a big fan of defensive programming. :-)

lgritz:
Also passes our renderer's testsuite.

AlexMWells:
>  I intend to migrate a good portion of the "sfm" functions back into OIIO fmath.h and OpenEXR's headers, and then undo some of these changes to use those versions rather than replicating that code here

Yes, that is a solid plan.

PR #1111, opened 2020-02-06
Adding support for OptiX 7

cmstein:
## Description

This patch adds support for building against OptiX7.  OptiX6 support is still there and everything is controlled by  `#if (OPTIX_VERSION)`-styled directives. 

The trickiest part was setting the global DeviceString variables.  I found I needed to use two different methods for the globals:  (1) create a simple "library" of globals (this is created on the fly as CUDA/PTX code by `testrender` and `testshade`); and (2) set all the `llvm::GlobalVariable` variables to the appropriate values after we create a module from the llvm bitcode (see `llvm_instance.cpp -- BackendLLVM::run()`.  The first method seems to resolve global variable linking for cuda programs which are compiled by `nvcc` (eg. `wraper.cu`);  the second method seems to resolve global variable linking for cuda programs compiled by `clang++` (eg. `rend_lib.cu`).  I believe the test failure I'm seeing is the inability to properly resolve global variable linking for `.cpp` files that are compiled into llvm bytecode and then transformed into PTX at runtime (like `opnoise.cpp`).

An extra set of eyes would be welcome.

### Notes
The Optix7 part requires llvm/clang-9.0.

## Tests
Compared I'm  getting one failure (`testoptix-noise/scene_02.xml`).  At this point I'm stumped so I welcome any advice on resolving this issue.

## Checklist:


cmstein:
A few notes on how I'm building (NB: I think you need to use llvm-9.0):

cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
    -Dextra_nvcc_args="-g -G"\
    -DUSE_OPTIX=1 \
    -DOptiX_ROOT=/path/to/OptiX7 \
    -DCUDA_TOOLKIT_ROOT_DIR=/path/to/cuda \
    -DOPTIX_EXTRA_LIBS=lzma \
    -DCUDA_EXTRA_LIBS=/path/to/cuda/lib64/libnvrtc.so \
    -DLLVM_DIRECTORY=/path/to/llvm_9.0/ \
    ....

timgrant:
Thanks, Cliff. I have it building and will continue to read through the changes. Looking good!

src/testrender/optixraytracer.cpp
```diff
@@ -359,7 +1001,125 @@ OptixRaytracer::finalize_scene()
...
+    // Copy Quads to cuda device
+    CUDA_CHECK (cudaMalloc (reinterpret_cast<void **>(&d_quads_list)  ,                     sizeof(QuadParams  ) * scene.quads.size()));
+    CUDA_CHECK (cudaMemcpy (reinterpret_cast<void  *>( d_quads_list)  , quadsParams.data(), sizeof(QuadParams  ) * scene.quads.size(), cudaMemcpyHostToDevice));
+
+    // Fill in Quad shaders
+    CUdeviceptr d_quadsIndexOffsetBuffer;
+    CUDA_CHECK (cudaMalloc (reinterpret_cast<void **>(&d_quadsIndexOffsetBuffer), scene.quads.size() * sizeof(int)));
+
+    unsigned int  quadSbtRecord;
+    quadSbtRecord = OPTIX_GEOMETRY_FLAG_NONE;
```
timgrant: This flag should probably be set to `DISABLE_ANY_HIT` instead of `NONE`. Otherwise visibility depends on the traversal order (which can vary between architectures), and we might end up with "missing" geometry. An alternative is to supply `OPTIX_RAY_FLAG_DISABLE_ANYHIT` to the `optixTraceCall`, but it's generally better for performance to disable any-hit using the geometry flag if possible.

lgritz:
Where are we on this PR?

timgrant:
Sorry, Larry. I got a bit distracted. I'm making another pass over the changes now.

timgrant:
The change looks good overall. If there are rough edges, we can sand them down in future reviews. "Smooth as glass," as my shop teacher used to say.

One thing that I'm a _little_ worried about is the change to string handling. The abstraction through the rtVariable mechanism was key to making the compiled shaders portable and cacheable. With this change, the string addresses end up in the generated PTX, and those addresses change from run to run. That's okay for the test suite, but it might not be so good for production shaders.

Replicating the desired behavior with OptiX 7 will take a bit of engineering. Rather than holding up this review, I think that work should be deferred for a follow-up PR. In the meantime, it would be better if we could preserve the old behavior when building against OptiX 6.

lgritz:
@cmstein and @timgrant : Merge, or is there still tinkering going on here?

cmstein:
Hmmmm, I somehow missed Tim's comment.  

I'll try and restore the old behavior with Optix6.

cmstein:
Just pushed an updated PR that's rebased on master and restores the old OptiX6 behavior.

src/include/OSL/sfmath.h
```diff
@@ -94,7 +94,8 @@ namespace sfm
 
         // NOTE: negation happens outside of conditional, then is blended based on the condition
 #if OPENIMAGEIO_VERSION >= 20112
-        Dual2<float> neg_x = OIIO::fast_neg(x);
+        //Dual2<float> neg_x = OIIO::fast_neg(x);
+        Dual2<float> neg_x = fast_neg(x);
```
fpsunflower, line 98: It seems to me like this is fixing a separate bug (code in master is resetting derivs to 0 here, unless I'm missing something).

Maybe worth splitting into its own PR ?

I also think the `ifdef` is not needed since `fast_neg` is defined on duals inside OSL already.

fpsunflower: On second look, this code doesn't appear to be used at all. Maybe its safe to just nuke it?

lgritz: used in code you haven't seen in PRs yet

lgritz: Why did Cliff's PR have any changes at all to sfmath? Some sort of rebase confusion?

src/liboslexec/llvm_util.cpp
```diff
@@ -8,11 +8,16 @@
...
+#ifdef OSL_USE_OPTIX
+#include <optix.h>
+#endif
+
 #include <OSL/oslconfig.h>
 #include <OSL/llvm_util.h>
 
 #if OSL_LLVM_VERSION < 70
 #error "LLVM minimum version required for OSL is 7.0"
+#include "oslexec_pvt.h"
```
lgritz, line 20: What does llvm_util need from oslexec_pvt.h?

In theory, llvm_util is standalone (-ish). ShadingSystem depends on llvmutil, not the other way around.

src/include/OSL/sfmath.h
cmstein: Change has been reverted.  I'm having trouble building locally without it, however, and am not sure why.

lgritz:
So can you summarize where are we on this now?

OptiX 6 works as always.
OptiX 7 ... works except a couple test failures? Or is everything fine now?

src/include/OSL/sfmath.h
lgritz: Let's take that offline, I'll give you a hand, it builds for me.

cmstein:
OptiX6 works as always.  OptiX7 has a failing case in testoptix-noise/scene02.

src/liboslexec/llvm_util.cpp
cmstein: That must have been leftover cruft/debugging-diagnostics.  Just updated the PR with these removed.

timgrant:
My build is out of data and unfortunately I'm away from my main development machine. Can you remind me what the failure with the noise test is?

timgrant:
I got a build up and running, and when I render scene_02 the center sphere that should have a gabor noise texture is black. Is that consistent with the failure you're seeing, Cliff?

If that's the case, it looks an issue with the userdata string param for the noise type. I put a print statement in the generic noise function, and the name is `gabor` as expected. So for whatever reason, the comparison between `name` and `StringParams::gabor` is failing. I guess there are two definitions for the canonical string "gabor"?

cmstein:
Yes, that's the failure I'm seeing.  And yes, I think you're on the right track as to why it's failing.  I'm not sure why it seems there are two definitions of the 'gabor' instead of one being an 'extern' of the other.

PR #1112, opened 2020-02-12
Add shader level annotation to disable range checking

fpsunflower:
This patch adds the metadata annotation `"int range_checking=0"` at the shader level to mean that the shader author is certain that range checks can be safely omitted for this particular shader. This can help speedup particularly expensive shaders without needing to globally disable the checks in other shaders.

### DIFF
--- a/src/liboslexec/oslexec_pvt.h
+++ b/src/liboslexec/oslexec_pvt.h
@@ -342,7 +342,7 @@ inline void stlfree (T &v)
 class ShaderMaster : public RefCnt {
 public:
     typedef OIIO::intrusive_ptr<ShaderMaster> ref;
-    ShaderMaster (ShadingSystemImpl &shadingsys) : m_shadingsys(shadingsys) { }
+    ShaderMaster (ShadingSystemImpl &shadingsys);
     ~ShaderMaster ();
 
     std::string print ();  // Debugging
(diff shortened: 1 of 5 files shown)


src/liboslexec/oslexec_pvt.h
```diff
@@ -342,7 +342,7 @@ inline void stlfree (T &v)
 class ShaderMaster : public RefCnt {
 public:
     typedef OIIO::intrusive_ptr<ShaderMaster> ref;
-    ShaderMaster (ShadingSystemImpl &shadingsys) : m_shadingsys(shadingsys) { }
+    ShaderMaster (ShadingSystemImpl &shadingsys) : m_shadingsys(shadingsys), m_range_checking(true) { }
```
lgritz: Observation: If you do this initialization as `m_range_checking(shadingsys->range_checking())`, then all of the other places you can simply check `rop.inst()->master()->range_checking()` and not have to check both the shader master and the shading system.

The only down side is that this would mean that if you change the global range_checking after the first Shader() call that reads a particular master the first time, it will have no effect. But of course, it's already the case that changing it after a group is compiled has no effect, so in either case users should beware to set the global range_checking (and many other options) up front when the SS is created and before any shader groups are declared, and not think they can change it back and forth at arbitrary times and control individual groups that way.

Up to you whether you think that's already clear and that this would be a nice simplification of the logic when you need to know when range checking is turned on.

fpsunflower: Yeah I think this would be a good cleanup. I'll give it a go.

fpsunflower: Updated with your suggestion.

PR #1113, opened 2020-02-12
Llvm gen for dot product

fpsunflower:
This patch replaces the `osl_dot_*` calls with straight-line IR like we do for other simple ops like add/mul/etc ...

In my tests on a particularly expensive shader, this give a 25% speedup. I can apply the same treatment to some of the other simple vector ops if this looks good.


ee33:
What does GPU perf look like?

fpsunflower:
Good question. I don't have a good way to test this particular shader on the GPU at the moment, but I can only assume that this will generate nicer code for the GPU as well?

lgritz:
On GPU, the implementation of dot is just expansion of 

```
template <class T> IMATH_HOSTDEVICE
inline T
Vec3<T>::dot (const Vec3 &v) const
{
    return x * v.x + y * v.y + z * v.z;
}
```

as found in ImathVec_cuda.h.

Would you have any reason to think that directly generating this IR could be any slower, @ee33  ?

AlexMWells:
When using the LLVM BITCODE for llvm ops, I would expect inlining/optimization to create equivalent performing code.  Of course if not using the BITCODE avoiding a real function call would be profitable.  So when you see a this give a 25% speedup, which configuration is it?
Also are there tests for the derivative versions to ensure the IR implementation matches the C++ version?

fpsunflower:
I am using `USE_LLVM_BITCODE=1`.

We assumed this inlining was taking place for a long time, but vtune definitely reported `osl_dot_fvv` as a non-inlined standalone function call in my particular test setup. Likewise for lots of others I'd like to tackle next, such as `osl_cross_*`, `osl_normalize_*`, `osl_sqrt_*`, etc ...

I think we have unit tests for all these foundational ops? This code does pass our renderer testsuite.

lgritz:
I'm totally ok with this, after all, the choice about which functions to inline versus which to put in llvm_ops.cpp was arbitrary all along. But before doing this to a whole bunch of functions, I wonder if we should first just try tweaking the inlining threshold in the LLVM optimizer?

fpsunflower:
I just tried this (threshold=250 in `llvm_util.cpp:544`) and it seems to have no effect on performance (though it did take 4s instead of 3s to JIT the shader).

lgritz:
OK, must be hitting some other heuristic inside LLVM that is preventing it from inlining fully.

AlexMWells:
Hmm, if inlining is broken, that is an issue.  Is the test shader simple or is this dot inside of a much larger shader.  What is the shading system's attribute "llvm_optimize" set to?  What is the target platform?  Which llvm version are you building against?

fpsunflower:
I'm on LLVM 8.0 at the moment. This was with `llvm_optimize=0`, but I tried 1,2,3 which didn't make any difference here either.

This is a pathological shader network that does a huge number of dot products per invocation. But I have seen the same issue on "regular" sized shaders as well. It just seemed like a good idea to use this stress test to poke at the topic and measure its impact.

Do you see these types of calls get inlined on your end?

fpsunflower:
Update: I _can_ get these calls to inline by simply adding the `always_inline` attribute to `OSL_SHADEOP`. The calls vanish from the vtune trace as expected, but sadly there is no performance gain at all.

I am starting to wonder if the problem is that all these calls are using `void*` as arguments, which surely must be creating extra obstacles to LLVM understanding what is going on ...

ee33:
@lgritz Re GPU, no I would not expect direct LLVM to change the GPU perf.
But it's a good thing to check ...  I wouldn't have expected 25% CPU speedup either :-)

lgritz:
Take the 25% with a big grain of salt, @ee33, it's a specific, somewhat unusual, shader.

lgritz:
So you get a perf increase by directly generating the IR here, but if you use always_inline, you avoid the inlining but nonetheless don't see any perf?

Hmmm... that's interesting. May be worth some investigation into how the IR, and the resulting assembly, differs in those two cases which we would think are the same. Your theory about the `void*` passing is interesting.

AlexMWells:
btw, the commits LGTM, but it brings up the bigger picture of getting precompiled IR to inline effectively, and maybe pass arguments/results to/from precompiled IR differently.   When I was trying out vector IR directly, I had added an OpDescriptor to indicate always_inlined and had the generic function code generator pass by value instead of void * and actually return a result instead of it being the 1st parameter.  As it got inlined before being lowered to x86-64, it worked out.  I eventually abandoned trying to write all the library functions directly in llvm IR, but the idea for inlined library functions still might be useful.

fpsunflower:
So what do you think? Should I look into changing the argument passing conventions and see if that works better? It certainly would be easier to write some of the more complicated cases that involve branches and calls to `sqrt` ...

Is there an easy way to dump the x86 assembly we are jitting to see more directly what LLVM is giving us?

AlexMWells:
In llvm::TargetOptions you can turn on PrintMachineCode (warning it spits out all assembly through each lowering pass, just look at the final output).  IE:
```
diff --git a/src/liboslexec/llvm_util.cpp b/src/liboslexec/llvm_util.cpp
index 85175cb..8296361 100644
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -447,6 +447,44 @@ LLVM_Util::make_jit_execengine (std::string *err)
         (new MemoryManager(m_llvm_jitmm)));
 
     engine_builder.setOptLevel (llvm::CodeGenOpt::Default);
+#if 1
+    const char * oslDumpAsmString = std::getenv("OSL_DUMP_ASM");
+    bool dumpAsm = (oslDumpAsmString != NULL);
+
+    llvm::TargetOptions options;
+    options.AllowFPOpFusion = llvm::FPOpFusion::Fast;
+    options.UnsafeFPMath = true;
+    options.NoInfsFPMath = true;
+    options.NoNaNsFPMath = true;
+    options.HonorSignDependentRoundingFPMathOption = false;
+    options.NoZerosInBSS = false;
+    options.GuaranteedTailCallOpt = false;
+    options.StackAlignmentOverride = 0;
+    options.FunctionSections = true;
+    options.UseInitArray = false;
+    bool use_soft_float_abi = false;
+    options.FloatABIType =
+        use_soft_float_abi ? llvm::FloatABI::Soft : llvm::FloatABI::Hard;
+    options.RelaxELFRelocations = false;
+
+    //options.DebuggerTuning = llvm::DebuggerKind::GDB;
+
+    if (dumpAsm) {
+        options.PrintMachineCode = true;
+    }
+    engine_builder.setTargetOptions(options);
+#endif
```

And to try this out, you will probably need a slightly different interface than 
```
OSL_SHADEOP float
osl_dot_fvv (void *a, void *b)
{
    return VEC(a).dot (VEC(b));
}
```
mabye

```
OSL_SHADEOP OSL_FORCEINLINE float
osl_dot_fvv (const Vec3 a, const Vec3 b)
{
    return a.dot (b);
}
```

However there might be type safety issues going from the LLVM types the OSL runtime has built and the data types created inside the BITCODE like Vec3.  And there will need to be some type of casting, or get both sides to use the exact same type (from LLVM's point of view at least).

fpsunflower:
After some more digging, I was able to get `always_inline` to do the right thing. It currently requires `llvm_optimize=1` or higher, but we can probably figure out the minimal set of extra passes that gives us the faster perf if the llvm optimization passes are too slow.

I am running some of our big production tests through to measure the impact of `llvm_optimize` with the `always_inline` keyword in place.

Since I am able to get the fast perf without changing anything about `llvm_ops.cpp`, I will abandon this patch.

PR #1114, opened 2020-02-14
Improve shader performance with always_inline

fpsunflower:
This patch annotates the ops in llvm_ops.cpp with the `always_inline` attribute. In combination with raising the default `llvm_optimize` setting to 1, this allows these ops to get properly inlined.

In one pathological shader I am seeing gains of 25%. In more typical use cases, there is a 1-2% gain (production benchmarks still running).

Surprisingly, the O1 setting in combination with this `always_inline` flag JITs faster than before (by 2-5% on average). Therefore I've removed the old hand picked set of passes.

This is an alternative proposal to the discarded PR #1113.


src/liboslexec/llvm_ops.cpp
```diff
@@ -148,9 +148,9 @@ void * __dso_handle = 0; // necessary to avoid linkage issues in bitcode
 
 #ifndef OSL_SHADEOP
 #  ifdef __CUDACC__
-#    define OSL_SHADEOP extern "C" __device__ OSL_LLVM_EXPORT
+#    define OSL_SHADEOP extern "C" __device__ OSL_LLVM_EXPORT __attribute__((always_inline))
```
lgritz: How about `OSL_FORCEINLINE` instead? It's already got multi-platform/compiler logic about how to express the attribute.

fpsunflower: `OSL_FORCEINLINE` includes `inline` as well which changes the linking semantics and causes a cascade of other errors.

Since this file is only compiled by `clang`, it seemed safe to rely on their syntax directly.

lgritz: OK, N/M

What happens with inline and __attribue__((always_inline)) are combined? does that mean something different?

fpsunflower: `inline` in C++ has more to do with changing the linking rules. I don't think there is anything special about the combination, its just that in this particular context, only the "stronger" variant is needed.

In regular C++ code you would probably rarely want to use one without the other because the `always_inline` keyword would typically apply to functions whose body is in a header file and you'd want to let the linker know that its ok to see multiple copies of it. The `always_inline` keyword is more of a hint directly to the guts of the optimizer, without any special linker semantics.

lgritz: I see. All good then.

fpsunflower:
So there is an issue with the current Optix code. It appears that (independent of these changes), simply changing `llvm_optimize=1` breaks one of the unit tests. Adding `always_inline` but keeping the old passes seems to be ok.

Somehow the current handpicked passes avoid this issue. @timgrant @ee33 - any ideas?

The pragmatic fix would be to leave the old set of passes for the GPU for now. There is probably a more principled way to fix this though. My guess is LLVM is working ok, but the optimizer is exposing a latent bug somewhere.

fpsunflower:
In case it rings any bells - the one test that fails is `vararray-default.optix`.

timgrant:
The failure is due to a misaligned store after the `float` value is converted to `double` for printing.

The loop in the `print_array_contents` function from test.osl is unrolled when `always_inline` is enabled. The first iteration of the unrolled loop completes successfully (by accident), but there is a `Misaligned address` error when executing the second iteration.

This is the PTX for the second iteration:
```asm
	ld.global.f32 	%f2, [_$newconst0_unnamed_group_1_1_test_small_0_0+4];
	cvt.ftz.f64.f32 	%fd2, %f2;
	ld.global.u64 	%rd9, [ds_1b80eb65a694fb10_000c];
	add.u64 	%rd10, %SP, 24;
	add.u64 	%rd11, %SPL, 24;
	mov.u32 	%r3, 1;
	st.local.u32 	[%rd11], %r3;
	st.f64 	[%SP+28], %fd2; // <-- misaligned store
	{ // callseq 2, 0
	.reg .b32 temp_param_reg;
	.param .b64 param0;
	st.param.b64 	[param0+0], %rd1;
	.param .b64 param1;
	st.param.b64 	[param1+0], %rd9;
	.param .b64 param2;
	st.param.b64 	[param2+0], %rd10;
	call.uni 
	osl_printf, 
	(
	param0, 
	param1, 
	param2
	);
```
The `st.f64 [%SP+28], %fd2` instruction is causing the error, because stores need to be aligned to a multiple of the data size.

I tested this theory by modifying `llvm_gen_printf` to always use `optix_size += sizeof(uint64_t)` when assembling the args for the print call, and `vararray-default.optix` passes. This might be a reasonable patch for printf in the short term, and we'll know what to look for if something similar crops up down the road.

AlexMWells:
Did you try just "inline" vs. "always_inline"?  Just curious. 
Wondering if there are any llvm_ops that one would NOT want to have inlined?  
Perhaps it shouldn't be an all or nothing setting and actually each op in llvm_ops.cpp should "opt-in" to being inlined.

fpsunflower:
(Sorry for the delay, back from vacation)

@AlexMWells I did try `inline` only, and IIRC I ran into trouble because this implies different linkage rules.

Benchmarks I ran with this patch indicate a 2% average rendering speedup and 3% JIT time speedup across a broad set of scenes. LLVM opt levels > 1 are roughly the same rendering speed, but slow down JIT time noticeably.


@timgrant Thanks for the detective work! I will incorporate your patch to the print function.

fpsunflower:
@timgrant your suggestions causes the broken `vararray-default.optix` test to pass, but many other tests to fail.

Did I understand correctly that I just needed to change `optix_size += sizeof(int)` to `optix_size += sizeof(uint64_t)` in two places? Or should I insert an actual conversion to a 64 bit int as well?

fpsunflower:
> Did I understand correctly that I just needed to change `optix_size += sizeof(int)` to `optix_size += sizeof(uint64_t)` in two places? Or should I insert an actual conversion to a 64 bit int as well?

@timgrant I tried inserting a conversion to `long long` without much success either ...

But I am still a bit confused about how this small piece of logic to codegen the args to `osl_printf` is getting impacted by this at all ...

timgrant:
@fpsunflower, I'm a bit baffled myself. Are these new new failures you're seeing?
- hex
- incdec
- intbits
- logic
- struct-array-mixture
- struct-nested-deep

fpsunflower:
Yes, those look like the same failures. It looks like its related to the set of passes that LLVM picks. At `O1`, just the one test fails, at `O0`, a bunch more fail. The old hand-picked list of optimization passes somehow lets everything pass. 

I think the expedient thing to do is to keep the old passes around for now and force their use when generating code for the GPU (with a comment that it should be revisited), but still update the default to get the perf win on the CPU.

I'll post an updated review in a second.

fpsunflower:
Pushed update that passes Optix tests and still gets the perf win for the CPU.

timgrant:
The two-line patch makes the stack slot for each parameter 8 bytes, but that doesn't reflect the format string. When the values are printed out the values aren't at the expected spots in the stack.

Since it's the float->double promotion that's causing the issue with `vararray-default`, fixing the 8-byte loads/stores and leaving the others untouched should hopefully clear up the new errors.

timgrant:
I've got a slightly more involved patch that clears up the new failures I'm seeing. It aligns the alloca for the arg array, and aligns `double` arguments. Apologies for plopping a big diff in a comment.

```diff --git a/src/include/OSL/llvm_util.h b/src/include/OSL/llvm_util.h
index d49e8b5..258544a 100644
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -348,6 +348,12 @@ public:
     }
 
     /// Generate an alloca instruction to allocate space for n copies of the
+    /// given llvm type, and return its pointer.
+    llvm::Value *op_alloca_align (llvm::Type *llvmtype, int n=1,
+                                  const std::string &name=std::string(),
+                                  int align=1);
+
+    /// Generate an alloca instruction to allocate space for n copies of the
     /// given type, and return its pointer.
     llvm::Value *op_alloca (const OIIO::TypeDesc &type, int n=1,
                             const std::string &name=std::string());
diff --git a/src/liboslexec/llvm_gen.cpp b/src/liboslexec/llvm_gen.cpp
index 4264f8a..a6fc09d 100644
--- a/src/liboslexec/llvm_gen.cpp
+++ b/src/liboslexec/llvm_gen.cpp
@@ -383,6 +383,9 @@ LLVMGEN (llvm_gen_printf)
                             // C varargs convention upconverts float->double.
                             loaded = rop.ll.op_float_to_double(loaded);
                             optix_size += sizeof(double);
+
+                            // Ensure that 64-bit values are aligned to 8-byte boundaries
+                            optix_size = (optix_size + sizeof(double) - 1) & ~(sizeof(double)-1);
                         }
                         else if (simpletype.basetype == TypeDesc::INT)
                             optix_size += sizeof(int);
@@ -430,11 +433,15 @@ LLVMGEN (llvm_gen_printf)
         call_args[new_format_slot] = rop.llvm_load_device_string (sym, /*follow*/ true);
 
         size_t nargs = call_args.size() - (new_format_slot+1);
-        llvm::Value *voids = rop.ll.op_alloca (rop.ll.type_char(), optix_size);
+        llvm::Value *voids = rop.ll.op_alloca_align (rop.ll.type_char(), optix_size, /*name*/ "", 8);
         optix_size = 0;
         for (size_t i = 0; i < nargs; ++i) {
-            llvm::Value* memptr = rop.ll.offset_ptr (voids, optix_size);
+            // Ensure that 64-bit values are aligned to 8-byte boundaries
             llvm::Value* arg = call_args[new_format_slot+1+i];
+            if (arg->getType()->isFloatingPointTy())
+                optix_size = (optix_size + sizeof(double) - 1) & ~(sizeof(double)-1);
+
+            llvm::Value* memptr = rop.ll.offset_ptr (voids, optix_size);
             if (arg->getType()->isIntegerTy()) {
                 llvm::Value* iptr = rop.ll.ptr_cast(memptr, rop.ll.type_int_ptr());
                 rop.ll.op_store (arg, iptr);
diff --git a/src/liboslexec/llvm_util.cpp b/src/liboslexec/llvm_util.cpp
index 9e4593b..3ddefe2 100644
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -1030,6 +1030,17 @@ LLVM_Util::op_alloca (llvm::Type *llvmtype, int n, const std::string &name)
 
 
 llvm::Value *
+LLVM_Util::op_alloca_align (llvm::Type *llvmtype, int n, const std::string &name, int align)
+{
+    llvm::ConstantInt* numalloc = (llvm::ConstantInt*)constant(n);
+    llvm::AllocaInst* alloca = builder().CreateAlloca (llvmtype, numalloc, name);
+    alloca->setAlignment (align);
+    return alloca;
+}
+
+
+
+llvm::Value *
 LLVM_Util::op_alloca (const TypeDesc &type, int n, const std::string &name)
 {
     return op_alloca (llvm_type(type.elementtype()), n*type.numelements(), name);
```

fpsunflower:
Nice! Let me try to incorporate this.

fpsunflower:
@timgrant your patch worked great!

I've pulled the old optimization passes back out and incorporated your suggested fix. 

I tweaked things a bit to instead add an `align` argument to the existing `op_alloca` calls (default value does not set any particular alignment like before). I'm guessing we may need this trick in other places in the future ...

timgrant:
Nice, I'm glad that it worked out! I figured there'd be a tweak or two. And I think your hunch is right...

PR #1115, opened 2020-02-18
Disentangle namespace declaration from use to avoid double decl.

lgritz:
For some reason, the namespace declaration was attached to the end of
OSL_NAMESPACE_EXIT. Which would make an error in any source module that
had multiple OSL_NAMESPACE_ENTER/EXIT blocks.


aconty:
LGTM!

PR #1116, opened 2020-02-18
Cleanup of platform.h

lgritz:
* Move OSL_NODISCARD by porting OIIO's OIIO_DISCARD.

* Also port OIIO_FALLTHROUGH.

* Move osl_assume_aligned to the end so it can be inside the usual
  namespace. Also rename it assume_aligned to match C++20 and use a
  `using` so that for C++20 it just aliases to the std one.


lgritz:
@AlexMWells Look ok to you?

Note the renaming of osl_assume_aligned to assume_aligned (so that it can just alias to std::assume_aligned when it's available). Was there a reason it really needed the `osl_` prefix?

AlexMWells:
LGTM.  The osl_ prefix was probably leftover when it was a global function not in the osl namespace yet.  Can always add it back if creates  problem later.

PR #1117, opened 2020-02-19
Minor Dual fixes

lgritz:
* Fix `<<` stream output that broke its pretty formatting in a recent
  overhaul.

* Supply missing Dual/float comparison operators: for example, we had
  Dual<float, but not float<Dual.

* Add equalAll that is like equalVal, but also compares whether the
  derivs are equal (when both items being compared have derivatives).


aconty:
LGTM

AlexMWells:
LGTM

PR #1118, opened 2020-02-22
Move Imath replicated headers into include/OSL/Imathx

lgritz:
* Rename them also, take away the `_cuda` suffix because they contain more than
  just Cuda decorations, restore their original names.

* Copy a bit of updated material from the openexr versions (as of 2.4)
  back into ours, to make a someday reconciliation an easier merge.

* This is all more clearly indicating the semantics of what we intend here:
  these are Imath functions that we had to modify (for whatever reason) but that
  we intend to push back to OpenEXR and then eventually they can disappear
  here once our minimum OpenEXR version requirements are known to include the
  fixes.


lgritz:
@timgrant @fpsunflower ?
This is mainly just renaming headers and doing some diff/reconcile with the newer versions of these headers in OpenEXR.

fpsunflower:
LGTM

PR #1119, opened 2020-02-22
Allow testsuite to run in parallel

lgritz:
The tricky bit was that we have variants of tests: mytest, mytest.opt,
and when using OptiX, also mytest.optix and mytest.optix.opt.

Those previously all ran from the same working directory, which was
fine if run in serial. When they run at the same time, they will
clobber each other's outputs. So the testsuite logic had to be changed
so that each of the variants runs in its own special test directory.

On my workstation at work, this makes running the whole testsuite
(including all the optix tests) go from 5 mins to 30 seconds.  Woo!


lgritz:
OK?

fpsunflower:
LGTM!

PR #1124, opened 2020-03-09
Refactor of sfmath.h and movement of most back into OIIO fmath.h

lgritz:
There was a huge amount of duplication between sfmath.h and OIIO's
fmath.h, with the ones here having only very minor modifications. With
almost no exceptions, I was able to port the fixes here back to
fmath. This now eliminates the redundant copies on the OSL side.

* Remove sfm::ifloor in favor of OIIO::ifloor
* Remove sfm::clamp in favor of improved OIIO::clamp
* Remove sfm::madd in favor of the improved OIIO::madd
* Remove sfm::lerp, bilerp, trilerp in favor of the fixed OIIO fmath version.
* Remove sfm::bit-cast in favor of the improved OIIO fmath version
* Remove sfm::negate in favor of OIIO fast_neg, add Dual fst_neg
* Remove sfm:log functions in favor of improved OIIO fmath.
* Remove unnecessary sfm::absi
* Eliminate most svm transcendentals, use improved OIIO ones instead
* Rearrange safe_fmod
* Remove sfm::absf which did nothing but call std::abs
* Keep sfm::isinf, but only for Intel

The requisite changes should be in both OIIO master ("2.2.2-dev" and later)
and the current release branch 2.1.12 and later. It should also build against
earlier versions of OIIO, but they won't have all the performance additions
for the vectorized loops.


lgritz:
@AlexMWells This all builds and passes tests for me. I hope I didn't break too much on your end, the code you haven't submitted yet will probably need a bunch of changes to redirect back to the fmath versions.

AlexMWells:
Direction is good, just would be nice if there they could be conditionally exist if an older OIIO is there.

lgritz:
That seems like the wrong approach to me. I would rather avoid the duplication and just bump the OIIO requirement to be new enough to have the updated fmath. I've already backported it to last month's 2.1 release. In fact, I *think* it will build properly with even older versions, but only be fully performant in SIMD mode if it's a new enough fmath.

lgritz:
Is this acceptable, @AlexMWells ? The current release OIIO 2.1.12 incorporates all the changes you need, there is no need to be on oiio master if you don't want to be.

AlexMWells:
Yeah, will be fine, can work around locally if need be, there is history to refer to if needed as well.

PR #1125, opened 2020-03-09
Partio finding fixes

lgritz:
A little black magic to ensure that a true cmake config file with
imported targets is preferred over my home grown partio finding module.

Rename FindPartio.cmake -> Findpartio.cmake to match the project naming.

Modify Findpartio.cmake to set up partio::partio target, use it properly.

This all paves the way for an eventual future in which Partio correctly
exports its targets.


aconty:
LGTM

PR #1128, opened 2020-03-12
Allow OSL to compile with LLVM 10.x

fpsunflower:
The release of LLVM 10.x is imminent, there are just a few API changes that impact OSL.

### DIFF
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -364,7 +364,11 @@ BackendLLVM::addCUDAVariable(const std::string& name, int size, int alignment,
 
     OSL_DASSERT (g_var && "Unable to create GlobalVariable");
 
+#if OSL_LLVM_VERSION >= 100
+    g_var->setAlignment  (llvm::MaybeAlign(alignment));
+#else
     g_var->setAlignment  (alignment);
+#endif
     g_var->setLinkage    (llvm::GlobalValue::ExternalLinkage);
     g_var->setVisibility (llvm::GlobalValue::DefaultVisibility);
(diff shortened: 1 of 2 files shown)


aconty:
Funny naming MaybeAlign :)

aconty:
LGTM

src/liboslexec/backendllvm.cpp
```diff
@@ -364,7 +364,11 @@ BackendLLVM::addCUDAVariable(const std::string& name, int size, int alignment,
 
     OSL_DASSERT (g_var && "Unable to create GlobalVariable");
 
+#if OSL_LLVM_VERSION >= 100
+    g_var->setAlignment  (llvm::MaybeAlign(alignment));
```
lgritz: Aside: what does that mean?

fpsunflower: No idea! It seems like an odd name to me too. But this does work and passes renderer tests.

fpsunflower:
Sure! Let me give that a whirl.

fpsunflower:
Actually `LLVM_Util` is a public header. It doesn't feel that great to introduce that there.

lgritz:
OK, whatever you think is best. LGTM.

src/liboslexec/backendllvm.cpp
fpsunflower: From the LLVM source:
```
// - Align represents an alignment in bytes, it is always set and always a valid
// power of two, its minimum value is 1 which means no alignment requirements.
//
// - MaybeAlign is an optional type, it may be undefined or set. When it's set
// you can get the underlying Align type by using the getValue() method.
```

PR #1130, opened 2020-03-21
Avoid compilation error -- can't use always_inline with USE_LLVM_BITCODE=0

lgritz:
This was broken by #1114. 

In a second PR coming later, I will add a testsuite entry where we build with `USE_LLVM_BITCODE=0` just so that this code path cannot be easily broken again.


AlexMWells:
Like the noinline addition.

LGTM

PR #1131, opened 2020-03-21
Fix broken fmod(triple,float) when USE_LLVM_BITCODE=0

lgritz:
It was a missing declaration in buildindecl.h (now accomplished by
the line

    BINARY_VF_OP_IMPL(fmod)

Do pow using the same macros, it had been done longhand.
Also use existing PERCOMP2 and PERCOMP2F shorthand for fmod and pow in
stdosl.h -- this was not related to the bug, but in the course of
investigation I saw that I could simplify.

Beef up fmod testing: Modify testsuite/miscmath to test the
fmod(triple,float) case, and also to choose parameter values that are
better to make sure fmod is working properly.

Change one of the testsuite entries to build with `USE_LLVM_BITCODE=0`
to be sure nothing is broken there.


lgritz:
Please note that this PR also incorporates the changes from #1130, because they were necessary to fully enable the new CI test that verifies that this works for the `USE_LLVM_BITCODE=0` case.

MasterZap:
LGTM (tho I admit to not having actually _tested_ it yet :) )

PR #1132, opened 2020-03-22
Port recent fixes / improvements from OIIO cmake scripts here

lgritz:
* build_foo scripts fail immediately upon error.

* Don't auto-download pybind11 (but do provide a script to do dl/build it)

* Developer snapshot builds from master should have so name include
  the patch number to help avoid linking incompatible shared libs.

* REQUIRED_DEPS and OPTIONAL_DEPS let you more easily turn off dependencies
  by name (beware unintended consequences!)

* Make the build area layout more closely resemble the dist area layout.


lgritz:
Any objections?

PR #1133, opened 2020-03-24
Fix CMAKE Unknown arguments specified Error for Boost

etheory:
When you use ${Boost_VERSION} instead of "${Boost_VERSION}" you get the following error when boost is explicitly undefined:

[cmake] -- [31mNot using Boost -- disabled  [m
[cmake] CMake Error at src/cmake/externalpackages.cmake:171 (if):
[cmake]   if given arguments:
[cmake] 
[cmake]     "CMAKE_SYSTEM_NAME" "MATCHES" "Linux" "AND" "VERSION_GREATER_EQUAL" "105500"
[cmake] 
[cmake]   Unknown arguments specified
[cmake] Call Stack (most recent call first):
[cmake]   CMakeLists.txt:124 (include)

This PR fixes this issue.

## Checklist:

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -169,7 +169,7 @@ endif ()
 
 # On Linux, Boost 1.55 and higher seems to need to link against -lrt
 if (CMAKE_SYSTEM_NAME MATCHES "Linux"
-      AND ${Boost_VERSION} VERSION_GREATER_EQUAL 105500)
+      AND "${Boost_VERSION}" VERSION_GREATER_EQUAL 105500)
     list (APPEND Boost_LIBRARIES "rt")
 endif ()
 


lgritz:
LGTM, but under what conditions would Boost_VERSION be undefined?

etheory:
I think this PR isn't quite as simple as I'd first thought.

I was trying to do a quick build with USE_Boost false and came across this.

But other issues appear down the line like including Boost_INCLUDE_DIRS etc.

So yeah, whilst this PR is technically correct, it does seem rather pointless in hindsight.

lgritz:
Boost is strictly required, I'm afraid.

Utilizing USE_FOO=0 is strictly "buyer beware", and is just not supposed to work properly for a required dependency, only for potentially optional ones. (I should probably modify the check_is_enabled tests to not allow those USE variables to disable any REQUIRED packages.)

I'm going to close this. Please re-open if you think there is a real use for this for cases where Boost is found.

lgritz:
Aside: If the goal is to remove the boost dependency, note that there are only three boost classes we currently use in the OSL code base:

1. boost regex, and that's only needed as long as we support gcc 4.8 / C++11, because std::regex is broken in gcc 4.8.

2. boost thread, used only for thread_specific_pointer, which I suppose we could write a substitute from scratch using thread_local and tricks.

3. boost flat_set and flat_map, which I suppose we could write a substitute from scratch.

(1) will take care of itself in time. (2) and (3) could be done with work, but we have to ask if the work (and potential bugs) are worth the trouble just to eliminate boost.

etheory:
Yeah, thanks @lgritz, totally understood. Out of annoyance with vcpkg on my home windows machine I tried to remove it to see if it would build, not really fully considering the consequences.

Thanks for the full explanation.

PR #1135, opened 2020-03-31
LLVM 10 odds and ends

lgritz:
* Use llvm 10 in some CI tests

* build_llvm: New download location for llvm 10

* Build-time check that LLVM 10+ requires C++ >= 14


fpsunflower:
LGTM

PR #1139, opened 2020-04-03
Pr/wide util

AlexMWells:
## Description

Summary:

This improves the time it takes to generate LLVM IR, optimization, and JIT.  It also adds host cpu features set and instruction set targeting.  LLVM_Util features set is significantly expanded enabling vector data types and SIMD operations and concept of masked execution.  It also adds support for shader source file and line number debug information during code gen.  There are some other features/changes to enable future pull requests that will utilize the wide code gen features.  Also includes refactoring of plumbing for scalar vs. SIMD perlin noise.

Details :

Introduce include/OSL/oslcontainers.h to enable users to customize
the allocator user for dynamic memory allocations.
Update all internal containers used from std::??? OSL::pvt::???.
The OSL::pvt:: namespaced containers use OSL::pvt::CustomAllocator.
Users can define OSL_CUSTOM_ALLOCATOR .  If undefined
CustomAllocator defaults to std::allocator<T>, but users
could modify oslcontainers.h to user their own.

Added build option USE_TBB_SCALABLE_MALLOC (defaults to OFF).
When OSL_TBB_SCALABLE_MALLOC=ON, the CustomAllocator will be
tbb::scalable_allocator<T> and add a dynamic link dependency
to libtbbmalloc.so.  The tbb::scalable_allocator<T> manages
per thread pools of memory so smaller allocations will reuse
previously freed memory from the same thread.  This reuse avoids
multiple threads contending for OS resources and avoids the
overhead of going to the OS for memory requests when possible.
Intent is to reduce overhead incurred in the OSL runtime
compilation/optimization stages as much as possible,
especially when used in a multi-threaded environment.

Separate OSL optimization from JIT by adding optional parameter
	bool do_jit = true
to all optimize_group* methods.  This will allow optimization
without JIT'ing for the host.  It will be necessary if one
will be JIT'ing only for batched execution, we would not want
to pay the cost of JIT'ng a scalar version as well (but still
could if we wanted both scalar and batched versions
simultaneously)

Major functionality improvements added to LLVM_Util:
1.  Add support for generating debug information to track
    source file and line number inside generated LLVM module.
    When enabled, can debug with source and line numbers.
    Also Intel(r) VTune(tm) Profiler associated assembly to
    shader source and line number.
2.  Add autodetection of Host cpu features and match against
    known TargetISA.  Now jit code gen can utilize the full
    featureset of the host.
	LLVM_Util::make_jit_execengine now optionally accepts
        bool debugging_symbols=false,
             // To control if debug info will be generated
        bool profiling_events=false,
             // To control if profiling events will be generated
        OIIO::ustring requested_isa_string = OIIO::ustring(),
             // To request a particular ISA.
             // e.g. "AVX" even though host is AVX512
        bool no_fma = false
             // To disable fused math operations during JIT
    TargetISA LLVM_Util::target_isa() will report on what ISA
    the make_jit_execengine chose.
    In order to chosen if batched execution should be used,
    an application may explicitly test with
        static bool LLVM_Util::supports_isa(TargetISA target);

3.  Expose LLVM_Util::PerThreadInfo (separate from its
    implementation so it can be embedded in OSL::PerThreadInfo
    vs. managed separately.
4.  Improve memory cleanup for JIT'd code by adding
    LLVM_Util::ScopedJitMemoryUser.  JIT'd code needs to
    exist with a longer lifetime than the LLVM_Util object.
    To enable better cleanup at shutdown, the lifetime of all
    JIT'd code is controlled by the the existence of
    ScopedJitMemoryUser objects.  When the last
    ScopedJitMemoryUser goes out of scope or is deleted,
    then the underlying memory managers will be deleted.
    Each ShadyingSystem instance has a ScopedJitMemoryUser.
5.  Development Debugging features to dump struct layouts and
    validate LLVM struct representation based off expected offsets
    populated from host versions of the data structures:
    	void dump_struct_data_layout(llvm::Type *Ty);
    	void validate_struct_data_layout(llvm::Type *Ty,
            const vector<unsigned int> & expected_offset_by_index);
6.  Replaced internalize_module_functions with
        void prune_and_internalize_module (
            unordered_set<llvm::Function*> external_functions,
            Linkage default_linkage = Linkage::Internal,
            std::string *out_err = NULL);
    Identify exactly which functions need to exist in the module based
    on actual usage from a set of external_functions.
    All unneeded functions are removed before we move to optimization
    which is much faster.  The external_functions will be set to have
    external linkage and the remaining functions set to internal linkage.
    GPU's might want to choose LinkOnceODR vs the default Internal.
7.  Improve LLVM optimization passes.
    	void setup_optimization_passes (int optlevel, bool target_host);
    If targetHost is true, passes to target the host will be added.
    Added support for experimental optlevel: 10,11,12,13.
    Optlevels 10, 11, 12, 13 explicitly create optimization passes.
    They are stripped down versions of clang's -O0, -O1, -O2, -O3.
    They try to provide similar results with improved optimization time by
    removing some expensive passes that were repeated many times and
    omitting other passes that are not applicable or not profitable.
    Useful for debugging, optlevel 10 adds next to no additional passes.
8.  Added custom optimization passes to improve AVX/AVX2 code generation
    which are automatically added when targeting the host:
	    PrePromoteLogicalOpsOnBitMasks
  	    PreventBitMasksFromBeingLiveinsToBasicBlocks
9.  Added support for wide data types and vector operations.
    Existing op's work when passed wide data types.
10. Added support for masked operations by managing a stack of masks
    that were created by conditionals or control flow.  Detailed
    explanation of the Masking Strategy in llvm_util.h
11. Added support for logically masked functions
    (handling of masked return from inside OSL functions)
12. Added support for masked loops
    (handling of masked break and or masked continue).
13. Added support for masked exit
    (handling of masked exit from entire shader).
14. Added support for management of when masking is required with a
    ScopedMasking object to allow nesting during code gen.
15. op_store (and other store variations) respect the current masking requirement
    and current top of the mask stack (not just the top entry, but a combined view
    of the entire mask stack).
16. Added support for results of conditional operations to remain boolean.
    Most masking operations are expecting vector of bool values, not a vector of integers.
    e.g. when vector width is 16,  <i1> for scalar or <16 x i1> for wide
    When a conditional result is used by another operation a numerical value
    it is automatically converted to an integer.
17. Added
        llvm::Type *type_native_mask() const;
    because Different ISA's may have different representations of a
    mask from llvm's vector of bits that comparison operations emit.
    And we need varying boolean symbols to have the
    correct data type (size) on the stack and in data structures.
18. Added
        void assume_ptr_is_aligned(llvm::Value *ptr, unsigned alignment);
        llvm::Value *op_gather(llvm::Value *ptr, llvm::Value *index);
        void op_scatter(llvm::Value *wide_val, llvm::Value *ptr, llvm::Value *wide_index);
19. Added helpers to convert from bool<->int, native_mask<->llvm_mask
20. Added vector data lane management helpers to extract and
    insert data for a particular lane as well as test for the
    1st active lane and which lanes match a value.
21. Added math ops:  fabs(v), is_not_finite(v), op_zero_if(cond, v)
22. Added ability to reuse allocas accross a function body.
    Avoid emitting any alloca's inside loops
    and reuse temporaries accross the body of a function
    which means we should not emit them in conditional branches either.
    So always place alloca's at the very beginning of a function.
    To do that we save the current insert point, change it to the begining
    of the function, emit the alloca, then restore the inserting point
    to where it was previously
23. To improve compile time, choose to avoid overhead
    of optimizing functions independently from
    the module level
24. Added environment variable OSL_DUMP_ASM that when set will cause
    the assembly during instruction lowering passes to be dumped
25. Set and document floating point code gen target choices
26. Move minimum LLVM version up to 7.  Support for 6 could be added back
    in with conditional compilation for intrinsics whose name, parameters,
    and behavior changed from 6 to 7.  Alternatively the ops
    who use those intrinsics could be conditionally compiled it if LLVM < 7.

Added op_functioncall_nr to as a noop with the name of the
function that was inlined.  Previously constfolding would
eliminate a op_functioncall if it had no return value.
Now we transmute into a op_functioncall_nr.  There should
be no real shader execution cost to this, but now can debug
through nested OSL shader level function calls.  We needed
the op_functioncall_nr to trigger the creation of debug info
for an inlined function call.

Added Developer only macro OSL_DEV_ONLY(STUFF)
to leave developer debugging/tracing code around and only have
it active if compiled with -DOSL_DEV

Add ShadingSystem attributes bool "llvm_debugging_symbols" and
bool "llvm_profiling_events" to enable/disable llvm code gen
options

Refactored perlin noise scalar vs SIMD code generation.
Previously both scalar and SIMD versions existed in a single
function and a template parameter enabled one version or
the other relying on dead code elimination.  Apparently
when using the scalar version from inside a "#pragma omp simd"
loop the compile cost/complexity from the extra intrinsics
based branch inside the perlin noise made it unusuable.
The simple change to having 2 separate functions chosen
by passing the CodeGenT as function parameter vs. template
parameter creates a cleaner representation that doesn't
mix n match intrinsics and scalar code inside the same
function body.  This improved compile times and allowed
use of the preferred Dual code implementation with generic
lambdas.

For Dual code generation on the Intel(r) C++ compiler,
enable use generic lambda (C++14+) based to loop to repeat code
sequences with compile time known ConstIndex (vs. MACROS).

Avoid potential aliasing issues by avoiding use of Vec3::operator[] which does a reinterpret_cast and instead explicitly use data member names .x, .y, .z

Added     void Opcode::transmute_opname (ustring opname)
    Runtime optimizer may have case to transmute an op to a
    different form.  Only opname is changed.

Separated contents of <OSL/strdecls.h> into private and public
portions as there will be a set of strings that would need
to be exposed as public in order to be accessed by dynamically
loaded shared libraries who wish to utilize the OSL::Strings
namespace.

Added aligned_allocator.h to implement OSL::pvt::aligned_allocator<T>
for use with the OSL runtime's heap allocation to ensure that the start
of the OSL heap is 64 byte aligned.

Added various OSL_MAYBE_UNUSED to make compiler builds happy.
Replaced various iterator types in for loops with the "auto" keyword.

Removed extraneous alignas(64) in sfm_simplex.h that conflicted
with the alignment of the variable type OSL::Block<Vec4,32> whose
natural alignment was 512 bytes.

Removed sfm::fast_rint in favor of OIIO::fast_rint

Removed osl_mul_m_ff and osl_div_m_ff as code gen never calls them,
instead code gen has been constructing a matrix from the
result of a mul_ff and div_ff.

To conform to other shader function names renamed mul and div
functions to encode the return type in their names:
    osl_mul_mm -> osl_mul_mmm
    osl_mul_mf -> osl_mul_mmf
    osl_div_mm -> osl_div_mmm
    osl_div_mf -> osl_div_mmf
    osl_div_fm -> osl_div_mfm

Due to OIIO::cspan constructor for vector not accepting custom allocators had to add make_cpsan helper to various locations.  Could be removed later when OIIO::cspan is fixed.

Inside BackendLLVM::build_llvm_instance fixed bug/optimized copying of data between layers
If connection is to a node not used in the next layer
then it may not have been analyzed properly
and more importantly can be skipped

When -llvm_debug is on, feel it is more useful to get a dump of the entire optimized module
vs. individual layer functions.  Especially now because we have pruned all
unused function declarations and functions out of the the module.
Big benefit is that the module output can be cut and pasted into
https://godbolt.org/ compiler explorer as LLVM IR with a LLC target
and -mcpu= options to see what machine code will be generated by
different LLC versions and cpu targets

Removed cleanup code that deletes individual functions as were were deleting the entire module (which will do that for us).

NullNoise pulled into its own header "null_noise.h" to enable reuse in Wide library functions.

Clarified implementation of some of the opstring operations by separating algorithms into helpers operating on char * and leaving casting to/from ustring in the shader ops.

For use in wide code gen which builds its texture options on the LLVM stack, added shadeops
    int osl_texture_decode_wrapmode(void * name);
    int osl_texture_decode_interpmode(void * name);

Added macro OSL_ONLY_TEXTURE_REQUIRED_CHANNELS which when defined will call the TextureSubsystem with the actual number of channels required vs. the harcoded 4 channels it has been.

In spline implementation separate out helper
    OSL_HOSTDEVICE static int basis_type_of(StringParam basis_name)
from SplineInterp::create so that it can be called by itself to enable code generation/runtime specialized spline implementations.

Moved filter_gabor_kernel_2d from gabornoise.h to gabornoise.cpp as it was not being reused elsewhere.

In testshade added environment variable overrides:
    TESTSHADE_DEBUG_NAN
    TESTSHADE_DEBUG_UNINIT
    TESTSHADE_RANGE_CHECK
    TESTSHADE_NO_NOISE

In testshade added command line options:
    "--llvm_debug", "Turn on LLVM debugging info"
    "--vary_pdxdy", "populate Dx(P) & Dy(P) with varying values (vs. uniform)"
    "--vary_udxdy", "populate Dx(U) & Dy(U) with varying values (vs. uniform)"
    "--vary_vdxdy", "populate Dx(V) & Dy(V) with varying values (vs. uniform)"
    "-llvm_O0", "Do no JIT shader optimization"
    "-llvm_O1", "Do a little JIT shader optimization"
    "-llvm_O2", "Do lots of JIT shader optimization"
    "-llvm_O3", "Do lots more of JIT shader optimization"
    "-llvm_O10", "Experimental:  Do no JIT shader optimization"
    "-llvm_O11", "Experimental:  Do a little JIT shader optimization"
    "-llvm_O12", "Experimental:  Do lots of JIT shader optimization"
    "-llvm_O13", "Experimental:  Do lots more of JIT shader optimization"

In testshade, explicitly set the OIIO attribute "threads" to ensure we don't create more threads in the thread pool than we will be using.
In testshade, switched from using std::bind to c++11 lambda functions.

## Checklist:


lgritz:
There is so much here. It's really overwhelming to review.

I want to try to rebase, and maybe split a few things off into discrete commits. Are your hands off for the moment? I'd like to take a stab at it.

lgritz:
When I build on my Mac (llvm 10), builds fine but when I run, I get lots of this:

Two passes with the same argument (-PreventBitMasksFromBeingLiveinsToBasicBlocks<8>) attempted to be registered!

Ring a bell?

AlexMWells:
Yes my hands are off.

LLVM 10 hasn't been extensively tested, however 
PreventBitMasksFromBeingLiveinsToBasicBlocks<8> is a newly added custom optimization pass.  We register it in llvm_util.cpp:
static llvm::RegisterPass<PreventBitMasksFromBeingLiveinsToBasicBlocks<16>> sRegCustomPass1("PreventBitMasksFromBeingLiveinsToBasicBlocks<16>", "Prevent Bit Masks <16xi1> From Being Liveins To Basic Blocks Pass",
                             false /* Only looks at CFG */,
                             false /* Analysis Pass */);

Based on your error message, it sounds like maybe to copies of llvm_util.o are ending up in your final executable or combination of executable and shared library.  Maybe try a nuke and rebuild.

AlexMWells:
Larry, ignore my previous suggestion.  You must be using a debug build of LLVM that fires asserts.  Looks like there might have a been a behavior change in LLVM, I was leaving the pass::ID's to 0 which 
previously assigned a new value during registration.  Let me look into this.

AlexMWells:
I just reviewed LLVM 10 behavior, it hasn't changed.  So it should be working.  So I will revive my earlier analysis/suggestion that llvm_util.o exists twice in your process and/or that static constructors for 
static llvm::RegisterPass<PreventBitMasksFromBeingLiveinsToBasicBlocks<8>> sRegCustomPass0("PreventBitMasksFromBeingLiveinsToBasicBlocks<8>", "Prevent Bit Masks <8xi1> From Being Liveins To Basic Blocks Pass",
                             false /* Only looks at CFG */,
                             false /* Analysis Pass */);
is being run twice.  The unique key in LLVM's map is the address of the passname:  &passName::ID
so for that assert to fire, the same address must be getting passed twice.

lgritz:
Let's leave this open as a placeholder, but what I'm currently doing is breaking this up into digestible chunks and submitting each one as a separate PR that can be independently reviewed, evaluated, benchmarked, and merged.

lgritz:
Closing, as we are breaking this up into lots of smaller PRs.

PR #1141, opened 2020-04-08
Change Vec3 [] to .xyz

lgritz:
Avoid potential aliasing issues by avoiding use of Vec3::operator[]
which does a reinterpret_cast and instead explicitly use data member
names .x, .y, .z.

This doesn't matter much for scalar code, but when loops are
explicitly or automatically vectorized, using myvec[0] can confuse the
compiler enough to inhibit vectorization, whereas myvec.x does not.

### DIFF
--- a/src/liboslcomp/oslcomp.cpp
+++ b/src/liboslcomp/oslcomp.cpp
@@ -729,8 +729,8 @@ OSLCompilerImpl::write_oso_const_value (const ConstantSymbol *sym) const
             osof("%.9g%s", sym->floatval(i), nelements>1 ? " " : "");
     else if (equivalent (elemtype, TypeDesc::TypeVector))
         for (int i = 0;  i < nelements;  ++i)
-            osof("%.9g %.9g %.9g%s", sym->vecval(i)[0], sym->vecval(i)[1],
-                 sym->vecval(i)[2], nelements>1 ? " " : "");
+            osof("%.9g %.9g %.9g%s", sym->vecval(i).x, sym->vecval(i).y,
+                 sym->vecval(i).z, nelements>1 ? " " : "");
     else {
         OSL_ASSERT (0 && "Don't know how to output this constant type");
     }
--- a/src/liboslexec/closure.cpp
+++ b/src/liboslexec/closure.cpp
@@ -87,13 +87,13 @@ print_component (std::ostream &out, const ClosureComponent *comp, ShadingSystemI
 {
     const ClosureRegistry::ClosureEntry *clentry = ss->find_closure(comp->id);
     OSL_ASSERT(clentry);
-    out << "(" << weight[0]*comp->w[0] << ", " << weight[1]*comp->w[1] << ", " << weight[2]*comp->w[2] << ") * ";
+    out << "(" << weight.x*comp->w.x << ", " << weight.y*comp->w.y << ", " << weight.z*comp->w.z << ") * ";
     out << clentry->name.c_str() << " (";
     for (int i = 0, nparams = clentry->params.size() - 1; i < nparams; ++i) {
         if (i) out << ", ";
         const ClosureParam& param = clentry->params[i];
         if (param.key != 0)
-        	out << "\"" << param.key << "\", ";
+            out << "\"" << param.key << "\", ";
         if (param.type.numelements() > 1) out << "[";
         for (size_t j = 0; j < param.type.numelements(); ++j) {
             if (j) out << ", ";
--- a/src/liboslexec/llvm_ops.cpp
+++ b/src/liboslexec/llvm_ops.cpp
@@ -179,9 +179,9 @@ osl_##name##_vv (void *r_, void *a_)                                \
 {                                                                   \
     Vec3 &r (VEC(r_));                                              \
     Vec3 &a (VEC(a_));                                              \
-    r[0] = floatfunc (a[0]);                                        \
-    r[1] = floatfunc (a[1]);                                        \
-    r[2] = floatfunc (a[2]);                                        \
+    r.x = floatfunc (a.x);                                          \
+    r.y = floatfunc (a.y);                                          \
+    r.z = floatfunc (a.z);                                          \
 }                                                                   \
                                                                     \
 OSL_SHADEOP void                                                    \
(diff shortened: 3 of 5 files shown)


lgritz:
Authored by @AlexMWells

fpsunflower:
LGTM

AlexMWells:
LGTM.  
Just a quick note on unrolling of some of the loops, if a target has issues because it actually inlined big chunks of code like sin/cos or other helper, the loop could be returned if necessary over a local array comps[3] = { val.x, val.y, val.z}

lgritz:
Alex, I'm not sure I understand your last comment. Can you clarify?

AlexMWells:
I was trying to add a backup plan if any issues (code bloat) arise from unrolling the loop because the array subscript is deemed illegal.

Before
```
    for (int i = 0; i < 3; i++)
        OIIO::fast_sincos(VEC(x_)[i], &VEC(s_)[i], &VEC(c_)[i]);

```
After unrolling removing array subscript
```
    OIIO::fast_sincos(VEC(x_).x, &VEC(s_).x, &VEC(c_).x);
    OIIO::fast_sincos(VEC(x_).y, &VEC(s_).y, &VEC(c_).y);
    OIIO::fast_sincos(VEC(x_).z, &VEC(s_).z, &VEC(c_).z);

```
If fast_sincos implementation inlined 3 times generates too much code, causing performance problems, etc.  The array subscript on Vec3 can be avoided by unpacking to a local array then looping over its components so that only 1 copy of the fast_sincos is generated (with some care to prevent loop unrolling)
```
    const auto & x = VEC(x_);
    float x_comps[3] = { x.x, x.y, x.z };
    float s_comps[3];
    float c_comps[3];
    #pragma nounroll
    for (int i = 0; i < 3; i++) {
        OIIO::fast_sincos(x_comps[i], &s_comps[i], &c_comps[i]);
    }
    VEC(s_) = Vec3(s_comps[0], s_comps[1], s_comps[2]);
    VEC(c_) = Vec3(c_comps[0], c_comps[1], c_comps[2]);
```

lgritz:
I see. I'll try to explain that in a comment there.

lgritz:
Alex, correct me if I'm wrong, but this is all because the vector is defined like this (schematically):

    struct vec3 {
        float x, y, z;
        float operator[](int i) { return (&x)[i]; }  //  <--- confusing trick
    };

And so the cast confuses the compiler's understanding of what can alias, etc. So one solution is to use .x, .y, .z directly whenever we can and eschew the use of operator[], then it knows exactly which memory you are talking about.

So can I assume that the following would have been fine all along:

    struct vec3 {
        float vals[3];
        float operator[](int i) { return vals[i]; }
    };

Would that have worked? Because there's no cast and no confusing what is in the memory. Though at the cost of not having the .x .y .z notation.

And if so, what about this implementation:

    struct vec3 {
        union {
            float vals[3];
            struct { float x, y, z; };
        };
        float operator[](int i) { return vals[i]; }
    };

I'm not 100% sure that's even valid syntax without trying to compile it, but would it work for the aliasing problem? I mean, basically, it is always an array, but still gives you the shortcut of being able to say .y when you mean [2], for those freaks who prefer that notation (I've always preferred the array notation myself). Or would the use of union just screw things up even more?

AlexMWells:
This PR is specific to avoiding the array subscript because the current & past implementations of Vec3 have that "trick" you mentioned that aliases Vec3::x,y,z with a float *.  
```
struct vec3 {
    float x, y, z;
    float operator[](int i) { return (&x)[i]; }  //  <--- confusing trick
};
```
Technically that float pointer is only valid to deference up to the size of a float, meaning float[0] is the only legal deference, [1] and [2] are undefined behavior.   This is a correctness issue.
```
struct vec3 {
    float x, y, z;
    float operator[](int i) { return reinterpret_cast<float *>(this)[i]; }  //  <--- safer confusing trick
};
```
I "think" this might be safer, in that this is the size of a Vec3 so compiler should be able to expect dat accesses withing that memory range from the result of the reinterpret_cast.

Now on the performance side, there are other considerations.  For performance our goal is for loop local variables that the data layout of a Vec3 be privatized.  By privatized, I mean the compiler is allowed to alter its data layout how it sees fit.  A simple example would be allowing Vec3::x and Vec3::y and Vec3::z to be represented by individual floats disjoint in memory.  This would allow allow a compiler to keep Vec3::x in a register and update it independently of the other 2 components.  Also once a compiler has privatized a structure and broken it down into built-in types it can optimize much more efficiently.  For SIMD code generation it can keep 4,8,or16 floats representing the Vec3:x for different data lanes side by side contiguous in stack memory and perform wide loads/store to/from registers as needed avoiding the need to perform a strided load/store.  

So what would disqualify a structure from being privatized?  Usually taking the address of it or any of its data members and passing it to a non-inlined function or performance pointer arithmetic locally.  In that case the standard data layout must be used.  And basic code gen for accessing an array element, Vec3::vals[i], would use the base address of (Vec3::&vals[0] + i) which technically uses array based math and would require that vals[0],vals[1], and vals[2] be linear in memory which "could" disqualify them from being privatized. 

The union approach has the same issue, a union may "appear" to solve the correctness issue, but it would clearly disqualify privatization when either data representation of a union is used as the memory layout must be adhered to.  Also reading inactive members of a union is undefined behavior, however most compilers have an extension allowing it.  Nevertheless current guidance from C++ Core Guidelines is to not use unions for type-punning only for storage reduction with intended usage being reading only the last data member that was written to.

In practicality, if going with a Vec3::vals[3] approach and if only compile time constants are used to index it then it a compiler "could" potentially figure it out and treat val[0], val[1], and val[2] as independent and privatize the layout.  However all it takes is a single dynamic index access val[i] to disqualify it.  It would definitely depend on the quality of the compiler.  And as you can see a lot of the existing code uses for(int i=0; i< 3; ++i) ....  Whether that "i" is known at compile time depends on whether a compiler chose to unroll a loop, which can depend on a ton of factors.  So it's very performance fragile.  Also if converting to LLVM IR, it maybe difficult to recover the original fact that the array accesses could be privatized after the fact, at that point it might just all be pointer arithmetic.

By forcing only .x, .y, .z accesses we setup compilers for success in privatization and can vastly improve code generation.  This comes at the cost of not being able to use the array subscript operator.  In practice this means loops get unrolled manually by the programmer, and usually provides much better performance as a result.  However, there maybe cases where we don't want a block of code repeated.  If we hit those, the one work around is to bring the array back, but bring it back in locals.  And possibly a better performing option at least for reading elements would be
```
struct vec3 {
    float x, y, z;
    float operator[](int i) const { 
         float r = x;
         if(i == 1) 
             r = y;
         if(i == 2) 
              r = z;
         return r;}
};
```
Now dead code elimination would handle cases when "i" is known at compile time, and privatization would still be allowed.  If "i" is dynamic, then the conditionals/ternaries/blends/masked assignments would be generated, but again privatization would still be allowed.  But this could be messy if only [i] where used everywhere vs. .x, .y, .z.  So if going this approach perhaps it should be more explicit to indicate it is not the preferred way to access.  
```
    float component(int i, const Vec3 &v) const { 
         float r = v.x;
         if(i == 1) 
             r = v.y;
         if(i == 2) 
              r = v.z;
         return r;
    }
};
```

PR #1142, opened 2020-04-08
Modernize some loops in LPE

lgritz:


lgritz:
Submitted on behalf of @AlexMWells  (though LG modified them beyond what Alex did, changing them into "range for" style loops).

src/liboslexec/automata.cpp
```diff
@@ -240,10 +240,9 @@ NdfAutomata::lambdaClosure(IntSet &states)const
...
         frontier.clear();
         // we do the same as in the above loop but with discovered instead of states
-        for (std::vector<int>::iterator i = discovered.begin(); i != discovered.end(); ++i) {
-            const State *state = m_states[*i];
-            std::pair <IntSet::const_iterator, IntSet::const_iterator> lr;
-            for (lr = state->getLambdaTransitions(); lr.first != lr.second; lr.first++) {
+        for (auto i : discovered) {
+            const State *state = m_states[i];
+            for (auto lr = state->getLambdaTransitions(); lr.first != lr.second; lr.first++) {
                 std::pair<IntSet::iterator, bool> rec = states.insert(*(lr.first));
```
AlexMWells, line 246: while your at it,
`std::pair<IntSet::iterator, bool> rec = states.insert(*(lr.first));`
could be changed to
`auto rec = states.insert(*(lr.first));`

AlexMWells: LGTM

src/liboslexec/accum.cpp
```diff
@@ -130,10 +130,10 @@ AccumAutomata::accum(int state, const Color3 &color, std::vector<AovOutput> &out
 
 Accumulator::Accumulator(const AccumAutomata *accauto):m_accum_automata(accauto)
 {
-    const std::list<AccumRule> &rules = m_accum_automata->getRuleList();
+    const auto &rules = m_accum_automata->getRuleList();
     // Make sure we have as many outputs as the rules need
     int maxouts = 0;
-    for (std::list<AccumRule>::const_iterator i =  rules.begin(); i != rules.end(); ++i)
+    for (auto i = rules.cbegin(); i != rules.cend(); ++i)
         maxouts = i->getOutputIndex() > maxouts ? i->getOutputIndex() : maxouts;
```
aconty, line 137: Range based iteration? Also the second line looks like an std::max

aconty:
With my comment above, LGTM

src/liboslexec/accum.cpp
lgritz: There are LOTS more fixes we could do. I will tackle in a separate PR. See explanation I'm about to write.

src/liboslexec/automata.cpp
lgritz: I have a second PR that fully modernizes. This set is *just* the things you did, which were directly needed to excise certain container names.

lgritz:
So, let me provide some more context for what's going on here.

Part of @AlexMWells big PR, which I'm now splitting up into logical pieces, is a build-time configurable option to use TBB's custom STL allocator that has better thread performance. (Aside: I will isolate and benchmark this so we know how much and where it helps.) Remember that `std::vector<T>` is *actually* `std::vector<T,Alloctator=std::allocator>`. So he defines alternate vector, list, stack, etc.

These were a few places in this code where he changed loops from looking like (I'm making this one up, but this is the gist):

    for (std::vector<int>::iterator i = myvec.begin(); i != myvec.end(); ++i)

to

    for (auto i = myvec.begin() ...)

So instead of changing the reference from `std::vector<int>::iterator` to `OSL::vector<int>::iterator` if he had changed the type of the vector, he just used `auto` to avoid having to fully name the vector type at all. Nice!

But before passing off the PR, I spotted that those same loops can be converted all the way to the "range" style:

    for (auto i : myvec)

Now then, of the many loops in this file, Alex **only** changed the particular loops that named STL types he wanted to give custom allocators, which he had to do to get a clean build after changing the declaration of those data structures. And so, in my modernizing even further, I also **only** changed the ones he already altered, in order to not tangle any possible unnecessary changes/bugs in this set of reviews.

I have another PR, which I want to hold onto until after Alex's work is all merged, in which I more fully take a pass over this whole file and clean it up. I did that first, getting carried away, then realized that was unwise and likely to introduce bugs, backed out, and stuck to changing just the few loops Alex had originally changed.

aconty:
Sounds good, go ahead!

PR #1143, opened 2020-04-08
Add missing oslc and oslinfo to exported targets in config

lgritz:


lgritz:
@lkerley

lgritz:
Lee reports that this works correctly now. Merging.

PR #1144, opened 2020-04-09
Make oslconfig.h be the result of a cmake configure_file step.

lgritz:
This is much like how oslversion.h is generated from oslversion.h.in,
which allows things to be "baked in" from cmake control (version
numbers and namespaces, for example, in the case of oslversion.h).

Now with this patch, oslconfig.h is generated in a similar manner. I
don't actually have any build-time changes as part of this patch, but
some are coming in future planned patches, so I wanted this in place first.


fpsunflower:
LGTM

PR #1145, opened 2020-04-10
Big copyright and license cleanup to follow ASWF conventions

lgritz:
This brings us into conformance with ASWF standards for license text
and how the copyright notices read. (And was cleared by Sony and Linux
Foundation.) This is the same set of changes made to all the other
projects coming into the foundation, including OpenColorIO.

Please note that none of the licenses are changing -- it's the same
3-clause BSD license we have always used. Also, this doesn't change
any of the copyright holders -- it has always been the case that the
authors of each patch retain the copyright to their specific additions
or changes. OSL has never required copyright assignment for
contributions.

* Add a LICENSE-THIRD-PARTY.md file that documents imported software.

* Update documentation license to CC-BY-4.0 (from 3.0).

* Fully reproduce the Qt license in the two files derived from Qt
  example code.

* Change copyright notice to the new wording of 'Contributors to the
  Open Shading Language project', per ASWF convention.

* Replace lengthy per-file copyright boilerplate with SPDX short form.
  (This is more readable and is easily scanned by automated software to
  detect which licenses are used.)

* For any files lacking a copyright notice, license text, or SPDX
  identifier, add them.

* Fix test output for many unit tests where line numbers changed after
  license info was added to .osl files in the testsuite.


lgritz:
No objections were voiced here, nor at the TSC meeting when the topic was raised. Merging.

PR #1148, opened 2020-04-15
Matrix function naming cleanup

lgritz:
Some of the internal matrix ops didn't use the usual naming
convention.  Fix. Also remove at least one that is never used because
it's a combination that oslc will never output.

Submitted on behalf of Alex Wells


lgritz:
Submitted on behalf of @AlexMWells

fpsunflower:
LGTM

AlexMWells:
LGTM

PR #1150, opened 2020-04-15
Fix warnings about unused varables

lgritz:
They aren't used in release builds, and some compilers will warn about
that.

By Alex Wells, with edits by Larry Gritz.


AlexMWells:
LGTM

PR #1152, opened 2020-04-16
Bump LLVM minimum to 7

lgritz:
Four major releases of LLVM (7, 8, 9, and 10) ought to be enough, so let's drop support for llvm 6 and clean up the cruft. That also simplifies some upcoming changes from Alex if it doesn't need to support llvm6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,7 +103,7 @@ matrix:
         os: osx
         compiler: clang
         env: PYTHON_VERSION=3.7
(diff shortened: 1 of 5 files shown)


fpsunflower:
LGTM!

AlexMWells:
LGTM

lgritz:
At today's TSC meeting, we raised this topic and nobody objected. Attendees were unsure if any of them still relied on LLVM 7, and were quite sure they did not require llvm 6. 

Also, I had asked if this was objectionable a few weeks ago on the osl-dev mail list, and nobody spoke up.

So I will merge this tomorrow if no objections are voiced before then.

PR #1157, opened 2020-04-21
Support partio string attributes on particles

lgritz:
We previously did not support indexed strings (only int, float, vector).

Also some SPI adjustments to partio finding, the rest of you can
ignore that part.

### DIFF
--- a/src/liboslexec/pointcloud.cpp
+++ b/src/liboslexec/pointcloud.cpp
@@ -194,6 +194,9 @@ TypeDescOfPartioType (const Partio::ParticleAttribute *ptype)
         if (ptype->count != 3)
             type = TypeDesc::UNKNOWN;  // Must be 3: punt
         break;
+    case Partio::INDEXEDSTR:
+        type = TypeDesc::STRING;
+        break;
     default:
         break;   // Any other future types -- return UNKNOWN
     }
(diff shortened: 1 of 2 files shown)


src/liboslexec/pointcloud.cpp
```diff
@@ -366,8 +369,24 @@ RendererServices::pointcloud_get (ShaderGlobals *sg,
     // then copy them back to the caller's indices.
 
     // Actual data query
-    const_cast<Partio::ParticlesData *>(cloud)->data (*attr, count, (Partio::ParticleIndex *)indices,
-                 true, out_data);
+    if (partio_type == OIIO::TypeString) {
+        // strings are special cases because they are stored as int index
+        std::unique_ptr<int[]> strindices(new int[count]);
```
fpsunflower, line 374: We have to malloc storage for all indices on every lookup?

lgritz: Well, let me take another pass at that. I guess we can do it only if there are more than one, which will be unusual. 

I really need to start using a good short vector implementation.

lgritz: Pushed an update

fpsunflower: Are you sure? I think in this use case there could be many. `count` is proportional to the number of points you found within your radius, right ?

Can we use `alloca` instead? (on the assumption that `count` is a reasonable value?)

lgritz: Yes, you are correct. I was misinterpreting as the number of attributes, rather than the number of points. My bad.
Will fix...

src/liboslexec/pointcloud.cpp
```diff
@@ -366,8 +369,38 @@ RendererServices::pointcloud_get (ShaderGlobals *sg,
...
+            if (ind >= 0 && ind < sicount)
+                ((ustring *)out_data)[i] = ustring(strings[ind]);
+            else
+                ((ustring *)out_data)[i] = ustring();
+        }
+    } else {
+        // All cases aside from strings are simple.
+        const_cast<Partio::ParticlesData*>(cloud)->data (*attr, count,
+                    (const Partio::ParticleIndex *)indices, true, out_data);
+        // FIXME: it is regrettable that we need this const_cast (and the
```
fpsunflower, line 399: I believe this is fixed already but we are on an old version?

src/liboslexec/pointcloud.cpp
lgritz: Clearly I did not have enough coffee today. Update pushed.

src/liboslexec/pointcloud.cpp
lgritz: Maybe? We are on a... customized version. We need to submit our changes back to the project and then rebase.

lgritz: But in fact, this is not fixed in its current master: https://github.com/wdas/partio/blob/master/src/lib/Partio.h#L115

fpsunflower: Fair enough!

lgritz: All data() does it call dataInternalMultiple(), which is const. But data() is not marked as const. Oh well.

PR #1158, opened 2020-04-23
Aligned allocator for the shader execution heap

lgritz:
The runtime "heap" is modified to use an aligned allocator to be sure that
allocations on it are cache line aligned.

For sufficiently new versions of boost, we can use
boost::align::aligned_allocator. When we don't have a new enough
boost, we us a new aligned_allocator.h provided by Intel (which we can
delete once the minimum required Boost is high enough).

Ensuring the alignment here is helpful for other upcoming changes that
support the SIMD batched shading.

### DIFF
--- a/src/liboslexec/aligned_allocator.h
+++ b/src/liboslexec/aligned_allocator.h
@@ -1,0 +1,201 @@
+// Copyright Contributors to the Open Shading Language project.
+// SPDX-License-Identifier: BSD-3-Clause
+// https://github.com/imageworks/OpenShadingLanguage
+
+/* NOTE:  Can be replaced with boost 1.56 boost::alligned_allocator
+ * Current boost dependency is only boost 1.55, so consider
+ * removing this when our boost minimum becomes >= 1.56.
+ * No expected performance difference, just one less file
+ * NOTE: pointer_or_number and is_aligned appear useful and should be
+ * pulled into their own files
+ */
+#pragma once
+
+#include <limits>
+#include <type_traits>
+#include <utility>
+
+#include <OSL/oslversion.h>
+
+OSL_NAMESPACE_ENTER
+
+namespace pvt {
+
+// For converting pointers to numbers for alignment or other math operations
+template<typename T = void>
+union pointer_or_number {
+    pointer_or_number() {}
+    pointer_or_number(const T * pointer_)
+    :pointer(pointer_)
+    {}
+    pointer_or_number(size_t number_)
+    :number(number_)
+    {}
+
+    const T * pointer;
+    size_t number;
+};
+
+
+
+template <size_t ByteAlignmentT, typename T>
+inline bool is_aligned(const T *a_pointer)
+{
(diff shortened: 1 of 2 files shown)


src/liboslexec/oslexec_pvt.h
```diff
@@ -32,6 +33,13 @@
 # include <regex>
 #endif
 
+#if BOOST_VERSION >= 105600
```
fpsunflower, line 36: Since you went to the trouble of writing your own - why bring boost into this at all ?

src/liboslexec/aligned_allocator.h
```diff
@@ -0,0 +1,201 @@
...
+template <class ObjT>
+typename aligned_allocator<ObjT>::const_pointer
+aligned_allocator<ObjT>::address( const_reference a_ref) const
+{
+    return &a_ref;
+}
+
+template <class ObjT>
+typename aligned_allocator<ObjT>::pointer
+aligned_allocator<ObjT>::allocate(size_type a_count, std::allocator<void>::const_pointer /*a_hint*/)
```
fpsunflower, line 111: On *nix platforms you can just call `posix_memalign` which works with plain old `free`.

On windows you can do `_aligned_malloc` and `_aligned_free`.

src/liboslexec/oslexec_pvt.h
```diff
@@ -1750,7 +1759,8 @@ class OSLEXECPUBLIC ShadingContext {
     PerThreadInfo *m_threadinfo;        ///< Ptr to our thread's info
     mutable TextureSystem::Perthread *m_texture_thread_info; ///< Ptr to texture thread info
     ShaderGroup *m_group;               ///< Ptr to shader group
-    std::vector<char> m_heap;           ///< Heap memory
+    /// Heap memory
+    std::vector<char, pvt::aligned_allocator<char,OIIO_CACHE_LINE_SIZE>> m_heap;
```
fpsunflower, line 1763: Are you sure this needs to be an `std::vector` ? Just having a plain `unique_ptr` would get rid of all the allocator boilerplate.

`std::vector` in this case has extra costs, like 0-init and copying on resize (which AFAICT is not important to this use case)

src/liboslexec/oslexec_pvt.h
lgritz: Well, I figured it would let us jettison the custom code once our boost floor raises.

src/liboslexec/oslexec_pvt.h
lgritz: As a practical matter, I don't know that those costs hurt -- this tends to get resized a couple of times until you've shaded the biggest material, then it just sits there.

But, hmmm, you raise an excellent point. Maybe I'll just rewrite it to use unique_ptr and a simple aligned malloc. Let me see if that looks cleaner.

lgritz:
Chris is correct, this is the wrong approach and overly complicates things. Withdrawing the PR, a new and simpler one will follow.

PR #1159, opened 2020-04-23
Make sure the shader execution heap is cache aligned

lgritz:


lgritz:
@AlexMWells

AlexMWells:
if (shadingsys().debug())
            infof("  ShadingContext %p growing heap to %d",
                  static_cast<void*>(this), heap_size_needed);

Do you care/mind if you no longer receive this debug info about heap growing?  Could add it back to the new helper function.  I have no need, just noticed it won't be there anymore.

lgritz:
No, I thought about that when I deleted it. I don't think it's served any real purpose, so I was ok letting it go.

AlexMWells:
LGTM

fpsunflower:
LGTM2! I like this way better!

PR #1160, opened 2020-04-24
CI on GitHub Actions improvements

lgritz:
* Fix newly broken Mac CI -- GHA's default homebrew now has llvm 10, which
  requires C++14.

* Schedule a nightly cron job for CI.

* Do not launch full builds and tests for pushes that only change
  documentation (md, rst, and tex files).


fpsunflower:
LGTM

PR #1164, opened 2020-05-06
Add cube root function to the language

fpsunflower:
Language was missing cbrt(x) which is part of the C99 math library.

Also added a special case to allow pow(x,1.0/3.0) to be automatically turned into cbrt during runtime optimziation.


fpsunflower:
The CI seems to be having trouble checking out my branch.

Did I do something wrong on my end?

lgritz:
Temp github glitch? I'll give it a while and then re-kick the CI tests.

fpsunflower:
Added a few lines to the testsuite.

Getting a different set of errors in CI now, though again it seems unrelated to this patch.

lgritz:
The CI errors are surely not related to your changes. I'll look into that separately. Could be a temporary problem on github's end, or something temporarily down at llvm.org (the one I looked at had trouble with an llvm download).

lgritz:
Oh, I see, some are having an error in testsuite/function-redef.

That's because of your change! Your additional line in stdosl.h changed the line number reported in an error message.

lgritz:
Just run the testsuite on your end, it should be easily to update that one ref output.

fpsunflower:
I think its all good now. The two CI failures seem unrelated to me (unless I'm missing something else).

lgritz:
I think those are unrelated, I'm looking at them separately. Merging.

lgritz:
As an aside, it's probably worth starting to use `git commit -s` to attach your DCO sign-off. Once the repo finally moves, PRs won't be able to merge without them. May as well establish the habit while there's stll not yet any penalty for forgetting.

fpsunflower:
Ah I was wondering what that was about! :)

PR #1166, opened 2020-05-07
Fix cube root derivative sign

fpsunflower:
I had noticed the comment didn't match the code and updated the code not realizing it was the comment that was wrong ...

### DIFF
--- a/src/include/OSL/dual.h
+++ b/src/include/OSL/dual.h
@@ -1237,13 +1237,13 @@ OSL_HOSTDEVICE OSL_FORCEINLINE Dual<T,P> inversesqrt (const Dual<T,P> &a)
     return result;
 }
 
-// f(x) = cbrt(x), f'(x) = -1/(3*x^(2/3))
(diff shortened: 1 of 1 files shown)


lgritz:
oops

fpsunflower:
This kind of brings up a larger topic for me (outside the scope of this review of course).

Testing math functions with printf like in the `miscmath` test is probably too limited. I think ideally we would have a testsuite entry per function that makes some kind of image. Possibly even outputs 3 images: value, Dx, Dy (not sure if our tetsuite allows multiple output images for a single test).

This would test functions on a broader range of arguments and exercise the dual versions.

AlexMWells:
@fpsunflower , for SIMD tests I have bunch of shell scripts that produce images and idiff them instead of using printf's.  printf's really didn't work with different precision results.  It would be great to see the normal test suite start having tests like that.  Also added some -vary_pdx, -vary-pdy,  etc. flags to test shade so that some different values flowed through the math code.  The trick is getting a meaningful image out, testshade might need an additional scaling step to bring results back into a visible range.

PR #1172, opened 2020-05-22
LLVM_Util::prune_and_internalize_module

lgritz:
Previously, we had `LLVM_Util::internalize_module_functions` that
reduced the LLVM optimize + JIT work by finding function symbols in an
IR module that started with "osl_" and were not our known entry points
and marking them with "internal" visibility. This lets LLVM recognize
them, if they are not called internally to the module, as dead code
and discard them before spending any time optimizing or jitting those
functions.

This patch from Alex Wells adds an additional method,
`prune_and_internalize_module`, that is even more aggressive, pruning
all symbols that we don't think can be reached with our generated code
(are not "materialized", in LLVM parlance), and then marking
everything that is not a known entry point as "internal".

N.B. Some care is needed not to mark as internal a few symbols that
Cuda adds and which if removed will cause the OptiX shader to
fail. These seem to all contain the substring "rti_internal_".

This is indeed slightly better, and in extensive benchmarks of
production shaders, we're seeing about a 10% average reduction in LLVM
opt+jit time.  This is a tiny portion of a fully rendered scene, but
for interactive renders where you care about "time to first pixel", it
is more significant to reduce this compilation overhead as much as
possible.

I (LG) am a little hesitant to deprecate the old method right away, in
case any as-yet undiscovered flaw is found or just to aid
benchmarking. So I added a ShadingSystem string attribute
"llvm_prune_ir_strategy" that can take on any of the values "prune"
(default, the new way, full prune and aggressively internalize),
"internalize" (old way, just internalize function symbols that aren't
entry points), and "none" (don't do either -- just for benchmarking or
debugging).

I also benchmarked the "none" strategy, and it spends an average of over
2x longer in LLVM opt+jit than our old "internalize." So either strategy
helps substantially cut down on opt+jit time, though "prune" seems the
fastest, and is thus the new default. As we gain more confidence in it
in production, we may eventually deprecate and remove the "internalize"
strategy, leaving only "prune."

### DIFF
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -9,6 +9,7 @@
 #include <OSL/oslconfig.h>
 
 #include <vector>
+#include <unordered_set>
 
 #ifdef LLVM_NAMESPACE
 namespace llvm = LLVM_NAMESPACE;
@@ -72,7 +73,10 @@ public:
     }
 
     /// Set the current module to m.
-    void module (llvm::Module *m) { m_llvm_module = m; }
+    void module (llvm::Module *m) {
+        m_llvm_module = m;
+        m_ModuleIsPruned = false;
+    }
 
     /// Create a new empty module.
     llvm::Module *new_module (const char *id = "default");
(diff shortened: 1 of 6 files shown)


lgritz:
TL;DR: speeds up LLVM optimize+JIT of large shader networks by around 10% on production scenes.

AlexMWells:
LGTM

PR #1173, opened 2020-05-23
Make debug_nan and debug_uninit work more accurately for matrices

lgritz:
Both were plagued by checking the entirety of the matrix, when in fact
only one element needs to be checked. This is wasteful because it
checks elements that aren't read or written by the op we are checking,
and it can also lead to "false positives", errors about elements that
are not touched by the present op.

Refinement of the debug_nan and debug_uninit cases was by Alex Wells.
LG added comments to clarify the implementation of these functions.

LG also amended the testsuite/debug-uninit to specifically test
arrays, triples, and matrices (checking for true positives and check
against false positives). In the process, he noticed another bug (!)
where float-based arrays were not being correctly initialized with
their uninitialized value sentinels, and thus were not detecting use
of uninitialized values therein, because instead of calling
is_float_based, we called is_floatbased, which excludes arrays.
(The similar names are yet ANOTHER bug, or at least unintended to be
so confusing, and this will be addressed in a separate PR).

### DIFF
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -110,8 +110,10 @@ static ustring op_end("end");
 static ustring op_nop("nop");
 static ustring op_aassign("aassign");
 static ustring op_compassign("compassign");
+static ustring op_mxcompassign("mxcompassign");
 static ustring op_aref("aref");
 static ustring op_compref("compref");
+static ustring op_mxcompref("mxcompref");
 static ustring op_useparam("useparam");
 
 
@@ -388,7 +390,7 @@ BackendLLVM::llvm_assign_initial_value (const Symbol& sym, bool force)
         if (sym.typespec().is_closure_based()) {
             // skip closures
         }
-        else if (sym.typespec().is_floatbased())
+        else if (sym.typespec().is_float_based())
             u = ll.constant (std::numeric_limits<float>::quiet_NaN());
         else if (sym.typespec().is_int_based())
             u = ll.constant (std::numeric_limits<int>::min());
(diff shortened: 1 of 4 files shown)


src/liboslexec/llvm_instance.cpp
```diff
@@ -592,32 +623,60 @@ BackendLLVM::llvm_generate_debugnan (const Opcode &op)
...
+        // There are a few special cases where an op reads a partial value:
+        // one element of an array, aggregate (like a point), or matrix. In
+        // those cases, don't check the other pieces that haven't been
+        // touched by this op, because (a) it's unnecessary work and code
+        // generation, and (b) they might generate a false positive error.
+        // An example would be:
+        //     float A[3];     // 1   Elements are uninitialized
+        //     A[1] = 1;       // 2
+        //     float x = A[1]; // 3   <--- this is where we are
+        // Line 3 only reads element [1]. It ss NOT an uninitialized read,
```
mprater, line 674: "It ss NOT" -> "It is NOT"

fpsunflower:
LGTM!

AlexMWells:
LGTM

lgritz:
Typo fixed, thanks for the look-over

lgritz:
Aside: Years ago we started adding "LGTM" as an ordinary comment. But nowadays there is a real "review changes" UI under the "files changed" tab (look in the upper right), where you can click an "approved" button. I think that's how we should do it. (You can still write "LGTM" in those comments if you wish.) If we ever enable to the restriction that merges require an "approval" (as many projects have), that's what it would be looking at, so I think it's a good habit to make.

PR #1174, opened 2020-05-24
Only range check matrix component access when necessary

lgritz:
Point/color and array accesses were range checked at runtime EXCEPT
when the indices were constant and known to be within legal range.

But matrix component access (m[r][c] = x, x = m[r][c]) were always
range checked.

This patch from Alex Wells (and then enhanced by LG) skips the range
checking on the row or column indices, or both, when they are constant
and >= 0 and < 4, which is probably a lot of the time, thus speeding
up matrix component access a little by skipping the range checking
when it's safe to do so.


AlexMWells:
LGTM

PR #1175, opened 2020-05-24
Add several wide data types and wide operations to LLVM_Util.

lgritz:
These aren't used anywhere in OSL's codebase yet. But they will be.

By Alex Wells.


AlexMWells:
LGTM

PR #1176, opened 2020-05-25
Fix confusing TypeSpec::is_float_based vs is_floatbased

lgritz:
I'm not sure how this happened, but TypeSpec had `is_float_based` and
`is_floatbased` methods, and they were subtly different (the latter is
false for float-based arrays). This has led to at least one bug.

A careful examination reveals that the vast majority of uses of
`is_floatbased` really intended `is_float_based`, so I changed those.

For the remainder, I renamed `is_floatbased` to `is_float_based_notarray`
for clarity.

### DIFF
--- a/src/liboslcomp/codegen.cpp
+++ b/src/liboslcomp/codegen.cpp
@@ -295,7 +295,7 @@ ASTNode::coerce (Symbol *sym, const TypeSpec &type, bool acceptfloat)
     }
 
     if (sym->symtype() == SymTypeConst && sym->typespec().is_int() &&
-            type.is_floatbased()) {
+            type.is_float_based() && !type.is_array()) {
         // It's not only the wrong type, it's a constant of the wrong
         // type. We need a new constant of the right type.
         ConstantSymbol *constsym = (ConstantSymbol *) sym;
@@ -1539,7 +1539,7 @@ ASTbinary_expression::codegen (Symbol *dest)
 
     // Promote ints to float-like types, for mixed arithmetic
     if ((m_op == Mul || m_op == Div || m_op == Add || m_op == Sub)) {
-        if (lsym->typespec().is_floatbased() && rsym->typespec().is_int()) {
+        if (lsym->typespec().is_float_based() && rsym->typespec().is_int()) {
             if (rsym->symtype() == SymTypeConst) {
                 float val = ((ConstantSymbol *)rsym)->floatval();
                 rsym = m_compiler->make_constant (val);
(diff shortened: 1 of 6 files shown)


src/liboslcomp/codegen.cpp
```diff
@@ -295,7 +295,7 @@ ASTNode::coerce (Symbol *sym, const TypeSpec &type, bool acceptfloat)
     }
 
     if (sym->symtype() == SymTypeConst && sym->typespec().is_int() &&
-            type.is_floatbased()) {
+            type.is_float_based() && !type.is_array()) {
```
fpsunflower, line 298: Did you mean to use `is_float_based_not_array()` here?

src/liboslexec/backendllvm.cpp
```diff
@@ -645,7 +645,7 @@ BackendLLVM::llvm_load_value (llvm::Value *ptr, const TypeSpec &type,
         return result;
 
     // Handle int<->float type casting
-    if (type.is_floatbased() && cast == TypeDesc::TypeInt)
+    if (type.is_float_based() && !type.is_array() && cast == TypeDesc::TypeInt)
```
fpsunflower, line 648: And here?

src/liboslcomp/codegen.cpp
lgritz: Nothing subtle. There were just two ways to express this and I chose based on aesthetics (and forgetfulness).

I think when I visited this file, I was toying with seeing whether I could get rid of is_floatbased entirely. But there were some others files I visted later where the complexity of the expression made it more awkward to add another clause, so I chose the "principle of least change" and just changed the one call to is_floatbased to is_float_based_notarray.

I can be more consistent if you think it's better -- either always using is_float_based_notarray whenever that's applicable (never letting it be is_float_based && !is_array), or else always doing the latter and eliminating is_float_based_notarray entirely.

lgritz: For this particular case, as with many others, I think I realized that it could never be called for the array case, so I just changed is_floatbased to is_float_based.

AlexMWells:
What is the scenario where
            (type.is_float_based() && !type.is_array())
would not suffice and you need
            type.is_float_based_notarray()
?
Just wondering if less is more, and as a developer I would just kinda of expect float[] to be true for is_float_based.

lgritz:
I suppose I can get rid of is_floatbased entirely, and test the arrayness explicitly when that is needed.

fpsunflower:
+1 to that idea, I think the non-array case is rare enough that the more explicit checking is clearer.

lgritz:
OK, the _notarray (only used in 3 places) is removed and just added `&& !type.is_array()` where needed.

AlexMWells:
LGTM

PR #1183, opened 2020-05-27
Added various OSL_MAYBE_UNUSED to make compiler builds happy.

lgritz:
From Alex Wells

### DIFF
--- a/src/liboslexec/constfold.cpp
+++ b/src/liboslexec/constfold.cpp
@@ -801,7 +801,7 @@ DECLFOLDER(constfold_aref)
 DECLFOLDER(constfold_arraylength)
 {
     Opcode &op (rop.inst()->ops()[opnum]);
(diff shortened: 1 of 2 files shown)


fpsunflower:
Why is this needed? Is it because some asserts go away entirely in some compilation modes?

AlexMWells:
Some compilers (don't ask my which) were complaining of unused variables.  And yes they were unused because OSL_DASSERT(STUFF) doesn't emit stuff when in profile/release builds

lgritz:
I've seen that problem, too (maybe on the OIIO side for analogous reasons) when experimenting with even more pedantic warning levels.

fpsunflower:
Defining `DASSERT(expr)` as `(void) sizeof(expr)` was an attempt at "fixing" this -- but I guess some compilers are smart enough to notice that isn't a real use of the expression.

LGTM

PR #1184, opened 2020-05-27
Make the strings of <OSL/strdecls.h> into public symbols.

lgritz:
Make the strings of <OSL/strdecls.h> into public symbols.

It's not officially part of the public API, but making the symbols
public is necessary for some upcoming work where Alex has separate DLLs
for different CPU/SIMD implementation modules and the symbols will need
to be visible to the contents of those DLLs.


lgritz:
Question for Alex: is it worth separating these strings into "visible" and "not visible"? Since they are behind namespaces anyway, is it simpler to just make all of these strings externally visible, and have a less complicated setup?

AlexMWells:
Hmm, if you make them all public, then that would work.  Typically one might view "public" symbols as being part of the API, and be versioned, have guarantees of existing in the next version.  This case is a bit different in that the shared libraries are an implementation detail that we need some public symbols to make happen.  I think I was striving the having the smallest public symbol space exposed, but that may not be a worthy goal.

lgritz:
Yeah, I think I like that simplification. Stand by for update.

lgritz:
Updated with a simpler implementation that just makes them all public, with a macro structure that is less likely to confuse the next person in this part of the code.

The new description is:

    Make the strings of <OSL/strdecls.h> into public symbols.

    It's not officially part of the public API, but making the symbols
    public is necessary for some upcoming work where Alex has separate DLLs
    for different CPU/SIMD implementation modules and the symbols will need
    to be visible to the contents of those DLLs.

lgritz:
I'm not crazy about making them all public, but that seems like a better alternative than having the mix, and forever having little patches where, oh, here's one more string that needs to be public. Oh, here's another, etc.

AlexMWells:
Can always revisit in future if a good enough reason arises, they were just all NOT public before.

PR #1185, opened 2020-05-28
Fix normalize() sqrt call to be sure it's using floats

lgritz:
By Alex Wells

### DIFF
--- a/src/liboslexec/llvm_ops.cpp
+++ b/src/liboslexec/llvm_ops.cpp
@@ -750,17 +750,16 @@ osl_distance_dfvdv (void *result, void *a, void *b)
 OSL_SHADEOP void
 osl_normalize_vv (void *result, void *a)
 {
-    using std::sqrt;
     // NOTE: must match with the Dual version of normalize used below
     Vec3 v = VEC(a);
-    float len = sqrt(v.x * v.x + v.y * v.y + v.z * v.z);
-    if (len > 0) {
-        float invlen = 1 / len;
+    float len = sqrtf(v.x * v.x + v.y * v.y + v.z * v.z);
+    if (OSL_LIKELY(len > 0.0f)) {
+        float invlen = 1.0f / len;
         v.x *= invlen;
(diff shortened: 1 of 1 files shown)


src/liboslexec/llvm_ops.cpp
```diff
@@ -753,14 +753,14 @@ osl_normalize_vv (void *result, void *a)
     using std::sqrt;
```
fpsunflower, line 753: The `using std::sqrt` line can be removed then?

lgritz: Well what do you know? Somehow we both missed this, I guess.
I'm tempted to abandon this. Alex, do you have a reason why we should keep these changes?

AlexMWells:
either way then, std::sqrt has overload for float.  So take your pick sqrtf or std::sqrt.

AlexMWells:
Looking elsewhere in the file sqrtf was used vs. std::sqrt.  The additional benefit of using sqrtf is the possibility of a compiler warning about loss of precision should a double be passed to it vs. std::sqrt which will happily just call the overloaded double version.

lgritz:
OK, I'll just remove the original "using".

lgritz:
The Dual version has OSL_LIKELY to explain that the `len > 0.0f` is the likely path. Should we do that here as well?

lgritz:
I'm torn. `sqrtf` means that we'll never have the problem of inadvertently converting to double. But std::sqrt means that if anybody wants to build OSL using doubles (changing the definition of Vec3), it will work properly without needing to modify this function.

lgritz:
std::sqrt I think is my preference. If it chose the "wrong" sqrtf (double), it will still make a warning when it assigns that back to a float result.

AlexMWells:
If you really want to respect Vec3::value_type then alot more changes are needed.  For this, std::sqrt would be fine, just pointing out that you might want to change all of them one way or the other.

About the OSL_LIKELY sure, it could be used there.

PR #1188, opened 2020-05-30
Change all bare std::ofstream/ifstream to OIIO::ofstream/ifstream

### LINKED ISSUE #1187: Can't build. The version of OpenImageIO is 2.1.12.0-1, build system is mingw64.
OpenShadingLanguage\src\liboslexec\shadingsys.cpp:3267:5: error: no matching function for call to 'open'
    OIIO::Filesystem::open(groupfile, groupfilename);
    ^~~~~~~~~~~~~~~~~~~~~~
D:\Develop\Library\msys2\mingw64\include\OpenImageIO/filesystem.h:204:15: note: candidate function not viable: no known conversion from 'std::ofstream' (aka 'basic_ofstream<char>') to 'OIIO::ifstream &' (aka 'basic_ifstream<char> &') for 1st argument
OIIO_API void open (OIIO::ifstream &stream, string_view path,

lgritz:
Because they are UTF-8 safe and work properly on MinGW.

Simplified a bit more as I went -- some reading idioms could be
replaced by a single call to Filesystem::read_text_file, some close()
calls were redundant because streams close upon destruction / scope
exit, some open flags unnecessary because the default write mode
already will truncate an existing file.

Closes #1187 

### DIFF
--- a/src/oslc/oslcmain.cpp
+++ b/src/oslc/oslcmain.cpp
@@ -188,12 +188,11 @@ main (int argc, const char *argv[])
             ok = compiler.compile_buffer (sourcecode, osobuffer, args, "",
                                           shader_path);
         if (ok) {
-            std::ofstream file;
+            OIIO::ofstream file;
             OIIO::Filesystem::open (file, compiler.output_filename());
-            if (file.good()) {
+            if (file)
                 file << osobuffer;
-                file.close ();
-            }
+            ok = file.good();
         }
     } else {
         // Ordinary compile from file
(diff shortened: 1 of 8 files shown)


src/oslc/oslcmain.cpp
```diff
@@ -188,12 +188,12 @@ main (int argc, const char *argv[])
...
         if (ok) {
-            std::ofstream file;
+            OIIO::ofstream file;
             OIIO::Filesystem::open (file, compiler.output_filename());
-            if (file.good()) {
+            if (file)
                 file << osobuffer;
-                file.close ();
-            }
+            else
```
fpsunflower, line 195: `ok = file.good()` to catch write errors in addition to open errors?

lgritz: Yes, good idea, will fix.

lgritz: updated

PR #1191, opened 2020-06-03
lazyerror option

lgritz:
By default, shader group layers execute lazily, meaning the node will
not execute until/unless one of its outputs is needed by a downstream
layer, and if in the course of optimization all of its outgoing
connections are eliminated, then the entire node will be pruned from
the shader graph. But if the node contained error() calls (say, to do
sanity check on parameters or other conditions), those error calls
will never be issued, possibly masking a problem.

This patch introduces a new ShadingSystem option:

"lazyerror" (default: 1) -- when set to 0, it means that the presence
in a shader layer of an error() call that *can't* be constant-folded
away will make the layer will execute unconditionally, thereby making
the error() call run even if the layer's outputs are never "pulled" by
any downstream layer.

Furthermore, if lazyerror=0, if you have shaders that after
optimization still have error calls (and thus will run the layer
unditionally), then if option "opt_warnings" is nonzero, you will get
a warning about the presence of a not-optimized-away error call (you
may want to know about this for performance reasons, if your intent
was that all error calls should be resolveable by the optimizer).

Along the way I also added a few more optimization-debugging logging
messages, including understanding exactly when and why unused connections
are pruned.


lgritz:
@fpsunflower @lkerley any comments?

src/liboslexec/oslexec_pvt.h
```diff
@@ -1057,6 +1063,10 @@ class ShaderInstance {
         // lazyunconnected is on.
         if (!outgoing_connections() && !shadingsys().m_lazyunconnected)
             return false;
+        // Shaders without any downstream connections are not lazy unless
+        // lazyunconnected is on.
```
lgritz, line 1067: I realize this comment is a cut and paste error and doesn't match the code. Will fix momentarily...

fpsunflower:
Code is looking ok to me. How did the benchmarks turn out?

lgritz:
Will run them tonight to be sure. Just wanted code to look ok so that if benchmarks are neutral I can do a merge.

fpsunflower:
@lkerley may need to give you some new scenes since they've peppered many more error messages everywhere in the shaders compared to the scenes we have

lgritz:
The way I imagined this working is that I'd add it as an option to our renderer, with the default being (just like in OSL) the old behavior. So the shading team can turn it on for individual development to get themselves to the point where all the shaders are policed up. Then in a second round of changes, we can change the renderer default to enable it all the time.

fpsunflower:
Makes sense. Should be a safe change for now then.

lgritz:
Update on benchmarks:

I tested a 3-way comparison across a wide variety of production scenes:
1. Current renderer (before these OSL changes).
2. These changes, but with default values (*should* match old behavior and perf).
3. New options enabled: lazyerror=0, opt_warnings=1 (note: our renderer exits immediately upon a true error)

I'm pleased to report that (1) and (2) have identical performance, so it seems safe to merge.

(3) turned out to be worthless -- because the shaders used by the scenes haven't been scrubbed with this change in mind, some tests ran much slower (because of layers that run unconditionally now, though no errors were hit), and some have much shorter runtime (because -- they trip errors that were elided before).

PR #1192, opened 2020-06-05
Add gcc9 and gcc10 tests to CI matrix

lgritz:
* Use the new ASWF VFX Platform 2021 docker image for gcc9/py37/llvm10/C++17

* Alter the "bleeding edge" CI test to use gcc10

* Fix some new warnings that pop up with gcc9 and gcc10

### DIFF
--- a/src/liboslexec/osolex.l
+++ b/src/liboslexec/osolex.l
@@ -96,13 +96,16 @@ using namespace OSL::pvt;
 
 // flex itself will generate fatal warnings about signed vs unsigned.
 // Bypass that nonsense.
-#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 6
+#if OSL_GNUC_VERSION >= 60000
 #pragma GCC diagnostic ignored "-Wsign-compare"
 #endif
 
 // flex uses the 'register' keyword, warned because it's deprecated in C++17.
 #if defined(__clang__)
 #pragma GCC diagnostic ignored "-Wdeprecated-register"
+#endif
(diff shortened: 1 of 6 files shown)


src/liboslexec/osolex.l
```diff
@@ -96,14 +96,17 @@ using namespace OSL::pvt;
...
 // Bypass that nonsense.
-#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 6
+#if OSL_GNUC_VERSION >= 60000
 #pragma GCC diagnostic ignored "-Wsign-compare"
 #endif
 
 // flex uses the 'register' keyword, warned because it's deprecated in C++17.
 #if defined(__clang__)
 #pragma GCC diagnostic ignored "-Wdeprecated-register"
 #endif
```
AlexMWells, line 106: Should usage of "register" be removed vs. disabling the warning?  Or is it still of value?

lgritz: I'm not using `register`, it's littered through the parser code that's generated by flex/bison.

lgritz: The best thing would be for flex/bison to remove use of register in their code base and any code they generate, and for everybody in the world who might want to build OSL to upgrade their system install of flex/bison to the new release containing the fix. But that's harder to arrange than suppressing the warning. :-)

AlexMWells: Sounds like that is out of scope,  so disable the warning LGTM ;)

PR #1194, opened 2020-06-06
Propagate rtti compile options to exported cmake config

lgritz:
People linking against oslexec needed to know to use -fno-rtti when
building their apps. Very brittle. This correctly adds the flag as
an interface compile flag in our exported CMake config files.

### DIFF
--- a/src/liboslexec/CMakeLists.txt
+++ b/src/liboslexec/CMakeLists.txt
@@ -190,6 +190,10 @@ if (partio_FOUND)
     target_compile_definitions (${local_lib} PRIVATE USE_PARTIO=1)
 endif ()
 
+# Propagate our library's RTTI options to downstream clients of this library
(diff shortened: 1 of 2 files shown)


src/liboslexec/CMakeLists.txt
```diff
@@ -190,6 +190,10 @@ if (partio_FOUND)
     target_compile_definitions (${local_lib} PRIVATE USE_PARTIO=1)
 endif ()
 
+# Propagate our library's RTTI options to downstream clients of this library
+# This is very importnat for proper linkage with LLVM.
```
AlexMWells, line 194: typo "importnat"

lgritz: fixed and amended

lgritz:
Any objections?

lgritz:
Chris okayed

PR #1195, opened 2020-06-06
Helper Parameter and ReParameter varieties for common types

lgritz:
For attribute and getattribute, we had some specializations that make
it extra easy to do the common cases of a single float, int, string.
This is just adding the same kind of convenience functions for
Parameter() and ReParameter().


lgritz:
Oh dear, you are correct. Update coming in seconds...

PR #1198, opened 2020-06-09
static_foreach does not compile in GPU code

DeclanRussell:
## Description
static_foreach doesn't compile in device code so I've just ifdefed this out as such. Without this you'll get errors similar to the following.
`
dual.h:523:5: error: reference to __host__ function 'static_foreach<ConstIndex, 2, (lambda at C:/Users/russeld/Projects/osl-arnold/src/include\OSL/dual.h:523:5)>' in __host__ __device__ function
    OSL_INDEX_LOOP(i, P, {
    ^
`

## Checklist:


lgritz:
Is the more correct solution to add OSL_HOSTDEVICE to the lambda (`auto functor = [&]...`) a dozen lines farther down?

PR #1200, opened 2020-06-10
Allow output params with lockgeom=0 to have auto userdata binding

lgritz:
Shader params declared with lockgeom=0

    surface foo (float bar = 0 [[ int lockgeom=0 ]]) { }

will generate code making underlying calls to RendererServices->get_userdata
to automatically bind userdata (a.k.a. interpolated "primitive variables")
that share the same name.

But, oddly, this was not true for output parameters:

    surface foo (output float bar = 0 [[ int lockgeom=0 ]]) { }

This was tripping some people up, and in retrospect seems strange.

This patch removes this restriction, allowing both input and output
parameters tagged with lockgeom=0 to attempt to retrieve userdata to get
their initial values.

Add test of userdata retrieval to make sure this works.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -460,6 +460,7 @@ TESTSUITE ( aastep allowconnect-err and-or-not-synonyms arithmetic
             transitive-assign
             transform transformc trig typecast
             unknown-instruction
+            userdata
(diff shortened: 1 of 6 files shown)


aconty:
LGTM!

PR #1201, opened 2020-06-10
"error: 'ListT': parameter pack must be expanded in this context" on windows

DeclanRussell:
## Description
For some reason this line doesn't compile on windows (at least with MSVC14) and throws the following error.
`
oslnoise.h(204): error C3520: 'ListT': parameter pack must be expanded in this context
`
The modification allows this to build and doesn't cause any issues in the testing I've done. I'm not entirely clear what the logic does so would be good if somebody could verify the new logic is correct.

## Checklist:

### DIFF
--- a/src/include/OSL/oslnoise.h
+++ b/src/include/OSL/oslnoise.h
@@ -199,15 +199,20 @@ reference_inthash (const unsigned int k[N]) {
     return c;
 }
 
-
+#if defined(_MSC_VER) && (_MSC_VER >= 1920)
 template<typename HeadT, typename... ListT>
-using RequireAllSame = typename std::enable_if<conjunction<std::is_same<ListT, HeadT> >::value...>::type;
+using RequireAllSame = typename std::enable_if<conjunction<std::is_same<ListT, HeadT>...>::value>::type;
+#endif
 
 // Allow any number of arguments to be adapted to the array based reference_inthash
 // and leave door open for overloads of inthash with specific parameters.
 // NOTE: only accept unsigned int's so that overloads with unsigned int's will be chosen.
 // Callers with "int" parameters will need to static_cast<unsigned int>(param)
+#if defined(_MSC_VER) && (_MSC_VER >= 1920)
 template<typename ...ListT, typename = RequireAllSame<unsigned int, ListT...> >
+#else
+template<typename ...ListT, typename = typename std::enable_if<conjunction<std::is_same<ListT, unsigned int>...>::value>::type >
+#endif
 OSL_FORCEINLINE OSL_HOSTDEVICE unsigned int
 inthash (ListT... uintList)  {
     const int count = sizeof...(uintList);
--- a/src/include/OSL/oslnoise.h
+++ b/src/include/OSL/oslnoise.h
@@ -199,27 +199,6 @@ reference_inthash (const unsigned int k[N]) {
     return c;
 }
 
-#if defined(_MSC_VER) && (_MSC_VER >= 1920)
-template<typename HeadT, typename... ListT>
-using RequireAllSame = typename std::enable_if<conjunction<std::is_same<ListT, HeadT>...>::value>::type;
-#endif
-
-// Allow any number of arguments to be adapted to the array based reference_inthash
-// and leave door open for overloads of inthash with specific parameters.
-// NOTE: only accept unsigned int's so that overloads with unsigned int's will be chosen.
-// Callers with "int" parameters will need to static_cast<unsigned int>(param)
-#if defined(_MSC_VER) && (_MSC_VER >= 1920)
-template<typename ...ListT, typename = RequireAllSame<unsigned int, ListT...> >
-#else
-template<typename ...ListT, typename = typename std::enable_if<conjunction<std::is_same<ListT, unsigned int>...>::value>::type >
-#endif
-OSL_FORCEINLINE OSL_HOSTDEVICE unsigned int
-inthash (ListT... uintList)  {
-    const int count = sizeof...(uintList);
-	const unsigned int iv[count] = {uintList...};
-	return reference_inthash<count>(iv);
-}
-
 #if (__OSL_USE_REFERENCE_INT_HASH == 0)
 	// Do not rely on compilers to fully optimizing the
 	// reference_inthash<int N> template above.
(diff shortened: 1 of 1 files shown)


AlexMWells:
Looks like a compiler bug, 
https://godbolt.org/z/vLpwqG
shows it failing with MSVC 19.16 and then moving to MSVC 19.20 it works fine.
Can you move to new MSVC?

DeclanRussell:
Ouch that sucks! Sounds like that would be the only way. This isn't super blocking for us right now. Our actual current work around is just to ifdef this out for _MSV_VER <= 1900.

How do you guys want to deal with this on your end. I could commit something similar and have it ifdefed out for certain MSVC versions which but this breaks if you start using the function. Another option would be to enforce a minimum MSVC version.

AlexMWells:
@DeclanRussell can you try addition conditional compilation based on MSVC version < 19.20 to not use the "RequireAllSame" alias and just use the conjunction directly (which appears to work), ie: https://godbolt.org/z/j9Wrou
The RequireAllSame is an alias to provide documentation to the intent, could remove it entirely but still like it there, so thinking some kind of compiler version check to use it or not.  (Larry feel free to overule)
```
#if defined(_MSV_VER) && _MSV_VER < 1920
template<typename ...ListT, typename = typename std::enable_if<conjunction<std::is_same<ListT, unsigned int>...>::value>::type >
#else
template<typename ...ListT, typename = RequireAllSame<unsigned int, ListT...> >
#endif
OSL_FORCEINLINE OSL_HOSTDEVICE unsigned int
inthash (ListT... uintList)  {
```

I don't have that compiler or build environment, so best if you try it out and update PR.

fpsunflower:
The only place using this fancy template expression is `inthash`? Can we find a different way to write this code?

Also -- aren't all the used cases explicitly specified right below? What purpose does the template serve?

AlexMWells:
The problem appears to be a particular compilers support for parameter pack expansion in a template alias.  In general, its purpose is to only allow "unsigned int" to be parameters to the variadic template.  As this is the only place it is used, we can just remove the template alias RequireAllSame and use the conjunction<> directly.  And then there is no conditional compilation.  Intent of the alias is to be self documenting, int has requires that all of its parameters be the same type and that type is "unsigned int":

```
// Require all parameters to variadic template inthash be unsigned int
template<typename ...ListT, typename = typename std::enable_if<conjunction<std::is_same<ListT, unsigned int>...>::value>::type >
```

fpsunflower:
But is the variadic template actually invoked anywhere?

There are explicit versions for `uint` and `simd_int` below that seem to cover all values of `N` that we care about.

What am I missing?

AlexMWells:
@fpsunflower , background for why its there.  Previously there was a 
```
template <int N>
unsigned int inthash (const unsigned int k[N]);
```
in a public facing header with unknown external uses as well as internal uses.  A goal of adding new optimized versions was to not break existing external callers as well as be able to fallback to the original version if necessary.  Rather than having conditional compilation at all the internal call sites to switch between the old version and optimized version, choose to have a new int_hash that is a variadic template taking any number of parameters vs. an array (most call sites just built an array on the stack then called the reference version).  Well that pattern is encapsulated in the int_hash variadic template.  The nice part is overloaded int_has can exist and will be called in favor of the template version.  This provides a clean way to switch between reference and optimized versions.  Absence  of the optimized overloads will cause the reference version to be used.  As a variadic template can accept any type, the "RequireAllSame" is just a way to require unsigned int parameters vs. accepting any parameter type (future C++ standards are addressing this shortcoming).  It has advantage of not allowing implicit conversion which is better self documentation for call sites who would need to explicitly convert to unsigned int.

fpsunflower:
I see. I personally would be fine with just saying that `inthash` is an implementation detail that is not really meant to be part of the public API.

If its not used by OSL, then I don't think the `inthash(uint x[N])` version is worth exposing to anyone else.

lgritz:
Agreed, I think we can assume that is an implementation detail.

lgritz:
@DeclanRussell Have you followed the train of thought above and are comfortable making that change?

DeclanRussell:
I tested @AlexMWells 's fix and it seems to build fine. Thanks @AlexMWells! I've gone ahead and committed this.

So do you guys instead want me to remove the `inthash (ListT... uintList)` version completely or are you happy with the fix Alex came up with?

fpsunflower:
I'd vote for removing it entirely.

lgritz:
Agree, toss it.

DeclanRussell:
Done!

src/include/OSL/oslnoise.h
```diff
@@ -199,22 +199,6 @@ reference_inthash (const unsigned int k[N]) {
...
-// Callers with "int" parameters will need to static_cast<unsigned int>(param)
-template<typename ...ListT, typename = RequireAllSame<unsigned int, ListT...> >
-OSL_FORCEINLINE OSL_HOSTDEVICE unsigned int
-inthash (ListT... uintList)  {
-    const int count = sizeof...(uintList);
-	const unsigned int iv[count] = {uintList...};
-	return reference_inthash<count>(iv);
-}
-
 #if (__OSL_USE_REFERENCE_INT_HASH == 0)
```
lgritz, line 202: Sorry to burden you with this, Declan, but can/should we eliminate these guards and use of the `__OSL_USE_REFERENCE_INT_HASH` symbol?

lgritz: I'm just thinking, if this serves no purpose, then let's not make the next person to come through this section of code get confused need to hunt for what that's supposed to be doing and under what circumstances it might be nonzero.

DeclanRussell: Hmm it could be worthwhile getting rid of the `__OSL_USE_REFERENCE_INT_HASH` given that if this is non zero then it will fail to compile anyway because it removes all the inthash functions. So at this point your going to have to do some big modifications to the header anyway. We would have to decide what to do with the second definition of the variadic `inthashVec` function though as these would clash.

lgritz: I'll merge this now, and then I can fix the definition in a second pass.

AlexMWells: I guess this is fine, the purpose of the inthash (ListT... uintList) was to be easily fall back to the reference version in case of code gen or correctness issues.  If we are comfortable/confident then such a fallback mechanism may not be necessary, still suggest keeping reference_hash around (maybe inside a comment) so readers could trace back what exactly the optimized versions are doing.

PR #1208, opened 2020-06-16
Add LLVM_Util methods to detect appropriate ISA for the host CPU.

lgritz:
We don't use this quite yet, but are staging it for easy review.

Work by Alex Wells. Some refactoring and minor edits by LG.

### DIFF
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -452,7 +452,369 @@ LLVM_Util::end_builder ()
 }
 
 
+static llvm::StringMap<bool> sCpuFeatures;
 
+static bool populateCpuFeatures()
+{
+    return llvm::sys::getHostCPUFeatures(sCpuFeatures);
+}
+
+
+static bool initCpuFeatures()
+{
+    // Lazy singleton behavior, populateCpuFeatures() should
+    // only get called once by 1 thread per C++ static initialization rules
+    static bool is_initialized = populateCpuFeatures();
+    return is_initialized;
+}
+
+
+// The list of cpu features should correspond to the target architecture
+// or feature set that the corresponding wide library.
+// So if you change the target cpu or features in liboslexec/CMakeList.txt
+// for any of the wide libraries, please update here to match
+static const char * target_isa_names[] = {
+    "UNKNOWN", "x64", "SSE4.2", "AVX", "AVX2", "AVX2_noFMA", "AVX512", "AVX512_noFMA"
+};
+
+
+// clang: default
+// icc: default
+static const char * required_cpu_features_by_x64[] = {
+    "fxsr", "mmx", "sse", "sse2", "x87"
+};
+
+// clang: -march=nehalem
+// icc: -xSSE4.2
+static const char * required_cpu_features_by_SSE4_2[] = {
+    "cx16","fxsr","mmx","popcnt",
+    // "sahf", // we shouldn't need/require this feature
+    "sse","sse2","sse3","sse4.1",
+    "sse4.2","ssse3", "x87"
+};
+
+// clang: -march=corei7-avx
+// icc: -xAVX
+static const char * required_cpu_features_by_AVX[] = {
+    "aes", "avx", "cx16", "fxsr", "mmx", "pclmul", "popcnt",
+    // "sahf", // we shouldn't need/require this feature
+    "sse",
+    "sse2", "sse3", "sse4.1", "sse4.2", "ssse3", "x87"
+    // ,"xsave","xsaveopt" // Save Processor Extended States, we don't use
+};
+
+// clang: -march=core-avx2
+// icc: -xCORE-AVX2
+static const char * required_cpu_features_by_AVX2[] = {
+    "aes", "avx", "avx2", "bmi", "bmi2", "cx16", "f16c", "fma",
+    // "fsgsbase", // we shouldn't need/require this feature
+    "fxsr",
+    // "invpcid", // Invalidate Process-Context Identifier, we don't use
+    "lzcnt", "mmx", "movbe", "pclmul", "popcnt",
+    // "rdrnd", // random # don't require unless we make use of it
+    // "sahf", // we shouldn't need/require this feature
+    "sse",
+    "sse2", "sse3", "sse4.1", "sse4.2", "ssse3", "x87"
+    // ,"xsave","xsaveopt" // // Save Processor Extended States, we don't use
+};
+
+// clang: -march=core-avx2 -mno-fma
+// icc: -xCORE-AVX2 -no-fma
+static const char * required_cpu_features_by_AVX2_noFMA[] = {
+    "aes", "avx", "avx2", "bmi", "bmi2", "cx16", "f16c",
+    // "fsgsbase", // we shouldn't need/require this feature
+    "fxsr",
+    // "invpcid", // Invalidate Process-Context Identifier, we don't use
+    "lzcnt", "mmx", "movbe", "pclmul", "popcnt",
+    // "rdrnd", // random # don't require unless we make use of it
+    // "sahf", // we shouldn't need/require this feature
+    "sse",
+    "sse2", "sse3", "sse4.1", "sse4.2", "ssse3", "x87"
+    // , "xsave", "xsaveopt" // Save Processor Extended States, we don't use
+};
+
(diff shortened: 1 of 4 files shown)


src/liboslexec/llvm_util.cpp
```diff
@@ -437,7 +437,369 @@ LLVM_Util::end_builder ()
...
+    return target_isa_names[static_cast<int>(isa)];
+}
+
+
+
+bool
+LLVM_Util::detect_cpu_features(string_view requested_isa_string,
+                               bool no_fma)
+{
+    TargetISA requestedISA = lookup_isa_by_name(requested_isa_string);
```
AlexMWells, line 641: I am wondering if maybe `requested_isa_string` parameter to detect_cpu_features, should just be `requestedISA `and make caller call` lookup_isa_by_name(requested_isa_string) `themselves?

src/liboslexec/llvm_util.cpp
```diff
@@ -437,7 +437,369 @@ LLVM_Util::end_builder ()
...
+    "pclmul",
+    // "pku" Memory Protection Keys we shouldn't need/require this feature,
+    "popcnt", "prfchw", "rdrnd", "rdseed", "rtm",
+#if OSL_LLVM_VERSION >= 70
+    "sahf",
+#endif
+    "sse", "sse2", "sse3", "sse4.1", "sse4.2", "ssse3", "x87"
+    // , "xsave", "xsavec", "xsaveopt", "xsaves" // Save Processor Extended States, we don't use
+};
+
```
AlexMWells, line 563: We have discovered the list of cpu features is a bit too specific to server parts and could have issues with client parts that may not have some of the cpu features, like "rtm".  As it turns out most of these server/OS geared instructions are only really accessible explicitly with intrinsics and we don't expect LLVM, GCC, CLang, or ICC to generate them by themselves, so we should be safe to remove those features which we are not explicitly using.  Suggest changing these lists to:
```
// clang: default
// icc: default
static const char * required_cpu_features_by_x64[] = {
    "fxsr", "mmx", "sse", "sse2", "x87"
};

// clang: -march=nehalem
// icc: -xSSE4.2
static const char * required_cpu_features_by_SSE4_2[] = {
    "cx16","fxsr","mmx","popcnt",
#if OSL_LLVM_VERSION >= 70
    // "sahf", // we shouldn't need/require this feature
#endif
    "sse","sse2","sse3","sse4.1",
    "sse4.2","ssse3", "x87"
};

// clang: -march=corei7-avx
// icc: -xAVX
static const char * required_cpu_features_by_AVX[] = {
    "aes", "avx", "cx16", "fxsr", "mmx", "pclmul", "popcnt",
#if OSL_LLVM_VERSION >= 70
    // "sahf", // we shouldn't need/require this feature
#endif
    "sse",
    "sse2", "sse3", "sse4.1", "sse4.2", "ssse3", "x87"
    // ,"xsave","xsaveopt" // Save Processor Extended States, we don't use
};

// clang: -march=core-avx2
// icc: -xCORE-AVX2
static const char * required_cpu_features_by_AVX2[] = {
    "aes", "avx", "avx2", "bmi", "bmi2", "cx16", "f16c", "fma",
    // "fsgsbase", // we shouldn't need/require this feature
    "fxsr",
    // "invpcid", // Invalidate Process-Context Identifier, we don't use
    "lzcnt", "mmx", "movbe", "pclmul", "popcnt",
    // "rdrnd", // random # don't require unless we make use of it
#if OSL_LLVM_VERSION >= 70
    // "sahf", // we shouldn't need/require this feature
#endif
    "sse",
    "sse2", "sse3", "sse4.1", "sse4.2", "ssse3", "x87"
    // ,"xsave","xsaveopt" // // Save Processor Extended States, we don't use
};

// clang: -march=core-avx2 -mno-fma
// icc: -xCORE-AVX2 -no-fma
static const char * required_cpu_features_by_AVX2_noFMA[] = {
    "aes", "avx", "avx2", "bmi", "bmi2", "cx16", "f16c",
    // "fsgsbase", // we shouldn't need/require this feature
    "fxsr",
    // "invpcid", // Invalidate Process-Context Identifier, we don't use
    "lzcnt", "mmx", "movbe", "pclmul", "popcnt",
    // "rdrnd", // random # don't require unless we make use of it
#if OSL_LLVM_VERSION >= 70
    // "sahf", // we shouldn't need/require this feature
#endif
    "sse",
    "sse2", "sse3", "sse4.1", "sse4.2", "ssse3", "x87"
    // , "xsave", "xsaveopt" // Save Processor Extended States, we don't use
};

// clang: -march=skylake-avx512
// icc: -xCORE-AVX512
static const char * required_cpu_features_by_AVX512[] = {
    "adx", "aes", "avx", "avx2", "avx512bw", "avx512cd", "avx512dq",
    "avx512f", "avx512vl", "bmi", "bmi2",
    // "clflushopt", "clwb", flushing for volatile/persistent memory we shouldn't need
    "cx16",
    "f16c", "fma",
    // "fsgsbase", // we shouldn't need/require this feature,
    "fxsr",
    // "invpcid", // Invalidate Process-Context Identifier, we don't use
    "lzcnt", "mmx", "movbe",
    //"mpx", // Memory Protection Extensions, we don't use
    "pclmul",
    // "pku"//  Memory Protection Keys we shouldn't need/require this feature,
    "popcnt",
    // "prfchw", // prefetch wide we shouldn't need/require this feature,
    // "rdrnd", "rdseed", // random # don't require unless we make use of it
    // "rtm", // transaction memory we shouldn't need/require this feature,
#if OSL_LLVM_VERSION >= 70
    // "sahf", // we shouldn't need/require this feature
#endif
    "sse", "sse2", "sse3", "sse4.1", "sse4.2", "ssse3", "x87"
    // , "xsave", "xsavec", "xsaveopt", "xsaves" // Save Processor Extended States, we don't use
};
// clang: -march=skylake-avx512 -mno-fma
// icc: -xCORE-AVX512 -no-fma
static const char * required_cpu_features_by_AVX512_noFMA[] = {
    "adx", "aes", "avx", "avx2", "avx512bw", "avx512cd", "avx512dq",
    "avx512f", "avx512vl", "bmi", "bmi2",
    // "clflushopt", "clwb", flushing for volatile/persistent memory we shouldn't need
    "cx16",
    "f16c",
    // "fsgsbase", // we shouldn't need/require this feature
    "fxsr",
    // "invpcid", // Invalidate Process-Context Identifier, we don't use
    "lzcnt", "mmx", "movbe",
    //"mpx", // Memory Protection Extensions, we don't use
    "pclmul",
    // "pku"//  Memory Protection Keys we shouldn't need/require this feature,
    "popcnt",
    // "prfchw", // prefetch wide we shouldn't need/require this feature,
    // "rdrnd", "rdseed", // random # don't require unless we make use of it
    // "rtm", // transaction memory we shouldn't need/require this feature,
#if OSL_LLVM_VERSION >= 70
    // "sahf", // we shouldn't need/require this feature
#endif
    "sse", "sse2", "sse3", "sse4.1", "sse4.2", "ssse3", "x87"
    // , "xsave", "xsavec", "xsaveopt", "xsaves" // Save Processor Extended States, we don't use
};
```

Also if our minimum LLVM version is 7+ now, we could drop the conditional compilation pre-processor stuff:
```
#if OSL_LLVM_VERSION >= 70
```

src/liboslexec/llvm_util.cpp
lgritz: I have no preference. Would you like me to change it?

src/liboslexec/llvm_util.cpp
lgritz: Thanks. Pushing an amended commit now with this change.

src/liboslexec/llvm_util.cpp
AlexMWells: Yes, I think so, it was bugging me in use, and now that you have busted out detect_cpu_features to its own function, it makes sense.

lgritz: OK, will do.

lgritz: Amended.

Also make llvmutil_test double check that what it's getting from these LLVM checks of the host ISA matches what OIIO thinks it's getting from cpuid queries.

AlexMWells:
LLVM_Util::LLVM_Util() constructor should probably set
m_target_isa(TargetISA::UNKNOWN)

PR #1211, opened 2020-06-21
Re-enable RTTI

lgritz:
We had disabled RTTI throughout the OSL code base to avoid linkage
problems against LLVM, which builds with RTTI disabled by default.

But it's actually ok to mix code modules with differing RTTI status --
that's fine. It's just that you can't mix RTTI status within a class
inheritance hierarchy.

We observe that

  * The only classes we inherit from RTTI-free LLVM are solely defined
    and used within llvm_util.cpp.

  * The only classes exposed in llvm_util.h have no virtual functions
    and aren't designed to be subclassed.

So in fact, it appears that we can get by with only using no-rtti when
compiling that one module, and don't need to mess with rtti anywhere
else (including downstream users). This is an important step to
supporting python bindings, since pybind11 requires RTTI.

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -446,26 +446,6 @@ endif ()
 
 
 ###########################################################################
-# Enable/disable RTTI
-#
-# NOTE: LLVM builds without RTTI by default so beware if you find the need
-# to turn this on, to use OSL in a project that requires RTTI for example,
-# you need to build LLVM with RRTI, otherwise OSL classes extending LLVM
-# ones will cause linker errors.
-#
-set (ENABLE_RTTI OFF CACHE BOOL "Build with RTTI. Requires a LLVM build with RTTI enabled.")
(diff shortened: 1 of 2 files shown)


fpsunflower:
This looks reasonable. End-users that have other reasons to disable exceptions (and don't need python bindings) will presumably be able to just pass in `-fno-rtti` via `CMAKE_CXX_FLAGS`?

lgritz:
It's not exceptions, it's RTTI.

Yes, they could still disable it with CMAKE_CXX_FLAGS, but it's hard see why. There's no harm in a dependent library using RTTI if your app doesn't want it. But if the dependent library disables RTTI and your app derives a class from it, then the derived class either also needs RTTI disabled, or else it will be a link error.

fpsunflower:
Yes sorry mistyped. The penalty is indeed not huge if you aren't using `dynamic_cast` calls, it's mostly just extra bloat in the vtable and some extra info in the executable.

In any case LGTM!

PR #1212, opened 2020-06-23
Python bindings for OSLQuery

lgritz:
Using pybind11. See testsuite/python-oslquery/test_oslquery.py for a
full example of how to use it.


lgritz:
Any comments or objections?

lgritz:
No objections --> merging

PR #1213, opened 2020-06-30
Fix memory leak when using the JIT Event listener

fpsunflower:
The execution engine does not take ownership of this object, so we need to be cleaning it up ourselves.

I discovered this when running address sanitizer on our renderer and noticed that every test using OSL was failing.

### DIFF
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -22,6 +22,7 @@ namespace llvm {
   class Function;
   class FunctionType;
   class SectionMemoryManager;
+  class JITEventListener;
   class Linker;
   class LLVMContext;
   class Module;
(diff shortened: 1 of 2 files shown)


src/liboslexec/llvm_util.cpp
```diff
@@ -866,6 +866,11 @@ LLVM_Util::execengine (llvm::ExecutionEngine *exec)
 {
     delete m_llvm_exec;
     m_llvm_exec = exec;
+
+    if (exec == NULL) {
```
lgritz, line 870: Maybe a comment here to explain why you need to delete it if and only if exec == nullptr?

src/include/OSL/llvm_util.h
```diff
@@ -615,6 +616,7 @@ class OSLEXECPUBLIC LLVM_Util {
     llvm::legacy::PassManager *m_llvm_module_passes;
     llvm::legacy::FunctionPassManager *m_llvm_func_passes;
     llvm::ExecutionEngine *m_llvm_exec;
+    lvm::JITEventListener *m_vtune_profiler;
```
lgritz, line 619: Seems to be a typo here

src/liboslexec/llvm_util.cpp
fpsunflower: Turns out argument is always `NULL` - not sure why I thought I needed the if. In any case, should be safe to always delete the profiler.

src/include/OSL/llvm_util.h
fpsunflower: Sorry bad rebase 😛 Fixed.

PR #1214, opened 2020-06-30
Add ISA-specific targeting and additional JIT control options

lgritz:
ISA means "instruction set architecture", by the way.

Heavy lifting by Alex Wells, some refactoring and adding options by LG.

A number of additional controls over the JIT process are added to the
ShadingSystem:

- "llvm_jit_target" Can be set to the name of an architecture ("none",
  "x64", "SSE4.2", "AVX2", "AVX2_noFMA", "AVX512", "AVX512_noFMA") to
  request that the JIT choose instructions for a specific machine ISA.
  Also "host" means to select whatever the host natively supports.
  It checks if the host CPU supports the requested architecture and
  will automatically restrict it to a choice your machine supports.
  The default is "none" meaning not to generate code for anything beyond
  the default Intel target. There seems to be little benefit to using
  this at this time for single-point shading, but we anticipate that
  it will be important to the forthcoming batch shading mode.

- "vector_width" Can set the vector width for SIMD instructions,
  independently of the JIT target (for example, you could JIT for
  AVX512 but still only make 8-wide SIMD instructions). The current
  default is 4, meaning it will not generate 8-wide or 16-wide
  vector ops, even when targeting and running on AVX512. We have found
  significant performance penalties when running single point shading
  mode on AVX512 hardware, unless we use this restriction (but no such
  slowdown on AVX2 hardware).

- "llvm_jit_fma" Enables use of fused multiply-add instructions on
  architectures that support them.  When enabled, resulting code is
  slightly faster improves rounding accuracy for some computations.
  But the default is 0, because enabling it can make results differ
  slightly on machines with different architetures (bad for testing,
  and maybe leading to problems if you have machines with differing
  architectures working on different frames of the same shot).

- "llvm_jit_aggressive" Enables the "aggressive" JIT option, whatever
  that means. Currently defaulted to 0.

- "llvm_dumpasm" a debugging mode that lets developers see the actual
  JIT code as assembly language.

The bottom line here is that the defaults are carefully chosen to be
performance-neutral on all architectures. We do not currently advise
adjusting these settings for single-point shading mode. When we add
batch-shading mode, we will probably have different defaults or
guidance to offer when using that mode (and the specific advice may
depend on the specific architecture you are running on).


src/liboslexec/shadingsys.cpp
```diff
@@ -715,13 +715,17 @@ ShadingSystemImpl::ShadingSystemImpl (RendererServices *renderer,
       m_opt_fold_getattribute(true),
       m_opt_middleman(true), m_opt_texture_handle(true),
       m_opt_seed_bblock_aliases(true),
+      m_llvm_jit_fma(false),
+      m_llvm_jit_aggressive(false),
       m_optimize_nondebug(false),
+      m_vector_width(8),
```
AlexMWells, line 721: @lgritz didn't you want 4 as the default m_vector_width?

src/liboslexec/llvmutil_test.cpp
```diff
@@ -220,7 +220,12 @@ test_isa_features()
     // Make sure it matches what OIIO's cpuid queries reveal
     OIIO_CHECK_EQUAL(ll.supports_avx(), OIIO::cpu_has_avx());
     OIIO_CHECK_EQUAL(ll.supports_avx2(), OIIO::cpu_has_avx2());
+#if OSL_LLVM_VERSION >= 80
     OIIO_CHECK_EQUAL(ll.supports_avx512f(), OIIO::cpu_has_avx512f());
+    // N.B. Skip this check because for LLVM < 8, we intentionally disable
+    // AVX512 support in order to avoid a performance problem, so we
+    // expect this to potentially not match.
+#endif
 }
```
AlexMWells, line 230: Suggest leaving supports alone and accurate for all LLVM versions and instead during backend's selection of a target isa limit it to SSE4.2 when OSL_LLVM_VERSION < 80.

src/liboslexec/llvm_util.cpp
```diff
@@ -294,15 +294,17 @@ LLVM_Util::LLVM_Util (int debuglevel, int vector_width)
     m_llvm_type_matrix_ptr = (llvm::PointerType *) llvm::PointerType::get (m_llvm_type_matrix, 0);
 
     // Setup up wide aliases
-    // TODO:  why are there casts to the base class llvm::Type *?  
+    // TODO:  why are there casts to the base class llvm::Type *?
+    if (m_vector_width != 4 && m_vector_width != 8 && m_vector_width != 16)
+        m_vector_width = 4;
```
AlexMWells, line 299: How would user know they passed illegal/unsupported vector width, maybe report an error if an assert is not desired.

src/liboslexec/llvm_util.cpp
```diff
@@ -666,13 +648,19 @@ LLVM_Util::detect_cpu_features(TargetISA requestedISA, bool no_fma)
         return false;  // Could not figure it out
     }
 
-    bool target_requested = (requestedISA != TargetISA::UNKNOWN);
+    // Try to match features to the combination of the requested ISA and
+    // what the host CPU is able to support. Note that we intentionally
+    // disable AVX512 ISA support on AVX512 hardware unless LLVM >= 8.0,
+    // to avoid some performance issues that we can't fix without LLVM 8+
+    // features.
```
AlexMWells, line 655: Suggest not limiting here, but in backend when choosing a targetISA

src/liboslexec/llvm_util.cpp
```diff
@@ -825,16 +796,67 @@ LLVM_Util::make_jit_execengine (std::string *err)
...
-    /* Future site of target options setting */
+    engine_builder.setOptLevel (jit_aggressive()
+                                ? llvm::CodeGenOpt::Aggressive
+                                : llvm::CodeGenOpt::Default);
+
+    llvm::TargetOptions options;
+    // Enables FMA's in IR generation.
+    // However cpu feature set may or may not support FMA's independently
+    options.AllowFPOpFusion = jit_fma() ? llvm::FPOpFusion::Fast :
+                                          llvm::FPOpFusion::Standard;
```
AlexMWells, line 815: Previous default of target options defaults to  llvm::FPOpFusion::Standard // Only allow fusion of 'blessed' ops (currently just fmuladd). 
To truly disable FMA, might need to replace llvm::FPOpFusion::Standard 
with llvm::FPOpFusion::Strict // Never fuse FP-ops.

src/liboslexec/shadingsys.cpp
AlexMWells, line 721: Shouldn't this be m_vector_width(4) by default?

AlexMWells:
Any movement on this?

src/liboslexec/llvm_util.cpp
```diff
@@ -474,7 +477,8 @@ static bool initCpuFeatures()
 // So if you change the target cpu or features in liboslexec/CMakeList.txt
 // for any of the wide libraries, please update here to match
 static const char * target_isa_names[] = {
-    "UNKNOWN", "x64", "SSE4.2", "AVX", "AVX2", "AVX2_noFMA", "AVX512", "AVX512_noFMA"
+    "UNKNOWN", "none", "x64", "SSE4.2", "AVX", "AVX2", "AVX2_noFMA",
+    "AVX512", "AVX512_noFMA", "host"
 };
```
sfriedmapixar, line 484: We've seen cases where the below requrement of "aes" has prevented the use of the SIMD path because the "aes" cpu feature was turned off in the BIOS.  Since OSL isn't actually using those instructions as far as I know, it should probably be dropped from the requirement list.

lgritz: Thanks for the tip! It sure seems unlikely that those ops would be needed in the JIT.

What would be the reason that somebody would want to turn this off in the BIOS?

sfriedmapixar: I think it was a mis-configuration, not intentional.

src/liboslexec/llvm_util.cpp
lgritz: switching to floor2(clamp(m_vector_width, 4, 16)), which effectively chooses the largest valid vector width less than or equal to what they requested.

src/liboslexec/llvm_util.cpp
lgritz: Aha, gotcha. I'm going to keep it extra simple, and not try to check or enforce this, since our current guidance is to not target 256 or 512 bit ops when doing single point shading, regardless. We'll measure separately for batch shading when we merge it, and it may be that at such time the guidance we give is that people should always use LLVM >= 8.0 if they want to employ the batch shading.

src/liboslexec/llvm_util.cpp
lgritz: I will simply note this as a comment.

src/liboslexec/llvm_util.cpp
lgritz: I'll comment out the AES features.

@AlexMWells Can you think of any reason that we would benefit (for OSL JIT purposes) from the AES instructions?

lgritz:
Pushed update that addresses all the review comments.

src/liboslexec/llvm_util.cpp
AlexMWells: Not at this moment, I'm pretty sure one would need to use intrinsics and until we have a use case, we can leave it off.

PR #1221, opened 2020-07-18
Option to generate full debugging and profiling info.

lgritz:
Add support for generating debug information to track function call
stack, source file, and line number inside generated LLVM module.
When enabled, can debug with source and line numbers.  Also Intel(r)
VTune(tm) Profiler associated assembly to shader source and line
number.

The big upshot here is that when this debugging is enabled,

  * gdb/lldb can debug a renderer app, all the way down to inside JITed
    shaders.

  * Stack dumps upon a crash will show you the full stack trace
    including where you were in a shader.

  * Profilers such as vtune can be used to accurately profile shaders.

The profiling only works if your LLVM was built with the option
`-DLLVM_USE_INTEL_JITEVENTS=ON`.

Due to some ABI changes on the LLVM side, the debugging symbols only
work properly if you are using LLVM >= 7.1.

Some other specifics that got rolled into this:

New ShadingSystem attributes to control this: `llvm_debugging_symbols`
and `llvm_profiling_events`. Both default to 0 ordinarily, a renderer
needs to enable them, which we only recommend doing if the renderer
itself is compiled for debugging or profiling.  The testshade and
testrender program automatically enable these options.

Debugging features to dump struct layouts and validate LLVM struct
representation based off expected offsets populated from host versions
of the data structures:

    void dump_struct_data_layout(llvm::Type *Ty);
    void validate_struct_data_layout(llvm::Type *Ty,
    const vector & expected_offset_by_index);

Added op_functioncall_nr to as a noop with the name of the function
that was inlined. Previously constfolding would eliminate a
op_functioncall if it had no return value.  Now we transmute into a
op_functioncall_nr. There should be no real shader execution cost to
this, but now can debug through nested OSL shader level function
calls. We needed the op_functioncall_nr to trigger the creation of
debug info for an inlined function call.

When not debugging, don't waste by naming basic blocks.

### DIFF
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -750,6 +750,8 @@ BackendLLVM::build_llvm_code (int beginop, int endop, llvm::BasicBlock *bb)
                 llvm_generate_debug_uninit (op);
             if (shadingsys().llvm_debug_ops())
                 llvm_generate_debug_op_printf (op);
+            if (ll.debug_is_enabled())
+                ll.debug_set_location(op.sourcefile(), std::max(op.sourceline(), 1));
             bool ok = (*opd->llvmgen) (*this, opnum);
             if (! ok)
                 return false;
@@ -782,11 +784,17 @@ BackendLLVM::build_llvm_init ()
 {
     // Make a group init function: void group_init(ShaderGlobals*, GroupData*)
     // Note that the GroupData* is passed as a void*.
-    std::string unique_name = Strutil::sprintf ("__direct_callable__group_%d_init", group().id());
+    std::string unique_name = Strutil::sprintf ("__direct_callable__group_%s_%d_init",
+                                                group().name(), group().id());
     ll.current_function (
            ll.make_function (unique_name, false,
                              ll.type_void(), // return type
                              llvm_type_sg_ptr(), llvm_type_groupdata_ptr()));
+
+    if (ll.debug_is_enabled()) {
+        ustring sourcefile = group()[0]->op(group()[0]->maincodebegin()).sourcefile();
+        ll.debug_push_function(unique_name, sourcefile, 0);
+    }
 
     // Get shader globals and groupdata pointers
     m_llvm_shaderglobals_ptr = ll.current_function_arg(0); //arg_it++;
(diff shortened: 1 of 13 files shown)


src/liboslexec/llvm_instance.cpp
```diff
@@ -782,12 +784,18 @@ BackendLLVM::build_llvm_init ()
 {
     // Make a group init function: void group_init(ShaderGlobals*, GroupData*)
     // Note that the GroupData* is passed as a void*.
-    std::string unique_name = Strutil::sprintf ("__direct_callable__group_%d_init", group().id());
+    std::string unique_name = Strutil::sprintf ("__direct_callable__group_%s_%d_init",
```
fpsunflower, line 787: `unique_name` is only used in the block below I think -- can it be moved inside there?

fpsunflower: Oh nevermind, I see its used for `make_function` as well.

lgritz:
@AlexMWells, in final testing to prepare for a merge, with OPTIX mode enabled, I'm seeing some problems here that the CI didn't turn up, specifically in the OptiX tests. I suspect it's mostly benign, but it's hitting some assertions and I want to be sure we understand them and are fixing what's necessary.

I'm only failing OptiX tests, and the first assertion I hit is in llvm_gen.cpp,

    llvm_gen_functioncall_nr: Assertion 'rop.ll.debug_is_enabled && "no return version should only exist when debug is enabled"' failed

Here's what I think is happening:

Ultimately, debugging symbols are only generated because of a cascade of choices made in LLVM_Util::make_jit_execengine, but that is only called when JITing code for x86, it is not called when generating PTX.

But the choice about whether to replace the functioncall with functioncall_nr is made in constfold_functioncall (by checking rop.keep_no_return_function_calls()), which is part of the runtime optimizer. 

So it's making the functioncall_nr choice based on whether debugging symbols were requested -- which they *always* are for testshade, but later it will be effectively disabled for the PTX path by virtue of never setting up a DIBuilder. Then when generating IR, we are failing assertions that are effectively checking the presence of the DIBuilder.

One way to fix is to disable several of those assertions, and just allow the functioncall_nr to persist anyway. There is no down side or harm to that, right? And there are a few other spots where we need to fix as well, such as avoiding the debug_push_inlined_function call in llvm_gen_functioncall_nr.

Another way to fix is farther upstream, very early in the RuntimeOptimizer constructor where we initialize m_keep_no_return_function_calls, just make sure it's always disabled for OptiX mode, regardless of whether the app requested debugging symbols. Though this raises two possible future issues:

1. Will we ever want to load and runtime optimize a shader group once, then potentially jit/execute it on BOTH OptiX and CPU? That doesn't sound unreasonable, and is related to the split of optimize vs jit that we have in an upcoming PR (which you did so that a renderer can use a shader group for both single-point and batch shading).

2. Is there any use for adding debug symbols in PTX mode? I'm not sure what the debugging options are in OptiX or if this is ever helpful or has any effect whatsoever when generating PTX.

lgritz:
OK, I will make that simple change for now to get us unstuck and able to merge this patch.

Late-breaking news: I'm led to understand that although the debugging information is useless to the PTX path at this very moment, at some point in the future it may make better use of it. So let's just make a mental note that we will probably want to refactor this all so that the DIBuilder and friends are not set up inside make_jit_execengine, but rather someplace else that can be included for the PTX/GPU path, so that it can be fully plumbed for debugging info as well.

PR #1223, opened 2020-07-22
Small improvement in materializing and llvm optimization passes

lgritz:
From Alex Wells

Signed-off-by: Alex M. Wells <alex.m.wells@intel.com>

]

### DIFF
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -966,17 +966,29 @@ LLVM_Util::do_optimize (std::string *out_err)
     OSL_ASSERT (m_llvm_module && "No module to optimize!");
 
 #if !defined(OSL_FORCE_BITCODE_PARSE)
-    LLVMErr err = m_llvm_module->materializeAll();
-    if (error_string(std::move(err), out_err))
-        return;
+    // Pruning would have already materialized all necessary functions and
+    // removed the rest from the module.
+    if (!m_ModuleIsPruned) {
+        LLVMErr err = m_llvm_module->materializeAll();
+        if (error_string(std::move(err), out_err))
+            return;
+     }
 #endif
 
+#if 1
+    // To improve compile time, choose to avoid overhead of optimizaing
+    // functions independently from the module level.
+    m_llvm_func_passes->doInitialization();
+    m_llvm_module_passes->run (*m_llvm_module);
+    m_llvm_func_passes->doFinalization();
+#else
     m_llvm_func_passes->doInitialization();
     for (auto&& I : m_llvm_module->functions())
         if (!I.isDeclaration())
             m_llvm_func_passes->run(I);
     m_llvm_func_passes->doFinalization();
     m_llvm_module_passes->run (*m_llvm_module);
+#endif
 }
 
 


src/liboslexec/llvm_util.cpp
```diff
@@ -965,17 +965,29 @@ LLVM_Util::do_optimize (std::string *out_err)
...
+        LLVMErr err = m_llvm_module->materializeAll();
+        if (error_string(std::move(err), out_err))
+            return;
+     }
 #endif
 
+#if 1
+    // To improve compile time, choose to avoid overhead of optimizaing
+    // functions independently from the module level.
+    m_llvm_func_passes->doInitialization();
```
fpsunflower, line 980: How much does this save?

The pattern here looks strange visually -- we init the function passes, run the *module* passes and then de-init the *function* passes. Do the module passes not need their own init? Do the function passes not need to be run at all? In which case, why init/de-init them at all ?

The previous code had the module passes running outside the function passes init/de-init, which makes me think the two are orthogonal to each other, but I don't really know exactly what happens under the hood in LLVM.

lgritz: I benchmarked (our current benchmarks, serial shading), and it's exactly performance neutral. I don't know exactly what's going on underneath, but they are clearly equivalent in results.

I'm assuming that @AlexMWells either felt like this was a slight simplification of this code, or else he had some particular example (maybe only after other material is added related to the batch shading?) where this was a certain improvement.

AlexMWells: If I remember correctly, when your module is optimized, it runs optimization passes against everything in it including all the functions.  The loop over functions separately repeats the same work.  So why do the work twice.  Admittedly, the 2nd time through the functions(@ the module level) there would be less work to do.  However looking forward to better global optimization passes (vs. function level) combined with desire to reduce optimization/compile time I thought the 2nd pass through functions independently should be avoided.  However I left the code here in case this is the wrong assumption.  Perhaps other LLVM/Clang aficionados can comment on the benefit of running the optimization passes at the function level then again at the module level (which turns around and runs them over the functions again).  However I measured no tangible benefits in my testing.  Because of the new improved pruning we added (that happens before any optimization passes) the measured benefit of this change may be minimized, but that doesn't make it not meaningful.

fpsunflower: Makes sense. Should we try to distill that rationale into a comment at least? When just reading through the code - it looks like it might just be a mis-use of the API instead of being intentional.

The previous code did:
```cpp
m_llvm_func_passes->doInitialization();
...
m_llvm_func_passes->doFinalization();	    
m_llvm_module_passes->run (*m_llvm_module);
```

Is there any reason to believe running the module passes "during" the function pass init/finalize is better? Or said another way, why not just comment out the loop over functions ? Does the re-ordering help? Why don't the module passes need their own `doInitialization`/`doFinalization` calls?

AlexMWells: Sure, drop the new part, and comment out the for loop with comment that the module pass will optimize all functions in it.  However I do not know (without research) the effect of  
m_llvm_func_passes->doInitialization();
m_llvm_func_passes->doFinalization();
and would tend to say that we shouldn't finalize until after the module passes has run (since it optimizes the functions).

fpsunflower: I'm also wondering -- since we aren't actually running the function passes, do we need them at all? Or does important work happen in the init/finalize steps?

Would it be enough to just populate the module pass manager and not even bother making a function pass manager?

fpsunflower: It looks like this works - I guess we'd need to run some profiles to see how it impacts performance (if at all).

AlexMWells: As we have some uncertainty around this, and the pruning system probably achieves the main benefit anyway, and we would need to investigate deeper into LLVM to confirm our expectations, I request we table this particular PR for now.

fpsunflower: The code change is small -- I can kick off a production benchmark just to see if there is any observable change from removing the function pass manager entirely.

lgritz: My benchmarks saw no benefit or time improvement at all from this patch (neither for shader execution, nor for JIT time). If Alex doesn't recall any specific gain for the batch shading case, maybe as he suggests, we should just abandon this PR for now and be confident that the old code is correct?

AlexMWells: Correct, I'm sure my initial analysis that whatever was in fpm was running on functions that were later dropped by mpm was wastefu and the reason for this code changedl.  I did measure an improvement in optimization time when I made this change.  However due to the new pruning, it is probably unnecessary.  If Larry didn't find any improvement, we can abaondon this we can concentrate on other PR's.

lgritz: OK, let's just put this PR on hold and move on. It's a small change and we can always re-submit this or a similar PR down the road if we discover any advantage or decide there's a good reason it's a better idiom.

lgritz:
Suspending this PR, as there is no empirical performance benefit, and lingering doubts about whether or exactly why this would be a better idiom. It's easy to re-open or re-submit in the future if we decide this is important.

src/liboslexec/llvm_util.cpp
fpsunflower: On the first few benchmark scenes I'm seeing the no_fpm version takes ~2% *longer* to JIT. Rendering time seems about the same. I'll let it run just out of curiosity, but I agree that this is probably not worth touching for now.

PR #1226, opened 2020-07-24
Split the opt and jit functionality of optimize_group

lgritz:
Work by Alex Wells.

Separate OSL optimization from JIT by adding optional parameter `bool do_jit = true` to all optimize_group* methods. This  will allow optimization without JIT'ing for the host, or optimizing once but JIT'ing for multiple back ends. It will be necessary if one will be JIT'ing only for batched execution, we  would not want to pay the cost of JIT'ing a scalar version as well (but still could if we wanted both scalar and batched versions simultaneously).


lgritz:
@fpsunflower I think we're playing it by ear for now. I wanted to keep this as one call rather than split into 2 for the sake of compatibility. But maybe we'll end up splitting when we understand the use cases better.

PR #1228, opened 2020-07-25
Tweak search for pugixml

lgritz:
Stay ahead of an anticipated change on the OIIO side, I want to move
these headers into a subdirectory.


lgritz:
Withdrawing. I am going for a different approach entirely.

PR #1234, opened 2020-07-30
Changes for LLVM-11.

marsupial:
For the most part, standard innocuous LLVM-style API changes, but:

`llvm::ConstantVector` constructor has changed, and while looking into how to migrate I came across `llvm::ConstantDataVector` which according to the docs uses less memory.  The only difference being that `llvm::ConstantVector` has a `llvm::Value` for each element, but it doesn't look like those individual `llvm::Value`s are used/needed ?


lgritz:
This seems to pass CI, which tests all of llvm 7, 8, 9, 10, so appears safe. Merging.

PR #1236, opened 2020-08-01
Proposal for debugging of invalid globals

etheory:
## Description

This PR is a suggestion to add something for debugging of invalid globals being added to shaders.

Currently, whenever an unknown global is encountered, the following is printed to the shell:
`llvm_global_symbol_ptr: Assertion 'sg_index >= 0' failed.`
and the entire process fails, which I think is probably too extreme.

It would be good to at least be able to identify when a global is invalid, and print some useful debug information to be able to track this down.

I doubt you'd want to put this in verbatim, however, consider it a simple identification of the issue, and a potential example of something that you might want to do.

## Tests

No, since this is just a suggestion.

## Checklist:

### DIFF
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -238,6 +238,8 @@ BackendLLVM::llvm_global_symbol_ptr (ustring name)
     // we use the name of the global to find the index of the field within
     // the ShaderGlobals struct.
     int sg_index = ShaderGlobalNameToIndex (name);
+    if (sg_index < 0)
+        std::cerr << "shader global: " << name << " could not be found...." << std::endl;
     OSL_ASSERT (sg_index >= 0);
     return ll.void_ptr (ll.GEP (sg_ptr(), 0, sg_index));
 }


src/liboslexec/backendllvm.cpp
```diff
@@ -238,6 +238,8 @@ BackendLLVM::llvm_global_symbol_ptr (ustring name)
     // we use the name of the global to find the index of the field within
     // the ShaderGlobals struct.
     int sg_index = ShaderGlobalNameToIndex (name);
+    if (sg_index < 0)
+        std::cerr << "shader global: " << name << " could not be found...." << std::endl;
```
lgritz, line 242: This should be done via a call to shadingsys().errorf(), not raw output to cerr.

etheory:
Hi @lgritz , I'll admit that it is an esoteric issue that possibly only we'll ever have, but the issue was due to have .oso files compiled with a customized version of OSL, being loaded with the standard version of OSL.

These .oso files were compiled with a version of OSL that had extra global definitions.

As a result, I think maybe I'll just close this ticket, since I was just wanted to add this check for debugging purposes, but I feel like this probably won't ever be useful to anyone else.

Thanks for taking a look.

I'll instead focus on proposing a mechanism to customize the OSL ShaderGlobals instead, which will, in a round about sort of way, address the reason for this issue in the first place.

lgritz:
There was already an assert there... was the problem just that its error messages didn't print enough information (like the name of the global)? With your patch, it does print a message, but then it continues on its merry way and it's not clear to me what the downstream consequences are.

I think that it would be a mistake to have studios customize oslc, subtly changing the language itself and what globals there were. (Chris in particular probably slept badly last night, he's been arguing for years that there shouldn't be any privileged "globals" at all. :-) 

Let's close this, I think we can grope a way toward a better solution for everybody.

fpsunflower:
Just for the record, I think its mainly the *output* globals that I think are not really necessary. Having globals for `P`, `N`, `u`, `v`, etc .. still makes sense to me.

Like we talked about a few meetings ago, I think it would be good to have a more flexible `ShaderGlobals` struct layout (including allowing some globals to be aliased to zero memory if you aren't planning to use them).

@etheory I'm curious about your use case for custom globals. Can you share examples of the kinds of things you had to add as globals?

etheory:
> There was already an assert there... was the problem just that its error messages didn't print enough information (like the name of the global)?

That's correct yes, it was impossible to figure out what was happening without changing the OSL code itself.

> I think that it would be a mistake to have studios customize oslc, subtly changing the language itself and what globals there were. (Chris in particular probably slept badly last night, he's been arguing for years that there shouldn't be any privileged "globals" at all. :-)

I totally agree, which is why I backed out after submitting this :smiley:.

> Let's close this, I think we can grope a way toward a better solution for everybody.

I completely agree, I'm going to give it some more thought and get back to you.

etheory:
> Like we talked about a few meetings ago, I think it would be good to have a more flexible `ShaderGlobals` struct layout (including allowing some globals to be aliased to zero memory if you aren't planning to use them).

I definitely agree with that.

> @etheory I'm curious about your use case for custom globals. Can you share examples of the kinds of things you had to add as globals?

Basically we were using it to pass in things we wanted a "zero-overhead" for, since the get_userdata and ge_attribute mechanism is unbelievably slow in comparison to reading directly from a known memory address.

Our customized variable set was:
```
float pu, pv, pdudx, pdvdx, pdudy, pdvdy
int pFaceId, int threadId
```
- ptex-related, and passed to our own customized ptex implementation through a customized oiio implementation, to replace the poor performance of the standard one (ptex was far too slow for us to use stock-standard). The user could see these, but in reality, I'd rather use a faster binding mechanism than keep these in the globals. It wasn't my design decision, I just have to maintain it for the moment, but I'm looking for alternatives, so these aren't good examples TBHO.

`int primType`
- an enum value for which primitive type we are hitting, i.e. quad, tri, subdiv, curve, point etc.
- can be useful for debugging, but there's probably better ways to pass this info than the globals.

`float curvature, edgeFactor`
- again there would be other ways to obtain information about curvature without using a custom global (i.e. us correctly implementing differentials would be equivalent to this).

`int rayDepth`
- We are still running an old version of OSL, and this wasn't available at the time, so we added this. I know where are probably better ways to pass in this info.

`float smpRoughness`
- A roughness value propagated along the way for automatic caustic-roughening.

`float textureRnd`
- freely define stochastic random number generated by the renderer

`float maskocclusion`
- we precompute the transmission component of all shaders above the current one in our stack (we support layered materials directly within our system) so this gives us a way to quickly know how much light should get to this point. Again, there's far better ways to do this with a redesign.

`int instancehash`
- Scene-consistent hash for objects that we use throughout our pipeline for identification purposes.


Full disclosure - I am not for a second suggesting that these values, or even maintaining them, makes the slightest amount of sense, however, I do have to maintain these purely because I've inherited design choices and code that are easier to keep as they are than change.

However, having said that, I can't help but think the reason that this was done, was to work around the performance of the get_userdata, get_attribute and get_array_attribute RendererServices functionality. The issue is that calling these at high frequency through a DLL-boundary and including all of their conditionals and multiple-levels of call stack (i.e. osl_get_attribute performs a conditional and then calls down to get_attribute and get_array_attribute) and indexing data via a ustring vs an index is a performance bottleneck we simply couldn't work with and had to work around.

I have a completely different idea I'd like to discuss about this which solve all of these issues in one. So I'll close this ticket and move such a discussion elsewhere.

Where should I put such a discussion? Do you want code examples in a ticket? Or do you want me to write the details into an issue?

Thanks for listening!

etheory:
> Just for the record, I think its mainly the _output_ globals that I think are not really necessary. Having globals for `P`, `N`, `u`, `v`, etc .. still makes sense to me.

But for a displacement or normal shader, aren't P and N considered output globals?
Is there a reason they should be considered an exception?
I might have misunderstood your statement, but trying to play devil's advocate.

fpsunflower:
> > Just for the record, I think its mainly the _output_ globals that I think are not really necessary. Having globals for `P`, `N`, `u`, `v`, etc .. still makes sense to me.
> 
> But for a displacement or normal shader, aren't P and N considered output globals?
> Is there a reason they should be considered an exception?

I think mutating globals that are otherwise read-only in other contexts is risky and can lead to confusing behavior when you connect shaders together. I prefer the style where a displacement shader would just have proper output of type `point` that the renderer can pull from. This way if the shader graph is re-wired, there are no side-effects to worry about.

In our shader library we wind up having to maintain tiny "root" nodes that are the only ones in the whole system allowed to write to the globals.

fpsunflower:
That set of globals makes sense. I can definitely picture how you arrived at that list over time.

In the SPI renderer we support many of these via `getattribute` (definitely slower, but an easy place to add things that are rarely used or that you think may change over time).

For cases like ptex information, my guess would be the shader doesn't really need to do anything with the coordinates/faceid. Perhaps we should consider adding a new call like `ptexture(filename)` and let the renderer just "know" how to index things behind the scenes.

`curvature` is an interesting one since OSL doesn't currently support derivatives on `N` (it did in the very beginning, but we never found a way to make those smooth on meshes, so we removed them). There is also the question of which type of curvature you want, etc ... but other than this, I think this one is a good candidate to be part of the native set of globals. There is some precedent for a built-in curvature metric in Renderman and Cycles I believe.

We should probably keep the discussion going in the mailing list. This is a good topic to hash out further.

lgritz:
@etheory since you have spilled most of the details about what you're doing from the off-list thread we had, is it ok if we bring this whole thing into the light by moving the conversation to osl-dev?

I haven't been ignoring your mail -- I was just taking the weekend to try to think carefully about the response, and I think I have some ideas about how to proceed.

etheory:
> @etheory since you have spilled most of the details about what you're doing from the off-list thread we had, is it ok if we bring this whole thing into the light by moving the conversation to osl-dev?

Absolutely.

> I haven't been ignoring your mail -- I was just taking the weekend to try to think carefully about the response, and I think I have some ideas about how to proceed.

No worries at all, I know this is a really tricky thing to design well, so I'm also taking some time to formulate my thoughts into something that is actually fruitful. I'll try and put a set of thoughts together to discuss on the forum, but I see a few specific areas:

1.) Globals - let's define them. What are they exactly, and what criteria do you use to choose an item to add to them?
2.) userdata - what exactly is it? Why is something considered userdata? Why does it seem like, by reading the OSL docs, that there is at least some level of implicit confusing between userdata and attributes? Why are they different? Why do a lot of implementations seem to fall back to calling get_attribute from get_userdata? Does this indicate some other deeper level of redundancy?
3.) attributes - what are they? What is a precise definition? Can they be constants only? Can they be varying primitive data? If so why aren't they defined as userdata?

I have my own thoughts on these things that have changed over time, but honestly, I've just recently read all the osl docs and the documentation in the code, and I don't really think there is a clear and concise explanation of what these things are, and how they should be used, which I think would be the best way to inform a design moving forward. I'll move this to the list later today. 

Thanks guys! Really appreciate your comments.

lgritz:
I'm going to reply on osl-dev

etheory:
Closing this as everything is now very well covered on osl-dev.

PR #1237, opened 2020-08-01
Remove null check from symbol_address for performance

etheory:
## Description

In the case that an implementation designer wants to use the following pattern:

- Define "renderer_outputs" for a group to stop them from being optimized away
- compile the group
- collect the output parameters using ShadingSystem::find_symbol and store them somewhere in the renderer
- at runtime, after calling ShaderGlobals::execute, we call ShadingSystem::symbol_address for each output parameter to obtain our results

This being the case, we already know due to calling find_symbol that the symbols we later call symbol_address with are valid, so the overhead of a conditional for each and every call to symbol_address could be extremely high considering in this particular use case, the conditional will never be false, especially in a tight render loop where we collect a lot of results.

## Tests

TODO: validate that this change doesn't break anything.

## Checklist:

### DIFF
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -387,6 +387,7 @@ const void*
 ShadingSystem::symbol_address (const ShadingContext &ctx,
                                const ShaderSymbol *sym) const
 {
+    OSL_DASSERT(sym != nullptr);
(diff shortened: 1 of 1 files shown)


etheory:
> This seems fine.
> 
> Let's add an
> 
> ```
> OSL_DASSERT(sym != nullptr);
> ```

I've now done this, thanks @lgritz for taking a look!

PR #1242, opened 2020-08-14
Fix fmt vs Cuda incompatibility

lgritz:
We used to disable fmt and fall back to tinyformat, when using Cuda.
The new OIIO 2.2 removes tinyformat and the OIIO_USE_FMT preprocessor
symbol. So make sure that we also tell fmt to avoid using the int128
type, not supported by Cuda. That seems to be enough to let it use
the rest of fmt without trouble.


lgritz:
@marsupial I'm shocked you haven't run into this build break!

marsupial:
I have, just hadn't found the time to do anything but a crazy work-around not worth submitting.
This looks good to me.

PR #1248, opened 2020-09-03
LLVM_Util: add support for masked operations

lgritz:
Work by Alex Wells.

This will be used extensively by the SIMD batched shading.

* Added support for masked operations by managing a stack of masks
  that were created by conditionals or control flow. Detailed
  explanation of the Masking Strategy in llvm_util.h.

* Added support for logically masked functions (handling of masked
  return from inside OSL functions)

* Added support for masked loops (handling of masked break and or
  masked continue).

* Added support for masked exit (handling of masked exit from entire
  shader).

* Added support for management of when masking is required with a
  ScopedMasking object to allow nesting during code gen.

* op_store (and other store variations) respect the current masking
  requirement and current top of the mask stack (not just the top
  entry, but a combined view of the entire mask stack).

* Added support for results of conditional operations to remain
  boolean.  Most masking operations are expecting vector of bool
  values, not a vector of integers., e.g. when vector width is 16, for
  scalar or <16 x i1> for wide. When a conditional result is used by
  another operation a numerical value it is automatically converted to
  an integer.

* Added `llvm::Type *type_native_mask() const;` because different
  ISA's may have different representations of a mask from llvm's
  vector of bits that comparison operations emit.  And we need varying
  boolean symbols to have the correct data type (size) on the stack
  and in data structures.

* Added gather and scatter operations.

* Added helpers to convert from bool<->int, native_mask<->llvm_mask

### DIFF
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -34,6 +34,7 @@ namespace llvm {
   class LLVMContext;
   class Module;
   class PointerType;
+  class StringRef;
   class Type;
   class Value;
   class VectorType;
@@ -188,6 +189,8 @@ public:
     /// Create a new JITing ExecutionEngine and make it the current one.
     /// Return a pointer to the new engine.  If err is not NULL, put any
     /// errors there.
+    /// Optionally enable debugging symbols (source file & line number)
+    /// Optionally enable profiling events
     /// Optionally request a specific ISA for JIT on the host
     ///     ["x64", "SSE4.2", "AVX", "AVX2", "AVX512"]
     ///     (ignored if requested ISA not valid for host)
@@ -230,6 +233,10 @@ public:
     bool detect_cpu_features(TargetISA requestedISA = TargetISA::UNKNOWN,
                              bool no_fma = false);
 
+    void dump_struct_data_layout(llvm::Type *Ty);
+    void validate_struct_data_layout(llvm::Type *Ty, const std::vector<unsigned int> & expected_offset_by_index);
+    
+    
     /// Return a pointer to the current ExecutionEngine.  Create a JITing
     /// ExecutionEngine if one isn't already set up.
     llvm::ExecutionEngine *execengine () {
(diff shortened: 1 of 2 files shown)


AlexMWells (approved):
LGTM

src/include/OSL/llvm_util.h
```diff
@@ -49,7 +50,7 @@ namespace llvm {
 OSL_NAMESPACE_ENTER
 
 namespace pvt {   // OSL::pvt
-
+class ShaderInstance;
```
fpsunflower, line 53: Is this forward declaration used? I thought `llvm_util.h` was trying to stay mostly agnostic about OSL specifics.

src/include/OSL/llvm_util.h
```diff
@@ -306,6 +309,189 @@ class OSLEXECPUBLIC LLVM_Util {
     /// Are we inside a function?
     bool inside_function() const;
 
+    // Overview of masking strategy
+    //
+    // For if/then based on a conditional, the result of the conditional is
+    // conditionals written to a symbol that is a wide boolean.  To handle
+    // nested we will keep a stack of conditional masks.  When a new
```
fpsunflower, line 316: typo: `nesting`

src/include/OSL/llvm_util.h
```diff
@@ -306,6 +309,189 @@ class OSLEXECPUBLIC LLVM_Util {
...
+    // conditional scope is exited, we just pop the top of the conditional
+    // mask stack.
+    //
+    // Loops are handled by pushing their conditional result onto the
+    // conditional mask stack, and popping it back off at the end of the
+    // loop.
+    //
+    // Handling of early outs (exit, return, break, continue) is done by
+    // storing a mask representing the lanes which are going to be
+    // deactivated (or continue executing) to an allocated location the
```
fpsunflower, line 329: typo: `in the stack`

fpsunflower (approved):
LGTM, just a few minor notes below

lgritz:
Quite right on the typos and other comments, @fpsunflower . Pushed an update and will merge after all the CI tests pass.

src/include/OSL/llvm_util.h
AlexMWells: Doesn't appear its used.

AlexMWells (approved):

PR #1250, opened 2020-09-16
Custom opt levels and passes

lgritz:
Work of Alex Wells, Intel.

Added custom optimization passes to improve AVX/AVX2 code generation
which are automatically added when targeting the host:

  * PrePromoteLogicalOpsOnBitMasks
  * PreventBitMasksFromBeingLiveinsToBasicBlocks

Added support for experimental llvm_optimize: 10,11,12,13.  Optlevels
10, 11, 12, 13 explicitly create optimization passes.  They are
stripped down versions of clang's -O0, -O1, -O2, -O3.  They try to
provide similar results with improved optimization time by removing
some expensive passes that were repeated many times and omitting other
passes that are not applicable or not profitable.  Useful for
debugging, optlevel 10 adds next to no additional passes.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -164,7 +164,7 @@ public:
     ///         opt_fold_getattribute, opt_middleman, opt_texture_handle
     ///         opt_seed_bblock_aliases
     ///    int opt_passes         Number of optimization passes per layer (10)
-    ///    int llvm_optimize      Which of several LLVM optimize strategies (0)
+    ///    int llvm_optimize      Which of several LLVM optimize strategies (1)
     ///    int llvm_debug         Set LLVM extra debug level (0)
     ///    int llvm_debug_layers  Extra printfs upon entering and leaving
     ///                              layer functions.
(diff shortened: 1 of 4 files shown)


AlexMWells (commented):
LGTM

lgritz:
Updated after spending some time chasing a bug I saw only on my Mac laptop (but, curiously, not on the CI Mac test), which was a crash and LLVM error messages about the custom passes being added twice. This took me a long time to track down and basically came down to the file-level static initializers in llvm_util.cpp being run twice in some circumstances. I still don't understand why, but I solved the problem by moving those llvm::RegisterPass from file level statics to statics inside LLVM_Util::SetupLLVM(), which definitely only run once.

Also, just to note here, that one way I altered Alex's code was in testshade, instead of many different options to control the llvm optimization level (-llvm_O1, -llvm_O....), I made just a single `--llvm_opt` taking an integer argument (also makes things easier if we make more opt levels later).

lgritz:
When this passes the CI tests, I will finish the merge.

PR #1251, opened 2020-09-19
Fix derivatives of texture calls when only Dalpha is needed

lgritz:
Luke Emrose noticed that for texture calls that retrieved alpha, if
derivatives were later taken *only* of the alpha but not the main
color result of the texture call, those derivs would always end up 0,
the texture call not noticing that it needed to call the version of
texture that returns derivatives.

This patch provides a fix. It was a simple matter of it testing only
whether main result derivs were needed, when determining which variety
of texture must be called.

Curiously, this was already done correctly for texture3d!

Added testsuite/texture-alpha-derivs to test this specific situation.


etheory:
Thanks @lgritz I really appreciate your time to confirm this for us.
I'll back port this to confirm it's working for us, then let you know.
Regards,
Luke

etheory:
@lgritz I've patched this into our internal version and it's working perfectly and has fixed the issue identified by one of our artists. Thank you so much! LGTM. Ship it! :smile:

PR #1253, opened 2020-09-22
Optimization: Skip copying of output params to used layers.

lgritz:
Work by Alex Wells.

When generating the code to copy one layer's outputs to a downstream
layer's inputs, we can skip the copy entirely if we already know that
the downstream layer will not be run.


AlexMWells (approved):
LGTM

PR #1254, opened 2020-09-23
Explicitly set the OptiX pipeline stack size.

timgrant:
## Description

`testrender` and `testshade` were both relying on the automatic pipeline stack size computation in OptiX 7. This was causing a handful of tests to fail for reasons that are still not clear. Computing and manually setting the required stack size clears up the failures.

## Tests

I didn't add any tests, but all existing tests pass with this change.

## Checklist:

### DIFF
--- a/src/testrender/optixraytracer.cpp
+++ b/src/testrender/optixraytracer.cpp
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 // https://github.com/imageworks/OpenShadingLanguage
 
+#include <vector>
 
 #include <OpenImageIO/filesystem.h>
 #include <OpenImageIO/sysutil.h>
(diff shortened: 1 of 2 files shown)


cmstein:
I've tried out this patch locally and it fixes the issues I was seeing with recent drivers.

LGTM!

src/testrender/optixraytracer.cpp
```diff
@@ -527,20 +529,18 @@ OptixRaytracer::make_optix_materials ()
     size_t sizeof_msg_log;
 
     // Make module that contains programs we'll use in this scene
-    OptixModuleCompileOptions module_compile_options;
+    OptixModuleCompileOptions module_compile_options = {};
```
lgritz, line 532: Can I ask... what is going on with this line?

timgrant: Just making sure the options are zero-initialized.

PR #1255, opened 2020-09-24
Fix osoparse & osolex from two independent liboslexec.so

marsupial:
## Description
When two OSL distributions are loaded, they will both use the same `osoparse` and `osolex`functions.

Problematic because the OSOReader::parse_XXX function from _liboslexec_A.so_ will set the static variable `osoreader = this`, but then call `osoparse` in _liboslexec_B.so_, where the `osoreader` static is still NULL.

https://github.com/marsupial/OpenShadingLanguage/blob/62797a9032a73162a62e19d03261d7d8bd3a8ded/src/liboslexec/osolex.l#L263
https://github.com/marsupial/OpenShadingLanguage/blob/62797a9032a73162a62e19d03261d7d8bd3a8ded/src/liboslexec/osolex.l#L287


lgritz (approved):

lgritz:
I'll merge as soon as CI is finished.

PR #1256, opened 2020-09-26
Raise OIIO dependency minimum to 2.1 going forward

lgritz:
This patch drops support for OIIO 2.0, making 2.1 the minimum as we
move forward on the way to OSL 1.12 (or 2.0?). The current public
supported release of OIIO is 2.2, so this means that we are still
supporting both the current and the previous year's OIIO major
releases, as is our usual policy (that is, for an OSL that won't
be an official release until some time in 2021).

This patch does not change the supported OIIO releases of prior OSL
releases. In particular, the current release family, OSL 1.11, will
continue to support OIIO 2.0. We never change minimum dependencies
during the course of a release series.

One way this becomes a big simplification is that OIIO 2.1 started
exporting cmake configurations directly, and thus we no longer need
our copy of FindOpenImageIO.cmake, and can jettison some old
conventions that it used for older OIIO releases.


lgritz:
Would this inconvenience anyone if we bumped OIIO minimum to 2.1 moving forward (master only)?

lgritz:
This was discussed at the last TSC meeting without objection, so I am merging.

PR #1257, opened 2020-09-26
Fix typo in version string

lgritz:
Extra char prevented the "release type" from being amended onto the
version string.

Also, change the OSL_INTRO_STRING to use the preferred convention
of "Open Shading Language" rather than "OpenShadingLanguage".


lgritz:
Anybody care to approve this?

fpsunflower (approved):

PR #1258, opened 2020-09-26
Print more build/runtime info in stats and help message

lgritz:
testshade --help or --runstats now print additional info about the
dependency versions, hardware capabilities enabled at build time, and
hardware capabilities detected at runtime. This should help debugging
of user experiences as an easy way to find out what OSL thinks it's
working with in terms of its environment.


lgritz:
Any comments?

fpsunflower (approved):

PR #1260, opened 2020-10-05
Use 'final' liberally

lgritz:
This is good for optimization, allowing "de-virtualization" of function
calls in cases when the compiler can be sure that there can be no further
subclassing.


aconty (approved):
LGTM!

PR #1262, opened 2020-10-05
pugixml build fixes

lgritz:
In OIIO land, some users pointed out deficiencies in our build
scripts and how they find (or fail to find) pugixml on some platforms and
circumstances. This is an OSL side copy of the approach that seems to
fix things.

* Bring back the idea of a Findpugixml.cmake file because it turns out
  that (for pugixml <= 1.10) the exported config files are not reliable.

* Fix CI scripts to build PugiXML (with a few different versions). We
  previously were relying on the embedded version for all CI tests, which
  masked problems associated with using an external PugiXML.


aconty (approved):
LGTM

PR #1266, opened 2020-10-12
Bug fix: is_nonzero was only considering the first element

lgritz:
No symptoms ever reported, but surely this is a typo bug.


AlexMWells (approved):
LGTM

PR #1267, opened 2020-10-13
Minor fixes/enhancements to TypeSpec & Symbol (OSL internal classes)

lgritz:
* Simplify implementation of TypeSpec::is_triple, more efficient.

* Add TypeSpec::is_triple_based(), analogous to all the other
  is_X/is_X_based.

* Bug fix: Symbol::get_string()'s assertion was checking whether it was
  a string, but should have checked if it was string_based.

* Add Symbol::get_vec3(), analogous to get_int/float/string.

* Add Symbol::coerce_float and coerce_vec3, like get but with some type
  coercions.

* Add Symbol::dataptr() as synonym for data(). This is for "forward
  compatibility", a future PR will make more widespread use of this.

* Lots and lots of places where error-prone and verbose casts are replaced
  with the new methods, and some logic simplified.

### DIFF
--- a/src/include/osl_pvt.h
+++ b/src/include/osl_pvt.h
@@ -310,11 +310,26 @@ public:
     ///
     bool is_triple() const
     {
+        return !is_closure() && m_simple.aggregate == TypeDesc::VEC3
+               && m_simple.basetype == TypeDesc::FLOAT && !m_simple.is_array();
+    }
+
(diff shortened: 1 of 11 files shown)


lgritz:
Does this seem ok to everyone?

aconty (approved):
Yeah, looks better!

PR #1270, opened 2020-10-15
Remove ustring::format references, they will change meaning someday

lgritz:


aconty (approved):
LGTM

PR #1272, opened 2020-10-15
Add interface to  ShadingSystem for Batched Execution 

AlexMWells:
Introduce template<int WidthT> struct BatchedShaderGlobals to store up to WidthT instances of ShaderGlobal data.  The data for a batch is divided into underlying uniform and varying data structures.  Uniform section contains a fixed set of shader global values that are shared between all instances.  Varying has Block<DataT, WidthT> members so each instance can hold a different value.  The Block data structure is specialized to hold the underlying data type in a SIMD friendly Structure of Arrays (SOA) data layout but provides an array subscript interface to a proxy that can import/export the DataT in its orignal data layout.

Introduce template<int WidthT> class ShadingSystem::BatchedExecutor as an interface to execute batches of shader globals through a shader network.  BatchedExecutor provides similar set of execute methods, but they accept a BatchedShaderGlobals and batchSize instead of individual ShaderGlobal(s). This is just the interface, batched backend implemention to follow in subsequent pull requests.

Introduce template<typename DataT, int WidthT> Wide to hold a reference to Block<DataT, WidthT> and provide access to the underlying SOA data layout.  Wide can handle array types and const correctness.  The Wide adapter is lightweight and can be copied by value, used as function parameter.  Future pull requests will add SIMD versions of OSL library functions and Wide will be the primary was to access SOA data.

Once batched execution finishes, the resulting symbol address can be passed to a Wide<DataT, Width> to access the resulting data in a type safe manner.

Updated testshade with command line option "--batched" causing batched execution of 16 or 8 instances through the shader network and to extract the results.  A later pull request will add batched interface to renderer services.

Replaced use of std::bind in testshade in favor of c++11 lambda function.

### DIFF
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -303,6 +303,51 @@ ShadingSystem::execute_layer (ShadingContext &ctx, ShaderGlobals &globals,
     return layernumber >= 0 ? ctx.execute_layer (globals, layernumber) : false;
 }
 
+template<int WidthT>
+bool
+ShadingSystem::BatchedExecutor<WidthT>::execute (ShadingContext &ctx, ShaderGroup &group,
+        int batch_size, BatchedShaderGlobals<WidthT> &globals_batch, bool run)
+{
+    OSL_ASSERT(0 && "To Be Implemented");
+    return false;
+}
+
+template<int WidthT>
+bool
+ShadingSystem::BatchedExecutor<WidthT>::execute_init (ShadingContext &ctx, ShaderGroup &group,
+        int batch_size, BatchedShaderGlobals<WidthT> &globals_batch, bool run)
+{
+    OSL_ASSERT(0 && "To Be Implemented");
+    return false;
+}
+
+
+template<int WidthT>
+bool
+ShadingSystem::BatchedExecutor<WidthT>::execute_layer (ShadingContext &ctx, int batch_size, BatchedShaderGlobals<WidthT> &globals_batch,
+                              int layernumber)
+{
+    OSL_ASSERT(0 && "To Be Implemented");
+    return false;
+}
+
+template<int WidthT>
+bool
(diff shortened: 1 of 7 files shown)


lgritz (approved):
I think this is all fine. I'll double check that it doesn't break the build at work (CI looks fine), then merge.

src/osltoy/qtutils.h
```diff
@@ -171,7 +171,7 @@ class DoubleSpinBox final : public QDoubleSpinBox {
         double ss = m_fixed_step;
         if (m_variable_step_digits) {
             ss        = m_variable_min_step;
-            double av = fabs(valsize);
+            double av = ::fabs(valsize);
```
lgritz, line 174: I think this would be more clear as std::fabs

src/liboslexec/shadingsys.cpp
```diff
@@ -394,7 +439,102 @@ ShadingSystem::symbol_address (const ShadingContext &ctx,
     return ctx.symbol_data (*(const Symbol *)sym);
 }
 
-
+bool
+ShadingSystem::supports_batch_execution_at(int width)
+{
+    auto requestedISA = LLVM_Util::lookup_isa_by_name(m_impl->llvm_jit_target().c_str());
```
lgritz, line 445: lookup_isa_by_name takes a string_view, so the `.c_str()` is unnecessary here.

lgritz:
If you don't object to the minor fix comments I made, I can add those changes in the process of doing the merge (and probably a teensy bit of formatting fixes as well).

src/osltoy/qtutils.h
AlexMWells: agreed, should I update that in PR?

AlexMWells:
Both those changes and any comments you want are fine, go ahead.

lgritz:
Yeah, sure, can you update the PR with those two small changes, then I'll just merge.

lgritz:
merged!

PR #1273, opened 2020-10-17
Update ref output after recent OIIO change

lgritz:
A change in OIIO master eliminated some spurious newlines in certain
error messages.


aconty (commented):
LGTM!

PR #1275, opened 2020-10-21
Remove rendererservices.h and batched_shaderglobals.h from oslexec.h

lgritz:
With just a few minor adjustments, we're able to eliminate these two
headers (and everything else they pull in) from every client of
oslexec.h.  For the purposes of oslexec.h's declarations, all uses are
either pointers or references, so we only need the names
forward-declared, but we don't need the implementations revealed to
everybody including oslexec.h.

In particular, this hides the RendererServices and BatchedShaderGlobals
from a lot of translation units, should cut down on compile time and
reduce exposure of certain semi-internals to only the modules that
need them.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -8,16 +8,13 @@
 
 #include <OSL/oslconfig.h>
 #include <OSL/shaderglobals.h>
-#include <OSL/rendererservices.h>
-
-#include <OSL/batched_shaderglobals.h>
 
 #include <OpenImageIO/refcnt.h>
-#include <OpenImageIO/ustring.h>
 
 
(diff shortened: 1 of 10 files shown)


aconty (approved):
Always good to remove includes. LGTM!

AlexMWells (approved):
LGTM

lgritz:
In particular, the philosophy I'm going for here is that nontrivial classes and templates...

*  ...specific to implementing renderer internals should only be needed by parts that implement the renderer, not the parts that just need to make calls to execute a shader.

* ...specific to batch SIMD shading should only be needed by renderers that call the batch shading interface.

* ...specific to GPU should only be needed by renderers that use GPU.

PR #1276, opened 2020-10-21
Pr/batched rs

AlexMWells:
Add Structure of Array data layout wrappers, see <OSL/wide.h> for detailed documentation and implementation:

	Wide<T> contains a T value for WidthT lanes and is accessed
	with [lane] where lane in [0-(WidthT-1)]

	Masked<T> bundles a T value for WidthT lanes with a bitmask representing
	the active lanes.  So you don't see a Mask or other representation of active
	lanes passed separately, it should be bundled with the data it applies to
	
	MaskedData bundles any type of value for WidthT lanes with a bitmask
	representing the active lanes.  MaskData can be invalid, may have derivs,
	and underlying data type can be queried or tested.  IE:
    	 void myFunction(MaskedData<16> any) {
    	     if (Masked<Vec3>::is(any) {
    	         Masked<Vec3> vecVal(any);
	MaskedDx<T> and MaskedDy<T> provide access to derivatives of MaskedData

	RefData bundles any type of value.  RefData can be invalid, may have derivs,
	and underlying data type can be queried or tested.  IE:
    	void myFunction(RefData any) {
    	    if (Ref<int[2]>::is(any) {
    	        Ref<int[2]> pairVal(any);
	RefDx<T> and RefDy<T> provide access to derivatives of RefData

Add BatchedRendererServices<int WidthT> interface for renderers to support batched execution:
  It mimics RendererServices but its methods parameters are Wide<DataT> and Masked<DataT> parameters where underlying DataT is stored in Structure Of Arrays (SOA) layout with WidthT entries.  Concrete BatchedRendererServices can implement the methods by looping over WidthT, extract per lane values from Wide & Masked parameters with val = wide_param[i], use the value to calculate/lookup a result, then assign results to Masked parameters with masked_param[i] = result.  Some BatchedRendererServices methods have a "_uniform" version which can be called which code generator has proved that the parameters cannot be varying.  Additional virtual methods exist to identify if a particular method is overriden, required to be able to generate calls to target specific optimized default implementation vs. calling a virtual function.

Add BatchedTextureOptions<int WidthT> which is binary compatible with OIIO::TextureOptBatch
Add BatchedTextureOutputs<int WidthT> which bundles result *, alpha *, errormessage*, bool resultHasDerivs, bool alphaHasDerivs, and a Mask<WidthT> into a single object which can produce correct MaskData for result(), alpha(), and message() all sharing a single Mask instance.

Workaround some QT "foreach" macro issues.  It is recommended any uses of QT's foreach be migrated to use C++11 range based loops.


src/include/OSL/batched_texture.h
```diff
@@ -0,0 +1,278 @@
...
+static_assert(offsetof(OIIO::TextureOptBatch, tblur) % 64 == 0,
+              "oops unaligned wide variable");
+static_assert(offsetof(OIIO::TextureOptBatch, rblur) % 64 == 0,
+              "oops unaligned wide variable");
+static_assert(offsetof(OIIO::TextureOptBatch, swidth) % 64 == 0,
+              "oops unaligned wide variable");
+static_assert(offsetof(OIIO::TextureOptBatch, twidth) % 64 == 0,
+              "oops unaligned wide variable");
+static_assert(offsetof(OIIO::TextureOptBatch, rwidth) % 64 == 0,
+              "oops unaligned wide variable");
```
lgritz, line 119: I don't even want to change it in this PR, but just food for future thought: Would it be better to tuck away these static checks in just one .cpp module (any one will do, right?) rather than have them parsed and evaluated for every translation unit that includes the header?

src/include/OSL/mask.h
```diff
@@ -0,0 +1,370 @@
...
+
+    // count number of active bits
+    OSL_FORCEINLINE int count() const
+    {
+#if __INTEL_COMPILER
+        if (value_width <= 32)
+            return _mm_popcnt_u32(m_value);
+        else
+            return _mm_popcnt_u64(m_value);
+#else
```
lgritz, line 133: Should we have a clause for gcc and clang to use the __builtin_popcount ?

src/include/OSL/mask.h
```diff
@@ -0,0 +1,370 @@
...
+            return __builtin_ctz(m_value);
+#endif
+        } else if (value_width <= 64) {
+#if (defined(_WIN32) || defined(_WIN64))
+            unsigned long index;
+            _BitScanForward64(&index, m_value);
+            return static_cast<int>(index);
+#else  // Linux, OS X, gcc or clang
+            return __builtin_ctzll(m_value);
+#endif
```
lgritz, line 165: Let's change this to an `#if defined(__GNUC__) || defined(__clang__)`, then have a final `#else` clause with an appropriate `#error` inside, just in case anybody tries to compile this with some unforeseen compiler.

src/include/OSL/wide.h
```diff
@@ -1335,565 +1391,1553 @@ struct alignas(VecReg<WidthT>) Block<Dual2<Color3>, WidthT>
...
+};
+
+
+}  // namespace pvt
+
+template<typename DataT, int WidthT>
+struct Masked : public pvt::MaskedImpl<DataT, WidthT> {
+    static_assert(std::is_const<DataT>::value == false,
+                  "Masked<> is only valid when DataT is NOT const");
+
```
lgritz, line 2503: I think that for this and other templates, maybe the overview of what they do and why should be here in doxygen-like comments? I know I saw you explain it elsewhere, but I think probably each of these classes should have their primary documentation right at their declaration.

src/osltoy/osltoyapp.cpp
```diff
@@ -33,6 +33,11 @@
 #include <QTextEdit>
 #include <QToolTip>
 
+// QT's extension foreach defines a foreach macro which interferes
+// with an OSL internal foreach method.  So we will undefine it here
+#undef foreach
+// It is recommended any uses ofQT's foreach be migrated
```
lgritz, line 39: ofQT -> of Qt

src/osltoy/osltoymain.cpp
```diff
@@ -7,6 +7,12 @@
 
 #include <QApplication>
 
+// QT's extension foreach defines a foreach macro which interferes
+// with an OSL internal foreach method.  So we will undefine it here
+#undef foreach
+// It is recommended any uses ofQT's foreach be migrated
```
lgritz, line 13: ofQT -> of Qt

lgritz (approved):
LGTM.

I've made several minor comments to consider, which can be tackled now, or in a subsequent review, or not at all. Just let me know when you want me to merge this one.

src/include/OSL/batched_texture.h
AlexMWells: Depends on one's confidence level, they could be moved into some type of unit test that builds and enforces this.  But that moves intent further away from enforcement.  Do we think there is a compile time cost (measurable, I'm sure its not free).  Honestly the motivation for moving it out would just be so reader doesn't have to see it there.

src/include/OSL/mask.h
AlexMWells: Feel free to add that, I didn't have all the gcc test matrix setup locally, does clang have a similar builtin?  However if gcc and clang recognize the idiom, that might be more maintainable.  Will have to test in godbolt.org

src/include/OSL/mask.h
AlexMWells: sounds reasonable

src/osltoy/osltoyapp.cpp
AlexMWells: ok

src/osltoy/osltoymain.cpp
AlexMWells: ok

src/include/OSL/mask.h
AlexMWells: How's this:
```
#if (defined(_WIN32) || defined(_WIN64))
            unsigned long index;
            _BitScanForward64(&index, m_value);
#elif defined(__GNUC__) || defined(__clang__)
            return __builtin_ctzll(m_value);
#else
#error unknown compiler, update Mask<int WidthT>::first_on() to support compiler
#endif
```

src/include/OSL/wide.h
AlexMWells: At the top there is a "overview" with documentation by each class's forward declaration, I could repeat (maybe more detailed) the specific portion above each class, or move it down above each class..

src/include/OSL/mask.h
lgritz: looks good

src/include/OSL/wide.h
lgritz: Yeah, I think maybe just move the explanations at the top to be near each class so that if we want to generate automatic docs for them later, it will all just work. I think people mostly expect to find the main documentation for a class at the point of its declaration.

This isn't really critical, we can move it around later, but I'm just wanting to be very careful to leave all the right breadcrumbs for whoever needs to modify or use these classes next.

src/include/OSL/mask.h
AlexMWells: Verified idiom is not detected, correct code though.  I'll try adding case for clang, gcc, then else to the idiom

lgritz: Looks like `__builtin_popcount` works with both gcc and clang, and ought to be able to be counted on to do the best thing on any architecture that those compilers support.

AlexMWells: I stand corrected, gcc and clang both recognized the idiom when targeting an ISA that supports popcount instruction.  I'll update it to be compiler intrinsics anyway, with a comment to use std::bitcount for c++20 in the future.

lgritz: Please include a comment

    // FIXME(C++20)

as a breadcrumb that will let us find those spots when it comes time to fix things up nicely for C++20.

AlexMWells:
I just updated the files incorporating the feedback so far.

lgritz:
All LGTM except for the one comment about __builtin_popcount vs __builtin_popcountll.

Don't worry about the Mac failure, that's unrelated to your changes and is something I'm tracking down separately.

AlexMWells:
__builtin_popcountll(m_value); added, so should be good to go

src/include/OSL/mask.h
madoodia: I tried to build OSL in Windows, from latest changes
this line broke the build
`OpenShadingLanguage\src\include\OSL/mask.h(166,20): error C3861: '__builtin_ctz': identifier not found `

lgritz: Oh yes, I see. Hang on, I will propose a fix...

madoodia: thanks @lgritz 
I can test it on Windows, any changes you apply, I am preparing Docs for building in windows, so can count on me on any OS based changes on the repo

lgritz: See #1291

PR #1277, opened 2020-10-23
Work around GHCI Mac problem with homebrew openssl

lgritz:


lgritz:
Sorry for the weird boilerplate. Was testing PR submission from the command line for the first time!
(But this is a real PR)

src/build-scripts/install_homebrew_deps.bash
```diff
@@ -20,7 +20,8 @@ if [[ `which brew` == "" ]] ; then
 fi
 
 
-brew update >/dev/null
+brew uninstall openssl
+#brew update >/dev/null
```
aconty, line 24: Why not updating anymore?

lgritz: The update re-downloads a broken package and the whole thing fails.

We'll go back to updating when the broken packages are fixed. It's a known problem affecting many projects.

aconty (approved):

PR #1279, opened 2020-10-23
Use cmake find_package to find LLVM rather than a custom setup

debaetsd:
## Description

This PR changes how CMake searches for LLVM.
Rather then having a custom FindLLVM setup, it is possible to rely on importable targets setup by the LLVM project.

This was proposed/discussed in `LLVM importable CMake targets` on the mailing list (initial post on Oct 16 2020).

Please note that I renamed the clang-format target into osl-clang-format to prevent name clashes with the "real" clang-format.

## Tests

Existing CI still passes.
Windows build verified locally.
Updated CI to look for `osl-clang-format`

## Checklist:


src/cmake/compiler.cmake
```diff
@@ -469,7 +469,7 @@ if (CLANG_FORMAT_EXE)
     # message (STATUS "clang-format file list: ${FILES_TO_FORMAT}")
     file (COPY ${CMAKE_CURRENT_SOURCE_DIR}/.clang-format
           DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
-    add_custom_target (clang-format
+    add_custom_target (osl-clang-format
```
lgritz, line 472: If you don't mind, let's call this "run-clang-format".

I tend to copy and/or synchronize the more generic cmake scripts between multiple projects, so it's very handy to not have too many project-specific names littering the code like this that could in theory be applied to any project.

src/cmake/externalpackages.cmake
```diff
@@ -100,19 +100,59 @@ checked_find_package (pugixml 1.8 REQUIRED)
...
+# link conflict with its dependency on the llvm libc++ and the system
+# libc++, both can end up dynamically linked and lead to very subtle and
+# frustrating behavior failures (in particular, osl's use of libclang will
+# botch include file parsing any time LD_LIBRARY_PATH doesn't have the llvm
+# libc++ first).
+#
+# It seems that this is not a problem when linking against the llvm and
+# libclang libraries statically. So on apple and when LLVM 10+ are involved,
+# just force that choice. Other than larger executables, it seems harmless,
+# and in any case a better choice than this beastly bug.
```
lgritz, line 115: Out of curiosity, have you checked that this is still necessary in the new LLVM 11?

src/cmake/externalpackages.cmake
```diff
@@ -100,19 +100,59 @@ checked_find_package (pugixml 1.8 REQUIRED)
...
+                     HINTS ${LLVM_DIRECTORY})
+
+# manual version check because "LLVM is API-compatible only with matching major.minor versions" aka it requires VERSION_EQUAL
+if ( ${LLVM_PACKAGE_VERSION} VERSION_LESS 7 )
+	message(FATAL_ERROR "LLVM 7.0+ is required but ${LLVM_PACKAGE_VERSION} was found")
+endif()
+
+include_directories(BEFORE SYSTEM ${LLVM_INCLUDE_DIRS})
+add_definitions(${LLVM_DEFINITIONS})
+link_directories(${LLVM_LIBRARY_DIRS})
```
lgritz, line 133: Are these still necessary to be declared here and therefore be exposed to the whole of the osl code base? Or are exported targets set up correctly by the LLVM config cmake? In other words, should these be deleted, and instead liboslexec and other components should have `target_link_libaries` say `PRIVATE llvm::foo` for whatever specific components they use?

lgritz (commented):
I like where this is going! It will be great to finally get rid of our FindLLVM.cmake, which I think we wrote in the early LLVM 3 days, long before LLVM contained exported config files. I pose some comments/questions for you to consider.

src/cmake/compiler.cmake
debaetsd: I have zero preference in this so `run-clang-format` is fine

src/cmake/externalpackages.cmake
debaetsd: I have not tested this on Mac (and just relied on the github CI). 
I could set up a build environment on my macbook but it will take some time (I don't even have brew installed :) ). Perhaps somebody else in the OSL community can help test this quicker ?

src/cmake/externalpackages.cmake
debaetsd: Kinda, for the clang libraries this is trivial. For LLVM itself, it is a bit more complicated.
It seems that OSL depends on 2 target independent libs (MCJIT and Passes) and 3 target dependent (Codegen, Disassembler, AsmParser). 
For the independent once, it is trivial to target them directly. 
For the target dependent libs, it is a bit harder as they contain the target in the lib name (for example `LLVMX86CodeGen`).
At first I just took the route of adding them explicit, but later reverted to the "target all what the installed llvm version supports".
That is why I use the `llvm_map_components_to_libnames` in order to support arbitrary targets. 
So I guess the question is, does OSL support anything other than x86 and NVPTX ? 
Ff not, it is trivial to directly target those components.
If we want to support all possible targets we needs some additional setup (though we could probably limit it to oslexec). Note that I think it can be further simplified  upon bumping the minimum required version to LLVM8 as then have all the special `allXXX` targets.

lgritz: Currently, OSL doesn't support anything but x86 and NVPTX, but I expect it will need to fairly soon (at the very least, the ARM-based Macs that are coming.

lgritz:
FYI, I'm fiddling around with this to get it totally working with the GPU mode and on my Mac. There are some minor edits I'll need to push to amend it.

debaetsd:
@lgritz  I hacked up a test to use `target_link_libaries ` directly: https://github.com/debaetsd/OpenShadingLanguage/commit/003168887f523f578c5d928233f17ae22ff0d202 All in all pretty clean imho. This will work for any target (installed in the linked LLVM that is). Note though that this has issues with llvm7, either we need to special case it or we drop support for it.

lgritz:
I think we let this languish for a while because it didn't work for llvm 7, which we were still supporting at the time. Since then, we have pulled (in main anyway) the minimum llvm to 9, so we should see if we can revive this.

PR #1281, opened 2020-10-27
Adjust for recent OIIO master change -- ImageInput no longer has m_mutex

lgritz:
While we're in there, get rid of some cruft that only applied to
OIIO 1.x.


lgritz:
Emergency merge without review because it's so obviously broken without this patch.

PR #1283, opened 2020-10-27
Add GPU CI test to build with USE_OPTIX=1

lgritz:
It can't run the tests, because there is no GPU on the GHA
instance. But at least we can tell if a PR will breake the build
itself. Later we will find a way to execute these via ASWF's AWS
instances that are for this purpose.

Includes the following changes:

* Download optix headers when needed. The ASWF docker container we're
  running on already has Cuda.

* Use CUDA_TARGET_ARCH CMake variable instead of hard coding in
  llvm_util.cpp.

* Bump cuda target default to sm_50.

* Make CI failure artifacts include CMakeCache.txt for inspection.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,9 @@ jobs:
         if: failure()
         with:
           name: ${{ github.job }}
-          path: build/*/testsuite/*/*.*
+          path: |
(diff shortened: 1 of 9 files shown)


aconty (approved):
LGTM!

PR #1285, opened 2020-10-31
Fix OSL::assume_aligned for broken MSVS clause

lgritz:
Bad cut and paste made this all wrong. Fix.

Also rename the template's AlignmentT -> Alignment to help reduce
confusion because we usually use the T end to indicate a template
parameter that's a type.

### DIFF
--- a/src/include/OSL/platform.h
+++ b/src/include/OSL/platform.h
@@ -454,7 +454,7 @@ using std::assume_aligned;
 
 // Until we have c++20, we can provide our own version of std::assume_aligned
 // http://open-std.org/JTC1/SC22/WG21/docs/papers/2018/p1007r0.pdf
-template<int AlignmentT, class DataT>
+template<int Alignment, class DataT>
 OSL_NODISCARD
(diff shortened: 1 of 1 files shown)


src/include/OSL/platform.h
```diff
@@ -454,26 +454,26 @@ using std::assume_aligned;
...
 #ifdef __INTEL_COMPILER
-    __assume_aligned(ptr, AlignmentT);
+    __assume_aligned(ptr, Alignment);
     return ptr;
 #elif defined(__clang__) || defined(__GNUC__)
-    return reinterpret_cast<DataT*>(__builtin_assume_aligned(ptr, AlignmentT));
+    return reinterpret_cast<DataT*>(__builtin_assume_aligned(ptr, Alignment));
 #elif defined(_MSC_VER)
-    constexpr std::uintptr_t unaligned_bit_mask = ((1 << N) - 1);
+    constexpr std::uintptr_t unaligned_bit_mask = (Alignment - 1);
```
madoodia, line 470: I just changed this line with `AlignmentT` and it worked fine without any error, and it is built successfully
(y)

PR #1291, opened 2020-11-05
Clean up Mask implementation of first_on

lgritz:
The version of Mask::first_on turned out not to have all the necessary
clauses fleshed out, breaking on Windows. So I refactored to get it
right.

### DIFF
--- a/src/include/OSL/mask.h
+++ b/src/include/OSL/mask.h
@@ -125,9 +125,9 @@ public:
     // count number of active bits
     OSL_FORCEINLINE int count() const
     {
-        // FIXME(C++20)
-        // return std::popcount(m_value);
-#if __INTEL_COMPILER
+#if OSL_CPLUSPLUS_VERSION >= 20
+        return std::popcount(m_value);
+#elif OSL_INTEL_COMPILER
         if (value_width <= 32)
             return _mm_popcnt_u32(m_value);
         else
@@ -156,35 +156,41 @@ public:
     // NOTE: undefined result if no bits are on
     OSL_FORCEINLINE int first_on() const
     {
-        // FIXME(C++20)
-        // return std::countr_zero(m_value);
-
-        if (value_width <= 32) {
-#if __INTEL_COMPILER
+#if OSL_CPLUSPLUS_VERSION >= 20
+        return std::countr_zero(m_value);
+#elif OSL_INTEL_COMPILER
+        if (value_width <= 32)
             return _bit_scan_forward(m_value);
-#else
+        else {
+            unsigned __int32 index;
+            _BitScanForward64(&index, m_value);
+            return static_cast<int>(index);
+        }
+#elif defined(__GNUC__) || defined(__clang__)
+        if (value_width <= 32)
             return __builtin_ctz(m_value);
-#endif
-        } else if (value_width <= 64) {
-#if (defined(_WIN32) || defined(_WIN64))
+        else
+            return __builtin_ctzll(m_value);
+#elif defined(_MSC_VER)
+        if (value_width <= 32) {
+            unsigned long index;
+            _BitScanForward(&index, m_value);
+            return static_cast<int>(index);
+        } else {
             unsigned long index;
             _BitScanForward64(&index, m_value);
             return static_cast<int>(index);
-#elif defined(__GNUC__) || defined(__clang__)
-            return __builtin_ctzll(m_value);
-#else
-#    error unknown compiler, update Mask<int WidthT>::first_on() to support compiler
-#endif
-        } else {
-            // reference implementation, use if ctzl is not available
-            ValueType m(m_value);
-            int lane = 0;
-            while ((m & 1) == 0) {
-                lane++;
-                m >>= 1;
-            }
-            return lane;
         }
+#else
+        // reference implementation, use if ctzl is not available
+        ValueType m(m_value);
+        int lane = 0;
+        while ((m & 1) == 0) {
+            lane++;
+            m >>= 1;
+        }
+        return lane;
+#endif
     }
 
     OSL_FORCEINLINE Mask invert() const


lgritz:
@madoodia Could you test?

@AlexMWells Can you review, please? I'm concerned about line 165. Not sure if the Intel compiler supports `_bit_scan_forward` for 64 bit values or if there was a different function for 64 bits. I'm having a hard time finding any documentation on this online. I think you can see that the old code would not have worked at all for this case.

More generally, I'm worried about whether this entire function is going to behave for 64 bit masks, or if it will be a syntax error on the 32 bit versions. Which function is called by 32 vs 64 is controlled by a normal `if`, not conditional compilation. It works now that masks are <= 32 bits, because it's harmless to pass a 32 bit value to the 64 bit version (will be auto upcast), but the other way around might be an error. 

Do we even foresee 64 bit masks in a near enough future that we should care? Should these functions be simplified to only  handle up to 32 bit wide masks (AFAIK we only require 16 at the moment?) and a static_assert to flag it if we ever try 64 in a distant future?

madoodia:
Of course I can Sir
I'm on it

src/include/OSL/mask.h
```diff
@@ -156,35 +156,38 @@ template<int WidthT> class Mask {
...
-        if (value_width <= 32) {
-#if __INTEL_COMPILER
+#if OSL_CPLUSPLUS_VERSION >= 20
+        return std::countr_zero(m_value);
+#elif OSL_INTEL_COMPILER
+        if (value_width <= 32)
             return _bit_scan_forward(m_value);
-#else
+        else
+            return _bit_scan_forward(m_value);  // Correct?
```
AlexMWells, line 165: probably not correct, that intrinsic only takes 32 bit int.  Looks like:
unsigned char _BitScanForward64 (unsigned __int32* index, unsigned __int64 a) is needed, 
```
            unsigned __int32 index;
            _BitScanForward64(&index, m_value);
```
should do it.

lgritz: Thanks!

lgritz: Wait, I thought _BitScanForward[64] for MSVS? Does that work on Intel, too?

AlexMWells: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=scan&expand=392,393,406,409

I think so

lgritz: Well, that even makes it look like I can collapse the INTEL and MSVC clauses to just use the same calls.

madoodia:
@lgritz 
Just wanted to say I replaced your changes on `mask.h` in my local repo with latest changes from master branch, it is built successfully on Windows.

src/include/OSL/mask.h
AlexMWells: Well, that would mean calling _BitScanForward(&index, m_value); instead of _bit_scan_forward
and in the end maybe not practical difference.  But as defined the intrinsics are different, _BitScanForward checks the incoming value for 0 and returns that, whereas _bit_scan_forward is undefined for 0, but because we don't do anything with the return value of _BitScanForward, the compiler optimizes the extra check away.

So it would work either way.  But this call "first_on" is undefined for 0, that is why I preferred the _bit_scan_forward.  In the end it probably doesn't make a difference one way or another.

AlexMWells:
"Which function is called by 32 vs 64 is controlled by a normal if, not conditional compilation",
`static constexpr int value_width = sizeof(ValueType) * 8;`
```
if (value_width <= 32) {
    ... 
} else {
   ...
}
```
So value_width is known at compile time, at any optimization level > 0, dead code elimination will remove the conditional and leave only one branch, just a bit tidier that have a bunch of template specializations or polymorphic helper functions with different overloads.

AlexMWells:
I suppose all of those 
`if (value_width <= 32) {`
could be changed to 
`if OSL_CONSTEXPR17 (value_width <= 32) {`
to be more clear, the constexpr version of if was added in c++17, and has some teeth if the expression in the if statement is not const expressible.

lgritz:
> So value_width is known at compile time, at any optimization level > 0, dead code elimination will remove the conditional and leave only one branch

Dead code elimination will generate code for only one branch, but both branches will need to be syntax checked, and in the 64 bit case I believe the 32 bit branch will possible have type errors. (The 64 bit branch is still acceptable in the 32 bit case because the parameters will be widened while being passed.)

The "constexpr if" will work for C++17, but since the constexpr will disappear for C++ < 17, we'll still hit the type errors during compilation if somebody tries to build 64 bit masks with a C++14 compiler, say. Now, that hardly matters if we never use 64 bit masks (we're only needing 16 lanes now, right?). But if we don't foresee the need to go beyond 32 lanes in the mask, then we maybe don't need the 64 bit branches at all.

lgritz:
Updated with Alex's advice on how to handle the Intel 64 bit case.

lgritz:
Fixes #1292

AlexMWells:
@lgritz , so you are right a 33-64bit mask when built with -Wconversion will generate a warning for that branch:
```
warning #1682: implicit conversion of a 64-bit integral type to a smaller integral type (potential portability problem)

              return _mm_popcnt_u32(m_value);
```
The codebase doesn't have a 64bit mask, yet.  We can fix this later or I can provide another example this afternoon which will just dispatch to protected type safe helpers: impl_first_on(uint32) and impl_first_on(uint64)

lgritz:
I'm happy to leave it as it is now and fix later. I was more musing about whether we want the extra complexity at all, versus the probability of ever using it.

My understanding -- correct me if I'm wrong -- is that we have 8 wide SIMD for AVX2, 16-wide SIMD for AVX-512, and we're allowing 32-wide because in some circumstances (some workloads? some specific CPU SKUs?) we may find that batches of 32 implemented as two AVX512 instructions for each operation may turn out to be more perforant. But the case for 64 is...? Do we anticipate that being used, or is it just a placeholder for very uncertain future expansion? And if the latter, is it better to build in the complexity for supporting it now, or leaving out 64-wide support and just clearly marking the places that will need to be revisited in a 64-wide SIMD world?

I mean, I know you wouldn't be allowed say even if it is happening, but I've heard no rumors about "AVX-1024" let alone "AVX-2048". Those have to be many years away, if ever.

An alternative is to make functions like this drastically simpler by assuming that a mask is represented by at most 32 bits wide, add a static_assert guaranteeing that and making it obvious where problems need to be addressed, and a short comment that says, "if we ever need to extend this to 64 bit masks, the logic is the same but needs to use the 64-bit wide versions of those intrinsics." In fact... it also may be a nearly sure bet that by the time that 64 bit masks are needed, we might have a C++20 minimum and the whole function would reduce to just a call to std::countr_zero.

Oh, that really makes me think.  Hang on, let me mock up an alternative implementation just to see what you think.

PR #1293, opened 2020-11-06
Build fixes

ix-dcourtois:
## Description

This pull request fixes 1 compiler error when building using MSVC, and a lot of warnings (which makes errors quite hard to see)

Note: I'm waiting for the CLA, in the meantime if you have any comment on the pull request, don't hesitate.

## Checklist:

### DIFF
--- a/src/liboslexec/CMakeLists.txt
+++ b/src/liboslexec/CMakeLists.txt
@@ -111,7 +111,10 @@ macro ( LLVM_COMPILE llvm_src srclist )
           "-I${ILMBASE_INCLUDES}"
           "-isystem ${Boost_INCLUDE_DIRS}"
           -DOSL_COMPILING_TO_BITCODE=1
-          -Wno-deprecated-register -Wno-ignored-attributes -Wno-unknown-attributes
+          -Wno-deprecated-register
+          # the following 2 warnings can be restored when all 3rd parties have fixed their export macros
+          # (dllimport attribute is not supported when compiling for Cuda and triggers a ton of warnings)
+          -Wno-ignored-attributes -Wno-unknown-attributes
           -O3 -fno-math-errno -S -emit-llvm -o ${llvm_asm} ${llvm_src}
       COMMAND "${LLVM_DIRECTORY}/bin/llvm-as" -f -o ${llvm_bc} ${llvm_asm}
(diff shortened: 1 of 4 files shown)


src/liboslexec/CMakeLists.txt
```diff
@@ -111,7 +111,7 @@ macro ( LLVM_COMPILE llvm_src srclist )
           "-I${ILMBASE_INCLUDES}"
           "-isystem ${Boost_INCLUDE_DIRS}"
           -DOSL_COMPILING_TO_BITCODE=1
-          -Wno-deprecated-register
+          -Wno-deprecated-register -Wno-ignored-attributes -Wno-unknown-attributes
```
lgritz, line 114: Rather than suppressing all warnings about unknown or ignored attributes, I wonder if the better fix is to modify export.h so that in the Cuda case, they are not added. That is, change this line (45)

    #    if defined(OSL_STATIC_BUILD) || defined(OSL_STATIC_DEFINE)

by adding `|| defined(__CUDA_ARCH__)`

???

ix-dcourtois: HI, yes you're absolutely right, and that's how I started fixing this warning. But this needs to be done in OIIO and OpenEXR at least (I don't remember exactly how many 3rd parties I had to patch like this before reverting and going the `-Wno-xx` way)
But yep, I should probably revert this, fix OSL's export macro, and submit PRs to 3rd parties.

lgritz: I see, of course. I wasn't thinking about the dependency headers.

Yes, I think that the best thing is:
* make the best changes to OSL's export macros (at the very least, so we can show by example how it should be done)
* submit PRs to the 3rd parties when you have a chance
* keep the warning suppression, but maybe put a comment somewhere nearby in that cmake file saying that the particular reason -Wno-x and -Wno-y are needed is because of those dllimport decorators and that the warnings can be restored once we're sure that any upstream projects have fixed their headers and we are guaranteed to be using at something newer than those fixed versions.

lgritz (approved):
LGTM, merging and will backport to the release branch

lgritz:
@ix-dcourtois Do we have a CLA from you?

ix-dcourtois:
> @ix-dcourtois Do we have a CLA from you?
Not yet, I'm currently checking with my boss (he seems to remember that we once signed one, but cannot find it back)
I'll update the pull request as soon as we find it back (or send a new one)

ix-dcourtois:
I just sent the CLA signed by my boss :)

PR #1294, opened 2020-11-06
Simplify popcount and countr_zero implementations

lgritz:
Rather than have complex per-platform implementations for the 64-bit
mask cases that we don't need for now, just assume that by the time we
need them, we can rely on the simple C++20 calls.

### DIFF
--- a/src/include/OSL/mask.h
+++ b/src/include/OSL/mask.h
@@ -12,6 +12,56 @@
 OSL_NAMESPACE_ENTER
 
 
+// clang-format off
+
+// Define popcount and countr_zero
+#if OSL_CPLUSPLUS_VERSION >= 20
+
+// For C++20 and beyond, these are in the standard library
+using std::popcount;
+using std::countr_zero;
+
+#elif OSL_INTEL_COMPILER
+
+OSL_FORCEINLINE int popcount(uint32_t x) noexcept { return _mm_popcnt_u32(x);}
+OSL_FORCEINLINE int popcount(uint64_t x) noexcept { return _mm_popcnt_u64(x); }
+OSL_FORCEINLINE int countr_zero(uint32_t x) noexcept { return _bit_scan_forward(x); }
+OSL_FORCEINLINE int countr_zero(uint64_t x) noexcept {
+    unsigned __int32 index;
+    _BitScanForward64(&index, x);
+    return static_cast<int>(index);
+}
+
+#elif defined(__GNUC__) || defined(__clang__)
+
+OSL_FORCEINLINE int popcount(uint32_t x) noexcept { return __builtin_popcount(x); }
+OSL_FORCEINLINE int popcount(uint64_t x) noexcept { return __builtin_popcountll(x); }
+OSL_FORCEINLINE int countr_zero(uint32_t x) noexcept { return __builtin_ctz(x); }
+OSL_FORCEINLINE int countr_zero(uint64_t x) noexcept { return __builtin_ctzll(x); }
+
+#elif defined(_MSC_VER)
+
+OSL_FORCEINLINE int popcount(uint32_t x) noexcept { return static_cast<int>(__popcnt(x)); }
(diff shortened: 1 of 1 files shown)


lgritz:
@AlexMWells This is just a proposal. I think this code is easier to read and understand, has no type errors, and is easier to maintain because it has no clauses that are untested and may never be needed.

AlexMWells:
I might offer a slightly different suggestion.
add OSL::popcount and OSL::countr_zero which house anything compiler specific or just using statement for c++20

lgritz:
That's fine, too, but it just pushes the question one level deeper -- should OSL_popcount/countr_zero (the full implementation, not the trivial `using std::popcount`) implement both 32 and 64 bit versions, or just 32 and hope that everybody's on C++20 if and when we ever care about 64-wide batches?

AlexMWells:
So no plans for 64 bit mask for "normal" SIMD processing, however as an optimized bit mask implementation we have used 64bit masks.  I just used one the other day to identify active texel samples in a texturing subsystem, other use cases might be loop unrolling where you loop over UNROLL_COUNT*SIMD_LANE_COUNT and could need a larger mask, even if not having wider SIMD hardware, in which case compilers could have 2 independent SIMD paths through the loop that have no dependencies which can reduce stalls at the trade off of register pressure.  So yes we could drop 64bit mask support, or leave it and just cleanup the mappings by adding something like this to platform.h:
```
namespace OSL 
{

#if OSL_CPLUSPLUS_VERSION >= 20
    using std::popcount;
    using std::countr_zero;
#else
#    if OSL_INTEL_COMPILER
        OSL_FORCEINLINE int popcount(uint32_t x) noexcept { return _mm_popcnt_u32(x);}
        OSL_FORCEINLINE int popcount(uint64_t x) noexcept { 
            return _mm_popcnt_u64(x);
            }
        OSL_FORCEINLINE int countr_zero(uint32_t x) noexcept { return _bit_scan_forward(x); }
        OSL_FORCEINLINE int countr_zero(uint64_t x) noexcept {  
            unsigned __int32 index;
            _BitScanForward64(&index, x);
        }
#    elif defined(__GNUC__) || defined(__clang__)
        OSL_FORCEINLINE int popcount(uint32_t x) noexcept { return __builtin_popcount(x); }
        OSL_FORCEINLINE int popcount(uint64_t x) noexcept { return __builtin_popcountll(x); }
        OSL_FORCEINLINE int countr_zero(uint32_t x) noexcept { return __builtin_ctz(x); }
        OSL_FORCEINLINE int countr_zero(uint64_t x) noexcept { return __builtin_ctzll(x); }
#    elif defined(_MSC_VER)
        OSL_FORCEINLINE int popcount(uint32_t x) noexcept { return static_cast<int>(__popcnt(x)); }
        OSL_FORCEINLINE int popcount(uint64_t x) noexcept { return static_cast<int>(__popcnt64(x)); }
        OSL_FORCEINLINE int countr_zero(uint32_t x) noexcept {  
            unsigned long index;
            _BitScanForward(&index, x);
        }
        OSL_FORCEINLINE int countr_zero(uint64_t x) noexcept {
            unsigned long index;
            _BitScanForward64(&index, x);
        }
#    else
#        error "popcount implementation needed for this compiler"
#    endif
#endif
} // end namespace OSL
```
No more messy m_value_width conditionals, just good old fashioned function overloads.  And usage in mask is really simple:
```
    OSL_FORCEINLINE int count() const
    {
        return OSL::popcount(m_value);
    }
    OSL_FORCEINLINE int first_on() const
    {
        return OSL::countr_zero(m_value);
    }

```

ICC:
 https://godbolt.org/z/r5jqYE
GCC:
  https://godbolt.org/z/ff39G4
CLANG:
  https://godbolt.org/z/4GsYc7
MSVC:
  https://godbolt.org/z/T6rsah

lgritz:
I kinda like that, Alex, and it also makes it easy for us to use popcount and countr_zero elsewhere if we need to. I'll incorporate that in to this PR.

lgritz:
Updated with suggestions from @AlexMWells

AlexMWells (approved):
LGTM

PR #1297, opened 2020-11-12
Fix compability break with OIIO < 2.2 (from change 1295)

lgritz:


lgritz:
I expect most of the CI tests to fail for unrelated reasons (it's because of changes in the ASWF docker images, which are in the processes of being fixed).

PR #1298, opened 2020-11-12
Pr/batched test shade

AlexMWells:
Implement BatchedSimpleRenderer for TestShade.

Derives from BatchedRendererServices and make use of `Masked<typename DataT, int WidthT>`wrappers and `MaskedData<WidthT>` and `RefData` wrappers for variant data.  Provides examples usages of these encapsulations, intent is actual data served up through the BatchedSimpleRenderer matches that of SimpleRenderer so the same unit tests could eventually work in -batched mode.  This means any changes made in SimpleRenderer shoud be mimicked in BatchedSimpleRenderer.  The BatchedSimpleRenderer does pull its values from the SimpleRenderer instance and adapts them to varying or uniform parameters of its methods.
A conscious decision was made to duplicate some code vs. refactoring SimpleRenderer to share.


lgritz (approved):
LGTM

PR #1299, opened 2020-11-16
Build against OpenEXR 3.0 + Imath 3.0

lgritz:
As OpenEXR works toward their 3.0 release, which includes splitting
Imath (the vector, matrix, math, and 'half' stuff) out into a separate
repo, their top of tree master finally breaks our "bleeding edge"
build.

This is a little tricky, because now our support of this dependency is
attempting to span 3 eras: 3.0 with the full split of openexr from
imath, recent openexr+ilmbase 2.x that supplies reliable exported
cmake configs, and older 2.x that predates cmake configs. Eventually
we'll phase out older versions and clean up.

Fixes in this patch generally include:

* Add a new CI test of VFX2021 + openexr/imath master. (This will go away
  when 3.0 is released and that's what we use for the main 2021 case.)

* For new enough versions of OpenEXR, use the exported config file and
  targets, rather than the old style OPENEXR_LIBRARIES style variables.
  We have kind of a mixed idiom use because we're still supporting older
  versions that predate use of the config files.

* Deal with the fact that the Imath headers now live inside
  `<Imath/ImathBlah.h>` rather than `<OpenEXR/ImathBlah.h>`. But of
  course, you need to start including things just to find out the
  version. It's tricky.  The logic is to look at
  `<OpenEXR/OpenEXRConfig.h>` to determine the version we're dealing
  with, then include Imath stuff from one or the other depending on
  what we found. I rearranged quite a few includes to ensure that all
  this logic only happens in a couple spots and the rest gets included
  transitively. As it turns out, quite a few of our `#include`s of
  Imath headers were redundant.

* OSL's exported config needs to look for OpenEXR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ on:
       - '**.tex'
   schedule:
     # Full nightly build
-    - cron: "0 4 * * *"
+    - cron: "0 8 * * *"
 
 
 jobs:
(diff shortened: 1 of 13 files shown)


fpsunflower (approved):
LGTM

The transition period will be messy - but this seems like a reasonable way to handle it 👍

PR #1302, opened 2020-11-18
LoadMemoryCompiledShader: ensure return false on parse failure

lgritz:
Most failures in this function returned false, but if it got as far as
parsing but failed that stage, it ended up returning true (success),
which is clearly wrong.


fpsunflower (approved):

MasterZap (commented):
LGTM!

PR #1309, opened 2020-12-04
change how strings are handled n OptiX 7

cmstein:
## Description

This patch changes how strings are handled in the OptiX 7 codepath.  Strings are now handled as hashes which are computed at compile time.  This means a few things:
   * strings are no longer copied to the GPU
   * we no longer have to create a special CUDA file that contains string pointers (which removes the libnvrtc dependency -- hooray!)
   * strings are now hashed with a `constexpr` hash function which requires c++14.  Because we don't have CUDA programs that depend on runtime values (ie. pointers to GPU memory),  I'm hoping compilation should be smoother overall.
   * `osl_printf()` output is now handled by the CPU.  `osl_printf()` copies shaders' `printf()` output to a GPU buffer.  After rendering, the CPU reads this buffer and handles all the output.  This means that all strings in the GPU code need to be "registered" with the host program.   See `testshade` and `testrender` for the appropriate changes.
   * because we no longer have a run-time compiled CUDA module to store global variables, I've now added extra `optixLaunch()` calls to `testshade` and `testrender` to set global variables via the pipeline launch parameters.

The OptiX 6 codepath should be untouched.

## Tests

Because strings are no longer copied to the CPU, we don't have access to their lengths nor can we reference individual characters.  As such, the `testoptix.optix` test now fails due to differing text output (the images are fine).

## Checklist:

### DIFF
--- a/src/cmake/cuda_macros.cmake
+++ b/src/cmake/cuda_macros.cmake
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # https://github.com/AcademySoftwareFoundation/OpenShadingLanguage
 
+set ($OSL_EXTRA_NVCC_ARGS "" CACHE STRING "Custom args passed to nvcc when compiling CUDA code")
+
 # Compile a CUDA file to PTX using NVCC
 function ( NVCC_COMPILE cuda_src ptx_generated extra_nvcc_args )
     get_filename_component ( cuda_src_we ${cuda_src} NAME_WE )
@@ -26,8 +28,9 @@ function ( NVCC_COMPILE cuda_src ptx_generated extra_nvcc_args )
             ${LLVM_COMPILE_FLAGS}
             -DOSL_USE_FAST_MATH=1
             -m64 -arch ${CUDA_TARGET_ARCH} -ptx
-            --std=c++11 -dc -O3 --use_fast_math --expt-relaxed-constexpr
+            --std=c++14 -dc -O3 --use_fast_math --expt-relaxed-constexpr
             ${extra_nvcc_args}
+            ${OSL_EXTRA_NVCC_ARGS}
             ${cuda_src} -o ${cuda_ptx}
         MAIN_DEPENDENCY ${cuda_src}
         DEPENDS ${cuda_src} ${cuda_headers} oslexec
--- a/src/include/OSL/device_string.h
+++ b/src/include/OSL/device_string.h
@@ -19,6 +19,14 @@
 OSL_NAMESPACE_ENTER
 
 
+// Strings are handled differently in OptiX 6 vs OptiX 7.
+//
+// In OptiX 7+:
+// Strings are not stored on the GPU at all.  Strings are represented by
+// their hash values which are computed at compile-time.  Because of this,
+// we have no access to either the strings' contents or their lengths.
+//
+// In OptiX 6:
 // Strings are stored in a block of CUDA device memory. String variables hold a
 // pointer to the start of the char array for each string. Each canonical string
 // has a unique entry in the table, so two strings can be tested for equality by
@@ -28,22 +36,43 @@ OSL_NAMESPACE_ENTER
 // stored in the table, in the 16 bytes preceding the characters.
 
 struct DeviceString {
+#if defined(__CUDA_ARCH__) && OPTIX_VERSION >= 70000
+    size_t      m_chars;
+#else
     const char* m_chars;
+#endif
+
+#if defined(__CUDA_ARCH__) && OPTIX_VERSION >= 70000
+    OSL_HOSTDEVICE DeviceString() {}
+    OSL_HOSTDEVICE DeviceString(uint64_t i) : m_chars(i) {}
+#endif
 
     OSL_HOSTDEVICE uint64_t hash() const
     {
+#if defined(__CUDA_ARCH__) && OPTIX_VERSION >= 70000
+        return m_chars;
+#else
         return *(uint64_t*)(m_chars - sizeof(uint64_t) - sizeof(uint64_t));
+#endif
     }
 
+    // In OptiX 7 we don't store the string's length. Make this a compile
+    // time error.
+#if ! (defined(__CUDA_ARCH__) && OPTIX_VERSION >= 70000)
     OSL_HOSTDEVICE uint64_t length() const
     {
         return *(uint64_t*)(m_chars - sizeof(uint64_t));
     }
+#endif
 
+    // In OptiX 7 we can't return the string's contents. Make this a compile
+    // time error.
+#if ! (defined(__CUDA_ARCH__) && OPTIX_VERSION >= 70000)
     OSL_HOSTDEVICE const char* c_str() const
     {
         return m_chars;
     }
+#endif
 
     OSL_HOSTDEVICE bool operator== (const DeviceString& other) const
     {
@@ -54,6 +83,22 @@ struct DeviceString {
     {
         return m_chars != other.m_chars;
     }
+
+#if defined(__CUDA_ARCH__) && OPTIX_VERSION >= 70000
+
+    OSL_HOSTDEVICE bool operator== (const size_t other) const
+    {
+        return hash() == other;
+    }
+
+    OSL_HOSTDEVICE bool operator!= (const size_t other) const
+    {
+        return hash() != other;
+    }
+
+    OSL_HOSTDEVICE operator size_t() const { return hash(); }
+
+#endif
 };
 
 
@@ -63,7 +108,7 @@ struct DeviceString {
 #ifndef __CUDA_ARCH__
 # define HDSTR(cstr) (*((ustring *)&cstr))
 #else
-# define HDSTR(cstr) (*(OSL::DeviceString*)&cstr)
+#define HDSTR(cstr) (*((OSL::DeviceString*)&cstr))
 #endif
 
 
@@ -81,9 +126,13 @@ typedef DeviceString StringParam;
 namespace DeviceStrings {
 #define STRDECL(str,var_name)                       \
     extern __device__ OSL_NAMESPACE::DeviceString var_name;
-#include <OSL/strdecls.h>
+#if OPTIX_VERSION < 70000
+# include <OSL/strdecls.h>
+#endif
 #undef STRDECL
 }
+#else
+#  define STRING_PARAMS(x)  StringParams::x
 #endif
 
 
@@ -91,7 +140,9 @@ OSL_NAMESPACE_EXIT
 
 
 #ifdef __CUDA_ARCH__
+#if OPTIX_VERSION < 70000
 namespace StringParams = OSL_NAMESPACE::DeviceStrings;
+#endif
 #else
 namespace StringParams = OSL_NAMESPACE::Strings;
 #endif
--- a/src/include/OSL/strdecls.h
+++ b/src/include/OSL/strdecls.h
@@ -142,3 +142,5 @@ STRDECL("AdobeRGB", AdobeRGB)
 STRDECL("colorsystem", colorsystem)
 STRDECL("arraylength", arraylength)
 STRDECL("unknown", unknown)
+STRDECL("ERROR: Unknown color space transformation \"%s\" -> \"%s\"\n",
+        ErrorColorTransform)
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -328,12 +328,21 @@ BackendLLVM::addCUDAVariable(const std::string& name, int size, int alignment,
         // char*.
         shadingsys().renderer()->register_string (((ustring*)data)->string(), name);
 
+#if OPTIX_VERSION < 70000
         // Leave the variable uninitialized to prevent raw pointers from
         // appearing in the generated code. The OptiX renderer will set the
         // variable to the string address before the kernel is launched.
         constant = llvm::ConstantInt::get (
             llvm::Type::getInt64Ty (ll.module()->getContext()), 0);
         m_varname_map [name] = ((ustring*)data)->string();
+#else
+        // DeviceStrings use the ustring hash to identify themselves.  These hashes will
+        // match on both the device and host side.
+        constant = llvm::ConstantInt::get (
+            llvm::Type::getInt64Ty (ll.module()->getContext()), ((ustring*)data)->hash());
+        m_varname_map [name] = ((ustring*)data)->string();
+#endif
+
     }
     else {
         // Handle unspecified types as generic byte arrays
@@ -354,17 +363,9 @@ BackendLLVM::addCUDAVariable(const std::string& name, int size, int alignment,
 
 #if (OPTIX_VERSION < 70000)
     g_var->setLinkage    (llvm::GlobalValue::ExternalLinkage);
+#endif
     g_var->setVisibility (llvm::GlobalValue::DefaultVisibility);
     g_var->setInitializer(constant);
-#else
-    if (type != TypeDesc::TypeString) {
-        g_var->setVisibility (llvm::GlobalValue::DefaultVisibility);
-        g_var->setInitializer(constant);
-    }
-    else {
-        g_var->setExternallyInitialized(true);
-    }
-#endif
     m_const_map[name] = g_var;
 
     return g_var;
@@ -457,6 +458,10 @@ BackendLLVM::getOrAllocateCUDAVariable (const Symbol& sym, bool addMetadata)
         //       more robust solution may be called for.
 
         ss << "ds_"
+#if OPTIX_VERSION >= 70000
+           << group().name() << "_"
+           << group().id() << "_"
+#endif
            << std::setbase (16) << std::setfill('0') << std::setw (16)
            << sym.get_string().hash()
            << "_"
--- a/src/liboslexec/opcolor.cpp
+++ b/src/liboslexec/opcolor.cpp
@@ -93,27 +93,27 @@ OSL_CONSTANT_DATA const static ColorSystem::Chroma k_color_systems[11] = {
 
 OSL_HOSTDEVICE const ColorSystem::Chroma*
 ColorSystem::fromString(StringParam colorspace) {
-    if (colorspace == StringParams::Rec709)
+    if (colorspace == STRING_PARAMS(Rec709))
         return &k_color_systems[0];
-    if (colorspace == StringParams::sRGB)
+    if (colorspace == STRING_PARAMS(sRGB))
         return &k_color_systems[1];
-    if (colorspace == StringParams::NTSC)
+    if (colorspace == STRING_PARAMS(NTSC))
         return &k_color_systems[2];
-    if (colorspace == StringParams::EBU)
+    if (colorspace == STRING_PARAMS(EBU))
         return &k_color_systems[3];
-    if (colorspace == StringParams::PAL)
+    if (colorspace == STRING_PARAMS(PAL))
         return &k_color_systems[4];
-    if (colorspace == StringParams::SECAM)
+    if (colorspace == STRING_PARAMS(SECAM))
         return &k_color_systems[5];
-    if (colorspace == StringParams::SMPTE)
+    if (colorspace == STRING_PARAMS(SMPTE))
         return &k_color_systems[6];
-    if (colorspace == StringParams::HDTV)
+    if (colorspace == STRING_PARAMS(HDTV))
         return &k_color_systems[7];
-    if (colorspace == StringParams::CIE)
+    if (colorspace == STRING_PARAMS(CIE))
         return &k_color_systems[8];
-    if (colorspace == StringParams::AdobeRGB)
+    if (colorspace == STRING_PARAMS(AdobeRGB))
         return &k_color_systems[9];
-    if (colorspace == StringParams::XYZ)
+    if (colorspace == STRING_PARAMS(XYZ))
         return &k_color_systems[10];
     return nullptr;
 }
@@ -576,16 +576,25 @@ ColorSystem::set_colorspace (StringParam colorspace)
 
     OSL_HOSTDEVICE void
     ColorSystem::error(StringParam src, StringParam dst, Context sg) {
+#if !defined(__CUDA_ARCH__) || OPTIX_VERSION < 70000
         const char* args[2] = { src.c_str(), dst.c_str() };
         osl_printf (sg,
             (char*) // FIXME!
             "ERROR: Unknown color space transformation \"%s\" -> \"%s\"\n",
             args);
+#else
+        uint64_t fmt_hash = UStringHash::Hash("ERROR: Unknown color space transformation \"%s\" -> \"%s\"\n");
+        uint64_t args[3]  = { 2 * sizeof(uint64_t), dst, src };
+        osl_printf (sg,
+        (char *)
+         fmt_hash,
+            args);
+#endif
     }
 
     __device__ static inline ColorSystem& op_color_colorsystem (void *sg) {
         void* ptr;
-        rend_get_userdata(StringParams::colorsystem, &ptr, 8, OSL::TypeDesc::PTR, 0);
+        rend_get_userdata(STRING_PARAMS(colorsystem), &ptr, 8, OSL::TypeDesc::PTR, 0);
         return *((ColorSystem*)ptr);
     }
 
@@ -629,21 +638,21 @@ ColorSystem::ocio_transform (StringParam fromspace, StringParam tospace,
 OSL_HOSTDEVICE Color3
 ColorSystem::to_rgb (StringParam fromspace, const Color3& C, Context context)
 {
-    if (fromspace == StringParams::RGB || fromspace == StringParams::rgb
+    if (fromspace == STRING_PARAMS(RGB) || fromspace == STRING_PARAMS(rgb)
          || fromspace == m_colorspace)
         return C;
-    if (fromspace == StringParams::hsv)
+    if (fromspace == STRING_PARAMS(hsv))
         return hsv_to_rgb (C);
-    if (fromspace == StringParams::hsl)
+    if (fromspace == STRING_PARAMS(hsl))
         return hsl_to_rgb (C);
-    if (fromspace == StringParams::YIQ)
+    if (fromspace == STRING_PARAMS(YIQ))
         return YIQ_to_rgb (C);
-    if (fromspace == StringParams::XYZ)
+    if (fromspace == STRING_PARAMS(XYZ))
         return XYZ_to_RGB (C);
-    if (fromspace == StringParams::xyY)
+    if (fromspace == STRING_PARAMS(xyY))
         return XYZ_to_RGB (xyY_to_XYZ (C));
     else
-        return ocio_transform (fromspace, StringParams::RGB, C, context);
+        return ocio_transform (fromspace, STRING_PARAMS(RGB), C, context);
 }
 
 
@@ -651,21 +660,21 @@ ColorSystem::to_rgb (StringParam fromspace, const Color3& C, Context context)
 OSL_HOSTDEVICE Color3
 ColorSystem::from_rgb (StringParam tospace, const Color3& C, Context context)
 {
-    if (tospace == StringParams::RGB || tospace == StringParams::rgb
+    if (tospace == STRING_PARAMS(RGB) || tospace == STRING_PARAMS(rgb)
          || tospace == m_colorspace)
         return C;
-    if (tospace == StringParams::hsv)
+    if (tospace == STRING_PARAMS(hsv))
         return rgb_to_hsv (C);
-    if (tospace == StringParams::hsl)
+    if (tospace == STRING_PARAMS(hsl))
         return rgb_to_hsl (C);
-    if (tospace == StringParams::YIQ)
+    if (tospace == STRING_PARAMS(YIQ))
         return rgb_to_YIQ (C);
-    if (tospace == StringParams::XYZ)
+    if (tospace == STRING_PARAMS(XYZ))
         return RGB_to_XYZ (C);
-    if (tospace == StringParams::xyY)
+    if (tospace == STRING_PARAMS(xyY))
         return RGB_to_XYZ (xyY_to_XYZ (C));
     else
-        return ocio_transform (StringParams::RGB, tospace, C, context);
+        return ocio_transform (STRING_PARAMS(RGB), tospace, C, context);
 }
 
 
@@ -676,20 +685,20 @@ ColorSystem::transformc (StringParam fromspace, StringParam tospace,
 {
     bool use_colorconfig = false;
     COLOR Crgb;
-    if (fromspace == StringParams::RGB || fromspace == StringParams::rgb
-         || fromspace == StringParams::linear || fromspace == m_colorspace)
+    if (fromspace == STRING_PARAMS(RGB) || fromspace == STRING_PARAMS(rgb)
+         || fromspace == STRING_PARAMS(linear) || fromspace == m_colorspace)
         Crgb = C;
-    else if (fromspace == StringParams::hsv)
+    else if (fromspace == STRING_PARAMS(hsv))
         Crgb = hsv_to_rgb (C);
-    else if (fromspace == StringParams::hsl)
+    else if (fromspace == STRING_PARAMS(hsl))
         Crgb = hsl_to_rgb (C);
-    else if (fromspace == StringParams::YIQ)
+    else if (fromspace == STRING_PARAMS(YIQ))
         Crgb = YIQ_to_rgb (C);
-    else if (fromspace == StringParams::XYZ)
+    else if (fromspace == STRING_PARAMS(XYZ))
         Crgb = XYZ_to_RGB (C);
-    else if (fromspace == StringParams::xyY)
+    else if (fromspace == STRING_PARAMS(xyY))
         Crgb = XYZ_to_RGB (xyY_to_XYZ (C));
-    else if (fromspace == StringParams::sRGB)
+    else if (fromspace == STRING_PARAMS(sRGB))
         Crgb = sRGB_to_linear (C);
     else {
         use_colorconfig = true;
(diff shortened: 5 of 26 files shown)


src/include/OSL/device_string.h
```diff
@@ -73,25 +110,35 @@ struct DeviceString {
 #ifndef __CUDA_ARCH__
 typedef ustring StringParam;
 #else
-typedef DeviceString StringParam;
+#if OPTIX_VERSION >= 70000
+  typedef DeviceString StringParam;
+# else
+  typedef DeviceString StringParam;
```
aconty, line 116: Isn't this line the same that above?

src/liboslexec/backendllvm.cpp
```diff
@@ -683,7 +691,11 @@ BackendLLVM::llvm_load_device_string (const Symbol& sym, bool follow)
     else if (userdata_index < 0) {
         // Handle non-varying variables
         OSL_DASSERT (sym.data() && "NULL data in non-varying string");
+#if OPTIX_VERSION < 70000
         val = getOrAllocateCUDAVariable (sym);
+#else
+        val = getOrAllocateCUDAVariable (sym);
```
aconty, line 697: Same line as above?

src/liboslexec/opcolor.cpp
```diff
@@ -576,16 +576,25 @@ ColorSystem::set_colorspace (StringParam colorspace)
...
     ColorSystem::error(StringParam src, StringParam dst, Context sg) {
+#if !defined(__CUDA_ARCH__) || OPTIX_VERSION < 70000
         const char* args[2] = { src.c_str(), dst.c_str() };
         osl_printf (sg,
             (char*) // FIXME!
             "ERROR: Unknown color space transformation \"%s\" -> \"%s\"\n",
             args);
+#else
+        uint64_t fmt_hash = UStringHash::Hash("ERROR: Unknown color space transformation \"%d\" -> \"%d\"\n");
+        uint64_t args[2]  = { 2 * sizeof(int), (uint64_t(dst & 0xffffffff) << 32) + uint64_t(src & 0xffffffff) };
```
aconty, line 587: Just trying to understand this. Seems the args are two strings but you are encoding them in 32 bits? And why do you use 2 * sizeof(int) instead of sizeof(uint64_t)? Don't you wanna pass two 64bit hashes? I guess I'm very confused :)

aconty:
Other than my questions this LGTM!

src/liboslexec/backendllvm.cpp
cmstein: They used to differ at some point during my work on this.  :-)  I'll clean those up

src/liboslexec/opcolor.cpp
cmstein: Yeah, I'm not sure what I was thinking.  I think I confused myself over strings being represented as ints.  I'll fix that up.

aconty:
This LGTM. Do you want to wait for @fpsunflower (whoever that is) to look at it? Or move the constexpr hash to oiio before checkin?

lgritz:
I'll look today as well. Maybe @timgrant has a chance to glance at it also?

src/cmake/cuda_macros.cmake
```diff
@@ -26,8 +28,9 @@ function ( NVCC_COMPILE cuda_src ptx_generated extra_nvcc_args )
             ${LLVM_COMPILE_FLAGS}
             -DOSL_USE_FAST_MATH=1
             -m64 -arch ${CUDA_TARGET_ARCH} -ptx
-            --std=c++11 -dc -O3 --use_fast_math --expt-relaxed-constexpr
+            --std=c++14 -dc -O3 --use_fast_math --expt-relaxed-constexpr
```
lgritz, line 31: Should this instead be the same version that is being used for the rest of the OSL build?

Or are you saying that from here on, the Cuda portions *must* be precisely C++14? Or is the whole system now C++14 minimum if one is building the OptiX support?

src/include/OSL/device_string.h
```diff
@@ -28,21 +28,42 @@ OSL_NAMESPACE_ENTER
 // stored in the table, in the 16 bytes preceding the characters.
 
 struct DeviceString {
+#if defined(__CUDA_ARCH__) && OPTIX_VERSION >= 70000
+    size_t      m_chars;
```
lgritz, line 32: Let's put a comment here:   `// For OptiX 7+, the ustring is represented as the hash`

Actually...

The whole comment above that documents DeviceString is incorrect. Can you rework this to correctly describe it how it works for OptiX 6 vs 7?

src/include/OSL/device_string.h
```diff
@@ -28,21 +28,42 @@ OSL_NAMESPACE_ENTER
...
+#endif
     }
 
     OSL_HOSTDEVICE const char* c_str() const
     {
+#if defined(__CUDA_ARCH__) && OPTIX_VERSION >= 70000
+        return nullptr;
+#else
         return m_chars;
+#endif
```
lgritz, line 66: I wonder if the better way to handle this (and length()) might be as:

    #if ! (defined(__CUDA_ARCH__) && OPTIX_VERSION >= 70000)
        OSL_HOSTDEVICE const char* c_str() const
        {
            return m_chars;
        }
    #endif 

... and then it's a compile-time error if you try to use c_str() or length() from Cuda+Optix7 code at all? Rather than returning 0/null and possibly having downstream logic errors?

src/include/OSL/device_string.h
```diff
@@ -73,25 +110,31 @@ struct DeviceString {
 #ifndef __CUDA_ARCH__
 typedef ustring StringParam;
 #else
-typedef DeviceString StringParam;
+  typedef DeviceString StringParam;
```
lgritz, line 113: unintended indent?

src/liboslexec/string_hash.h
```diff
@@ -0,0 +1,667 @@
...
+STATIC_INLINE uint64_t Hash128to64(uint128_t x) {
+  // Murmur-inspired hashing.
+  const uint64_t kMul = 0x9ddfea08eb382d69ULL;
+  uint64_t a = (Uint128Low64(x) ^ Uint128High64(x)) * kMul;
+  a ^= (a >> 47);
+  uint64_t b = (Uint128High64(x) ^ a) * kMul;
+  b ^= (b >> 47);
+  b *= kMul;
+  return b;
+}
```
lgritz, line 58: Should we eventually migrate these to the OIIO side, i.e. adorn hash.h with the proper decorators for Cuda for the things we wish to use?

lgritz (commented):
The basics LGTM. A handful of comments/questions inline.

I assume this passes the OSL testsuite when run in OptiX7 mode, or are there any of the previously-working tests that still fail?

timgrant:
Very cool. I will look this over ASAP, hopefully tomorrow.

src/testrender/optixraytracer.cpp
```diff
@@ -334,12 +316,15 @@ OptixRaytracer::synch_attributes ()
...
+    ustring userdata_str2("userdata string");
+
+    register_string(userdata_str1.string(), "");
+    register_string(userdata_str2.string(), "");
 
     // Store the user-data
-    register_global("test_str_1", addr1);
-    register_global("test_str_2", addr2);
+    d_test_str_1 = userdata_str1.hash();
+    d_test_str_2 = userdata_str2.hash();
```
timgrant, line 327: People tend to use `d_` to denote `CUdeviceptr` variables.

src/testrender/optixraytracer.cpp
```diff
@@ -1398,21 +1326,130 @@ OptixRaytracer::render(int xres OSL_MAYBE_UNUSED, int yres OSL_MAYBE_UNUSED)
...
+    // maybe send buffer size to CUDA instead of the buffer 'end'
+    params.osl_printf_buffer_end = d_osl_printf_buffer + OSL_PRINTF_BUFFER_SIZE;
+    params.color_system          = d_color_system;
+    params.test_str_1            = d_test_str_1;
+    params.test_str_2            = d_test_str_2;
 
     CUDA_CHECK (cudaMemcpy (reinterpret_cast<void *>(d_launch_params), &params, sizeof(RenderParams), cudaMemcpyHostToDevice));
 
+    // Set up global variables
+    OPTIX_CHECK (optixLaunch (m_optix_pipeline,
```
timgrant, line 1339: Is there some reason this can't be done in the regular raygen program? I tried it and all of the tests pass, except for the expected string ops failure.

src/testshade/cuda/optix_grid_renderer.cu
```diff
@@ -115,6 +115,20 @@ extern "C" __global__ void __anyhit__()
...
+    // Set global variables
+    OSL::pvt::osl_printf_buffer_start    = render_params.osl_printf_buffer_start;
+    OSL::pvt::osl_printf_buffer_end      = render_params.osl_printf_buffer_end;
+    OSL::pvt::s_color_system             = render_params.color_system;
+    OSL::pvt::test_str_1                 = render_params.test_str_1;
+    OSL::pvt::test_str_2                 = render_params.test_str_2;
+}
+extern "C" __global__ void __miss__setglobals() { }
+extern "C" __global__ void __closesthit__setglobals() { }
+extern "C" __global__ void __anyhit__setglobals() { }
```
timgrant, line 130: Do you anticipate different sets of global various for the different program types?

timgrant:
Overall, this change looks good to me. Maintaining `printf()` is clutch. Very nice.

I do worry a _little_ bit about the potential loss of functionality, but I'm confident that there should be some way to make the string contents available on the device, if that should prove to be necessary. We can cross that bridge **if** we get to it.

One last comment, I notice that the string table is still being constructed for O7, but as far as I can tell it's not used for anything.

src/cmake/cuda_macros.cmake
cmstein: Only the Cuda portions require C++14 for the compile-time evaluation of the string hash.  Technically, it only applies if we're compiling for OptiX7, so I can look into "lowering" the C++ requirements to C++11 for OptiX6.

src/include/OSL/device_string.h
cmstein: Ah yes, will do!

src/include/OSL/device_string.h
cmstein: I can look into that.

src/liboslexec/string_hash.h
cmstein: Already been looking into that.

src/testrender/optixraytracer.cpp
cmstein: In this particular case, I don't think we need the separate launch to set up globals.  I just figured I'd keep the same logic as from `testshade` in case someone wanted to use `testrender` as an example program on writing an optix-based ray tracer with OSL support.

src/testshade/cuda/optix_grid_renderer.cu
cmstein: Nope -- I just filled them in in case OptiX expects those to exist (even if they don't do anything).  I'll look into removing them if it shouldn't cause any issues.

src/cmake/cuda_macros.cmake
lgritz: No, I think C++14 mode for Cuda is fine, I don't see how that can hurt us, even on OptiX6. And we'll probably be bumping the floor to C++14 in the coming months, as C++11 is being more and more untenable over time (C++11 users already can't build against LLVM 10 or 11).

cmstein:
Addressing more comments, removed string table from OptiX7 code-path, and removed unnecessary 'hit' programs from the "setglobals" launch.

cmstein:
Any other feedback on this?

AlexMWells:
I wanted to walk through out a "possible" alternative to the hash.  I might be missing some of the requirements, but let me outline an approach and see what I'm missing.

Conditions:
1.  At rendertime enough shader network parameters have been set or connected such that any string operations in OSL shader networks can be constant folded and all ustring values are known by the time shader optimization finishes.  T

Requirements:
1.  Device side compile time constant value for a DeviceString be available for all STRING_PARAM() entries.  
2.  DeviceString be equal and not equal comparable on the device.
3.  At rendertime JIT, additional ustrings can be mapped to DeviceString values that can be embedded in JIT as LLVM IR constants.
4.  On host, DeviceString value can be used to lookup a ustring (in order to process printf's, texture filenames, etc.)

Proposal:
Let DeviceString encapsulate a 64bit unsigned integer.
Use strdecls.h with some macro magic to populate constexpr DeviceString for DeviceStrings namespace with a linearly increasing index.  IE:  

```
namespace DeviceStrings {
    __device__ OSL_NAMESPACE::DeviceString _emptystring_(0u);
    __device__ OSL_NAMESPACE::DeviceString camera(1u);
    __device__ OSL_NAMESPACE::DeviceString common(2u);
   // ...
    __device__ OSL_NAMESPACE::DeviceString unknown(127u);
}
```

The same initialization would happen on the host side.
In addition on the host side where rendertime jit code gen happens we need a way to dynamically create additional device strings dynamically from a ustring:

```
namespace { // Unnamed
struct DeviceStringManager {
    std::vector<ustring> m_ustring_by_device_string_index;
    std::unordered_map<ustring, uint64_t, ustringHash>m_device_string_index_by_ustring;
public:
    DeviceStringTable():m_device_count(0) {}
  
    DeviceString register_device_string(ustring val) {
        auto iter = m_device_string_index_by_ustring.find(val);
        if (iter == m_device_string_index_by_ustring.end()) {
            uint64_t device_string_index = m_ustring_by_device_string_index.size();
            m_ustring_by_device_string_index.push_back(val);
            iter = m_device_string_index_by_ustring.insert(std::make_pair(val, device_string_index)).first;
        }
        return DeviceString((*iter).second);
    }
    
    ustring lookup_ustring(DeviceString val) {
        OSL_ASSERT(static_cast<uint64_t>(val) < m_ustring_by_device_string_index.size());
        return m_ustring_by_device_string_index[static_cast<uint64_t>(val)];
    }
}

DeviceStringManager &get_device_string_manager() { 
    static DeviceStringManager  lazy_manager;
    return lazy_manager;
}
} // namespace

DeviceString register_device_string(ustring val) {
    return get_device_string_manager().register_device_string(val);
}

ustring lookup_ustring(DeviceString val) {
    return get_device_string_manager().lookup_ustring(val);
}

```

Additionally on the host side, process strdecls.h again.  This time to register all the StringParams and validate they match 
```
OSL_ASSERT(register_device_string("") == DeviceStrings::_emptystring_);
OSL_ASSERT(register_device_string("camera") == DeviceStrings::camera);
OSL_ASSERT(register_device_string("common") == DeviceStrings::common);
// ...
OSL_ASSERT(register_device_string("unknown") == DeviceStrings::unknown);
```
This could be done in a static object constructor so it happens only once.

So because strdecls is processed in the same order at compile time on the device and execution time registering device strings on the host, we have a set of comparable StringParams between the host and device.  Furthermore any new dynamically registered DeviceStrings will map to the existing device strings or create new unique entries with 0 collision potential.
And we still have the efficiency on the device side of using compile time constants.

Might suggest changing it from DeviceString to StringId (or something implying its just an identifier or handle, not a string itself).

AlexMWells:
I think I might have missed a requirement here.

Is it required that for a JIT'd shader that the same DeviceString(StringID) value be produced the next time the shader is JIT'd?
I suppose if an underlying driver wanted to cache binary version of a shader and match it up next run (assuming the same LLVM IR would be produced and could be matched to a previously built binary version) there could be a problem.  The StringID approach doesn't guarantee any dynamicall created ID's will be in the same order, only the StringParams would be consistent (same ID's) between runs, whereas a hashing approach would make such a guarantee.   Is this a realworld requirement?

lgritz:
@AlexMWells Yes, indeed. The post-JIT, NVIDIA driver-side compilation of the PTX and for that matter the Cuda of the renderer-side components to the actual device code and other things OptiX does with that block of code are *incredibly*,  *painfully* expensive. This is ameliorated by Optix's compile cache. So subsequent runs of the renderer can go much faster as long as the generated PTX/Cuda is absolutely identical. (Eric Enderton will, I'm sure correct any botched terminology or description I have given. I'm doing a little hand waving without consulting my notes.)

So the reason why we couldn't use the CPU-side *address* as the representation of the string on the Cuda side is that these addresses, while guaranteed to be unique, are not going to be identical from run to run, so we don't want them showing up anywhere in the generated PTX or it will foil the cache. (For now, it doesn't matter that the addresses would be host memory, because on the GPU side we are taking for granted that by the time we get on the other side of runtime optimization, we are only doing string assignment, ==, and != comparisons, all of which require only knowing the pointer value, not having access to the characters it points to.)

Using a unique integer index for each string doesn't really get us any farther, because the correspondence of these IDs to the strings is dependent on the order that they are created, which can vary from scene to scene and run to run. 

Another choice, and what we started with but thought was too big a PITA, is essentially to have a *variable* on the Cuda side for each ustring that will be used. Fill in the value (holding the host-side address) at runtime, but the PTX/Cuda thinks it's using a variable, so it's not polluted with the run-to-run-nonconstant value. But this requires a lot of coordination between the Cuda side, OSL internals, and the renderer itself to make sure all strings are predeclared so that those variables are set up. The burden of coordinating with the renderer was something we really wanted to avoid.

Hence, the hope that the hashes would be good enough. Because the hashes are stable from run to run for a particular string, it requires no pre-declaration, no allocation of extra variables, they are just constants to the PTX, but constants that are always the same for a given string, so the compile cache is happy. The only problem is what happens if there is a hash collision (and specifically, a collision between two things that will actually be compared -- a collision between a color space name, and a texture wrap mode name, say, is of no consequence).

aconty:
I wonder how do you ensure that strings from different compilation units
get different ids. Or the same one if they are equal.

AlexMWells:
@aconty , for the StringID approach the StringParams:: are the only ones that would have same ID's because the result of the macro magic just produces DeviceStrings with hard coded #'s and only on the host side are those actually registered as strings (and validated to have the correct #s.  Thus precompiled code (the user's of StringParams) would match between multiple compilation units on host or device.

Alas, it would be of no help for the JIT'd dynamic strings, unless the DeviceStringManager became persistent.

fpsunflower:
I think we have discussed the same design alternatives often enough that it would be worth writing a document summarizing the pros and cons of each and why we chose the current hashing strategy.

It doesn't have to be the final word on the topic, but it would help anyone trying to wrap their heads around why this approach was chosen over other ones.

lgritz:
@etheory [reports](https://github.com/OpenImageIO/oiio/pull/2786) that he is able to generate hash collisions. It's scary that they are "short", but also they are kinda random looking.

I think it's an open question whether we need to fix this right away, or whether we can let the hash approach stand for now, knowing that we'll need to come back and fix this for real, but in the mean time we can more carefully ponder how to do it without running afoul of the ptx cache problem.

Opinions?

lgritz:
As further background, we should spell out that there are three sources of strings, really:

(1) "pre-known" strings that we know that are referenced inside the OSL runtime implementation, for example all the ustrings in the `Strings::` namespace (examples: "RGB", "snoise", "wrap", etc.).

(2) strings that are specified in the shader as literals, or come from the scene as parameter instance values (such as most texture names), or are constructed in the shader in a way that can resolved during runtime optimization of the shader group. In other words, all the strings not in group 1 that the shader itself will try to reference.

(3) strings that are used by the *renderer* -- for example, we imagine that all the things that are currently supplied on the CPU side as RendererServices methods, probably will need corresponding sections of Cuda code, supplied by the renderer, and the internals of those functions need to be able to know the unique values of those strings.

fpsunflower:
I'm not quite convinced yet. Are those hash collisions @etheory mentioned for the current OIIO string hash? Were they *generated* to collide? Obviously this is possible since we don't use cryptographic hashes. The hope was that the odds of it happening by chance with real world strings would be low enough -- we're not making any claims that someone can't engineer strings to collide.

That being said - how bad would it be if we had to increase the size of the string type to 128 bits? At that point its definitely safe to ignore collisions.

etheory:
@fpsunflower hi!
Yes they were engineered to collide ;-). I'm still working on getting something into production for a real-world example.

Increasing the number of bits won't matter.

The fact still stands. There is a non-zero change on a string collision for ANY length of hash.

IMO the only sensible thing to do is just know this, and ensure the system can handle hash collisions.

A few specific points to add to this thread:
- The "pigeon-holing" that happens with hashes is not length-dependent, by design. It should be the same probability regardless of length. So the idea that longer strings or shorter strings are somehow more or less likely to collide, I think is misguided. Any decent hash function should ensure the probability of collision is effectively "independent" of string length.
- Even though my example doesn't seem indicative, what about the case where facilities use asset-management systems that abstract filenames on disc, then return references to these files from some kind of data-base lookup. In that scenario, this kind of filename (the one I gave above in my counter-example) is 100% plausible, both in terms of length, and alphanumeric content. Not all facilities use "sensible-looking filenames" (caveat, AL does, but I've worked in facilities that don't).
- Collisions simply do happen. I personally think building a system on the assumption that they won't, is not a good idea, and will at some point, lead to issues.

lgritz:
>     * Collisions simply do happen. I personally think building a system on the assumption that they won't, is not a good idea, and will at some point, lead to issues.

I don't believe any of us were putting this forward as a permanent solution. We know that using the hash is playing with fire. We were just hoping that it would be rare enough to give us breathing room to come back to this issue later with a better solution.

aconty:
Collision probability does depend on the hash length. If you use a 2 bit
hash you will have many collisions, not so much with 128bit. This frequent
enough discussion always reminds me this quote from Structure and
interpretation of computer programs (Abelson, Sussman), in a similar
context of primality testing:

"In testing primality of very large numbers chosen at random, the chance of
stumbling upon a value that fools the Fermat test is less than the chance
that cosmic radiation will cause the computer to make an error in carrying
out a “correct” algorithm. Considering an algorithm to be inadequate for
the first reason but not for the second illustrates the difference between
mathematics and engineering."

It is interesting that it's precisely the engineers losing sleep over this,
and not mathematicians :)

Like Chris said, one key question is: how bad is it if two strings collide
in the real world. On the probability of collision you also have to scale
by the probability of comparison.

lgritz:
We will have collisions, and we will have them on short, simple strings. That sounds counter-intuitive, but if you think about it, it MUST be that way.

Proof: The hash is 64 bits. EITHER (a) strings less than 8 chars (64 bits) have hash collisions against each other, or (b) the set of all <= 8 char strings perfectly hash to all possible 64 bit values with no collisions (that would be amazing). If (b) is the case, then they have *filled* the space of 64 bit values, so *every* string of >= 8 chars will be a hash collision, and it will collide with a string of <= 8 chars. So either way, you will find hash collisions copiously among even very short (< 8 chars) strings.

But memory is not big enough to have anywhere near 2^64 strings. A ustring table entry is 56 bytes (not counting the characters!), so filling even a huge workstation with 128GB RAM with nothing but the shortest ustrings possible will max out memory at only a couple billion strings.

The danger is that all the usual analysis assumes random strings, but that's not what we're dealing with. The insidious cases would be things like two texture names

    /path/to/texture/A/B/C/diffuse.exr
    /path/to/texture/C/B/A/diffuse.exr

producing the same hash value. The hash function may be pretty good for random input, or even "sequential" input, but for certain structured inputs may be sensitive to permutations that crop up typically in a production data set.

timgrant:
There's a collision hazard even in the current implementation, since the name of the CUDA variable is based on the hash (see the cute TODO in `getOrAllocateCUDAVariable()`. I guess kicking the can down the road has finally developed into street hockey. 😃

It should be possible to prove that there are no collisions in the set of static strings. The path-like strings are still problematic. At least for texture paths, isn't it the case that those must be known in advance and resolved to a texture handle? Are there other cases of string-specified resources we need to worry about? Named transforms and such?

If all strings were "laundered" through ustring, would it be possible to detect and warn about any collisions?

lgritz:
Oh, that's an interesting observation. If we have the same hash problem prior to this patch, then there's no reason not to merge it, right? Even if we keep it on our list to find a more robust long-term solution to this problem.

AlexMWells:
@timgrant (asked in a constructive manner)  So if we could "detect" the collisions, what would user do about it?  Start changing filenames or transform names until the warning goes away?

src/include/OSL/strdecls.h
```diff
@@ -141,3 +141,4 @@ STRDECL("CIE", CIE)
 STRDECL("AdobeRGB", AdobeRGB)
 STRDECL("colorsystem", colorsystem)
 STRDECL("unknown", unknown)
+STRDECL("ERROR: Unknown color space transformation \"%s\" -> \"%s\"\n", ErrorColorTransform)
```
lgritz, line 144: Can you add a line break here, this one line is causing the clang-format test to fail

timgrant:
Good question, Alex. I wasn't thinking of it as a solution, but more as a very rough indicator of how well (or poorly) the hashing approach is working. But it's maybe a bit like those "Days Since Last Accident" signs. It's the next accident we need to worry about, and it's anyone's guess how minor or severe it might be. I guess it might be useful for our initial testing, but probably isn't something a user would want to rely on.

lgritz:
OK, how about this crazy idea:

The ustring class can keep a second hash table internally -- a reverse map that maps hash values to ustrings. (Let's call this the ID map.)

When a new ustring table entry is created for the first time by make_unique(s), we start with h = hash(s) and nrh = 0.  (nrh stands for "number of re-hashes"). The IDmap[h] is checked, if there is no entry for that ID, it's added to the ID table as well as doing usual insertion into the ustring table, and we're done.  If there is already an entry for that ID and it's this very string, also we're done.

But if there is *different* string with that ID, we rehash (for example, `h = hash(h, ++nrh)`), and repeat the process until we either find our string or else get to an empty ID table entry, i.e. we have a string we've never seen before and a unique h value, and that becomes the hash of this new ustring.

I claim that this has a few consequences:

1. We'll never have two ustrings with identical final hash values, so the GPU side using the hash as a proxy for the ustring will not lead to any errors.

2. For the vast, vast majority of scenes that we will ever encounter, there will be no collisions, no strings that get rehashed, all strings will have precisely the same hash as every other frame we rendered, and so the PTX cache works wonderfully.

3. Every once in a great while (and for any particular user, shot, or production, very possibly never), there will be ustring collisions, and some string that is referenced in a shader will end up with a different hash value than it was assigned for the other frames we have previously rendered, and so the PTX cache will not be a match and the shader will be *correct*, but will take longer than usual to compile. (Though, the PTX cache will be helpful the *next* time we render that frame, or for that matter, any other frame that has the same set of collision resolutions.)

Costs: There is no extra cost for using constructed ustrings (=, ==, !=). There is a slight additional cost for constructing ustrings from characters -- checking or amending the ID table, and very very occasionally having to do the rehashing and multiple probes. And there is additional memory use to store the ID table itself (another 16 bytes or so per unique ustring, compared to I think 56 bytes currently for the internal table entry, so maybe the internal non-character, table-related storage behind the ustring implementation will increase by around 25%).

A couple of minor implementation notes: 

It might be kind to remove the extra expenses of all this from the non-OSL users of ustring (and even OSL doesn't need it when not being used by a GPU renderer), by having runtime-settable switch that tells ustring implementation whether to maintain the ID table.

We may or may not wish to care that the implementation of OSL's hash(string) function may -- very very occasionally -- not be 100% stable from scene to scene. I don't know if anybody uses OSL's hash function in such a way that they rely on its absolute stability across all possible scenes. But if this is important, we can probably figure something out, like storing in the internal ustring table both the "original" and "final" hashes, and OSL's hash() returns the original one. (Yeah, yeah, how do we get that on the GPU side? Maybe, like the contents of the characters, we just don't... for now.)

aconty:
The fact that there are collisions doesn’t mean you’re going to find one.
And structured input shouldn’t be an issue if the hash is good. A hash is
usually engineered so if you change a bit of the input, about half the bits
of the output change.

I suggest we put the alarms for collisions so we sleep better. Just for the
problem to not happen silently.

An emergency fix if it happens could be to have a global salt string for
the hash. If collisions are detected you can change the salt instead of
changing file paths.

aconty:
Rehashing is appealing, but it means it can no longer be a constexpr. Don’t
know how much Cliff will like the idea.

lgritz:
Aagh, dammit. I forgot about the constexpr. How badly does that hurt us?

lgritz:
So constexpr is important because that's how compiled Cuda code gets the hash value representing the string from the characters. It needs to be constexpr because it has no access to the CPU-side internal ustring tables.

But bear with me... it doesn't matter. 

Any compiled Cuda code (or liboslexec-side and renderer-side C++ code) will have strings that are (a) few, (b) fixed. All the "dynamic" strings -- meaning it could be different from run to run and possibly a large number of unanticipated strings -- will come into the system via the shaders themselves and in particular the parameters (such as texture names).

So the definition of the chars-to-ustring constructor on the Cuda side is: the old constexpr hash, just like in this PR.

Remember, that's always the first guess for the hash, and *almost always* the final hash, except in the rare, rare case that there is a collision. So we assume (and we can check to verify) that the small number of ustrings statically constructed in the renderer and the OSL library do not produce a collision against each other. If they do, we're in trouble, but the chances of that are minuscule for a small number of strings and a 64 bit hash space. We can have code that verifies that it's collision-free before we load the shaders.

Only the CPU side, subsequently constructed, strings have a chance at colliding, so it's only there that we use the non-constexpr, collision-handling constructor. Those strings just have their final hash values passed to the Cuda code, they don't need to be constructed in the Cuda code.

aconty:
How do we check the cuda side hashes? Aren't the chars lost at compile time?

lgritz:
The rule is: If you have a constructed ustring from a string literal on the Cuda side, you need a statically initialized ustring of the same string in your C++ side code. That ensures that the string is constructed early in the render, before any of the dynamic strings have been constructed.

To verify, OSL asks ustring if there were any collisions so far, fairly early in its operations (before any of the shaders have been loaded, so only the statically constructed ones are present). They you're home-free.

lgritz:
So in Cuda code, you can't say

    if (mysr == devustring("blah")) ...

but rather, (in essence, I know the syntax isn't right)

    OSL_HOSTDEVICE static devustring ublah("blah");
    if (mystr == ublah) ...

devustring is identical to ustring on the CPU side, but on the Cuda side its constructor from a string literal just computes the constexpr 64 bit hash. The HOSTDEVICE and static add up to the fact that the C++/CPU side will make the "correct" ustring with the same hash early enough that there will not be any collisions until later on when all the textures and what not are added.

AlexMWells:
@lgritz Larry, I like the rehash approach combined with the strdecls.h STRING_PARAM to seed the CPU side reverse look with the 1st hash.  Simple rule to never use devstring natively on CUDA side, just always use STRING_PARAM (which I think this existing PR does?).  Perhaps there could be a version of the constructor that declares the result must be the 1st hash.

So as long as all STRING_PARAMS don't collide with anything that already exists on the CPU side after a renderer is loaded. 
 NOTE: we may not be in control of what other 3rd party (aka the renderer) code has populated the ustring table with before our static constructors happen.  However we can certainly detect a collison for these STRING_PARAMS and any existing ustrings (with different string contents) and throw an exception.  Alternatively we could detect it, maybe report it, but NOT do the rehash just use the 1st hash value as the "final" hash that way at least.  We are back to having a "potential" incorrect equality comparison but only if those 2 ustring values share the same hash code are compared.
 
 And as Larry outlined, all dynamic JIT'd strings would use the rehashed "final" hash code which handles collisions.  The only downside being the resulting final hash code "could" be different run to run depending on order which ustrings are registered in the lookup table.  You still get "correct" results, just the shader's JIT'd source would have different final hash code values harded in it (defeating caching in drivers).  However this is just to handle collisions.   
 
 Seems like a decent trade off vs. incorrect results.

lgritz:
The other nice thing is that this idea builds on Cliff's PR here, it doesn't replace it. All it does is ensure no collisions for subsequently added ustrings. If others agree that this is the right approach, we can commit this PR now, and then add the collision-fixing layer at our leisure. It doesn't hold up any other progress.

cmstein:
formatting now clang-format friendly

lgritz:
We decided at the TSC meeting that we don't want to merge this quite yet, because of the hash collision problem. My proposal above was accepted, so I'm going to implement it (much of the work is on the OIIO side), and we'll merge this when my complement to it is ready to also be tried.

aconty:
I did the test

$ find /shots/show1 | tee /tmp/strings.txt | wc

I got bored when it got to 800k strings and stopped it. Then run the
farmhash 64bit on all those strings.

No collisions

That was an animated show. Now trying a vfx show, and only regular files:

$ find -L /shots/show2 -type f | tee /tmp/strings.txt | wc

Got tired of waiting at 2.5 million strings. Again, no collisions

Now for the math, which is exactly the same as the birthday problem
https://en.wikipedia.org/wiki/Birthday_problem
The probability of one or more collisions can be approximated by 1 -
e^(-n^2 / (2 m)) where n is the number of strings and m the hash space.

If farmhash were perfect the hash space would be 2^64, and the probability
of collision in 800k strings would be around 10e-8, 10e-7 for 2.5M. Let's
suppose that farmhash is 10bits away from being perfect (so pretty bad),
the hash space is 2^54 and the probability goes up to 10e-5, more or less
like winning the lottery.

Now take into account no frame is going to reference 800k - 2.5M strings,
and if you factor in the probability of comparison (on the GPU), the
probability of failure is just ridiculous. With a 32bit hash on 800k
strings, even if perfect, the probability of collision is almost 1.0, so
you have a problem there. But with 64bit it is very rare. Collisions do
exist, of course. There also exists a number that when squared gives you 2
:) But finding one in our small world is a different story.

Just wanted to run the numbers. I know you guys are 100% convinced the
simple hash is not acceptable. Or is it 99.999999% convinced? ;)

cmstein:
Updated a few things:
   * cleaned up `string_hash.h` to use the "constexpr" hash that now resides in OIIO
   * rebased this on the top of trunk so things don't get too out of date.

lgritz:
Hey, Cliff, can you add a DCO signoff to the latest commit, just

    git commit --amend -s

ought to do it, then re-push (with force).

lgritz:
This has waited far too long. I'm going to merge this today if nobody has a strong objection. We're very close to being ready to merge the OIIO side that makes hash collisions impossible. If you are super concerned about the extremely low chance of collisions, then, just don't update to this commit on OSL until we've committed the OIIO side.

PR #1310, opened 2020-12-05
Add ShadingSystem attribute "searchpath:library" (similar to searchpa…

AlexMWells:
(similar to searchpath:shader) it will be used to identify directories where target ISA optimized OSL shared libraries may exit.
Make testshade set "searchpath:library" to ../lib64 and ../liboslexec relative to the path testshade is executed from

## Checklist:

### DIFF
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -1405,6 +1405,11 @@ ShadingSystemImpl::attribute (string_view name, TypeDesc type,
         OIIO::Filesystem::searchpath_split (m_searchpath, m_searchpath_dirs);
         return true;
     }
+    if (name == "searchpath:library" && type == TypeDesc::STRING) {
+        m_library_searchpath = std::string (*(const char **)val);
+        OIIO::Filesystem::searchpath_split (m_library_searchpath, m_library_searchpath_dirs);
+        return true;
+    }
     if (name == "colorspace" && type == TypeDesc::STRING) {
         ustring c = ustring (*(const char **)val);
         if (colorsystem().set_colorspace(c))
@@ -1473,6 +1478,7 @@ ShadingSystemImpl::getattribute (string_view name, TypeDesc type,
     lock_guard guard (m_mutex);  // Thread safety
 
     ATTR_DECODE_STRING ("searchpath:shader", m_searchpath);
+    ATTR_DECODE_STRING ("searchpath:library", m_library_searchpath);
     ATTR_DECODE ("statistics:level", int, m_statslevel);
     ATTR_DECODE ("lazylayers", int, m_lazylayers);
     ATTR_DECODE ("lazyglobals", int, m_lazyglobals);
--- a/src/testshade/testshade.cpp
+++ b/src/testshade/testshade.cpp
@@ -46,6 +46,7 @@ static std::vector<std::string> outputvars;
 static std::vector<ustring> outputvarnames;
 static std::string dataformatname = "";
 static std::string shaderpath;
+static std::string librarypath;
 static std::vector<std::string> entrylayers;
 static std::vector<std::string> entryoutputs;
 static std::vector<int> entrylayer_index;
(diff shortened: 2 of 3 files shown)


src/testshade/testshade.cpp
```diff
@@ -571,6 +613,7 @@ getargs (int argc, const char *argv[])
                 "--saveptx", &saveptx, "Save the generated PTX (OptiX mode only)",
                 "--warmup", &warmup, "Perform a warmup launch",
                 "--path %s", &shaderpath, "Specify oso search path",
+                "--library %s", &librarypath, "Specify library search path",
```
lgritz, line 616: Now that we have multiple kinds of search paths, I think `--path` is ambiguous and confusing. Grep says that we never even use the `--path` option to testshade or testrender, so maybe we should eliminate it, and not bother adding `--library`.

FYI, testshade and testrender do not really need separate command line options added for every possible new SS attribute... the `--options` flag can set them:

    testshade --options searchpath:library "blah/blah" ...other args...

So let's add new command line options for true options for those programs, or cases where many tests need to set them easily/commonly. But for settings that merely pass values directly to ShadingSystem attributes, especially if they are not commonly needed, I think we can rely on `--options` for simplicity.

lgritz (changes requested):
I made some comments/suggestions

src/liboslexec/shadingsys.cpp
```diff
@@ -1473,6 +1478,7 @@ ShadingSystemImpl::getattribute (string_view name, TypeDesc type,
     lock_guard guard (m_mutex);  // Thread safety
 
     ATTR_DECODE_STRING ("searchpath:shader", m_searchpath);
+    ATTR_DECODE_STRING ("searchpath:library", m_library_searchpath);
```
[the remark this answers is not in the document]
AlexMWells: it's a library, so library path seems more applicable vs. what the library does (isa)

src/testshade/testshade.cpp
```diff
@@ -159,6 +189,18 @@ set_shadingsys_options ()
...
     if (! shaderpath.empty())
         shadingsys->attribute ("searchpath:shader", shaderpath);
+    if (librarypath.empty()) {
+        // Look in expected dist directory tree,
+        // also look in expected build tree for the unittests
+        // TODO: test on windows
+#if (defined(_WIN32) || defined(_WIN64))
+        static const char * relative_lib_dir = "\\..\\lib64:\\..\\liboslexec";
+#else
+        static const char * relative_lib_dir = "/../lib64:/../liboslexec";
```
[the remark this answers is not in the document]
AlexMWells: ok

src/testshade/testshade.cpp
AlexMWells: I understand for "--library %s" that you want to not list it.  OK.  Are you suggesting that we remove "--path %s" as well?

src/testshade/testshade.cpp
```diff
@@ -113,7 +114,36 @@ inject_params ()
                                pv.interp() == ParamValue::INTERP_CONSTANT);
 }
 
-
+static std::string get_executable_directory() {
```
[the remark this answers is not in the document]
AlexMWells: I an give that a try.

src/testshade/testshade.cpp
lgritz: Yes, I think we should get rid of both. The "path" one is particularly confusing now that there are two different kinds of searchpaths.

A lot of those command line options that merely set a value of a ShadingSystem option, but don't otherwise have any testshade-specific logic, were added to testshade before the --options catch-all flag existed. If I had thought to add --options first, perhaps most of them would never have been added, especially those that are rarely or never used in our testsuite.

src/liboslexec/shadingsys.cpp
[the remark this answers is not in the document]
lgritz: ok

src/testshade/testshade.cpp
```diff
@@ -159,6 +189,18 @@ set_shadingsys_options ()
...
+    if (librarypath.empty()) {
+        // Look in expected dist directory tree,
+        // also look in expected build tree for the unittests
+        // TODO: test on windows
+#if (defined(_WIN32) || defined(_WIN64))
+        static const char * relative_lib_dir = "\\..\\lib64:\\..\\liboslexec";
+#else
+        static const char * relative_lib_dir = "/../lib64:/../liboslexec";
+#endif
+        librarypath = get_executable_directory() + relative_lib_dir;
```
AlexMWells, line 201: Looking at this line now, looks like the 1st relative_lib_dir entry gets the parent directory appended.  I'll have to update this to a loop to build them.

src/testshade/testshade.cpp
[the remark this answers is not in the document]
AlexMWells: I think is the past the "/../liboslexec" worked when running testshade from inside the build subdir, but looks like nowadays "/../lib" is what is required, so I'll make that change

src/testshade/testshade.cpp
AlexMWells:
```
#if (defined(_WIN32) || defined(_WIN64))
        static const char * relative_lib_dirs[] = {"\\..\\lib64", "\\..\\lib"};
#else
        static const char * relative_lib_dirs[] = {"/../lib64", "/../lib"};
#endif
        auto executable_directory = OIIO::Filesystem::parent_path(OIIO::Sysutil::this_program_path());
        bool dirNum = 0;
        for(const char * relative_lib_dir:relative_lib_dirs) {
            if(dirNum++ > 0) 
                librarypath	+= ":";
            librarypath += executable_directory + relative_lib_dir;
        }
```
produces what I expect:
```
        librarypath = /nfs/pdx/home/amwells/OSL_Dev/github/OpenShadingLanguage/dist/linux64.profile/bin/../lib64:/nfs/pdx/home/amwells/OSL_Dev/github/OpenShadingLanguage/dist/linux64.profile/bin/../lib
```

src/testshade/testshade.cpp
AlexMWells: I've never used "library %s" option via command line, as the defaults do what I needed.  So probably a good reason.  Ok, I'll just remove both those command line options.

src/testshade/testshade.cpp
lgritz: great

AlexMWells:
Changes have been pushed, CI has passed(expect appveyor), so ready for review/approve/merge.

lgritz (approved):

PR #1313, opened 2020-12-08
Symbol and Opcode changes to support batched analysis

AlexMWells:
During batched analysis we will be figuring out which symbols need to be varying and which can be uniform during batched execution, we will also be figuring out the minimum number of operations which require masking as well as storing/caching some other operation specific flags used later in batched code generation.  We choose to store these analysis results inside Symbol and Opcode directly so that they are efficiently accessible and stable under optimization passes than might invalidate OpcodeVec.

## Checklist:

### DIFF
--- a/src/include/osl_pvt.h
+++ b/src/include/osl_pvt.h
@@ -481,6 +481,7 @@ public:
         , m_allowconnect(true)
         , m_renderer_output(false)
         , m_readonly(false)
+        , m_is_uniform(true)
         , m_valuesource(DefaultVal)
         , m_free_data(false)
         , m_fieldid(-1)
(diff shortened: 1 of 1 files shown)


src/include/osl_pvt.h
```diff
@@ -1028,6 +1039,18 @@ class Opcode {
     /// different form.  Only opname is changed.
     void transmute_opname(ustring opname) { m_op = opname; }
 
+    /// Op would require masking under batched execution
+    /// when its arguments are not uniform (varying)
+    bool requires_masking() const { return m_requires_masking; }
```
AlexMWells, line 1044: Looking at the names they might be too close to each other, despite proper meanings:
```
bool requires_masking() const
void require_masking()
```
Maybe this would be better
```
bool requires_masking() const
void make_masking_required()
```

lgritz: The idiom I tend to use is:

    bool requires_masking() const;   // return the current value
    void requires_masking(bool val);   // set new value

AlexMWells: I started with that and realized it was a one way ticket, only was being set to true with default of 0.  So changed the API to reflect that.  But its just data, and idiom used elsewhere so can change it back to that.

lgritz (approved):
LGTM

PR #1321, opened 2021-01-02
Disable Appveyor by renaming the file

lgritz:
Our Appveyor testing hasn't worked properly for a long time.  Instead
of trying to fix, we should instead put effort into making a full
Windows CI build and test via GitHub Actions CI.

Seeing the appveyor failures on every PR is not helpful. So just turn
it off by renaming the file (keeping it for reference while we work on
the GHA; we can remove it completely when the GHA Windows CI fully
works).


lgritz:
Withdrawing. I'm going to leave the file as-is, but just disable the web hook. We'll remove the file after the GHA is working for Windows.

PR #1331, opened 2021-02-05
LLVM 12 compatibility fixes

EuanFH:
## Description
Currently if you try and build with a system using LLVM 12 the build will fail. To fix this I have updated and removed deprecated features.
- Removed use of options.PrintMachineCode since this has been removed as of llvm 12.  [LLVM commit of removal](https://github.com/llvm/llvm-project/commit/589c646a7e5fb03223340476f7ffb67fd9628726#diff-bead32db0e4ea40b9f9f5a5149ea3e21ed0e815ffa25a8a6916b0d77f2eed61f)
- Converted uses of DebugLoc::get to DILocation::get since it has been depricated and remove as of llvm 12 [LLVM commit of removal](https://github.com/llvm/llvm-project/commit/41c3b271399229976d9885233db79ebbbe0383c6)

DebugLoc and DILocation are basically equivalent from what i can tell.
I'm unsure if print machine code is used seems like a debug option.

I originally made this fix for 1.11.11.0 and it compiled and worked fine with blender.

I'm not familiar with this project. Nor am I a C++ developer.
So feel free to ignore and fix another way.

## Tests

- I haven't written any new test cases.
- I have tested my 1.11.11.0 version of the fix with blender and it compiled and worked fine.
- I have tried compiling this fix for master on my own machine but hit other issues related to this file. But i don't think its caused by this change.

## Checklist:

- [x ] My code follows the prevailing code style of this project.

### DIFF
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -1389,6 +1389,11 @@ LLVM_Util::make_jit_execengine (std::string *err,
     options.RelaxELFRelocations = false;
     //options.DebuggerTuning = llvm::DebuggerKind::GDB;
 
+    // TODO: Find equivalent function for PrintMachineCode post LLVM 12
+    #if OSL_LLVM_VERSION < 120
+      options.PrintMachineCode = dumpasm();
+    #endif
+
     engine_builder.setTargetOptions(options);
 
     detect_cpu_features(requestedISA, !jit_fma());
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -701,7 +701,8 @@ LLVM_Util::debug_pop_function()
     // that has been finalized, point it back to the compilation unit
     OSL_ASSERT(m_builder);
     OSL_ASSERT(m_builder->getCurrentDebugLocation().get() != nullptr);
-    m_builder->SetCurrentDebugLocation(llvm::DebugLoc::get(static_cast<unsigned int>(1),
+    m_builder->SetCurrentDebugLocation(llvm::DILocation::get(getCurrentDebugScope()->getContext(),
+                static_cast<unsigned int>(1),
                 static_cast<unsigned int>(0), /* column?  we don't know it, may be worth tracking through osl->oso*/
                 getCurrentDebugScope()));
 
@@ -776,7 +777,8 @@ LLVM_Util::debug_set_location(ustring sourcefile, int sourceline)
     }
     if (newDebugLocation) {
         llvm::DebugLoc debug_location =
-                llvm::DebugLoc::get(static_cast<unsigned int>(sourceline),
+                llvm::DILocation::get(sp->getContext(),
+                        static_cast<unsigned int>(sourceline),
                         static_cast<unsigned int>(0), /* column?  we don't know it, may be worth tracking through osl->oso*/
                         sp,
                         inlineSite);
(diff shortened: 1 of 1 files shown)


lgritz (approved):
This is great, thanks so much.

I'm a little surprised, frankly, that these changes seem fine for older LLVM without needing any `#if` to handle multiple versions. But our CI tests have entries for LLVM 7, 8, 9, 10, 11, so apparently they are fine.

Don't worry about the two test failures you are seeing -- that's a totally separate issue related to Imath, and I am about to post a fix for them.

src/liboslexec/llvm_util.cpp
```diff
@@ -1386,7 +1389,6 @@ LLVM_Util::make_jit_execengine (std::string *err,
     options.RelaxELFRelocations = false;
     //options.DebuggerTuning = llvm::DebuggerKind::GDB;
 
-    options.PrintMachineCode = dumpasm();
```
AlexMWells, line 1389: Just because its not in 12, doesn't mean it should be available for people building with LLVM 7, 8, 9, 10, or 11.  Could you wrap this with some conditional compilation to exclude it for LLVM version 12?  Better yet, is there a way to achieve same behavior with 12 (dumping assembly of jit'd code).  If not sure, maybe exclude it for 12 with a TODO comment to renable... somehow.

AlexMWells:
Suggest verifying that debugging still works with testshade and marble.osl or something to make sure the DILocation change is all good.

src/liboslexec/llvm_util.cpp
lgritz: Good catch, Alex!

EuanFH: Will look into the if statement shouldn't be too hard to add again not mainly a c++ dev. Yes the same functionality is possible with other debug options of LLVM from what i read but i am pretty unfamiliar with it. Will see if i can find a solution. can't promise much be will add a todo if not.

lgritz:
@AlexMWells , do you think there is an automated way that our CI can test that we haven't (and don't in the future) break the debugging symbols?

src/liboslexec/llvm_util.cpp
AlexMWells: Thanks @EuanFH

lgritz (changes requested):
Just changing to "request changes" so I don't forget this discussion and accidentally prematurely merge.

AlexMWells:
@lgritz I had the same thought, my primary use of the llvm_debugging_symbols attribute in the ShadingSystem has been to interactively use gdb during development.  I suspect there might be something we could do here (inspect existence of known debugging symbols for a shader), just haven't looked into it before.

EuanFH:
Now have put in an if statement to include PrintMachineCode if lower than LLVM 12.
All the code used to set this will be dangling above version 12 for now.
I have had a look into how to set this option in LLVM 12.
There is a command line argument "print-after-isel" which replicates PrintMachineCode.
Once LLVM has the new pass manager this will also be removed and you would just used the command line arguments "print-after". 
[Discussion where i found this information](https://reviews.llvm.org/D83275#change-sppNqVLUofnL)

TargetPassConfig is adding a pass for print-after-isel from the command line arguments
```
//PM is PassManagerBase
void TargetPassConfig::addPrintPass(const std::string &Banner) {
   if (PrintAfterISel)
     PM->add(createMachineFunctionPrinterPass(dbgs(), Banner));
 }
 ```
So you can add in a pass to achieve the same goal.
But I don't know if createMachienFunctionPrinterPass is idempotent. So I'm worried if adding it from the command line and within the program would cause issues like it printing the machine code twice.

So possibly adding to this function would fix the issue.
```
//no idea where you would get banner from or where dbgs() comes from
LLVM_Util::setup_optimization_passes (int optlevel, bool target_host){
    ...
    #if OSL_LLVM_VERSION > 120
        if(dumpasm()){
            fpm.add(createMachineFunctionPrinterPass(dbgs(), Banner))
        }
    #endif
    ...
}
```

Hope this helps. Unsure if i have taken the wrong approach in my research.

lgritz:
I have more commits to add to this to make it 100% LLVM 12 correct. But I can't seem to push onto your branch. I may open a new PR and attribute it to you.

lgritz:
Closing this and replacing by #1351, which is a combination of Euan's original work and a few amendments by me.

PR #1332, opened 2021-02-06
Imath 3 support: ImathLimits.h no longer exists

lgritz:
Also change some refs to std::abs rather than deprecated Imath functions.


lgritz:
How is it even possible that it, just today, thinks I haven't confirmed my affiliation with the comany?

lgritz:
Any objections to this PR?

PR #1337, opened 2021-02-18
Typo prevented visibility hiding from happening properly

### LINKED ISSUE #1336: LLVM symbols not hidden in 1.11
### Problem

When statically linking with LLVM, it appears that the LLVM symbols are no longer hidden within `liboslexec.so`. This can cause conflicts with other LLVM libraries, like those linked by the Mesa drivers. 

I believe this is because the  [`if (VISIBILITY_COMMAND)`](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/blob/master/src/liboslexec/CMakeLists.txt#L251-L254) conditionals in the various CMakeLists.txt files are typos, and should be `if( VISIBILITY_MAP_COMMAND )`. At least, that would match the [equivalent](https://github.com/OpenImageIO/oiio/blob/6313484e35765df1adf8f40c217cebc4c9d9d954/src/libOpenImageIO/CMakeLists.txt#L196-L199) in OpenImageIO, and does appear to hide the symbols if fixed.

Sorry, it would probably have been quicker to open a PR than an issue, but I don't have a CLA sorted out with my current employer.

**Expected behavior:** 

```
> nm -D liboslexec.so | grep llvm | grep -v OSL | wc -l
0
```

**Actual behavior:**

```
> nm -D liboslexec.so | grep llvm | grep -v OSL | wc -l
25529
```

### Steps to Reproduce

1. Build OpenShadingLanguage against a static LLVM (https://github.com/GafferHQ/dependencies/blob/master/OpenShadingLanguage/config.py#L26-L40).
2. Check the number of exported LLVM symbols with `nm -D liboslexec.so | grep llvm | grep -v OSL | wc -l`

### Versions

* OSL branch/version: 1.11.10
* OS: CentOS 7
* C++ compiler: GCC 6.3.1 (devtoolset-6)
* LLVM version: 10.0.1
* OIIO version: 2.2.11.1

lgritz:
Fixes #1336

### DIFF
--- a/src/liboslcomp/CMakeLists.txt
+++ b/src/liboslcomp/CMakeLists.txt
@@ -54,7 +54,7 @@ set_target_properties (${local_lib}
     )
 
 
-if (VISIBILITY_COMMAND)
+if (VISIBILITY_MAP_COMMAND)
     set_property (TARGET ${local_lib}
                   APPEND PROPERTY LINK_FLAGS ${VISIBILITY_MAP_COMMAND})
 endif ()
--- a/src/liboslexec/CMakeLists.txt
+++ b/src/liboslexec/CMakeLists.txt
@@ -248,7 +248,7 @@ set_target_properties (${local_lib}
         OUTPUT_NAME ${local_lib}${OSL_LIBNAME_SUFFIX}
     )
 
-if (VISIBILITY_COMMAND)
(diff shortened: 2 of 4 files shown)


lgritz:
@johnhaddon

johnhaddon:
Thanks Larry, LGTM.

PR #1342, opened 2021-03-06
Update old Strutil::format to new fmt conventions

lgritz:
The old ones will someday be deprecated, so just doing some opportunistic cleanup.

### DIFF
--- a/src/liboslexec/shadeimage.cpp
+++ b/src/liboslexec/shadeimage.cpp
@@ -32,8 +32,13 @@ shade_image (ShadingSystem &shadingsys, ShaderGroup &group,
     if (! roi.defined())
         roi = buf.roi();
     if (buf.spec().format != TypeDesc::FLOAT) {
-        buf.error ("Cannot OSL::shade_image() into a %f buffer, float is required",
-                   buf.spec().format);
+#if OIIO_VERSION >= 20200
+        buf.errorfmt("Cannot OSL::shade_image() into a {} buffer, float is required",
+                     buf.spec().format);
+#else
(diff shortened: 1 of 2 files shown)


src/liboslexec/shadeimage.cpp
```diff
@@ -32,8 +32,13 @@ shade_image (ShadingSystem &shadingsys, ShaderGroup &group,
     if (! roi.defined())
         roi = buf.roi();
     if (buf.spec().format != TypeDesc::FLOAT) {
-        buf.error ("Cannot OSL::shade_image() into a %f buffer, float is required",
-                   buf.spec().format);
+#if OIIO_VERSION >= 20200
+        buf.errorfmt("Cannot OSL::shade_image() into a {} buffer, float is required",
+                     buf.spec().format);
```
AlexMWells, line 37: I think the OIIO_VERSION # is wrong.
https://github.com/OpenImageIO/oiio/blob/RB-2.2.3/src/include/OpenImageIO/imagebuf.h
As OIIOv2.2.3 doesn't contain errorfmt.
And my local build against 2.2.3 fails.

Evidence is RB-2.3.1 does contain it, suggest updating the OIIO_VERSION >= 20300

version

lgritz: Thanks, I think you are right, I will update. It was probably a typo.

lgritz: So, the current release is 2.2.12.  The FIRST public supported release of OIIO 2.2 was 2.2.6. So you are actually using some kind of development pre-release of OIIO.

PR #1346, opened 2021-03-11
Fully split oslquery and oslnoise libraries from oslexec

lgritz:
liboslquery and liboslnoise exist as separate libraries for apps that
only need to query shader parameters or use OSL's noise functions,
respectively, but dont want to pull in the rest of OSL and all its
heavy dependencies (including LLVM). But liboslexec also needs that
functionality and so separately put the individual modules inside.

This was causing some problems for a tricky use case, and I can't
think of a good reason for it to be done this way now. So in this
patch I remove the redundant modules and just make liboslquery and
liboslnoise into *link dependencies* of liboslexec. None of this
should affect downstream projects, I don't think; linking against
liboslexec ought to just automatically pull in those other libs.

### DIFF
--- a/src/liboslexec/CMakeLists.txt
+++ b/src/liboslexec/CMakeLists.txt
@@ -24,19 +24,16 @@ set (lib_src
           batched_backendllvm.cpp
           batched_llvm_gen.cpp
           batched_llvm_instance.cpp
-          ../liboslnoise/gabornoise.cpp
-          ../liboslnoise/simplexnoise.cpp
     )
 
 if (BUILD_SHARED_LIBS)
-    # oslcomp and oslquery symbols used in oslexec
+    # oslcomp symbols used in oslexec
     list(APPEND lib_src
         ../liboslcomp/ast.cpp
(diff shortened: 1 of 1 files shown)


lgritz:
@marsupial, does this look like it will be good for you?

lgritz:
So, just so everybody understands, the case that made us do this is if you have library A that links liboslexec (it needs a full shading system), and library B that links liboslquery (it just needs to query OSL shader parameters). Let's say that A and B are *independently developed* and they don't know anything about each other, but application C needs both A and B. Now you have two copies of the oslquery symbols, one coming from liboslexec via A and the other coming from liboslquery via B. Bad, bad.

marsupial:
Testing it now!

marsupial:
LGTM!

lgritz:
Oops, this actually has some problems. I forgot that we were relying on some OSLQuery methods being defined differently in oslquery and oslexec, and the conflict is causing problems. I'll try to have a correct fix by Monday.

lgritz:
2nd round of this fix is in #1348

PR #1347, opened 2021-03-12
Fix guard so errorfmt is not used with too early an OIIO version

lgritz:


AlexMWells (approved):
LGTM

PR #1348, opened 2021-03-15
Fix oslquery link conflict

lgritz:
In PR 1346, we overlooked the fact that OSLQuery::init() is defined
*differently* in liboslquery and the separate copy that used to be
in liboslexec.

In this patch, we remove the oslquery-side version. So now
instead of being replicated differently in the two libraries, it only
exists in oslexec. That constructor, since it takes a ShaderGroup* as
an argument, only makes sense to call if you have a full ShadingSystem
nearby.


fpsunflower (approved):
Something feels a bit odd about having the methods of a class being split among different libraries. I agree that this seems like the simplest fix though.

Just to suggest an alternative - maybe `ShaderGroup` could contain a method that returns an `OSLQuery` object? Would that make things any clearer?

lgritz:
@fpsunflower Maybe it would. I did think of that, too. I don't know if that's more or less confusing. I can mock that up pretty quickly if you want to look at it and render a judgment.

lgritz:
I can't get rid of this completely because it's an ABI change, and we may need to backport this fix to the release branch.

What do you think about me merging this PR as it stands, and then I'll submit a separate one (for master only) that proposes deprecating the OSLQuery direct support for the shader group and makes a new ShadingSystem method instead?

lgritz:
Updated with a proposed addition of ShadingSystem::oslquery(). Marked the OSLQuery methods as deprecated, but can't remove them until we're at an ABI-breaking version change.

lgritz:
Any comments?

lgritz:
One more update posted: avoid a debug mode assertion when the new ShadingSystem::oslquery encounters a parameter that is an array of indeterminate length.

fpsunflower (approved):
LGTM!

PR #1351, opened 2021-03-29
Various fixes for LLVM 12 compatibility

lgritz:
Removed use of options.PrintMachineCode since this has been removed as
of llvm 12. Converted uses of DebugLoc::get to DILocation::get since
it has been deprecated and removed as of llvm 12.

PrintMachine code will only be used if llvm is lower than version 12

Pass adjustment as various things have been deprecated or changed
names. (LG)

Signed-off-by: Euan Haahr <euanhaahrmail2@gmail.com>


lgritz:
This supersedes #1331. It is joint work of Euan Haahr and Larry Gritz. I was unable to push my amendments onto Euan's repo to correctly amend his original PR. But I have tested the combination with LLVM 12 and it works, so I will merge this as soon as the CI passes.

PR #1355, opened 2021-04-07
Added farmhash for lower OpenImageIO versions

pellerington:
Older versions of OpenImageIO don't have the farmhash which is now nessisary for optix 7 strings.

To fix this, I added the file into into OpenShadingLangauge wrapped around version guards. Some existing functions had to be included again with OIIO_HOSTDEVICE added (Uint128Low64, Uint128High64, Uint128, CopyUint128, Hash128to64) so I wrapped them in the "inlined" namespace to prevent confilcts.

I made these changes so that I could build OSL with my version of OIIO, I am opening this pull request incase there is any interest in supporting it in the official branch. Potentially the VFX platform might have moved far enough by the time of the OSL release.

### DIFF
--- a/src/include/OSL/farmhash.h
+++ b/src/include/OSL/farmhash.h
@@ -1,0 +1,2189 @@
+// Copyright (c) 2014 Google, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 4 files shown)


lgritz:
Hi, sorry for the delay in responding to this (the last couple weeks have been really busy for me).

I've backported these farmhash fixes to the OIIO release branch and it's included in OIIO 2.2.13.0. The in-development OSL (and in particular the OptiX support) is technically considered "experimental/optional", it's a pre-release, so I think it's not unreasonable to expect it to be built against the latest OIIO release. I think I would prefer to require using a sufficiently recent OIIO release than to have to duplicate this code in the OSL codebase as well. But if people are really inconvenienced  by that, I can certainly reconsider.

PR #1357, opened 2021-04-13
USE_BATCHED cmake option listing target library to build.  New BATCHED_REGRESSION testing mode.

AlexMWells:
## Description
Introduce cmake option USE_BATCHED to enabling building batched SIMD shader execution for a comma separated list of batch sizes, ISA's, and noFMA: (0, b8_AVX, b8_AVX2, b8_AVX2_noFMA, b8_AVX512, b8_AVX512_noFMA, b16_AVX512, b16_AVX512_noFMA)

When USE_BATCHED is set, -DOSL_USE_BATCHED=1 will be passed allowing for source files to conditionally compile in batched execution support.
When USE_BATCHED is set, cmake variable BUILD_BATCHED=True and projects can conditionally add batched sources to be built.
Updated OSL source with #ifdef OSL_USE_BATCHED/#endif were needed.
Improved/streamlined liboslexec/CMakeLists.txt to iterate over targets in USE_BATCHED vs. hardcoding all variants.
Updated _OSL_SUPPORTS_* defines to match USE_BATCHED values simplying creating compiler definitions in Cmake.

Added new BATCHED_REGRESSION testing mode, when a file named BATCHED_REGRESSION exists in a testsuite folder, the run.py will be executed and its output files copied to "./baseline".  The same run.py is then run again but with environment variable TESTSHADE_BATCHED=1, which cause testshade to execute in batched mode.  Then the results/image comparison happens between the batched output files and those in "./baseline".  The "./ref" dir is never used in a regression test.
Presence of BATCHED_REGRESSION prevents non-regression tests from being created for a folder.

Renamed ShadingSystem::supports_batch_execution_at(int width) to ShadingSystem::configure_batch_execution_at(int width) to better identify its behavior:  based on currently set attributes for llvm_jit_target and llvm_jit_fma, test if current machine is capable of supporting batched execution at the specified width.  If no specific llvm_jit_target was requested, sets llvm_jit_target and llvm_jit_fma to the supported configuration for the requested width.  Returns true if supported, false otherwise

Fix bug in batched Matrix*Matrix code generation where the order of arguments were being reversed (swap was intended to transform float*Matrix to a Matrix*float function call, but accidentally happened for Matrix*Matrix)

## Tests

Converted arithmetic-cov to be a BATCHED_REGRESSION test.
Removed testsuite/arithmetic-cov/ref image files.

## Checklist:


src/include/OSL/oslexec.h
```diff
@@ -23,7 +23,9 @@ struct PerThreadInfo;
 class ShadingContext;
 class ShaderSymbol;
 class OSLQuery;
+#ifdef OSL_USE_BATCHED
```
lgritz, line 26: This may not work the way you think. This header, oslexec.h, is read by client/app code, but OSL_USE_BATCHED is only defined by the OSL build system itself. So these definitions won't be visible to a renderer if they don't know to enable the preprocessor symbol.

Instead, what I would suggest is that in oslconfig.h.in, you add:

    #cmakedefine OSL_USE_BATCHED

And then that symbol will be defined (in the generated `oslconfig.h`) if the CMake variable OSL_USE_BATCHED is defined and nonzero, otherwise will omit the definition. (See https://cmake.org/cmake/help/latest/command/configure_file.html) Then the rest of the `#ifdef OSL_USE_BATCHED` will work as intended, even in client code that doesn't use the OSL build system -- it will appear defined in oslconfig.h any time that the OSL build itself enabled batched shading.

src/cmake/compiler.cmake
```diff
@@ -283,6 +283,33 @@ if (NOT USE_SIMD STREQUAL "")
...
+# The USE_BATCHED option may be set to indicate that support for batched
+# SIMD shader execution be compiled along with targe specific libraries
+set (USE_BATCHED "" CACHE STRING "Build batched SIMD shader execution for (0, b8_AVX, b8_AVX2, b8_AVX2_noFMA, b8_AVX512, b8_AVX512_noFMA, b16_AVX512, b16_AVX512_noFMA)")
+set (BATCHED_SUPPORT_DEFINES "")
+set (BATCHED_TARGET_LIBS "")
+set (BUILD_BATCHED False)
+if (NOT USE_BATCHED STREQUAL "")
+    message (STATUS "Compiling with batched SIMD targets ${USE_BATCHED}")
+    if (NOT USE_BATCHED STREQUAL "0")
+        add_compile_definitions ("OSL_USE_BATCHED=1")
```
lgritz, line 298: I think you won't need this add_compile_definitions if you do the other thing I suggested with oslconfig.h.

lgritz:
Also, I have recently fixed some problems with the cmake files that will make a (minor) merge conflict with you. See #1356, it's only a couple lines within your changes that need to be altered, and I would recommend rebasing this on top of the current master before we merge.

lgritz (approved):
LGTM

PR #1362, opened 2021-04-20
Bump the default C++ mode to 14

lgritz:
This leaves the minimum C++ at 11, but bumps the default (if not set by
the user via CMAKE_CXX_STANDARD) from 11 to 14.

This is meant to start identifying any problems people have switching
to 14. We anticipate a hard switch to 14 minimum before the next OSL
major release. This change is in master only (i.e. future OSL), we would
never change minimum or default C++ standard in the supported release
branches (currently OSL 1.11.x).

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -27,7 +27,7 @@ message (STATUS "CMAKE_CXX_COMPILER_ID  = ${CMAKE_CXX_COMPILER_ID}")
 ###########################################################################
 # C++ language standard
 #
-set (CMAKE_CXX_STANDARD 11 CACHE STRING
+set (CMAKE_CXX_STANDARD 14 CACHE STRING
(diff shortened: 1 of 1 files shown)


lgritz:
This would be a good time to let us know if you are currently dependent on the master branch of OSL but are also tied to a toolchain that requires gcc4.8 / C++11.

If you're using the released OSL 1.11 and are ok with that until you are able to upgrade to C++14/gcc6, then you have nothing to worry about, release branches are not going to change compatibility requirements.

lgritz:
My plan is to merge this change on Apr 29, after the OSL TSC meeting, if nobody has objected here or at the meeting.

Furthermore, absent strong objection, I will aim for June 1 to bump the actual minimum to C++11 (and therefore, for those using gcc, the minimum compiler to gcc 6).

PR #1369, opened 2021-05-06
Bump minimum C++ to 14 and update compiler mins to match

lgritz:
As discussed at the TSC meeting and on the mail list, we won't merge this
change until June 1. It will only be in master (future OSL 1.12), not
backported to any release branch.


lgritz:
I said I would merge this on June 1, and so here we are.

PR #1375, opened 2021-05-22
Build: llvm-config --system-libs needs --static for LLVM_STATIC

lgritz:


lgritz:
This fixes CI in the Mac test. Something about the Homebrew upgrade to llvm 12.0.0_1 made this fix necessary to build.

PR #1379, opened 2021-06-16
CI: add a clang11 test, also test batched shading with clang

lgritz:
Along the way, discovered and fixed a couple clang-specific warnings when
using C++17.


lgritz:
@AlexMWells

PR #1383, opened 2021-06-24
Sanitize filenames when using "archive_groupname"

lgritz:
This feature (used, very possibly, only by me, for debugging)
forces OSL to write an archive that contains the oso files and a
serialized version of the shader network.

If no explicit name is given (by the "archive_filename" attribute), it
picks a name based on the shader group name. But that may not be a
safe string for a valid filename. We already eliminated anything
before the last slash, but it turns out that a colon (':') can also
really mess things up. So this small change protects against that.

### DIFF
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -2455,10 +2455,12 @@ ShadingSystemImpl::ShaderGroupEnd (ShaderGroup& group)
         std::string filename = m_archive_filename.string();
         if (! filename.size()) {
             filename = OIIO::Filesystem::filename (groupname.string()) + ".tar.gz";
-            // Colons will ruin our day when passed to tar as a filename.
-            // Why, yes, some people name their shader groups with colons.
+            // Transform characters that will ruin our day when passed to
+            // tar as a filename.
             if (OIIO::Strutil::contains(filename, ":"))
                 filename = OIIO::Strutil::replace(filename, ":", "_");
+            if (OIIO::Strutil::contains(filename, "|"))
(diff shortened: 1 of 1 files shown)


src/liboslexec/shadingsys.cpp
```diff
@@ -2453,8 +2453,13 @@ ShadingSystemImpl::ShaderGroupEnd (ShaderGroup& group)
     ustring groupname = group.name();
     if (groupname.size() && groupname == m_archive_groupname) {
         std::string filename = m_archive_filename.string();
-        if (! filename.size())
+        if (! filename.size()) {
             filename = OIIO::Filesystem::filename (groupname.string()) + ".tar.gz";
+            // Colons will ruin our day when passed to tar as a filename.
+            // Why, yes, some people name their shader groups with colons.
+            if (OIIO::Strutil::contains(filename, ":"))
```
fpsunflower, line 2460: Maybe catch `/`, `\` and `|` as well?

lgritz: The call to `Filesystem::filename` already strips out anything prior to the last directory separator. Hmmm, maybe watching out for `|` isn't a bad idea, though.

lgritz:
Amended:

Be even more careful with how we assemble the strings used to form
shell commands for tar/zip'ing up group archives.

* Look out for `|` character.

* For the actual commands sent to system(), escape special characters
  in the strings and enclose any arguments that come from the user in
  double quotes. That prevents shenannigans like asking for an archive
  filename called "; rm -r *". By enclosing in double quotes, we'll
  end up with a badly named filename produced by tar, rather than
  possibly an arbitrary command being executed.

PR #1401, opened 2021-09-02
fix problem with unresolved symbols when multiple shaders call certain

cmstein:
functions on the GPU (ie. osl_printf()).

## Checklist:

### DIFF
--- a/src/liboslexec/opcolor.cpp
+++ b/src/liboslexec/opcolor.cpp
@@ -563,11 +563,11 @@ ColorSystem::set_colorspace (StringParam colorspace)
 // For Optix, this will be defined by the renderer. Otherwise inline a getter.
 #ifdef __CUDACC__
     extern "C"  __device__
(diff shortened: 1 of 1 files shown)


lgritz (approved):
LGTM, will merge after CI results are complete.

cmstein:
This fixes a problem I was seeingon the GPU:  if I had two shaders that contained a `printf()`, then the second shader group would report `osl_printf` as being an unresolved function (the first shader group was fine).  If I looked at the PTX for the second shader group, `osl_printf` was tagged as a "weak" function.  Removing the "weak" attribute resolves the problem.

I'm not sure why `osl_printf` and `rend_get_userdata` should be different from the other functions in `rend_lib.cu`.

This is easily reproduced in `testsuite/testoptix` by adding a `printf()` to `checkerboard.osl` and `microfacet_test.osl`.

PR #1405, opened 2021-09-14
Changes to OSL ownership of color systems

lgritz:
Patrick Hodoul pointed out that we were forcing OpenColorIO to
initialize and read configs as soon as the ShadingSystem was
created. In this patch, we're changing to a lazy creation approach.

We remove the `OCIOColorSystem m_ocio_system` member and the
ocio_transform() method from ShadingSystem. Put them in the
ShadingContext instead. This allows lock-free access because each
context is used by only one thread at a time.

Within the OCIOColorSystem, make the `OIIO::ColorConfig` be a
shared_ptr rather than a member, so it can be created lazily.  If the
Context's OCIOColorSystem needs a config and doesn't have its own yet,
it asks the ShadingSystem, which allocates one the first time (and
only that one time).

So, in short, SS allocates *one* shared ColorConfig, the first time
any context asks for one. The Context only asks the SS for this once
per context (i.e., once per thread). Individual shades just use the
ColorConfig of their context, not needing to lock for it (except for
that brief lock the first time any context asks for the very first
ColorConfig that's needed).

Some things were guarded by `OIIO_HAS_COLORPROCESSOR`, which is silly
because we no longer support OIIO versions so old that it's not
defined.  But then OSL `undef`ed it for Cuda. I replaced these guards
by a more straightforward direct test of whether it was Cuda code.

I added a sample OCIO config to testsuite so we can use it for tests.
In particular, the new testsuite/ocio -- test using OCIO for color
transforms. We never actually tested that in the test suite before!


lgritz:
Any objections and/or positive reviews?

src/liboslexec/oslexec_pvt.h
```diff
@@ -835,7 +833,7 @@ class ShadingSystemImpl
 
     ustring m_colorspace;                 ///< What RGB colors mean
     ColorSystem m_colorsystem;            ///< Data for current colorspace
-    OCIOColorSystem m_ocio_system;        ///< OCIO color system (when OIIO_HAS_COLORPROCESSOR=1)
+    std::shared_ptr<OIIO::ColorConfig> m_colorconfig;  ///< OIIO/OCIO color configuration
```
AlexMWells, line 836: Just checking... is any CUDA_ARCH needed around this?

src/liboslexec/oslexec_pvt.h
```diff
@@ -2081,6 +2083,8 @@ class OSLEXECPUBLIC ShadingContext {
 
     Dictionary *m_dictionary;
 
+    OCIOColorSystem m_ocio_system;
```
AlexMWells, line 2086: Just checking... is any CUDA_ARCH needed around this?

src/liboslexec/shadingsys.cpp
```diff
@@ -3710,67 +3710,46 @@ ShadingSystemImpl::merge_instances (ShaderGroup &group, bool post_opt)
...
-                  (CC[1] - CC[0]) * (1.0f / eps),
-                  (CC[2] - CC[0]) * (1.0f / eps));
-        return true;
-    }
-#endif
-    return false;
+std::shared_ptr<OIIO::ColorConfig>
+ShadingSystemImpl::colorconfig()
+{
+    lock_guard lock(m_mutex);
```
AlexMWells, line 3749: Is the guard only needed when it's not initialized?  Won't this lock a mutex everytime colorconfig is called?

sfriedmapixar (approved):

src/liboslexec/opcolor.h
```diff
@@ -18,9 +18,6 @@
 
 #include <OpenImageIO/color.h>
 
-#ifdef __CUDACC__
-  #undef OIIO_HAS_COLORPROCESSOR
-#endif
```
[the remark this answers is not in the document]
lgritz: You're misunderstanding what this means. OIIO_HAS_COLORPROCESSOR never meant "OIIO was built with OCIO." Rather, it meant "you are using an OIIO that is new enough to have the ColorProcessor class exposed in the public API", which was added between 1.9 and 2.0. Current OSL doesn't support OIIO versions old enough to predate this, so the guard is no longer needed.

OIIO's ColorProcessor exists even if the OIIO was built without OCIO support, or even if no OCIO config is found. In those cases, it reverts to a basic implementation that knows a few general color transforms like converting between sRGB and a linear transfer function.

src/liboslexec/oslexec_pvt.h
lgritz: I don't think so. Builds fine for me.

src/liboslexec/oslexec_pvt.h
lgritz: Builds fine for me.

src/liboslexec/shadingsys.cpp
lgritz: The mutex is needed every time ShadingSystemImpl::colorconfig() is called, but it will only be called the first time a ColorConfig is needed by each ShadingContext (that is, each thread) needs a context, after which a copy is retained in the context, where it does not need a lock because it's only used by one thread.

So if there are n threads in your renderer, this lock will be engaged at most n times during the whole render.

AlexMWells: sounds good, just the name of the method wouldn't make me think it has the potential cost of a lock involved

lgritz:
Did I answer everybody's questions on this adequately, or are there outstanding concerns?

src/liboslexec/opcolor.h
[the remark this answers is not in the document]
sfriedmapixar: Got it, then LGTM.

PR #1406, opened 2021-09-15
Raise OpenEXR/Imath minimum to 2.3

lgritz:
Remember that our current 'master' is really aiming for a release in
late 2021 at the earliest, and thus will be a version primarily
intended to be used widely in 2022. Release branches never drop
supported dependencies, so the 1.11 branch will continue to support
OpenEXR all the way back to 2.0.

VFX Platform guidance is to support the current year, plus up to 3 years
prior, so that means that for 2022-ish software, we should support back
to 2019 dependency ecosystem.

For OpenEXR, 2019 state of the art was 2.3 -- that's still dreadfully
old, and is no longer getting even critical or security bug fixes. But
it still means that we can comfortably upgrade our advertised minimum
OpenEXR from 2.0+ to 2.3+ and eliminate some build and dependency
logic that was dedicated to dealing with 2.0..2.2.

We're also raising the OpenEXR/Imath "recommended minimum" (build-time
warning, but not rejection) to 2.4. Even that is shaky, but the fact
that it's still in wide use and was officially (though IMHO unwisely)
the VFX Platform 2021 recommendation, means we can't yell at people
too much about it.

It is possible that by the time we make a supported OSL 1.12 release,
we will bump even further to OpenEXR 2.4 as the minimum. We'll make
that decision later. It would depend on our being confident that even
though 2.3 is technically within the 3 year window we try to support,
we cannot discern that anybody is actually using it (who would also
need the most recent OSL in the same app).


fpsunflower (approved):
LGTM!

PR #1408, opened 2021-09-21
PR/BatchedColorOps

AlexMWells:
## Description

Add batched SIMD implementation for all color operations.
Added batched llvm code gen for: llvm_gen_blackbody, llvm_gen_luminance, llvm_gen_transformc (including code gen of loop to identify unique "from" and "to" space combinations to call library function with).

Made ColorSystem const correct.
Pulled much of ColorSystem's implementation into src/liboslexec/opcolor_impl.h so it could be inlined in both opcolor.cpp and wide_opcolor.cpp.
To enable SIMD fast path for ColorSystem::blackbody_rgb added new methods to expose underlying lookup table optimization.
    bool can_lookup_blackbody(float T /*Kelvin*/) const;
    Color3 lookup_blackbody_rgb (float T /*Kelvin*/) const;
    Color3 compute_blackbody_rgb (float T /*Kelvin*/) const;
When can_lookup_blackbody() returns true
then lookup_blackbody_rgb can be safely called,
otherwise the expensive compute_blackbody_rgb is required.

Updated implementation of ColorSystem to successfully vectorize on multiple compilers, optimized lookup table data types and indexed dereferences to generate preferred 32bit indexed gathers vs 64bit gathers which require multiple instructions and registers.

To improve code generation and less chances of aliasing issues, changed nested M[i][j] operator access to Matrix in ImathMatrix.h to directly access underlying 2d array M.x[i][j] also changed Vector operator V[i] to directly access V.x, V.y, V.z in OSL/IMathx/IMath*.h

Moved helper testIfAnyLaneIsNonZero from wide_opmatrix.cpp to be testIfAnyLaneIsNonZero in OSL/wide.h

Added sfm::min_val to provide an implementation of min which returns values not references to the original through a ternary (which can cause issues with vectorization).  Added Clang specific sfm::min_val and sfm::max_val implementations to assist in vectorization.

## Tests

Enabled BATCHED execution of existing testsuites: blackbody, color, transformc, wavelength_color
Added new testsuite BATCHED regression tests:  blackbody-reg, color-reg, luminance-reg, transformc-reg, wavelength_color-reg
Added Clang specific sfm::min_val and sfm::max_val implementations to assist in vectorization.

## Checklist:


src/include/OSL/Imathx/ImathMatrix.h
```diff
@@ -1133,7 +1133,7 @@ Matrix33<T>::equalWithAbsError (const Matrix33<T> &m, T e) const
 {
     for (int i = 0; i < 3; i++)
         for (int j = 0; j < 3; j++)
-            if (!IMATH_INTERNAL_NAMESPACE::equalWithAbsError ((*this)[i][j], m[i][j], e))
+            if (!IMATH_INTERNAL_NAMESPACE::equalWithAbsError ((*this).x[i][j], m.x[i][j], e))
```
lgritz, line 1136: Hey, Alex, just going over this again: Is this one really necessary? Because operator[](i) is already overloaded to be x[i]. So does the insertion if this particular .x really do anything? Is the fact that the operator is a method (even if inlined?) enough to throw off the vectorization?

AlexMWells:
```
template <class T>
IMATH_HOSTDEVICE inline T *
Matrix33<T>::operator [] (int i)
{
    return x[i];
}
```
@lgritz , It inlines, it's really the loss of type information.  Take it very literally from a compiler's point of view.  m.x[i][j] is a native access to a data member that is a 2 dimensional C array.  Compiler can track that back for SROA (Scalar Replacement of Aggregates) optimization pass.  Whereas m[i][j] is 2 operator calls, 1st returns a float * to data member, the 2nd operator subscript is an array access with an index.  As soon as a pointer is involved, compiler may not be able to track it back to a specific data member but instead do the 2 step grab pointer, then dereference pointer.  The bigger implication is the data layout may not be privatizable and the array access would need to be against standard data layout on the stack or memory vs. the possibility of keeping everything in registers (or SIMD registers when vectorizing).

lgritz: Gotcha, thanks. I'm porting these changes back to upstream Imath and wanted to be sure it all served a purpose.

AlexMWells: For my own Matrix classes, I don't even use arrays at all, just 9(3x3) or 16(4x4) data members and only support array subscript access with ConstIndex (std::integral_constant<int, #>) that is overloaded to access individual data members.  Basically removes ability for aliasing and pointer access that could get in the way of optimizations.

AlexMWells: Use of memset and memcpy in constructors might be another area of concern/tuning.  I think in my versions I directly set/copy all data members and let the compiler decide if it wants to replace it with something.  Main point is to avoid taking address of object which would then require standard data layout and an actual stack representation of something that might of been able to kept in a private data layout (SOA) in registers and only spilling registers as needed vs. the entire object.

PR #1409, opened 2021-09-22
Fix missing hostdevice modifier on some matrix operators

ix-bruiz:
## Description

Four of the matrix operators were missing the IMATH_HOSTDEVICE modifer and were thus not compiling in device code.

## Checklist:

This is just a simple compilation fix, so no tests other than compiling it in cuda, which we did on all supported platforms.


lgritz (approved):
LGTM, thanks for the fix!

PR #1411, opened 2021-09-23
Removing race condition when compiling for OptiX 7

cmstein:
## Description

This removes a race condition when compiling PTX device code in a multithreaded environment.

The problems stems from OptiX6 requiring the `rend_lib.cu` device bitcode to be linked to each compiled shadergroup whereas we only need to link the `rend_lib.cu` device bitcode to one shadergroup in OptiX7.  When I added OptiX7 support, I put a static/global in OSL to ensure that we don't get multiple definitions from `rend_lib.cu` in the device code for OptiX7.  But that's a poor solution for a couple of reasons:
- it's not thread-safe
- it doesn't support recompiling shaders in an interactive environment.

This removes the requirement to supply the `rend_lib.cu` device code when using OptiX7.  But that also means  the renderer now needs to manually link it in.  Both `testshade` and `testrender` have been updated to reflect this.

## Tests

testsuite runs the same with this fix

## Checklist:

### DIFF
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -1500,14 +1500,10 @@ BackendLLVM::run ()
         }
     }
 
-#if (OPTIX_VERSION >= 70000)
-    // FIXME:  I'm thinking this probably shouldn't be necessary, but I'm having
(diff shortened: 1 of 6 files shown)


lgritz (approved):
LGTM, and it passed testsuite for me by hand as well.

PR #1413, opened 2021-09-27
Pr/batched gen ops

AlexMWells:
## Description
Implement batched llvm code generation for ops: 

- andor
- bitwise_binary_op
- clamp
- get_simple_SG_field
- isconstant
- select
- unary_op
- mix

NOTE: clamp maybe unreachable as stdosl.h removed the builtin in 2010, but unless clamp is removed an operation through all of oso and scalar implementation, thought it should be here.

NOTE: andor maybe unreachable because "and" and "or" built-ins were removed in commit f7e8ed3, May 2011, release 0.5.4., but unless "and" and "or" is removed an operation through all of oso and scalar implementation, thought it should be here.

Fixed bug in BatchedAnalysis where the complement operator was being treated as always having a boolean result.  In reality the result of complement or other bitwise operation would only be boolean if its input parameters were forced to be boolean (left TODO note for future improvement).

Fixed bug in printf where integer who is forced_llvm_bool() is formatted as a float was not converted to an integer first.

## Tests
Enabled BATCHED for execution for tests:  

- and-or-not-synonyms
- isconstant
- logic
- select

Expanded testsuite/shaderglobals to exercise/access all shader global data members.

Add regression tests: 

- andor-reg
- bitwise-and-reg
- bitwise-or-reg
- bitwise-shl-reg
- bitwise-shr-reg
- bitwise-xor-reg
- complement-reg
- mix-reg
- select-reg
 

 
## Checklist:


lgritz (approved):
LGTM

PR #1414, opened 2021-09-28
Remove direct access to ShaderGlobals members from llvm_ops

lgritz:
This is a stepping stone to "shader globals placement".

Turns out that there were only two functions in llvm_ops.cpp that
directly needed to know the layout of the ShaderGlobals struct: the
implementations of osl_calculatenormal, and osl_raytype_bit.

For osl_calculatenormal, just pass the flipHandedness flag directly,
instead of the ShaderGlobals ptr, from which the field was extracted.

For osl_raytype_bit, I changed the function to only retrieve the bit
corresponding to a raytype name (and it's not in llvm_ops any more),
and the "bitwise and" code is just directly generated as bitcode.

I managed to complete the batched shading implementation of
calculatenormal! But for batched raytype_bit, I handled the case with
a constant name, but for the unknown name I got a little confused and
I'm hoping Alex can bail me out with some hand-holding or tell me
what needs to be done exactly.


fpsunflower:
LGTM for the scalar path.

src/liboslexec/batched_llvm_gen.cpp
```diff
@@ -4054,8 +4054,8 @@ LLVMGEN (llvm_gen_calculatenormal)
 
     llvm::Value *args[] = {
         rop.llvm_void_ptr (Result),
-        rop.sg_void_ptr(),
         rop.llvm_load_arg (P, true /*derivs*/, false /*op_is_uniform*/),
+        rop.llvm_global_symbol_ptr(Strings::flipHandedness),
```
AlexMWells, line 4058: As a matter of sanity checking, might consider  pulling the global_symbol_ptr out and passing the optional is_uniform parameter and OSL_ASSERT(!is_uniform) as the function being called will interpret the pointer as varying.

src/liboslexec/batched_llvm_gen.cpp
```diff
@@ -4313,19 +4313,15 @@ LLVMGEN (llvm_gen_raytype)
...
-            sg,
-            rop.ll.constant (rop.shadingsys().raytype_bit (name))};
-
-        llvm::Value *ret = rop.ll.call_function (rop.build_name("raytype_bit"), args);
-
-        if(!result_is_uniform)
-        {
+        llvm::Value* bit;
+        bit = rop.ll.constant(rop.shadingsys().raytype_bit(Name.get_string()));
+        llvm::Value* raytype = rop.ll.op_load_int(
```
AlexMWells, line 4318: Wondering if less is more, meaning was op_load_int needed, in that LLVM_Util::ptr_to_cast or LLVM_Util::ptr_cast and LLVM_Util::op_load which could be used to provide the same functionality.

src/liboslexec/batched_llvm_gen.cpp
```diff
@@ -4313,19 +4313,15 @@ LLVMGEN (llvm_gen_raytype)
     llvm::Value * sg = rop.sg_void_ptr();
     if (Name.is_constant()) {
         // We can statically determine the bit pattern
-        ustring name = ((ustring *)Name.data())[0];
-        llvm::Value *args[] = {
-            sg,
-            rop.ll.constant (rop.shadingsys().raytype_bit (name))};
-
-        llvm::Value *ret = rop.ll.call_function (rop.build_name("raytype_bit"), args);
```
AlexMWells, line 4321: You should be able to remove
OSL_BATCHOP int __OSL_OP(raytype_bit)(void* bsg_, int bit)
from wide_shadingsys.cpp, and 
DECL(__OSL_OP(raytype_bit), "iXi")
from builtindecl_wide_xmacro.h

src/liboslexec/batched_llvm_gen.cpp
```diff
@@ -4313,19 +4313,15 @@ LLVMGEN (llvm_gen_raytype)
...
+        llvm::Value* bitanded = rop.ll.op_and(raytype, bit);
+        llvm::Value* ret = rop.ll.op_int_to_bool_to_int(bitanded);
+        if (!result_is_uniform)
             ret = rop.ll.widen_value(ret);
-        }
         rop.llvm_store_value (ret, Result);
-
     } else {
         FuncSpec func_spec("raytype_name");
         // No way to know which name is being asked for
```
AlexMWells, line 4327: For the non-const case suggest following the same pattern you did for llvm_gen, go into wide_shading.cpp and rename the uniform and varying versions of raytype_name to raytype_bit.

For the uniform verson of raytype_name just return the bit by removing the  bitwise AND `return (bsg->uniform.raytype & bit)`, and in the batched_llvm_gen add a bitwise AND to the raytype from batchedShaderGlobals (very similar to llvm_gen version).

For the varying version, leave the foreach_unique loop, but again drop the bitwise AND `int ray_is_named_type = ((bsg->uniform.raytype & bit) != 0)` and just pass the `bit` variable into `assign_all`.  But back in the batched_llvm_gen you will now need to allocate a varying temporary int to pass to varying version of raytype_bit.  IE:
```
BatchedBackendLLVM::TempScope temp_scope(rop);

// We need wide place to hold the results of raytype_bit 
llvm::Value * loc_raytype_bit = rop.getOrAllocateTemp (TypeSpec(TypeDesc::INT), false /*derivs*/, false /*is_uniform*/, false /*forceBool*/, std::string("raytype_bit"));
            
llvm::Value *args[] = {
                sg,
                rop.llvm_void_ptr (loc_raytype_bit ),
                rop.llvm_void_ptr (Name),
                rop.ll.mask_as_int(rop.ll.current_mask())};
rop.ll.call_function (rop.build_name(func_spec), args);

llvm::Value* raytype= rop.ll.op_load(loc_raytype_bit );
llvm::Value* wide_bit = rop.ll.widen_value(bit);
llvm::Value* bitanded = rop.ll.op_and(raytype, wide_bit );
llvm::Value* ret = rop.ll.op_int_to_bool_to_int(bitanded);
OSL_DASSERT(!result_is_uniform);
rop.llvm_store_value (ret, Result);
```

To simplify, you can can probably combine some of the code between the uniform and non-uniform branches

src/include/OSL/llvm_util.h
```diff
@@ -858,6 +861,10 @@ class OSLEXECPUBLIC LLVM_Util {
     llvm::Value *op_int_to_bool (llvm::Value *a);
     llvm::Value *op_float_to_double (llvm::Value *a);
     llvm::Value *op_int_to_longlong (llvm::Value *a);
+    // int_to_bool_to_int is essentially turning any nonzero -> 1
+    llvm::Value *op_int_to_bool_to_int (llvm::Value *a) {
+        return op_bool_to_int(op_int_to_bool(a));
```
AlexMWells, line 866: Does op_int_to_bool_to_int generate better code than opt_bool_to_int(op_ne(a, constant(0)))?
If so might be useful elsewhere.

src/liboslexec/builtindecl.h
```diff
@@ -376,11 +376,11 @@ DECL (osl_trace_set_traceset, "xXs")
 DECL (osl_trace, "iXXXXXXXX")
 
 #ifdef OSL_LLVM_NO_BITCODE
-DECL (osl_calculatenormal, "xXXX")
+DECL (osl_calculatenormal, "xXXi")
 DECL (osl_area, "fX")
 DECL (osl_filterwidth_fdf, "fX")
 DECL (osl_filterwidth_vdv, "xXX")
-DECL (osl_raytype_bit, "iXi")
+DECL (osl_raytype_bit_from_name, "iXi")
```
AlexMWells, line 383: I don't see `osl_raytype_bit_from_name` referenced anywhere?
Although I do support using that name instead of just osl_raytype_bit.

src/liboslexec/shadingsys.cpp
```diff
@@ -4174,15 +4174,17 @@ osl_range_check_err (int indexvalue, int length, const char *symname,
 
 
 
+// Return the raytype bit corresponding to a raytype name.
 // Asked if the raytype is a name we can't know until mid-shader.
-OSL_SHADEOP int osl_raytype_name (void *sg_, void *name)
+// FIXME: We should pass the context, not the sg.
```
AlexMWells, line 4179: Why a FIXME, any technical barrier to passing context?  For that matter, I've been pondering a ReadOnlyContext or DeviceContext that might be a stripped down version of the full context to answer just these types of questions.

PR #1417, opened 2021-10-01
Fix oslc missing a write error on function parameter

### LINKED ISSUE #1416: oslc doesn't notice a violation of function output param rules
Pointed out by Lee, consider this shader:

```
void g(output float b) {
    b = 1;  // ok, since b is marked as output
}

void f(float a) {
    g(a);  // <--- should be disallowed, a is not output, but the arg of g() is output
}

shader test (output color Out_Color = 0)
{
    float x = 2;
    f(x);
    printf("%f\n", x);
    Out_Color = x;
}
```

oslc is missing a check somewhere, it should flag this as an error.

lgritz:
Fixes #1416

In OSL, function parameters must be marked as `output` or they are not
supposed to be writable. In issuing errors when a write is performed on
something not writable, we missed an interesting case:

```
void g(output float b) {
    b = 1;  // ok, since b is marked as output
}

void f(float a) {
    g(a);  // <--- should be disallowed, g is writing to f's non-output param
}

float x;
f(x);
```

This patch catches cases like these. The new error looks like this:

`test.osl:22: error: cannot write to non-output parameter "a", as passed to function g() as arg 1 (output parameter "b")`

### DIFF
--- a/src/liboslcomp/ast.cpp
+++ b/src/liboslcomp/ast.cpp
@@ -1003,8 +1003,10 @@ const char* ASTcompound_initializer::childname(size_t /*i*/) const
 
 
 bool
-ASTNode::check_symbol_writeability(ASTNode* var)
+ASTNode::check_symbol_writeability(ASTNode* var, bool quiet, Symbol** dest_sym)
 {
+    if (dest_sym)
+        *dest_sym = nullptr;
     if (var->nodetype() == index_node)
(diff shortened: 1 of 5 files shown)


lgritz:
@lkerley or others, does this seem ok? Is the form of that error message helpful?

lgritz:
Lee said LGTH

PR #1436, opened 2021-11-18
Pr/batched texture

AlexMWells:
## Description
Add batched support for 2d texture calls, 3d texture calls, environment, and gettextureinfo.
    
To enable, added:
```
bool RenderServices::is_udim (TextureHandle *texture_handle);

TextureSystem::TextureHandle* 
BatchedRendererServices::resolve_udim_uniform(...,TextureSystem::TextureHandle*, float S, float T);

void 
BatchedRendererServices::resolve_udim(...,TextureSystem::TextureHandle*,Wide<const float> wS, Wide<const float> wT, Masked<TextureSystem::TextureHandle*> wresult);
```
which allows batched code gen to resolve_udim's for varying S & T and bin the resulting TextureHandle* and make calls to get_texture_info_uniform with a single resolved udim texture handle, then broadcast the results to data lanes with matching TextureHandle *'s.
    
 Batched texture options are implemented on the llvm code gen stack using a temporary and directly populating vs. calling out to helper functions.  Strategy is to have any varying inputs to uniform members of the batched texture options be binned, so calls to the underlying texture subystem only have to deal with an expected set of uniform options and subset of varying options.  This means that the batched code gen handles extracting the 1st active lane of varying data from variables populating texture options and reducing the mask to lanes that match the same options, we refer to this as binning.

NOTE:  The default BatchedRendererService implementation turns around and calls the texture subsystem once for each active lane, it will be up to the texture subsystem to add support to processes multiple texture requests in the future to better take advantage of outer loop SIMD.
## Tests
Added extensive new regression tests:
```
gettextureinfo-reg
gettextureinfo-udim-reg
texture-opts-reg
texture3d-opts-reg
texture-environment-opts-reg
```    
Enabled BATCHED execution of existing tests:
```
gettextureinfo-udim
gettextureinfo
texture3d
texture-alpha-derivs
texture-alpha
texture-blur
texture-connected-options
texture-derivs
texture-environment
texture-errormsg
texture-firstchannel
texture-interp
texture-missingalpha
texture-missingcolor
texture-simple
texture-smallderivs
texture-swirl
texture-udim
texture-width
texture-withderivs
texture-wrap
```
## Checklist:


AlexMWells:
Discovered mismatch between documentation, llvm_gen implementation, and oslc when passing user derivatives to texture3d.
oslc and documentation expect optional parameters "vector dpdx, vector dpdy, vector dpdz",
however llvm_gen and optexture implementation only accept "vector dpdx, vector dpdy"
and use 0 for dpdz when calling into the texture subsystem.

This commit includes changes to oslc to make the compiler match the implementation of accepting only "vector dpdx, vector dpdy".  This change could have (un)intended consequences, such as causing existing .osl shaders that explicitly pass dpdz fail to compile.  This is probably preferable to inform user that there is an issue.  The osl-language.pdf still needs to be updated, assuming that not supporting dpdz is the intended behavior.  Otherwise, llvm_gen and optexture would need to be updated to properly handle dpdz.

AlexMWells:
After conferring with lgritz, it was determined that user provided dPdz should be respected and llvm_gen should be fixed to properly handle it.  [commit ](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1436/commits/a23554566fd872942056898db2429923fda2b59f) reverts the changes made and fixes the behavior for texture3d when user provided derivates are passed.

lgritz (approved):
LGTM

PR #1437, opened 2021-11-19
Restore unit tests for texture3d

lgritz:
OIIO master recently removed support for Field3D, which is rarely used
and the project is effectively unmaintained and isn't able to build
against the newer OpenEXR releases that we want to encourage people to
use.

So the primary purpose of this patch is to switch our unit test for
texture3d from using a field3d file (and thus would not run when using
newer OIIO) to a roughly equivalent test using openvdb (which is
supported, assuming OIIO is built in the presence of OpenVDB).

But, ha, I discovered two things, either one of which means that we
had actually not been unit testing texture3d at all lately: (1) To
discover whether the installed OIIO supported openvdb (and therefore
whether to bother running the test), we were relying on a totally
nonexistant OPENIMAGEIO_BIN cmake varaible. I don't know what that was
supposed to be (did I mean OIIOTOOL_BIN? but that also is not a
thing), but anyway, it caused it to always skip the test. (2) Oh boy,
the OIIO that we were relying on for may tests, provided by the ASWF
docker containers, was an OIIO that was not built in the presence of
OpenVDB (even though our OSL testing containers DO have OpenVDB), so
again it skipped the test because it needs that OIIO-side support for
texture3d() to know how to read openvdb files. Fix this by building
OIIO from scratch with openvdb support when doing the "VFX Platform
2022" test, rather than relying on the one in the container.


lgritz:
@AlexMWells This reinstates a unit test for texture3d, but it's still enabled only if the OIIO you are using was built in the presence of OpenVDB.

AlexMWells (approved):
LGTM

PR #1438, opened 2021-11-24
Add testsuite coverage of environment()

lgritz:
Can you believe we never had a test?


AlexMWells (approved):
LGTM, note PR/BatchedTexture adds testsuite/environment which is pretty similar to texture-environment here.  I suggest they both live, then maybe one goes away after both PR's are merged.

lgritz:
Ha! I almost named this "environment", but the reason I didn't is that there is utility in having all the tests whose primarily purpose is to exercise the texture system have a common prefix. So I think that the easiest thing to do is after I merge this, you should cherry-pick this commit into your PR/BatchedTexture branch, and then you can eliminate the redundant "environment" test.

AlexMWells:
Sounds, good.  Lets get this merged then so I can rebase PR/BatchedTexture, then I will rename my environment-opts-reg to texture-environment-opts-reg to line up.

lgritz:
Already merged

lgritz:
I don't think you have to rebase, if that's a pain. It's sufficient to just cherry-pick this commit into your branch, then keep building on top of that. It will all get squashed together when we do the final merge of this PR.

PR #1439, opened 2021-11-24
Fix minor CI issues

lgritz:
* Increase testing timeout, some of the batch shading regression tests
  seem to take longer than the prior 240s limit.

* Turn off OpenVDB when using the osl:2021 in combination with OpenEXR
  3.x -- the version of OpenVDB in the 2021 container is too old to
  understand OpenEXR 3.


AlexMWells (approved):
LGTM.

PR #1440, opened 2021-11-24
Testing: Adjust timeout strategy

lgritz:
I recently bumped the ctest per-test timeout to 300s because one of the
batched shading regression tests was occasionally taking longer than the
prior limit of 240s.

But there is a better way: Individual tests can have their timeouts
set individually via set_tests_properties(TIMEOUT), so I raised the
timeout to 300s for all tests whose names contain "-reg", also kept it
rather long (the old 240s) for the long render tests that are all named
"render-". Then for everything else, we can actually lower the default
to 60s, since no other tests should be long-running.

(For the curious, the reason these batched regression tests are so
expensive is that they test a bunch of combinations, and they first
run the test in the non-batch scalar mode, then again in batched mode,
and compare the results to ensure that the batch code never deviates
from scalar. This makes them take at least twice as long as they
ordinarily would. But also remember that these are only run for the
few CI test matrix entries where we are specifically testing batch
shading.)

### DIFF
--- a/src/build-scripts/ci-test.bash
+++ b/src/build-scripts/ci-test.bash
@@ -13,7 +13,7 @@ $OSL_ROOT/bin/testshade --help
 echo "Parallel test " ${CTEST_PARALLEL_LEVEL}
 pushd build
 time ctest -C ${CMAKE_BUILD_TYPE} -E broken --force-new-ctest-process \
-    --output-on-failure --timeout ${CTEST_TEST_TIMEOUT:=300} ${CTEST_ARGS}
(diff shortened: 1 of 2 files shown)


AlexMWells (approved):
LGTM.
Out of curiosity, could these timeouts be overridden per test if necessary?

lgritz:
Absolutely. At the bottom of `osl_add_all_tests()` in testing.cmake, you can just add special cases like

    set_tests_properties (my_test_name PROPERTIES TIMEOUT 800)

PR #1442, opened 2021-11-24
Docs: fix missing "errormessage" explanation for environment and texture3d

lgritz:
It was documented for texture(), but not for the others, for which it
also has worked for many years.

Thanks for Alex Wells for pointing this out.


AlexMWells (approved):
LGTM

PR #1445, opened 2021-12-01
Pr/batched dict

AlexMWells:
## Description
Add batched support for dict find, dict next, dict value.
Implementation takes advantage of uniform values and only does lookup once and broadcasts results if necessary.
When nodeid is varying, it falls back to simple loop over active data lanes performing lookups.

## Tests
Added new regression test xml-reg.
Enabled BATCHED execution of existing xml test

## Checklist:


lgritz (approved):
LGTM

PR #1447, opened 2021-12-14
RuntimeOptimizer: Fix messages & unknown_message_sent initialization

lgritz:
Alex Wells pointed out that they were never explicitly initialized.


AlexMWells (approved):
LGTM

PR #1448, opened 2021-12-21
PR/BatchedMessage

AlexMWells:
## Description

Add batched support for trace, setmessage, and getmessage

Added MaskedData::assign_from(void *ptr_wide_data) to perform a masked copy of wide data at some address.

Added enum class RenderServices::TraceOpt::LLVMMemberIndex to correlate C++ data members with LLVM type generated by BatchedBackendLLVM::llvm_type_batched_trace_options()

Fixed bug in BatchedBackendLLVM::llvm_load_arg where array types where only properly broadcasting uniform to varying arguments for the 0th element of the array.

For batched implemented trace options as a LLVM type with direct code gen to alloca, init default values, and set optional values vs. callbacks.  Same code could be resused for non-batched impl.

Added BatchedMessageBuffer to context and BatchedMessageList to wide_opmessage.cpp to allow a shared buffer to be cleared by the ShadingSystem while allowing target specific compilation of SIMD import/export of message values into the list to compile in wide_opmessage.cpp.

## Tests

Extended testshade's SimpleRenderer and BatchedSimpleRenderer to implement trace and getmessage to allow new tests to exercise some basic functionality.

Added new batched regression tests:
-   message-reg
-   trace-reg

Added new batched test:  

- message-no-closure (which is a simplied version of testsuite/message)

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -548,6 +548,7 @@ jobs:
       PYTHON_VERSION: 3.9
       OSL_BUILD_MATERIALX: 1
       USE_BATCHED: b8_AVX2,b8_AVX512,b16_AVX512
+      CTEST_ARGS: -L message-reg
     steps:
       - uses: actions/checkout@v2
       - name: Prepare ccache timestamp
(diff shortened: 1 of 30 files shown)


lgritz (approved):
LGTM.

The failure on the GPU test is not due to you, I'll look into that separately.

lgritz:
I think something newer must be slightly interfering. @AlexMWells, can you do a rebase on top of the current main before we merge this?

AlexMWells:
yes, I will rebase again

PR #1450, opened 2022-01-03
Remove out of date MaterialX shaders

lgritz:
We used to build MaterialX shaders on our end. Now it's all done in
the MaterialX project itself, and along the way, we failed to keep in
sync with them. (And even if we had "kept up", what does that even mean?
We want OSL to be independent of any particular MX release, but the MX
shaders may change from release to release. Better to be entirely on
their side of the fence.)

We turned off building these shaders by default a long
time ago, now let's finally remove the last remnants and urge everyone
to rely on the MaterialX project for this material.


fpsunflower (approved):
LGTM!

lgritz:
oops, closed this by accident

PR #1451, opened 2022-01-04
windows compile fix

debaetsd:
## Description

This tiny PR fixes a compile error on Windows

## Tests

it compiles on windows while without this, it does not
CI should catch any regressions wrt other platforms

## Checklist:


lgritz (approved):
LGTM, thanks for the fix.

lgritz:
The CLA bot thinks you don't gave a CLA on file. If you think you already filed one, then are you maybe using a different github ID or email than usual?

debaetsd:
@lgritz I had indeed used the wrong email (hence the force pushed commit) but at the same time I hadn't signed the post-ASWF-move CLA yet so I think it should now be fine

lgritz:
Yep, CLA looks good now, thanks.

But now you need the DCO!  The fix for that is just to `git commit --amend -s` which will add the "signed-off-by" line, and then `git push --force` your branch to update the PR.

There's a failed test for "Linux GPU", but that's unrelated to you, so don't worry about it. It looks like maybe something has changed in the nvidia headers that we download in order to do an OptiX build. I won't let that hold up your patch.

PR #1453, opened 2022-01-12
Pr/enable batched tests

AlexMWells:
## Description
Enable BATCHED execution of most testsuite entries that don't use closures or point cloud.
Ran into missing hash function, and behavior differences between batched and scalar execution which this PR fixes:

Added batched implementation of hash function.

Updated batched implementation to use the symbol.get_int(), symbol.get_float(), and symbol.get_string() methods vs. *(int*)symbol.data() approach.

Fixed bug in batched code gen where mangled symbol names were output as part of debug uninit warnings and range checking.

Updated batched behavior for unlocked geom shader output parameters to function, like scalar version.
Fixed batched bug when non-entry layers are executed and execution masks are compared producing a <i1x16> instead of i1 for use in conditional operation.

Enabled command line assignable user data for batched execution in testshade, which fixes testsuite/userdata-passthrough

Changed print_outputs for batched execution to not use << operator which includes parenthesis, but instead output each componenet with spaces to match the non-batched print_outputs.

renamed MaskedData<...>::assign_from(void* ptr_wide_data) to assign_val_from(const void* ptr_wide_data) to better reflect that only the value is assigned, and Dx and Dy are NOT assigned.
Added MaskedData<...>::assign_all_from(const void* ptr_data) to take a scalar DataT and broadcast it to all lanes of the MaskedData, masked off lanes are NOT overwritten, and value, Dx, Dy are assigned.

## Tests

Enable BATCHED execution of all testsuite entries that don't use closures or point cloud.
Added more reparameter test cases to cover arrays and arrays of aggregate types.
Wrapped all printf arguments with pretty() in testsuite/geomath/test.osl to avoid small floating point differences from affecting printed results in test.

## Checklist:


lgritz (approved):
LGTM, and Alex and I went over it in detail on a conf call.

PR #1455, opened 2022-01-21
PR/BatchedPlacement

AlexMWells:
## Description

Implement batched version userdata and output placement.

Added Wide<const int, WidthT> wide_shadeindex, void* userdata_base_ptr, void* output_base_ptr to BatchedExecutor interface (and underlying implementation) that allows each data lane
of a batch to utilize a different shade index in combination with a uniform userdata and output base pointer.

Fixed bug in batched version of llvm_gen_compassign where index was being loaded as uniform value even when it was varying (really just caused assert/extra output in Release).
For scalar and batched, avoid generating code for init ops or default values when a sym_loc is found because unlike get_userdata, when supplied sym_loc can't fail.

Improved LLVM_Util::op_scatter to handle scalar destination data types, previously only wide destination were required but to handle scatter results from SOA wide data layout to AOS
output placement layout scalar destinations need to be supported.

Updated BatchedRendererServices to be less abstract replacing several pure virtual functions with default noop versions to better match behavior of RendererServices.

## Tests

Changed testshade to utilize output placement by default in batched mode.
Added NOSCALAR option to testing framework to allow for batched only tests.
Added testsuite/example-batched-deformer to exercise userdata placement and simple example of utilizing batched interface.

## Checklist:


lgritz (approved):
LGTM

PR #1456, opened 2022-01-24
PR/BatchedFpTests

AlexMWells:
## Description

Implement batched versions of isnan, isinf, and isfinite.

## Tests

Added batched regression testsuite ieee-fp-reg.

NOTE:  its really difficult to test isnan(value) == true,
as its unclear to actually generate a NAN in OSL because
most functions are "safe" and check/avoid NAN's

## Checklist:


lgritz (approved):
LGTM

PR #1457, opened 2022-01-26
oslc: detect using type name as an variable identifier

lgritz:
Here was the construct it was confused by:

    struct Data {
        float height;
    };

    Data foo(Data a, float b)
    {
        return Data;   // <----- what's this madness?
    }

Curiously, this revealed a source code error in one of our other tests,
oslc-err-struct-print, which had not been detected previously!


lgritz:
@lkerley

fpsunflower (approved):
LGTM

PR #1458, opened 2022-01-28
Wean ourselves off of OIIO::string_view::c_str()

lgritz:
OIIO's string_view::c_str() was cute, but it's a little dicey and it's
not present in the C++20 std::string_view. We didn't use it in a ton
of places, so replace with other idioms so that in the future we are
better able to interchange with or switch to std::string_view (when
our minimum supported C++ standard is high enough).

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -254,7 +254,8 @@ public:
         return attribute (name, TypeDesc::FLOAT, &f);
     }
     bool attribute (string_view name, string_view val) {
-        const char *s = val.c_str();
+        std::string valstr(val);
+        const char *s = valstr.c_str();
         return attribute (name, TypeDesc::STRING, &s);
     }
 
@@ -284,7 +285,8 @@ public:
         return attribute (group, name, TypeDesc::FLOAT, &f);
     }
     bool attribute (ShaderGroup *group, string_view name, string_view val) {
-        const char *s = val.c_str();
+        std::string valstr(val);
+        const char *s = valstr.c_str();
         return attribute (group, name, TypeDesc::STRING, &s);
     }
 
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -3912,7 +3912,7 @@ ClosureRegistry::register_closure (string_view name, int id,
              * because we will be allocating the real struct inside it. */
             OSL_ASSERT_MSG(params[i].field_size <= int(alignof(ClosureComponent)),
                 "Closure %s wants alignment of %d which is larger than that of ClosureComponent",
-                name.c_str(),
+                name.str().c_str(),
                 params[i].field_size);
             break;
         }
--- a/src/liboslcomp/codegen.cpp
+++ b/src/liboslcomp/codegen.cpp
@@ -695,12 +695,12 @@ ASTNode::one_default_literal(const Symbol* sym, ASTNode* init, std::string& out,
             float f = lit->intval();
             for (int c = 0; c < 16; ++c)
                 out += Strutil::sprintf("%.9g%s", (c / 4) == (c % 4) ? f : 0.0f,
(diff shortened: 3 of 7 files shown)


src/include/OSL/oslexec.h
```diff
@@ -254,7 +254,8 @@ class OSLEXECPUBLIC ShadingSystem
         return attribute (name, TypeDesc::FLOAT, &f);
     }
     bool attribute (string_view name, string_view val) {
-        const char *s = val.c_str();
+        std::string valstr(val);
```
AlexMWells, line 257: Will this create an extra copy for the local variable valstr?

lgritz: Yeah, it'll create a temporary std::string. That seems a little icky, but remember that what this is replacing was potentially creating a ustring, so it's probably not a lot worse.

This is called maybe a few dozen times, tops, to set up the shading system. So I don't think there is a legit perf issue here either way.

If this were called VERY frequently, I would add specializations for std::string and ustring that would avoid the temp string creation, but I'm not sure it's worth it for a method that's called a handful of times total.

fpsunflower: The same pattern happens a few times, but I think the concern is that the `void*` flavor of `attribute()` (which accepts a `TypeDesc` to say what the `void*` is -- assumes a C-style null-terminated string.

So unfortunately we can't support these cases without forcing a copy from the `string_view`. In the future we could redefine that case to accept a pointer to a `string_view`, but it would be a behavior change.

On a sufficiently modern c++ standard library, short strings would at least not go through the heap.

src/liboslexec/shadingsys.cpp
```diff
@@ -3912,7 +3912,7 @@ ClosureRegistry::register_closure (string_view name, int id,
              * because we will be allocating the real struct inside it. */
             OSL_ASSERT_MSG(params[i].field_size <= int(alignof(ClosureComponent)),
                 "Closure %s wants alignment of %d which is larger than that of ClosureComponent",
```
fpsunflower, line 3914: Beyond the scope of this review -- but would it be possible for these macros to use the fmt::printf under the hood?

fpsunflower (approved):
LGTM

src/include/OSL/oslexec.h
AlexMWells: Ok, so string_view (OIIO or C++17) doesn't require null terminated string.  So I guess the copy is necessary to ensure null termination.  Does make me think that the next call down might be better to add direct support for string_view (vs. const char *) assuming it's going to make its own copy anyway.

src/liboslexec/shadingsys.cpp
lgritz: Yes, I think so. I believe when I first wrote this macro, it may have been prior to use of the formatted output template magic, and also I was a little hesitant because fmt::printf is a complicated implementation that made me nervous about calling in the middle of a panic bad enough to cause us to assert. But maybe it would be fine.

src/include/OSL/oslexec.h
lgritz: I could make additional overloads for ustring and/or std::string that could avoid the copy. 

But these are not inner loop calls with performance implications. We're talking about at most tens of calls per execution of the program.

lgritz: For the record, current there are 11 documented ShadingSystem::attribute() settings whose value is a single string. Five of them are only useful for my debugging of the implementation of OSL.

This is not a performance issue.

sfriedmapixar: I agree this isn't a performance issue.  Worst remotely reasonable case is if you want to change all the renderer_outputs or the shadersearchpath per shader, and even then you're only going to get in the thousands of calls, maybe up to megabytes of allocations in the most extreme situations -- and in those situations that should be dwarfed by shader compile and execute time and space.

AlexMWells (approved):
LGTM

PR #1459, opened 2022-02-01
CI: Add test that uses the Intel icc/icpc compiler.

lgritz:
Current status:

* I have tried to set up two tests, one on the native Ubuntu, one on
  an ASWF CentOS-based container for VFX Platform 2022. (We may not
  keep both of them, I had an easier time getting the Ubuntu running,
  but I wanted to see if I could get a faster total test on the ASWF
  container since it would let me avoid building several dependencies
  from scratch.)

* The Ubuntu-based test (which uses apt-get) gets the Intel
  components, builds, runs tests... but fails a number of tests. Some
  appear to be differing reference outputs (LSB stuff), but a few are
  fast failures because it can't read the compiled .oso files -- and
  inspecting them reveals stray NUL characters in the file, where they
  shouldn't be, and always right in front of certain floating point
  constants (suspect maybe a bug in fmt library when using icpc?).

* The CentOS-based test (which uses yum) never gets as far as installing
  the Intel components -- I get an error message about not finding the
  correct GPG keys for the repo, and I'm not sure what's going wrong.
  I think I'm following the instructions correctly that I see here:
  https://www.intel.com/content/www/us/en/develop/documentation/installation-guide-for-intel-oneapi-toolkits-linux/top/installation/install-using-package-managers/yum-dnf-zypper.html


lgritz:
@AlexMWells I would love some insight from you about the things that are broken. Do any of the test case failures ring a bell for you? I'm also not sure how the version of icc that's in this package differs from what you may be using.

lgritz:
@AlexMWells Three other oddities that I noticed and worked around for now, but that we'll need to come back to and I'm curious about your thoughts:

* I pretty much had to use CMAKE_CXX_STANDARD=17, because when I used 14, icpc gave me errors about the OSL_FALLTHROUGH macro.

* I had to bend over backwards to make sure that the dependencies built with gcc even though OSL itself is building with icc, because icc was complaining about things in OIIO's simd.h while building OIIO (something about forceinline changing between a template's initial declaration and its later full definition). But that's weird, because OSL also uses OIIO's simd.h, but I didn't see the same errors when building OSL! Not sure what to make of that.

* I got errors when building OpenColorIO. I'm not sure if anybody's ever built OpenColorIO with icc before, so I don't know if this is something spurious about our build setup, or if OCIO source code is not icc friendly. It might be something to follow up with that project after we get the rest of our own house in order.

lgritz:
CMake identifies the compiler as `The CXX compiler identification is Intel 2021.5.0.20211109`

Is that icc classic, or is that the new compiler? If the latter, that might explain why I get lots of warnings and errors that you don't seem to get.

AlexMWells:
Great start Larry, I will have to try to duplicate to properly answer your questions.

src/build-scripts/gh-installdeps.bash
```diff
@@ -41,6 +39,14 @@ if [[ "$ASWF_ORG" != ""  ]] ; then
         export LLVM_ROOT=$PWD
         popd
     fi
+
+    if [[ "$CXX" == "icpc" || "$CC" == "icc" || "$USE_ICC" != "" ]] ; then
+        sudo mv src/build-scripts/oneAPI.repo /etc/yum.repos.d
```
AlexMWells, line 44: Any reason this is a "mv" instead of a "cp"?

lgritz: oops, no, probably just a cut and paste error

AlexMWells:
Just looking at the CI / Linux icc ubuntu, the build itself appears to be completing.
However as it as configured with only:
`USE_BATCHED: b16_AVX512`
Only an AVX512 build is happening.  And the machine it just happened to execute CI on was an:
`Intel(R) Xeon(R) CPU E5-2673 v3 @ 2.40GHz`
which only has `avx` and `avx2` as reported in the build setup.

The result is that when batched regression tests run, they fall back to single point execution but emit:
`-WARNING:  Hardware or library requirements to utilize batched execution llvm_jit_fma(1) are not met, ignoring batched and using single point interface to OSL`
which ultimately can cause tests to fail because now captured output doesn't match:
`NO MATCH for  out.txt
FAIL  out.txt`

Simplest fix is to modify ci.yml to add AVX and AVX2 builds so that it will hopefully pass CI when running on older machines.  
`USE_BATCHED: b8_AVX,b8_AVX2,b16_AVX512`
Or if anyway to control CI to only run on machines with AVX512.

lgritz:
UPDATED!

Fully builds and all tests pass except for one! (Still chipping away at it.)

lgritz:
Current status: WORKS!

@AlexMWells

testsuite/exponential/ref/out.alt-icc.txt
```diff
@@ -0,0 +1,67 @@
+Compiled test.osl -> test.oso
```
AlexMWells, line 1: There are several naming conventions used for the alternative results, should they be standardized some.

AlexMWells (approved):
LTGM

lgritz:
I think that all of your suggested changes were already incorporated into the latest version I had sent, Alex. Those diffs appear to be against an older version.

testsuite/exponential/ref/out.alt-icc.txt
AlexMWells: Examples of 3 differing naming conventions (assuming they are not out of date): 
[testsuite/geomath/ref/out.icc.txt]
[testsuite/exponential/ref/out.alt-icc.txt]
[testsuite/trig/ref/out.alt-icc.txt]
[testsuite/vector/ref/out-icc.txt]

PR #1460, opened 2022-02-05
Favor our choice of Imath includes over OIIO's.

lgritz:
New changes in OIIO guards the contents of its Imath.h with the
`OIIO_IMATH_H_INCLUDED` preprocessor symbol. This lets us in OSL
pre-define it to absolutely ensure that OSL headers are including
exactly the Imath headers they want, not accidentally getting the ones
that OIIO includes.

See https://github.com/OpenImageIO/oiio/pull/3322
This present PR is harmless for versions of OIIO prior to that PR being
integrated.

### DIFF
--- a/src/include/OSL/oslconfig.h.in
+++ b/src/include/OSL/oslconfig.h.in
@@ -4,9 +4,9 @@
 
 #pragma once
 
-#include <OSL/platform.h>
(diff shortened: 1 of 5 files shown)


lgritz:
Does this seem ok?

PR #1462, opened 2022-02-11
Add N/I to globals needed list when backfacing() shadeop is used

sfriedmapixar:
## Description

Fix a bug where utilizing the backfacing() shadeop would not mark the N and I globals as needed.

During the OSL runtime optimizations, if we find a backfacing() shadeop, we now add both N and I
to the list of needed globals, as the underlying implementation of that shadeop relies on them.

## Tests

This is tested by the globals-needed test.  Additionally it adds the ability to filter
out.txt in a test before diff'ing by including a regular expression for the filter in the filter_re variable.
This was so that the output from the debug mode of testshade could be limited to only
the line of interest to this test.

## Checklist:

- [ x] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [ x] I have previously submitted a [Contributor License Agreement](http://opensource.imageworks.com/cla/).
- [ x] I have ensured that the change is tested somewhere in the testsuite (adding new test cases if necessary).
- [x ] My code follows the prevailing code style of this project.

### DIFF
--- a/src/cmake/testing.cmake
+++ b/src/cmake/testing.cmake
@@ -209,6 +209,7 @@ macro (osl_add_all_tests)
                 function-overloads function-redef
                 geomath getattribute-camera getattribute-shader
                 getsymbol-nonheap gettextureinfo gettextureinfo-reg
+                globals-needed
                 group-outputs groupstring
                 hash hashnoise hex hyperb
                 ieee_fp ieee_fp-reg if if-reg incdec initlist initops intbits isconnected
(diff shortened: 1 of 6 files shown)


sfriedmapixar:
My EasyCLA manager is out on vacation for the next week, so this may be in limbo for a bit.

lgritz:
Its been like this for a LONG time, I don't think there's a rush.

But I like the fixes. And nice trick with the testsuite option to filter the output!

sfriedmapixar:
Ok, got my EasyCLA approval through.  The bleeding-edge CI failure looks like a main/master rename issue.  Do I need to rebase this before you can pull it, or is it good-to-go?

lgritz:
The bleeding edge failure was already fixed by #1463, no need to worry about that. I'll just do the merge.

PR #1464, opened 2022-02-15
Pr/batched point cloud

AlexMWells:
## Description

Implement Batched pointcloud_search, pointcloud_get, and pointcloud_write.

Updated behavior of scalar pointcloud_search and pointcloud_get to not code gen branch with code gen runtime error for maxpoints/num_points larger than array size.  Instead now it codgens a clamp of the maxpoints/num_points to the arraysize.  This allows code to continue executing, and if accessed outside of the range, existing codgen range checks will still emit a runtime error.

Extracted class PointCloud definition and some helper functions to pointcloud.h so it can be shared between batched and scalar implementation.
Moved PointCloud out of anonymous namespace to OSL::pvt so a single implementation (including static maps) could be shared between batched and scalar implementation.

Refactored implementation of 
template<typename DataT, int WidthT> struct Masked;
template<typename DataT> struct Ref;

to flatten class hierarchy and fix bug where derivatives of unbounded arrays could not be constructed, IE:
MaskedDx<int[]> myDeriv(ptr, arraylength);

Added                                  MaskedData::assign_val_lane_from_scalar
Renamed MaskedData::assign_val_from to MaskedData::assign_val_from_wide
Renamed MaskedData::assign_all_from to MaskedData::assign_all_from_scalar

Updated BatchedAnalysis to indicate that results of pointcloud_search, pointcloud_get, and pointcloud_write are always varying regardless of the symbols passed in.
Updated debugging in BatchedAnalysis to emit unmangled symbol names.

Properly implement (was broken previously) BatchedBackendLLVM::llvm_zero_derivs(const Symbol& sym, llvm::Value* count) including using code gen loop to reduce code bloat.

Fixed bug in llvm_gen for raytpye where the wrong type was passed as a name parameter .  Only showed up when trying regression tests with optimizations disabled.
## Tests

Expanded testsuite/pointcloud adding 2 additional write cloud tests and 10 additional read cloud including conditionally reading/writing and varying filename, maxpoints, and sort arguments.

## Checklist:


src/cmake/testing.cmake
```diff
@@ -167,11 +167,18 @@ macro ( TESTSUITE )
...
-            if (EXISTS "${_testsrcdir}/BATCHED_REGRESSION")
+            if ((EXISTS "${_testsrcdir}/BATCHED_REGRESSION")
+                AND NOT EXISTS "${_testsrcdir}/NOOPTIMIZE")
                 # optimized for right now
                 add_one_testsuite ("${_testname}.regress.batched.opt" "${_testsrcdir}"
                                    ENV TESTSHADE_OPT=2 OSL_REGRESSION_TEST=BATCHED )
             endif ()
+            if ((EXISTS "${_testsrcdir}/BATCHED_REGRESSION")
+                AND EXISTS "${_testsrcdir}/NOOPTIMIZE")
+                # optimized for right now
```
lgritz, line 178:
```suggestion
                # not optimized
```

lgritz (approved):
LGTM. I'm going to merge now so you can move on. There was one place I spotted an incorrect comment in testing.cmake, we can just fix that on the next PR.

lgritz:
Oh neat, if a committer expresses a review comment as a "suggestion", it lets you merge the suggestions directly!

PR #1468, opened 2022-02-25
CI: Fix botched handling of OPENIMAGEIO_CMAKE_FLAGS during CI builds

lgritz:
The gh-installdeps.bash file accidentally squashed any value that
the env variable already had. Oops.


lgritz:
@AlexMWells This is what was preventing proper setting of the fmt version by the CI scripts!

PR #1472, opened 2022-03-01
Testsuite: allow using oiiotool for image diffs instead of idiff

lgritz:
Additionally, allow custom command line arguments to run after each
image is loaded, but before the comparison.


AlexMWells (approved):
LGTM

PR #1476, opened 2022-03-15
Fixes for recent OIIO TextureSystem API changes

lgritz:
Changed a function signature in OIIO master that threw us off here
just a tad, as there was no direct type conversion the compiler could
figure out on its own. Easy adjustment gives it the right hint.


lgritz:
Merging without review because things are super broken without this (if you build from OIIO master).

PR #1479, opened 2022-03-28
Fix GPU codegen crash

lgritz:
llvm_load_device_string wasn't considering temporary string variables,
only declared local variables.


cmstein (approved):
LGTM!

PR #1480, opened 2022-03-29
Hide Matrix22 definition better -- it's not part of any public API

lgritz:
Also change the one exception mentioned to a std one, matching
Imath 3.x and insulating us from any of the old Iex custom exceptions
from Imath 2.x.


AlexMWells (approved):
LGTM

PR #1483, opened 2022-04-08
Add Visual Studio Code folder to gitignore

fpsunflower:
Just testing out commit workflow from windows 😄

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ build
 \#*\#
 *.pyc
 *.log
+.vscode
+


lgritz (approved):

lgritz:
Now all we need is Windows in our CI!

fpsunflower:
I'll certainly try to think about this. The big one as you said is figuring out an easy way to pull the LLVM dependency.

It looks like I don't have write access to the repo to merge the PR ?

lgritz:
Oops, I'm not sure how that happened. Maybe when you were removed from the imageworks org, we neglected to explicitly add you also as a committer to osl? Huge oversight. Invite sent; let me know if you're able to accept it and merge.

fpsunflower:
All good! That's why I tested the waters with a trivial change :)

PR #1491, opened 2022-04-18
WIP Refactor common parts of [Batched]BackendLLVM to BackendLLVMCommon

lgritz:
I've been thinking about the long-term maintenance burden of the
separate scalar and batched LLVM back ends, and trying to ease the
work burden and tendency to make errors that will result from the
large amount of replicated and nearly-replicated code.

The scalar (single-point) shading back end is more established and
well understood, is much simpler, and is more commonly used (and
probably always will be -- it will be a minority of renderers that are
structured to take advantage of batched execution and that run on
architectures where it makes sense). So I think that looking toward
the future, many contributors and maintainers of this code base will
naturally be more familiar with the single-point internals. I'd like
to make it so that most changes can be made in one place, and by
people who know the single point code well, with the batch code
occupying as small a footprint as practical, to make it more rare to
need to change it, and easier / smaller changes when it does also need
modification.

In this patch, I introduce a BackendLLVMCommon class, derived from OSO
OSOProcessorBase, and inherited by both BackendLLVM and
BatchedBackendLLVM. I try to move a bunch of the replicated code from
those two back ends into the common superclass.

I'm not "done", just did enough to give the gist so you can see where
I'm going with it and enough to see that a bunch of code replication
has been eliminated. I think a lot more can be done in this direction,
ideally with BatchedBackendLLVM eventually containing *ONLY* the parts
that are relevant to batch only or must be significantly different
than the scalar back end. (I don't mind the scalar back end having a
little bit of incidental support for simd types and the like; it's
unobtrusive where it's simple, and it may come in handy for other
things.)

It's possible that the end game is that BackendLLVMCommon and
BackendLLVM simply become one class -- that is, the general/common
llvm back end contains the full implementation of single-point
shading, and and then the batched back end is a derived class that
overloads virtual methods for the things that need to be different for
batched execution. I'm not sure yet. But for now, I'm structuring it
as separate scalar and batched classes, with a parent class that
contains the methods and data that are common to both of them.

While we're at it, I submit that it's probably wise to change
terminology and internal naming so that we talk about "single point"
shading versus "batch" shading, but don't use the term "scalar" to
describe shading one point at a time, because we also, and for quite
some time now, have used "scalar" to describe any single data item
that isn't part of an array or an aggregate. I realized while doing
some of this refactoring that the use of "scalar" for both concepts
made some of the comments a quite confusing. Similarly, we are already
pretty good at saying "batch shading" rather than "SIMD" or "Vector",
both of which are used widely (excuse the pun) elsewhere to have other
meanings. But there are a few places where we use vector to describe
batch shading, and I aim to clean that up over time.

### DIFF
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -64,10 +64,11 @@ check_cwd (ShadingSystemImpl &shadingsys)
 
 
 
-BackendLLVM::BackendLLVM (ShadingSystemImpl &shadingsys,
-                          ShaderGroup &group, ShadingContext *ctx)
-    : OSOProcessorBase (shadingsys, group, ctx),
-      ll(ctx->llvm_thread_info(), llvm_debug(), shadingsys.m_vector_width),
+BackendLLVMCommon::BackendLLVMCommon(ShadingSystemImpl& shadingsys,
+                                     ShaderGroup& group, ShadingContext* ctx,
+                                     int width)
+    : OSOProcessorBase(shadingsys, group, ctx),
+      ll(ctx->llvm_thread_info(), llvm_debug(), width),
       m_stat_total_llvm_time(0), m_stat_llvm_setup_time(0),
       m_stat_llvm_irgen_time(0), m_stat_llvm_opt_time(0),
       m_stat_llvm_jit_time(0)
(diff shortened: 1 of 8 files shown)


lgritz:
Withdrawing. I need to rebase this and will take another stab at a later time.

PR #1497, opened 2022-04-29
More comprehensive warnings about things that don't optimize away

lgritz:
New classes of warnings triggered when enabling option "opt_warnings":

* Creation of new strings mid-shader.
* Access to the characters of strings mid-shader.

We already warned about texture access by name that could not be
turned into a handle.

In all of these cases, we only warn *after* runtime optimization, so that shader
idioms in the source code that appear to do these forbidden things will not warn
as long as they can be reduced to safe operations in the course of the runtime
optimization step.

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -137,10 +137,11 @@ public:
     ///    int no_noise           Replace noise with constant value. (0)
     ///    int no_pointcloud      Skip pointcloud lookups. (0)
     ///    int exec_repeat        How many times to run each group (1).
-    ///    int opt_warnings       Warn on certain failure to runtime-optimize
-    ///                              certain shader constructs. (0)
-    ///    int gpu_opt_error      Consider a hard error if certain shader
-    ///                              constructs cannot be optimized away. (0)
+    ///    int opt_warnings       Warn on failure to runtime-optimize certain
+    ///                              shader constructs. (0)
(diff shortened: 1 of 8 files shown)


fpsunflower (approved):

src/include/OSL/oslexec.h
```diff
@@ -137,10 +137,11 @@ class OSLEXECPUBLIC ShadingSystem
...
     ///    int no_pointcloud      Skip pointcloud lookups. (0)
     ///    int exec_repeat        How many times to run each group (1).
-    ///    int opt_warnings       Warn on certain failure to runtime-optimize
-    ///                              certain shader constructs. (0)
-    ///    int gpu_opt_error      Consider a hard error if certain shader
-    ///                              constructs cannot be optimized away. (0)
+    ///    int opt_warnings       Warn on failure to runtime-optimize certain
+    ///                              shader constructs. (0)
+    ///    int gpu_opt_error      Issue a hard error if certain shader
+    ///                              constructs cannot be optimized away, which
```
AlexMWells, line 143: Can we turn on gpu_opt_error even if we are running on the cpu?  (In order to get a shader into shape before trying the GPU)

AlexMWells (approved):
LGTM

src/include/OSL/oslexec.h
lgritz: That is 100% the point, yes.

lgritz:
@lkerley FYI

PR #1502, opened 2022-05-09
Windows build fixes for recent free function render services

fpsunflower:
## Description

The latest changes to leverage llvm-bitcode for renderer services did not build properly on windows. This set of patches should fix this along with a few other minor things.

I have a feeling there is probably a more "Cmake native" way of getting the path to the llvm-as and llvm-link binaries. I'm definitely open to suggestions here.


lgritz (approved):
LGTM, thanks!

PR #1504, opened 2022-05-11
More conversion of old printf style to std::format style

lgritz:
Another semi-random collection of conversions. Bit by bit.

These are all internals, none of it involves public interfaces.


AlexMWells (approved):
LGTM

PR #1505, opened 2022-05-11
A couple tweaks in how we build OIIO for OSL's CI

lgritz:
* Use unity build to speed up OIIO build (for sufficiently new OIIO).
* Build OIIO with STOP_ON_WARNING=OFF, so that if a compiler upgrades
  and makes new warnings for OIIO, that doesn't break OSL's CI build.


fpsunflower (approved):
LGTM

PR #1508, opened 2022-05-14
clang-format more of the code base

lgritz:
Bring more files into the clang-format fold, primarily things related to batch shading,
which we had held off on reformatting while they were being actively worked on.

Because this is just reformatting and not changing any code, I've attributed the
commits to the individual people who were the predominant authors of the files
in question, to preserve the amount of LOC authorship they get if somebody
runs statistics on the files. I hope that's ok, Alex and Stephen. It's 100% your code as
far as I'm concerned, I'm just preventing this automated process from changing
apparent authorship to me (which I feel would be more dishonest than this extremely
minor way that I am sort of forging a commit in your name.)

In addition to the clang-format reformatting, I've also taken the liberty of making most of
these files conform to our blank space conventions -- three blank lines between freestanding
function implementations, never more than one blank line in a row INSIDE a function,
and no blank lines immediately after an opening brace or immediately before a closing
brace. That is beyond the domain of clang-format, but I think those conventions promote
clear visual parsing (at least by me) of the code layout.

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -479,7 +479,9 @@ if (PROJECT_IS_TOP_LEVEL)
                                 "src/include/*.h"
                                 "src/liboslcomp/*.cpp"
                                 "src/liboslcomp/*.h"
-                                "src/liboslexec/batched_rendservices.cpp"
+                                "src/liboslexec/batch*.cpp"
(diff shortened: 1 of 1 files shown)


sfriedmapixar (approved):
LGTM

PR #1509, opened 2022-05-17
clang-format

lgritz:
The next round of adding files to the clang-format regime.

Like last time, I've broken it into a few distinct commits so that I can attribute
the changes to whomever was the dominant contributor to the changed files,
rather than make it seem like I touched everything. So there are commits here
for each of Cliff, Alex, and Tim. (And also a bunch legitimately for me, in a
separate commit -- for those files that either I was the primary author of, or
which had sufficiently mixed authorship already.)


lgritz:
Any objections?

fpsunflower (approved):

PR #1510, opened 2022-05-22
Adjust string formatters after change in fmt master

lgritz:
fmt master seems to have just deprecated the default ability of
fmt::sprintf("%s") to printing anything that allows stream
output. This breaks use of Strutil::sprintf formatting of ustrings.

This isn't yet in any tagged release of fmtlib, but I'm trying to get
ahead of it.

The quick fix is to just change those handful of places that still use
sprintf convention with nonstandard types to use std::format style
calls, which we are slowly converting to anyway. (And maybe a couple
other spots while I was at it.)


fpsunflower:
Do you have a pointer to the change on the fmtlib side? I'm curious about the rationale for the change and if they suggest any workarounds.

lgritz:
It was this commit: https://github.com/fmtlib/fmt/commit/edeb3d80918b1ff361b12f1b1ac30ce5e530d099

There is no specific explanation, and near as I can detect, this particular commit was not part of a PR with a formal review. It is merely getting rid of some things that had been marked deprecated for a while. It's part of a gradual trend to conform to the accepted std::format, which (unlike fmtlib) does not have "automatic" support for any type that has an `ostream& operator<<`, and rather expects custom formatters for custom types. 

With this particular change, I think what's happening is that they just cut the final bit of tape that kept `fmt::sprintf("%s", foo)` working for any type of foo that had an `operator<<`. Now (I mean, in their master, not yet in any real release, which well may bump versions in such a way as to indicate a major compatibility break), `%s` requires a string type of some kind. I think they are pushing for fmt::format to be the place for custom types, and fmt::sprintf to have fewer behaviors that aren't more or less interchangeable with std::sprintf.

It's no big deal. We, too, had been doing gradual conversion away from sprintf and to fmt::format underneath, the new std::format conventions and custom formatters for custom types. There were only a tiny handful of places in our code base that were still using the old calls that were removed.

fpsunflower:
That makes sense. I guess the handling of the printf like functions in OSL will be one of the few places we actually need support for the old printf style syntax which is probably fine (its not performance critical here anyway).

lgritz:
Yeah, that's one thing that will look like more of an anachronism over time. We'll probably want to expose new functions to OSL itself, just to up keep with what people expect from their experience with python, rust, and now C++ moving forward. But I don't think there's a big rush.

lgritz:
does this look ok, then?

fpsunflower (approved):
Yep - LGTM!

PR #1512, opened 2022-05-28
More clang-format

lgritz:


lgritz:
I'm going to merge this today if there are no objections. Seems safe, since it's just reformatting and no code changed.

PR #1513, opened 2022-05-28
Fix incorrect transform(m44,v4) in vector4.h

lgritz:
I think we were multiplying by the transpose. Can people please check my
math? This is obviously easy to get confused about. Remember that by
our conventions, points are column vectors and matrices have the
bottom row containing the translation (like RenderMan, not like math/physics),
and they are stored in memory in row-major order (like C, not like Fortran).


jstone-lucasfilm (approved):
Looks good to me!  Just adding a self-reminder that we to make the corresponding change in the MaterialX `genosl` library as well.

PR #1520, opened 2022-06-15
Fix recent ArgParse change

lgritz:


lgritz:
Critical fix, merging.

PR #1522, opened 2022-06-17
Branch for eventual release, call it alpha

lgritz:
Aiming to declare a stable beta soon and release circa SIGGRAPH.

Primarily, this branch is to allow us to continue changing main in
ways that would be destabilizing before the release.


lgritz:
Note: I called it RB-1.12 but forgot that we are changing notation. I will immediately rename the branch "dev-1.12", that's the staging area for things that will go into 1.12.

PR #1527, opened 2022-06-24
Fix permission error and build errors when running serialize-bc.py on Linux

AdrienHerubel:
Signed-off-by: Adrien Herubel <adrien.herubel@autodesk.com>

## Description

This PR fixes the following two issues I encountered while building the head of the main branch on Linux (Centos 7)

`/bin/sh: /home/herubea/dev/vanilla_osl/OpenShadingLanguage/src/build-scripts/serialize-bc.py: Permission denied
`

It looks like the serialize-bc.py script is missing the execution permission, and a shebang declaration for running the script as python.

## Checklist:

### DIFF
--- a/src/build-scripts/serialize-bc.py
+++ b/src/build-scripts/serialize-bc.py
@@ -1,3 +1,4 @@
+#!/bin/env python
 # Copyright Contributors to the Open Shading Language project.
 # SPDX-License-Identifier: BSD-3-Clause
 # https://github.com/AcademySoftwareFoundation/OpenShadingLanguage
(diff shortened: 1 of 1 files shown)


src/build-scripts/serialize-bc.py
```diff
@@ -1,3 +1,4 @@
+#!/bin/env python
```
lgritz, line 1: Isn't this traditionally `/usr/bin/env` ?

AdrienHerubel: Correct, this is more in line with the other builds scripts in the project.

PR #1530, opened 2022-06-28
Allow naming of llvm symbols for many ops

lgritz:
Most LLVM API IR calls that generate ops allow an optional name of the
llvm symbol that will hold the result of the operation. By default, it
will make one up, such as "%37", "%38", etc. Needless to say, these
are difficult to understand when reading volumes of IR dumped for
debugging purposes.

So in this patch, I'm starting to expose the ability for our own code
to pass names through our wrapper functions. I needed this for some
deep debugging.

I haven't instrumented the entire set, just trying it on for some of
the specific areas I needed to debug. But this is a useful first step.

### DIFF
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -259,7 +259,8 @@ BackendLLVM::llvm_global_symbol_ptr(ustring name)
     // the ShaderGlobals struct.
     int sg_index = ShaderGlobalNameToIndex(name);
     OSL_ASSERT(sg_index >= 0);
-    return ll.void_ptr(ll.GEP(sg_ptr(), 0, sg_index));
+    return ll.void_ptr(ll.GEP(sg_ptr(), 0, sg_index),
+                       fmtformat("glob_{}_voidptr", name));
 }
 
 
@@ -560,7 +561,7 @@ BackendLLVM::llvm_load_value(const Symbol& sym, int deriv,
     }
 
     return llvm_load_value(llvm_get_pointer(sym), sym.typespec(), deriv,
-                           arrayindex, component, cast);
+                           arrayindex, component, cast, sym.name().string());
 }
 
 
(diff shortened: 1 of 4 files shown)


src/liboslexec/backendllvm.cpp
```diff
@@ -590,7 +591,7 @@ BackendLLVM::llvm_load_value(llvm::Value* ptr, const TypeSpec& type, int deriv,
         ptr = ll.GEP(ptr, 0, component);
 
     // Now grab the value
-    llvm::Value* result = ll.op_load(ptr);
+    llvm::Value* result = ll.op_load(ptr, llname);
```
sfriedmapixar, line 595: For these more complicated entry points, I'm wondering if it would be more useful to bake things like deriv, arrayindex, component into the name so they are there in a consistent way.  Also, naming the intermediate results, such as the offset here so they are easily tied together as optimizations shuffle them around may also be helpful.

sfriedmapixar (commented):
Overall LGTM.  I am assuming that we'll still get some postfix of the _## variety to keep the single-static-assignment properties, is that correct? If not that would be problematic in any but the simplest cases and we'd need to add one ourselves.  
Also, I wouldn't mind seeing this on the wide ops as well.  I'm glad you started this...it should be immensely valuable in identifying missed optimization opportunities and debugging IR gen.

src/liboslexec/backendllvm.cpp
lgritz: LLVM automatically appends an incrementing counter to the string, if it is not unique.

How these are used will be a little more clear in my next PR, which is an overhaul of how strings are handled on GPU (and makes use of this). This is just setting up some of the debugging infrastructure that looked like I could cleave it off into a separate self-contained PR. I've just scratched the surface, I think this will continue to evolve as we realize what information is useful to embed in the names.

lgritz: @sfriedmapixar I definitely will extend to wide ops and maybe universally, but not as part of this first PR. I need to put the string stuff to bed first.

sfriedmapixar: Sounds completely resonable.

PR #1531, opened 2022-07-07
GPU string overhaul

lgritz:
Big string cleanup, yet again, for GPU. We still had this tricky
double indirection happening with strings for OptiX -- the string was
a pointer to a "global", which in turn held the hash. No, throw all
that out.  Just directly store strings as the ustring hash on
OptiX. No extra indirection. Many fewer differences between how things
are handled on OptiX versus CPU (except for the fact itself that the
"string" holds the hash rather than the char*).

Some details worth noting:

* LLVM_Util loses "type_string()" and calls it "type_ustring()" to
  emphasize it's the type we're using to represent a ustring, and
  definitely not a C or C++std string. Also, LLVM_Util itself now lets
  the caller tell it whether it prefers that the representation of a
  ustring be the charptr or the hash, and then handles most of the
  details internally. The BackendLLVM, then, makes that decision early
  on, charptr for CPU and hash for GPU.

* Several more spots in LLVM_Util and BackendLLVM have been plumbed to
  pass around llvm IR SSA variable names to try to make the resulting
  LLVM IR more readable when I'm trying to debug it. (This was
  strictly debugging scaffalding aid for me, but comes along for the
  ride because it's useful to keep.) I also added a
  BackendLLVM::llnamefmt to format strings, which only does the
  formatting when the shading system knows it's going to save the IR
  for inspection, and otherwise (that is, in production) it does not
  do the string formatting or pass the names along, so it's not doing
  all that string manipulation when no human is going to be looking at
  the IR dumps.

* A lot of clunky OptiX-only code disappears.

* Some side effects of my reducing complexity of the IR as I debugged:
  there were times we were generating redundant pointer casting in the
  IR, such as grabbing a pointer, then casting it to a void*, then
  casting back to some other kind of pointer. In some cases I reduced
  that quite a bit. Again, this was mostly just a scramble to make the
  IR more readable while I debugged, but it probably has a minor
  performance gain as well since it reduced somewhat the sheer amount
  of IR we generate and pass to the optimizer and JIT.

### DIFF
--- a/src/include/OSL/rendererservices.h
+++ b/src/include/OSL/rendererservices.h
@@ -405,16 +405,6 @@ public:
     /// Return a pointer to the texture system (if available).
     virtual TextureSystem *texturesys () const;
 
-    /// Register a string with the renderer, so that the renderer can arrange
-    /// to make the string available in GPU memory as needed. Optionally specify
-    /// a variable name to associate with the string value.
-    virtual uint64_t register_string (const std::string& str,
-                                      const std::string& var_name)
-    {
-        ustring ustr = ustring(str);
-        return ustr.hash();
-    }
-
     virtual uint64_t register_global (const std::string& var_name,
                                       uint64_t           value)
     {
--- a/src/include/OSL/rendererservices.h
+++ b/src/include/OSL/rendererservices.h
@@ -411,7 +411,8 @@ public:
     virtual uint64_t register_string (const std::string& str,
                                       const std::string& var_name)
     {
-        return 0;
+        ustring ustr = ustring(str);
+        return ustr.hash();
     }
 
     virtual uint64_t register_global (const std::string& var_name,
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -459,7 +459,7 @@ BackendLLVM::llvm_assign_initial_value (const Symbol& sym, bool force)
 
     // Handle interpolated params by calling osl_bind_interpolated_param,
     // which will check if userdata is already retrieved, if not it will
-    // call RendererServices::get_userdata to retrive it. In either case,
+    // call RendererServices::get_userdata to retrieve it. In either case,
     // it will return 1 if it put the userdata in the right spot (either
     // retrieved de novo or copied from a previous retrieval), or 0 if no
     // such userdata was available.
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -244,6 +244,7 @@ BackendLLVM::llvm_type_groupdata ()
     std::vector<llvm::Type*> fields;
     int offset = 0;
     int order = 0;
+    m_groupdata_field_names.clear();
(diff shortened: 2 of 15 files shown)


lgritz:
@timgrant Would be great to get your eyeballs on this, too.

tgrant-nv:
Taking a look now.

src/include/OSL/rendererservices.h
```diff
@@ -411,7 +411,8 @@ class OSLEXECPUBLIC RendererServices {
     virtual uint64_t register_string (const std::string& str,
```
tgrant-nv, line 411: It looks like `register_string` is vestigial now. Does it make sense to remove it? Or maybe there is some potential use beyond a fancy wrapper for `ustring::hash`?

lgritz: I think you're right, but I wasn't sure about the compatibility break that would be incurred to removing a virtual method from the parent class. Maybe I'll just mark it as deprecated first?

cmstein:
That'd be my vote.  I think we'd be ok with it removed, but I'd rather wait
for the dust to settle a little bit once the new functionality is pushed
out to make sure all the pieces are in place on our side to make sure we
won't need it.

src/include/OSL/rendererservices.h
lgritz: YOLO, removed it entirely.

fpsunflower (approved):

src/include/OSL/rendererservices.h
lgritz: Oops, yeah, fixing.

lgritz: Other than this, did the overall simplification look sane to  you, @tgrant-nv ?

tgrant-nv: Yes, definitely. It's a huge improvement in readability and maintainability for GPU strings. I don't think anyone will miss trying to reason about `load_device_string`.

src/liboslexec/llvm_instance.cpp
```diff
@@ -440,14 +448,18 @@ BackendLLVM::llvm_assign_initial_value (const Symbol& sym, bool force)
...
+
     // From here on, everything we are dealing with is a shader parameter
     // (either ordinary or output).
     OSL_ASSERT_MSG (sym.symtype() == SymTypeParam || sym.symtype() == SymTypeOutputParam,
                     "symtype was %d, data type was %s", (int)sym.symtype(), sym.typespec().c_str());
 
     // Handle interpolated params by calling osl_bind_interpolated_param,
     // which will check if userdata is already retrieved, if not it will
-    // call RendererServices::get_userdata to retrived it. In either case,
+    // call RendererServices::get_userdata to retrive it. In either case,
```
tgrant-nv, line 462: I believe this should be "retrieve".

lgritz: updated

lgritz:
Will merge this at the end of the day. Register any final objections now.

PR #1533, opened 2022-07-16
Implement basic support for MaterialX closures in testrender

fpsunflower:
* Extend the bsdf interface in testrender to use Color3 instead of float in preparation for MaterialX closures which are almost all tinted

  * Add stub parameters and IDs for all MaterialX closures

  * Register all MaterialX closure types

  * Implement most MaterialX closures, reusing the existing implementations for now

This commit is mainly intended to do the busy work of adding the new closures, no new BSDFs are actually implemented yet. This means that some of the closures are handled in a very rudimentary way, ignoring some key parameters. The intent is to improve support for these over time.


src/testrender/shading.cpp
```diff
@@ -192,13 +393,13 @@ struct OrenNayar final : public BSDF, OrenNayarParams {
...
+            return Color3(A + B * stinv);
         }
-        return pdf = 0;
+        return Color3(pdf = 0);
     }
-    virtual float sample(const OSL::ShaderGlobals& sg, float rx, float ry,
+    Color3 sample(const OSL::ShaderGlobals& sg, float rx, float ry,
                          float /*rz*/, OSL::Dual2<OSL::Vec3>& wi,
-                         float& pdf) const
+                         float& pdf) const override
```
aconty, line 402: This patch looks good to me, thanks! But I'm going to abuse it to open a discussion on the BSDF API:

  1) Should we return a struct like { weight, pdf } to avoid output args?
  2) Can we drop the full sg as argument and instead pass a Vec3 wo (meaning I) so you can evaluate them in an SG-less context?
  3) Does anybody need accurate derivs from wi? If not, can we stop pretending we can compute them?
  4) Should we add roughness to the output? I believe most renders track it. You can guess it from pdf following a convention (like reversing an isotropic GGX), but wanted to get opinions. A normalized [0-1] roughness can be used as a measure of specularity for AOVs. If we sign off on BSDFs being single lobe the roughness can be just a BSDF property and not an output from sample()/eval().

I volunteer to making any of these changes after you merge this, of course.

src/testrender/shading.cpp
```diff
@@ -860,6 +1063,92 @@ process_closure(ShadingResult& result, const ClosureColor* closure,
...
+                // TODO: implement BSSRDF support?
+                const MxSubsurfaceParams* srcparams = comp->as<MxSubsurfaceParams>();
+                DiffuseParams params = {};
+                params.N = srcparams->N;
+                ok = result.bsdf.add_bsdf<Diffuse<0>, DiffuseParams>(cw * srcparams->albedo, params);
+                break;
+            }
+            case MX_SHEEN_ID:
+            case MX_LAYER_ID: {
+                OSL_ASSERT(false && "MaterialX closure not yet implemented");
```
aconty, line 1149: I have a suggestion on how to implement this. Add a 

virtual Color3 filter_base(Vec3 wo);

You call it here, scale the weight of base and just add_bsdf() both. Recurse where needed. Works for us by making assumptions on wi. Eliminates the need of recursion at eval/sample time. De-virtualize and you are ready for GPU.

src/testrender/shading.cpp
fpsunflower: 1. Yes - good idea. I can do that refactor in another PR. Would you have a { weight, pdf, dir } struct for sample() then?
2. Yes - good idea.
3. Yes - definitely. Only the reflection() and refraction() cases do anything sensible. Do you want to submit a patch that switches to ray cones? That would be a nice simplification for the bsdf API.
4. I wouldn't bother with this for now. I prefer keeping testrender minimal for now.

src/testrender/shading.cpp
fpsunflower: Thats what I was thinking too. Maybe it can even just use the existing "get_albedo()" method?

src/testrender/shading.cpp
aconty: Yes { color, wi, pdf } in any order you like works for me. I will do the cone tracing.

src/testrender/shading.cpp
aconty: Good idea, although the coating may want to add absorption according to some thickness. So maybe a different method that by default does 1 - get_albedo().

fpsunflower:
I haven't found any workaround yet ... I'll keep updating `render-microfacet`'s reference image for now. Hopefully the next few commits won't require touching any related code so it will be stable for a while.

aconty (approved):

PR #1534, opened 2022-07-17
Modernize sampler in testrender

fpsunflower:
The sampler implementation in testrender was very close to the 2020 article from Brent Burley. This patch
catches up to best practices, including a faster owen scrambler from Nathan Vegdahl and recognizing that
the index scrambling can be achived by the same owen scrambler method (which saves a bunch of logic and runs
faster). Finally, avoid the redundant bit reversals and use compiler intrinsics when available to benefit
from hardware instructions (for ARM or GPUs). Finally the integer hash constants were updated to the latest
from the hash prospector project.

While more optimizations are certainly possible, this is a reasonable balance between quality and speed and saves a few lines of code over the previous implementation.

On my laptop, running the microfacet sample from the testsuite at 1024x768 with 4x4 samples goes from 10.2s down to 8.9s (which is roughly 12% faster).


lgritz:
Wow, the sample scrambler itself was accounting for 10% of runtime??

fpsunflower:
Yep! Now that the faster methods are well documented in the literature, might as well use them :)

aconty (approved):

lgritz:
It's failing one of the unit tests. I know you're looking into that separately, but in the mean time, should we update the ref images (you can have more than one ref image for a test -- the test passes if it matches any of the choices, so sometimes we have multiple that we consider "correct" if we just can't make all platforms match exactly)? Or if it's just one or two pixels we can adjust the amount of allowable difference? Or would you like to simply hold off on merging this until you fully solve #1535?

fpsunflower:
It seems its always the same test giving us trouble. I have one or two more things to try, otherwise I'll just commit this with an extra reference image.

lgritz:
I've had trouble with that test before, too. I was never able to figure out why that one, in particular, was so brittle to minor platform differences. And not just a couple LSB errors, it was something that was changing where they rays were going.

PR #1535, opened 2022-07-17
Try to make the output of render-microfacet more consistent across platforms

fpsunflower:
For some reason this test fails somewhat constantly on only the newer compilers. I am not really sure what is special about it yet.

Wild hunch is that it is maybe related to this constant in the sampling routine getting constant folded differently across compilers. So just manually constant fold it to see if it helps at all.


fpsunflower:
No luck yet. Switching to ggx doesn't help either, so its not related to the shading model it seems ...

PR #1542, opened 2022-07-27
Add some furnace tests for some of the MaterialX closures

fpsunflower:
This is to quantify the energy loss/gain of the individual closures. I've also added a way for testrender to show the results of the `get_albedo()` method directly, for comparison with furnace tests.

I've also added a rational fit for the albedo of sheen, which is taken from the MaterialX project. It doesn't look like an exact match, but it should be good enough for now.


fpsunflower:
The mx-layer test was added in a previous PR, but was not actually being run. I added it to the list of tests to run but didn't re-generate the image yet.

lgritz:
The one remaining that fails is the Debug build, and at least some of the test were failing simply by timing out -- some of the render tests take a godawful long time when not optimized. Like that touch-up patch I added earlier today, we may just want to bump up the ctest timeout value for the individual tests higher in debug mode.

lgritz (approved):

lgritz:
I'm tempted to merge this and just bump the timeouts for those failing tests individually, as a separate PR. Shall I do that, or do you really want to do that yourself, @fpsunflower ?

fpsunflower:
Yeah I'm fine with ignoring the timeouts, but there is one legit failure in debug mode -- an assert is being triggered. I was just able to repro locally, so let me try to sort that out first ...

PR #1543, opened 2022-07-29
Switch testrender to cone tracing

aconty:
## Description

The idea is to stop propagating derivs through BSDFs, which we
never did properly. Instead we trace rays with radius + spread,
which we turn into OSL derivs in SG when we hit a surface. Then
when bouncing off a surface we use the BSDF roughness to set the
spread. In addition we also:

  * Return BSDF samples/eval as a small struct Sample(). This
    avoids output parameters in the methods.
  * Stop passing a full ShaderGlobals to the BSDF methods. Pass
    a viewing direction wo instead so they could run in a simpler
    context.

## Tests

Updated testrender tests because of very minimal diffs.

## Checklist:


src/testrender/shading.cpp
```diff
@@ -972,165 +949,168 @@ struct MxBurleyDiffuse final : public BSDF, MxBurleyDiffuseParams {
...
 
 
-Color3 clamp(const Color3& c, float min, float max)
+Color3
+clamp(const Color3& c, float min, float max)
 {
-    return Color3(c.x < min ? min : max < c.x ? max : c.x,
-		  c.y < min ? min : max < c.y ? max : c.y,
-		  c.z < min ? min : max < c.z ? max : c.z);
+    return Color3(c.x < min   ? min
```
fpsunflower, line 1014: This line looks like it got formatted a bit strangely?

fpsunflower (approved):
LGTM!

There are a few spots where it looks like clang-format had too narrow of a width that you might want to double check ...

aconty:
Maybe, but that's the project configuration as far as I know.

fpsunflower (approved):

fpsunflower (approved):
LGTM

Sorry about the merge conflicts 😅

aconty:
No problem! Thanks

PR #1548, opened 2022-08-02
CI: Increase timeout for some render tests for debug builds

lgritz:


fpsunflower:
This looks good. I just tried another approach in #1547, to set an increased timeout for all render tests (we'll see if I got the cmake syntax right ...)

PR #1552, opened 2022-08-14
Fix typo error that prevented correct typecheck of ternary

lgritz:
Didn't properly detect the disallowed use of arrays for the false
clause because we checked `t` twice.

Yay for static analysis!


fpsunflower (approved):

AlexMWells (commented):
LGTM

PR #1553, opened 2022-08-18
GPU string fix -- needed missing cast

lgritz:
Fix slightly incorrect code generation when getting a string address
for GPU.


lgritz:
Look ok to everyone?

(The CI failures are unrelated, I will address those separately.)

PR #1559, opened 2022-08-24
Fixes needed to support OptiX 7.4

lgritz:


lgritz:
@chellmuth (and I'll make an MR for the same changes on the renderer side internally)

aconty (approved):

PR #1561, opened 2022-08-27
Fix OptiX + clang14 issue with duplicate symbols

lgritz:
We had been using clang/llvm 12 and when trying out 14 I started
getting test failures with OptiX compliaining about duplicate
symbols. They all seemed to be the things we declared as
OSL_CONSTANT_DATA. It looks like there's a difference in llvm 12 vs 14
in how the generated PTX marks its symbol visibility.

The solution (thanks to a clever suggestion by Alex Conty) was just to
enclose those declarations in anonymous namespaces. Works like a
charm!


aconty (approved):

aconty:
static should have the same effect, don't know if it is being drop for a legitimate reason (bad __device__ declaration) or a bug in the front/backend. Anyway I'm glad it helped.

PR #1564, opened 2022-08-29
New closure docs

lgritz:
* Change the languagespec to document the new MaterialX-synchronized
  closures.

* Document the old ones as "deprecated" -- still supported for 1.x but
  assumed to be on the slate for removal in OSL 2.x.

* Change stdosl.h so that the new closures, instead of being seen only
  if MATERIALX_CLOSURES is defined, to being enabled by default, but
  (for now) can be disabled if NO_MATERIALX_CLOSURES is defined.

### DIFF
--- a/src/doc/languagespec.tex
+++ b/src/doc/languagespec.tex
@@ -66,7 +66,7 @@ All rights reserved.
 Editor: Larry Gritz \\
 \emph{lg@imageworks.com}
 }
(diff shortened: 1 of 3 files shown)


lgritz:
I think this is the very last thing to merge before declaring a 1.12 beta (other than rebuilding the PDF itself and some touch-ups on the release notes).

PR #1566, opened 2022-08-30
Generate deterministic ptx for OptiX cache hits

chellmuth:
## Description

In multithreaded environments, use of global counter ids inside function and variable identifiers in generated ptx code leads to cache misses during OptiX module compilation.

With this change we assume unique shadergroup names, and remove the non-deterministic ids. In my testing, everything is still OptiX-compatible even with duplicate shadergroup names, but I'm not sure I've considered all the valid cases.

## Tests

Existing tests pass; I can add tests that run shaders twice and grep the logs for cache misses if desired.

## Checklist:

### DIFF
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -858,8 +858,8 @@ BackendLLVM::build_llvm_init()
 {
     // Make a group init function: void group_init(ShaderGlobals*, GroupData*)
     // Note that the GroupData* is passed as a void*.
-    std::string unique_name = fmtformat("__direct_callable__group_{}_{}_init",
-                                        group().name(), group().id());
+    std::string unique_name = fmtformat("__direct_callable__group_{}_init",
+                                        group().name());
     ll.current_function(
         ll.make_function(unique_name, false,
                          ll.type_void(),  // return type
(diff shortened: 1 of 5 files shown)


src/liboslexec/llvm_instance.cpp
```diff
@@ -1620,8 +1620,7 @@ BackendLLVM::run()
     }
 
     if (use_optix()) {
-        std::string name = fmtformat("{}_{}", group().name(), group().id());
-        ll.ptx_compile_group(nullptr, name,
+        ll.ptx_compile_group(nullptr, group().name().c_str(),
```
lgritz, line 1623:
```suggestion
        ll.ptx_compile_group(nullptr, group().name().string(),
```

Extremely minor nit: I think this should be `group().name().string()` to return the string (which is part of the ustring internals, it doesn't make a new string). But if you say `.c_str()`, that'll give you the character pointer, which will then have to make another string to pass to ptx_compile_group().

lgritz (approved):
LGTM, pending that one extremely minor change I suggested, and squaring away the CLA. (Is this really your first OSL patch?)

lgritz:
@timgrant @ee33 You may find this interesting

tgrant-nv:
Thanks for the heads-up, Larry. Stripping the `callseq` comments from the PTX string is something Chris and I had discussed. It seems like the best solution short of patching LLVM, since as far as we can tell there is no way to manage these comments through the LLVM API.

PR #1569, opened 2022-09-01
Deal with deprecated bit_cast function in OIIO 2.5

lgritz:
Today's branching of OIIO 2.4 and subsequent changing of its master
branch to think it's "2.5" seems to have triggered an automatic fuse
that annotates a particular function in OIIO as deprecated for version
2.5 and beyond. OSL was still calling the old version, so some
immediate changes are necessary to call the correct new name when
using a new OIIO.

### DIFF
--- a/src/include/OSL/oslnoise.h
+++ b/src/include/OSL/oslnoise.h
@@ -141,6 +141,19 @@ inline OSL_HOSTDEVICE float bits_to_01 (unsigned int bits) {
 }
 
 
+OSL_FORCEINLINE OSL_HOSTDEVICE int
+bitcast_to_uint (float x)
+{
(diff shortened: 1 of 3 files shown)


fpsunflower (approved):
I'm just curious why the function was renamed from `bit_cast` to `bitcast`? The C++20 version is `bit_cast`.

fpsunflower:
Oh nevermind - I dug up the OIIO PR, I forgot about the argument ordering difference 😅

lgritz:
Yeah, seemed like having the identically named function with opposite template argument order was bound to be trouble, so I'm trying to deprecate and phase ours out before anybody is using C++20 too widely.

PR #1572, opened 2022-09-05
Docs: RELEASING.md

lgritz:
I've tried to write a complete set of step-by-step instructions for
everything about how to coordinate and deploy a release. This is a
living document, we will amend over time. I'm sure there must be
things I've forgotten.

### DIFF
--- a/doc/RELEASING.md
+++ b/doc/RELEASING.md
@@ -1,0 +1,341 @@
+<!-- SPDX-License-Identifier: CC-BY-4.0 -->
+<!-- Copyright Contributors to the Open Shading Language Project. -->
+
+# Release Procedures for Open Shading Language
+
+## Version nomenclature and release cadence
+
+### The meaning of the version parts
+
+Our release numbers are of the form `MAJOR.MINOR.PATCH.TWEAK`, and we follow
+an almost-semantic-versioning scheme.
+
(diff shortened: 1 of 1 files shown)


lgritz:
I put this in doc, to keep the top level from being too cluttered. But if people think it should go in the top level, please say so, that would also be fine.

lgritz:
Anybody have comments on this? Did I miss anything or is it unclear?

fpsunflower (approved):
This looks good!

Do we have the process for building the pdf docs documented somewhere?

I am guessing we want to move to readthedocs at some point, so maybe its not worth spending the time documenting it if it will change soon.

lgritz:
The process is just `cd src/doc ; make`, and then if it looks good, `mv languagespec.pdf osl-languagespec.pdf`. But we don't have documented exactly what versions of latex or whatever are needed to make somebody produces precisely the same result as I do with whatever versions I use.

I'm tempted to sweep this under the rug and just work on switching the whole thing to readthedocs. Having just rebuilt the docs for the new release that's in beta now, I decided that I want that to hopefully be the last time that do that.

lgritz:
I'm extremely satisfied with the way the readthedocs turned out for OpenImageIO. I think there's no question that we should do it for OSL. Will get to it soon.

PR #1576, opened 2022-09-06
Fix memory leak in ASTvariable_declaration

lgritz:


fpsunflower (approved):
LGTM

Though the fact that this node sometimes owns the memory and sometimes does not seems a bit confusing.

Would there be a way to rig this to use unique_ptr? Or is the pointer potentially valid but non-owning in some cases?

lgritz:
If it puts it into the symbol table (which it does if it's not metadata), this pointer is non-owning and the symbol table itself is responsible for deleting it. But for metadata, this is the only pointer to it.

Let me take another look at this, maybe there is a more clear way to organize it to not be so confusing.

fpsunflower (approved):
I like it!

PR #1577, opened 2022-09-08
Fix overrunning allocation region in debug common_ancestor_between

lgritz:
Bug found by LG running address sanitizer.
Fix by Alex. Tested by LG (works) and submitted on AW's behalf.

Signed-off-by: Alex Wells <alex.m.wells@intel.com>


lgritz:
Note to @AlexMWells: I left the DASSERT statements, but removed all the OSL_DEV print messages to reduce clutter. The assertions ought to be enough, we can easily add prints if we ever need to debug this again.

PR #1578, opened 2022-09-10
Bug fix: assigning init-op param its default value as instance got lost

lgritz:
This is a tricky situation to describe. Consider this shader:

    shader test (float A = 0.5,
                 float B = A)
    {
        printf("A = %g, B = %g\n", A, B);
    }

By default, A is 0.5 and B gets A's value, also 0.5. If you give A a different instance value, like 1.0, then B will also get the same value, they'll both be 1.0.  If you give B an instance value, that will be used for B rather than copying A's value, like this:

    $ testshade  -param B 1.0 -shader test test
    A = 0.5, B = 1

But there is a corner case bug: if we assign B an instance value of 0.0, it will incorrectly fall into a special case where it thinks it gets a default value, only coincidentally because 0.0 is in an unused slot for the default value (unused because it has no default -- its default comes from "init ops", that assignment).

    $ testshade  -param B 0.0 -shader test test
    A = 0.5, B = 0.5

!!!

Reading the patch should be self-explanatory for what the bug is and how we fix it.

### DIFF
--- a/src/cmake/testing.cmake
+++ b/src/cmake/testing.cmake
@@ -246,7 +246,9 @@ macro (osl_add_all_tests)
                 globals-needed
                 group-outputs groupstring
                 hash hashnoise hex hyperb
-                ieee_fp ieee_fp-reg if if-reg incdec initlist initops intbits isconnected
+                ieee_fp ieee_fp-reg if if-reg incdec initlist
+                initops initops-instance-clash
+                intbits isconnected
                 isconstant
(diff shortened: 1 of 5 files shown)


lgritz:
@lkerley @sandipshukla This is the weird bug we talked about a couple days ago.

sandipshukla:
Cool.
It's fixed now.

fpsunflower (approved):

PR #1581, opened 2022-09-19
CI: Set up a dynamic analysis test using address and leak sanitizers

lgritz:
- There are a handful of tests we disable entirely with because they are irrelevant or just cannot be made to be clean (and that's ok).

- Tests with sanitizers run slow, so we increase timeouts, and also set AA to 1 for render tests to speed them up.  We rig it so a non-matching output won't fail, and that's ok because every other CI test is checking the results; here we have no reason to believe results will differ from other tests, we just need to run with sanitizers turned on.

- In testing.cmake, we rig texture tests to disable the leak sanitizer, because we can't eliminate some completely benign leaks from the texture system, and because they are called from JITed code, we don't get a clean stack trace of things that happen in OIIO and therefore we just can't enumerate the allowed exceptions. So we trust that OIIO itself tests the texture system adequately and is able to ignore the benign "leaks."

- We already submitted PRs and merged the little fixes that were discovered by enabling address and leak sanitizers. This is just the final bit to turn it on for CI runs from here on out.


fpsunflower (approved):

lgritz:
With this PR merged, we seem to now be at 100% of the OpenSSF Core Infrastructure Initiative Best Practices badge at the "passing" level.

fpsunflower:
Nice!

PR #1583, opened 2022-09-20
add .envrc to .gitignore

marksisson:
## Description

direnv (https://github.com/direnv/direnv) can be used for local developer workflow enhancements.  direnv uses .envrc file for configuration.  Because .envrc is usually specific to a developer workflow, it is generally not committed.

## Tests

None

## Checklist:

*just emailed my signed CLA today*

Signed-off-by: Mark Sisson <5761292+marksisson@users.noreply.github.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ build
 *.pyc
 *.log
 .vscode
+.envrc
 


lgritz:
It seems like it's going to be very tedious to have to make essentially the identical trivial patch to every single project you might ever interact with.

Is there a way in your ./gitconfig to give patterns for gitignore that apply to all repos, so users of direnv can handle it on their end without patching all open source projects?

marksisson:
There is a way to do a global ignore, but some projects do commit the .envrc file (especially in some internal projects where there are standardized project specific environments).  It helps to ignore it in projects that do not use it.  The global ignore breaks workflows where .envrc file is committed.

PR #1589, opened 2022-09-27
OSL_ALLOCA

lgritz:
We use OIIO_ALLOCA in many places. Make an OSL version and put it in platform.h.

### DIFF
--- a/src/include/OSL/platform.h
+++ b/src/include/OSL/platform.h
@@ -482,11 +482,19 @@
 #    define OSL_DASSERT_MSG(x, ...) ((void)sizeof(x)) /*NOLINT*/
 #endif
 
-#if defined(_MSC_VER) && !defined(__INTEL_COMPILER)
-    #define OSL_STACK_ARRAY(TYPE, NAME, DYNAMIC_COUNT) TYPE * NAME = reinterpret_cast<TYPE *>(alloca(sizeof(TYPE)*(DYNAMIC_COUNT)))
+
+/// OSL_ALLOCA is used to allocate smallish amount of memory on the stack,
+/// equivalent of C99 type var_name[size].
+///
+/// NOTE: in a debug build, this will assert for allocations >= 1MB, which is
+/// much too big. Hopefully this will keep us from abusing alloca and having
+/// stack overflows. The rule of thumb is that it's ok to use alloca for small
+/// things of bounded size, but not for anything that could be arbitrarily
+/// big, which should instead use heap allocation.
+#if defined(__GNUC__)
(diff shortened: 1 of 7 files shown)


aconty (approved):

src/include/OSL/platform.h
```diff
@@ -482,11 +482,19 @@
...
+/// OSL_ALLOCA is used to allocate smallish amount of memory on the stack,
+/// equivalent of C99 type var_name[size].
+///
+/// NOTE: in a debug build, this will assert for allocations >= 1MB, which is
+/// much too big. Hopefully this will keep us from abusing alloca and having
+/// stack overflows. The rule of thumb is that it's ok to use alloca for small
+/// things of bounded size, but not for anything that could be arbitrarily
+/// big, which should instead use heap allocation.
+#if defined(__GNUC__)
+#    define OSL_ALLOCA(type, size) (assert(size < (1<<20)), (size) != 0 ? ((type*)__builtin_alloca((size) * sizeof(type))) : nullptr)
```
AlexMWells, line 495: Any style guidelines here and casting, should it be 
`(reinterpret_cast<type*>(__builtin_alloca((size) * sizeof(type))))`

lgritz: I think I am just an old programmer whose fingers tend to reflexively use C casts and forget about the C++ style until somebody points out my error. To be honest, I don't really feel that there are many places where the more verbose C++ syntax confers any advantage in practice.

I understand their purpose and see the point: const_cast *only* casts away constness while changing nothing else, static_cast is a value conversion from one type to another, reinterpret_cast turns a pointer into a different pointer type (but referring to the same memory). The use of the right specialized cast is a semantic breadcrumb conveying intent, whereas the C `(cast)` style is just a big hammer that clobbers everything. But the return value of alloca is already just raw memory that is semantic-free, so there's nothing to clobber, it's a truly arbitrary cast. In this case, what does `reinterpret_cast<type*>` do or what information does it convey that `(type*)` does not?

I'm just thinking out loud. I have no objection to the suggestion to change this to more idiomatic C++.

AlexMWells: To be clear, I don't have any objection to the old C cast, just using C++ as a matter of habit and was curious if there was a style guideline in practice for the project.  Although I think I like the in your face verbosity of the reinterpret_cast<>() and the wrapping parenthesis around the item to cast.

AlexMWells (approved):
LGTM

fpsunflower (approved):

src/include/OSL/platform.h
lgritz: I did update with the new style cast

PR #1591, opened 2022-09-28
Raise OIIO minimum to 2.3

lgritz:
Traditionally, OSL has supported building against the current and also the previous supported version of OIIO, but not necessarily any older.

With OIIO 2.4 released (any day now), and OSL "main" is actually going to be next year's supported release, it seems reasonable to change the forward-looking development of OSL to have a minimum of OIIO 2.3, which means that OSL-next will support OIIO 2.3, 2.4, and the currently-in-progress 2.5 (or whatever it's called by that point).

The main reason to do this bump is because we'd like to depend on the ustringhash class, and possibly other things we will probably be able to easily backport to 2.3 but not to 2.2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
             vfxyear: 2019
             vfxsuffix: -clang9
             cxx_std: 14
-            openimageio_ver: v2.2.17.0
+            openimageio_ver: v2.3.17.0
             python_ver: 2.7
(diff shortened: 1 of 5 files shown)


lgritz:
Does anybody object to this? Is anyone relying on OSL directly from main somehow stuck with only being able to use OIIO 2.2 and can't upgrade even to 2.3?

lgritz:
I will merge this at the end of the day today if there are no objections.

PR #1592, opened 2022-09-29
LLVM 15.0 support

### LINKED ISSUE #1587: Couldn't build against LLVM 15
### Problem

OSL couldn't be built against LLVM 15

**Expected behavior:**  

To build against LLVM 15

**Actual behavior:**  

https://gist.github.com/MehdiChinoune/e34157d779094c516d5b003901f59d1b

### Steps to Reproduce

1. Install LLVM-15
2. Configure OSL `cmake ..`
3. Build OSL

### Versions

* OSL branch/version: 1.12.6.1-RC1
* OS: Windows 10
* C++ compiler: Clang 15.0.1
* LLVM version: 15.0.1
* OIIO version: 2.3.19.0

lgritz:
* A variety of changes to get a clean build and passing tests when using LLVM 15.0.

* I've noticed problems when using LLVM 15 but building with earlier clang, so the cmake scripts now print a warning in that case, so if users encounter trouble they have a hint about what to do to fix it.

* For our CI tests on Mac, force the MacOS-11 test to use llvm 14, and the MacOS-12 test to use llvm 15.

IMPORTANT TO-DO / CAVEATS:

1. When doing JIT at optlevel 12 or 13, I had to disable a number of passes that don't seem to exist anymore in LLVM 15. This is enough to get it working, and to be honest, I don't know if anybody uses these opt levels.  But we need to revisit this, because I don't know if there these are cases where the names of the passes merely changed or that new passes take their place (in which case we should add the new passes, not stop after merely disabling the deprecated ones). For that matter, optlevel modes 11, 12, 13 are supposed to match what clang does for -O1, -O2, -O3, and that changes from one release to the next, so we should probably revisit this list and make sure it's matching clang's current choices (which I assume are crafted and periodically revised by clang's authors).

2. LLVM 15 switches to "opaque pointers". It still supports typed pointers...  for now. But as you can see in the diffs, I had to disable a variety of deprecation warnings and take some other actions to put LLVM 15 in the old "opaque ptr" mode to match our use of LLVM <= 14. But this is only temporary, because the typed pointer mode is expected to be removed entirely in LLVM 16. So at some point in the next few months, we are going to need to support opaque pointers in our use of LLVM. (Also note: for a while, we're going to have a bunch of ugly `#if` guards or other logic to support both opaque pointers for users with llvm >= 16 and also typed pointers for users with llvm <= 14, until such time as our LLVM minimum is at least 15, which I expect is not a reasonable assumption for at least a couple years.)

Closes #1587

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -426,20 +426,25 @@ jobs:
           - desc: MacOS-11
             os: macos-11
             nametag: macos11-py39
-            cxx_std: 17
-            python_ver: 3.9
-            aclang: 13
-            openimageio_ver: release
-          - desc: MacOS-12
-            os: macos-12
-            nametag: macos11-p310
+            cc_compiler: /usr/local/opt/llvm@14/bin/clang
+            cxx_compiler: /usr/local/opt/llvm@14/bin/clang++
             cxx_std: 17
             python_ver: "3.10"
             aclang: 13
+            setenvs: export LLVMBREWVER="@14" DO_BREW_UPDATE=1
+          - desc: MacOS-12
+            os: macos-12
+            nametag: macos11-p310
+            cc_compiler: /usr/local/opt/llvm@15/bin/clang
+            cxx_compiler: /usr/local/opt/llvm@15/bin/clang++
+            cxx_std: 17
+            python_ver: "3.10"
+            aclang: 13
+            setenvs: export LLVMBREWVER="@15" DO_BREW_UPDATE=1
     runs-on: ${{matrix.os}}
     env:
-      CC: clang
-      CXX: clang++
+      CXX: ${{matrix.cxx_compiler}}
+      CC: ${{matrix.cc_compiler}}
       CMAKE_CXX_STANDARD: ${{matrix.cxx_std}}
       PYTHON_VERSION: ${{matrix.python_ver}}
       OPENIMAGEIO_VERSION: ${{matrix.openimageio_ver}}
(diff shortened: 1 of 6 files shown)


lgritz:
I want to get this merged ASAP because a few days ago, Homebrew upgraded their default llvm to 15, so a large swath of Mac users now can't build OSL (without figuring out the arcane way to install and use a non-default llvm 14 or earlier version). I expect to do the merge tomorrow unless somebody raises a serious objection to the contents of this PR.

But as I noted, although this patch works as-is, there are a few things we need to return to. One is the issue of updating the set of passes we use for optlevel 12 and 13, which I hope @AlexMWells or @steenax86 might tackle, because it is they who wrote that section (which isn't exactly a perfect replica of the clang logic as I oversimplified above, but includes some curation and fine tuning of the passes based on their specific knowledge).

Another caveat worth mentioning is that I've tested this only on Mac, because my own use of Homebrew to manage an llvm install forced the issue upon me. I made a good effort today to build llvm 15 on Linux at work, but haven't been able to get it built, it may be that our compilers or some other component of our build environment are too old to do it properly. It might take some time to dig into. Anyway, this is a long-winded way of saying that additional problems crop up with llvm 15 on Windows or Linux specifically, since I haven't personally tested them yet and they aren't part of the CI, and we'll address those when and if they do. (I will separately look into how we get an llvm15 into our Linux CI; it obviously already is there for Mac.)

aconty (approved):

PR #1593, opened 2022-09-29
Fix memory leaks found by static analysis in leparse.cpp

lgritz:
Several places in leparse.cpp were found by static analysis to have memory leaks -- results of calling _parse() that, in the case of early exists in error conditions, never got deleted.

I also did some other modernization: changed NULL to nullptr, changed some loops to more compact range-for, utilize some Strutil helpers that will be more efficient than the alternative std::string manipulation.

### DIFF
--- a/src/liboslexec/lpeparse.cpp
+++ b/src/liboslexec/lpeparse.cpp
@@ -47,14 +47,14 @@ Parser::Parser(const std::vector<ustring>* user_events,
     m_minus_stop.insert(Labels::STOP);
 
     if (user_events)
-        for (size_t i = 0; i < user_events->size(); ++i) {
-            m_label_position[(*user_events)[i]] = 0;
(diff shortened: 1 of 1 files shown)


aconty (approved):
Back when I wrote this there weren't range loops, so very good case of bit rot. The leaks are obvious enough that I should buy you a beer or lose my driving license :) Looking at this code makes me wanna rewrite it all with less new/delete. 
Thanks!

lgritz:
If I was smart, I would just have fixed the leak and moved on. (Yay for static analysis!)

I couldn't control myself on the range for loops, they were just sitting there calling to me.

I started to do a bigger refactor with std::unique_ptr and eliminating all the delete calls entirely, but stopped myself. There is zero need to go down that rathole, the code works as it is and the fact that it has barely been touched for 10 years attests to that.

PR #1594, opened 2022-09-30
Fix LPE leak in error condition

lgritz:


lgritz:
One more found by static analysis that I missed in the last PR related to this.

sfriedmapixar (approved):
LGTM

PR #1604, opened 2022-10-10
CI: Disabling dso testshade to fix coverage missing testshade.cpp

lgritz:
We compile testshade.cpp twice -- one for a special "does it compile as a DSO" check but that isn't run from the testsuite. This was causing the coverage analysis to think nothing in testshade.cpp (at least one copy of it) was getting run, which is technically true, and counting all of its lines as untested for the coverage analysis. Just exclude that funny dso build from the coverage case.

Also exclude anything in src/build-scripts from coverage analysis, that's not code that ships, so to speak.

We're trying to chip away at untested code bit by bit to get our coverage to over 80% of lines. We're getting close.

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -18,7 +18,7 @@ sonar.links.issue=https://github.com/AcademySoftwareFoundation/OpenShadingLangua
 # Source properties
 sonar.sources=src
 sonar.sourceEncoding=UTF-8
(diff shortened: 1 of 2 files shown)


lgritz:
Post script: this worked -- after this change, our testing coverage is now up to 79.1%, with no more spurious "0 lines executed in testshade.cpp"!

That last .9% is my next target.

PR #1605, opened 2022-10-10
include the immintrin.h header only when needed

sharkcz:
## Description

The immintrin.h header is required for the implementation of the popcount and coutr_zero functions only when the Intel compiler is used. Move the inclusion accordingly. The current state breaks build on platforms that don't provide the header, but provide the alternative means (gcc/clang built-ins, etc).

## Tests

N/A

## Checklist:

This is not a copyrightable work.

### DIFF
--- a/src/include/OSL/mask.h
+++ b/src/include/OSL/mask.h
@@ -4,7 +4,6 @@
 
 #pragma once
 
-#include <immintrin.h>
 #include <type_traits>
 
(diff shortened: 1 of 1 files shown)


fpsunflower (approved):

AlexMWells (approved):
If all CI pass (including icc and icx batched), then LGTM

lgritz:
Passes CI, so I'll merge it.

PR #1607, opened 2022-10-11
Improve testing coverage

lgritz:
Looking at the code coverage report and trying to get a higher coverage amount, employing two strategies: (a) remove unused code that is not covered because it's... unused; and (b) test things that are used but aren't in the test run. Details follow.

Remove deprecated version of LLVM_Util::make_function.

Disable contents of dump_struct_data_layout when OSL_DEV is not defined, it will never be called in any other circumstances -- it's strictly a debugging tool, not part of the shipping library.

When OSL_USE_OPTIX is not enabled, exclude compiling contents of ptx_compile_group.

get_alignment can also be OSL_DEV only as well as simplified

Arbitrarily pick a couple of render tests to use llvm_opt levels 12 and 13 to verify that they work correctly. (That's a couple hundred lines of pass setup code that wouldn't otherwise be included in the coverage analysis, nor tested in the testsuite.) This also required us to add the --llvm_opt command to testrender, as it is in testshade.

Make the analysis run build for avx512 as well for more code coverage.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - desc: sonary gcc9/C++14 llvm13 py3.9 boost1.76 exr3.1 oiio2.3 avx2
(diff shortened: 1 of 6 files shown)


AlexMWells (approved):
LGTM

PR #1608, opened 2022-10-14
Try to have more optix-only parts hidden behind OSL_USE_OPTIX

lgritz:


lgritz:
This seem ok?

PR #1610, opened 2022-10-19
OSL::bitcast utility

lgritz:


aconty (approved):

src/include/OSL/platform.h
```diff
@@ -499,6 +500,24 @@
 
 OSL_NAMESPACE_ENTER
 
+
```
AlexMWells, line 503: If you compiler/headers do support c++20 and not compiling for CUDA, should this conditionally compiled to be
template <typename To, typename From>
using bitcast = std::bitcast<To, From>;

lgritz: Yeah, but that's the tricky part -- in addition to very few users ready to rely on C++20, the std one still wouldn't have the HOSTDEVICE decorators, so I'm not sure the extra clutter of conditionally choosing between this and the std one gives us any tangible advantage. To the best of my knowledge, this code optimizes to the same thing anyway.

AlexMWells: Ok, it could always be added later should there be a need

PR #1611, opened 2022-10-19
Testing: run pointcloud tests when partio is found

lgritz:
Wow, this has been lurking for a while. Our Findpartio.cmake tries looking for a partio exported config, but if not found, falls back to a simple code to find it manually.

All was fine with the fallback code, which if found would set both "partio_FOUND" and "PARTIO_FOUND". But the config file, if found -- only for the newer partio versions -- only set partio_FOUND. And in that circumstance, we were not seting up the pointcloud tests!

Although the partio tests were running in CI (thankfully), I was not running them locally on my machine with a new-ish partio.

### DIFF
--- a/src/cmake/testing.cmake
+++ b/src/cmake/testing.cmake
@@ -383,7 +383,7 @@ macro (osl_add_all_tests)
     endif()
 
     # Only run pointcloud tests if Partio is found
-    if (PARTIO_FOUND)
+    if (partio_FOUND)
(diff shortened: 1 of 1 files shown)


aconty (approved):

lgritz:
Just to clarify: If Partio was found, we did enable the pointcloud support, regardless of whether it was from an exported config or the find module. It was not a question of disabling the OSL code. But if it was from the exported config, we neglected to run the pointcloud-related tests in the testsuite.

PR #1612, opened 2022-10-20
Checkpoint to phase 2 of the great ustringhash conversion

lgritz:
Reminder: phase 1 changed the RendererServices APIs to be based completely on ustringhash for public method calls, not ustring.

This PR starts along the path of phase 2.

In oslconfig.h, I define a new type `ustringrep` which is the in-memory/in-shader CPU-side representation of a ustring. Currently, it's still equivalent to a ustring.

I also define a type called `ustring_pod`, which is the "plain old data" representation inside a ustringrep (i.e., a const char* when it's a ustring, and a uint64_t when it's a ustringhash). For reasons not important, it's exceedingly hard to pass a true ustring/ustringhash as parameters to or from LLVM-jitted code without running afoul of LLVM's type checking, so it's just easier to pass the underlying data type.

So the main work of phase 2 is to change everything related to shader data from a ustring into a ustringrep, and things passed as const char* into passing a ustring_pod. In some sense, it's just renaming and type conversion plumbing.  When that's done, phase 3 is to throw the switch and change the definition of ustringrep to ustringhash (and of course fix everything that's broken).

This PR does not complete phase 2, but takes a big bite out of it. It's an arbitrary bite, not much rhyme or reason to which specific parts I did so far. Just working on it piece by piece and felt that this was a place where much has been done, everything still works, and it's a reviewable chunk without getting too much out of control.

Areas I worked on are related to range_check, raytype, getattribute, bind_interpolated_param, uninit_check, naninf_check, getmessage/setmessage, some texture setup, pointcloud.

### DIFF
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -1296,8 +1296,13 @@ BackendLLVM::initialize_llvm_group()
         TypeSpec rettype = OSLCompilerImpl::type_from_code(types, &advance);
         types += advance;
         std::vector<llvm::Type*> params;
+        // bool debug = OIIO::Strutil::contains(funcname, "range");
+        // if (debug)
+        //     print("defining func {}:\n", funcname);
         while (*types) {
             TypeSpec t = OSLCompilerImpl::type_from_code(types, &advance);
+            // if (debug)
+            //     print("\t{}\n", t);
             if (t.simpletype().basetype == TypeDesc::UNKNOWN) {
                 OSL_DASSERT(*types == '*');
                 if (*types == '*')
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -1296,13 +1296,8 @@ BackendLLVM::initialize_llvm_group()
         TypeSpec rettype = OSLCompilerImpl::type_from_code(types, &advance);
         types += advance;
         std::vector<llvm::Type*> params;
-        // bool debug = OIIO::Strutil::contains(funcname, "range");
-        // if (debug)
-        //     print("defining func {}:\n", funcname);
         while (*types) {
             TypeSpec t = OSLCompilerImpl::type_from_code(types, &advance);
-            // if (debug)
-            //     print("\t{}\n", t);
             if (t.simpletype().basetype == TypeDesc::UNKNOWN) {
                 OSL_DASSERT(*types == '*');
                 if (*types == '*')
--- a/src/liboslexec/opmessage.cpp
+++ b/src/liboslexec/opmessage.cpp
@@ -27,33 +27,34 @@ namespace pvt {
 
 
 OSL_SHADEOP void
-osl_setmessage(ShaderGlobals* sg, const char* name_, long long type_, void* val,
-               int layeridx, const char* sourcefile_, int sourceline)
+osl_setmessage(ShaderGlobals* sg, ustring_pod name_, long long type_, void* val,
+               int layeridx, ustring_pod sourcefile_, int sourceline)
 {
-    const ustring& name(USTR(name_));
-    const ustring& sourcefile(USTR(sourcefile_));
+    ustringhash name       = ustringhash_from(USTR(name_));
+    ustringhash sourcefile = ustringhash_from(USTR(sourcefile_));
     // recreate TypeDesc -- we just crammed it into an int!
     TypeDesc type   = TYPEDESC(type_);
-    bool is_closure = (type.basetype
-                       == TypeDesc::UNKNOWN);  // secret code for closure
+    bool is_closure = type.basetype == TypeDesc::UNKNOWN;  // indicates closure
     if (is_closure)
         type.basetype = TypeDesc::PTR;  // for closures, we store a pointer
 
     MessageList& messages(sg->context->messages());
     const Message* m = messages.find(name);
-    if (m != NULL) {
+    if (m) {
         if (m->name == name) {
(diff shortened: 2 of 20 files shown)


src/liboslexec/llvm_instance.cpp
```diff
@@ -1294,8 +1294,13 @@ BackendLLVM::initialize_llvm_group()
         TypeSpec rettype = OSLCompilerImpl::type_from_code(types, &advance);
         types += advance;
         std::vector<llvm::Type*> params;
+        // bool debug = OIIO::Strutil::contains(funcname, "range");
```
AlexMWells, line 1297: Did you mean to leave these commented out lines here?

src/liboslexec/opmessage.cpp
```diff
@@ -27,33 +27,34 @@ namespace pvt {
...
 
 OSL_SHADEOP void
-osl_setmessage(ShaderGlobals* sg, const char* name_, long long type_, void* val,
-               int layeridx, const char* sourcefile_, int sourceline)
+osl_setmessage(ShaderGlobals* sg, ustring_pod name_, long long type_, void* val,
+               int layeridx, ustring_pod sourcefile_, int sourceline)
 {
-    const ustring& name(USTR(name_));
-    const ustring& sourcefile(USTR(sourcefile_));
+    ustringhash name       = ustringhash_from(USTR(name_));
```
AlexMWells, line 33: Does USTR(...) casting need to be involved anymore?  I thought a benefit of these changes is to not need that very pointer/reinterpret_cast'ish and just use the ustring_rep or ustring_pod?

src/liboslexec/oslexec_pvt.h
```diff
@@ -234,18 +234,33 @@ struct AttributeNeeded {
...
-#endif
+inline ustringrep
+USTR(ustring_pod s) noexcept
+{
+    return OSL::bitcast<ustringrep>(s);
+}
+inline ustringrep
+USTREP(ustring_pod s) noexcept
+{
+    return OSL::bitcast<ustringrep>(s);
```
AlexMWells, line 245: Debatable here, but instead of a bitcast, can't we construct a ustringrep, ie:
return ustringrep{s};

For that matter, perhaps remove these casts all together replacing them at callsites with
ustringrep{string_param_pod}

AlexMWells:
When all this is totally done (past stage 3 with the switch thrown), would the idea be to remove ustringrep and just have ustringhash?

src/liboslexec/llvm_instance.cpp
lgritz: no, that's a mistake, will remove

src/liboslexec/opmessage.cpp
lgritz: I think in most cases, that's where we'll eventually get to, yes.

src/liboslexec/oslexec_pvt.h
```diff
@@ -234,18 +234,33 @@ struct AttributeNeeded {
...
-#if OPENIMAGEIO_VERSION >= 20500
-#    define TYPEDESC(x) OIIO::bitcast<TypeDesc, long long>(x)
-#else
-#    define TYPEDESC(x) OIIO::bit_cast<long long, TypeDesc>(x)
-#endif
+inline ustringrep
+USTR(ustring_pod s) noexcept
+{
+    return OSL::bitcast<ustringrep>(s);
+}
```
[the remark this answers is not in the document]
lgritz: I think I'm leaning that way as I touch things. Mostly I was trying USTREP on for size because USTR was making me continually wonder whether it was a ustring or a ustringrep.
Will transition this over time.

lgritz:
> Consider changing name of `ustring_pod `to `ustring_param` as that makes it more clear where its going to be used.

Ha, it was ustring_param until the last minute, when I thought that was unclear and renamed it ustring_pod. I'm still on the fence about the final name. I think where my head was can be summarized as that "param" made me think "how is this different from passing a ustring as the param?" whereas the pod was to emphasize that it's the non-class, plain-old-data representation of what's inside a ustring.

lgritz:
> When all this is totally done (past stage 3 with the switch thrown), would the idea be to remove ustringrep and just have ustringhash?

Yeah, I think so. I wanted a temporary name in the interim that I can easy switch back and forth. Once everything is working, it may just get wholesale replaced by ustringhash.

lgritz:
I think we should treat all this nomenclature as temporary. When this work gets completed, some of them will no longer be necessary at all, and for others we can then choose what names we want on a permanent basis.

AlexMWells:
> it was ustring_param until the last minute, when I thought that was unclear and renamed it ustring_pod. I'm still on the fence about the final name. I think where my head was can be summarized as that "param" made me think "how is this different from passing a ustring as the param?" whereas the pod was to emphasize that it's the non-class, plain-old-data representation of what's inside a ustring.

I had similar issue, mostly when do I use ustring_pod vs. ustring vs ustringhash

AlexMWells:
I think it just might be a bit weird because its in an intermediate state.  Forward looking, 
ustringrep would just be gone, replaced with ustringhash
ustring_pod would be gone, replaced with ustringhash::hash_t

lgritz:
Yeah, I think something like that is the final state.

AlexMWells:
Maybe don't worry about the names then (as they are intermediate) but I do recommend killing off the casting USTR macros and using helper functions.

lgritz:
Ok, I'm going to merge as it is and move on, cleaning up these things we've mentioned as I go.

Issue #1619, opened 2022-11-14
Varying Closure Keyword Parameters Not Being Correctly Evaluated During Batched Execution


danieldresser-ie:
### Problem

It appears that closure parameters declared with CLOSURE_*_KEYPARAM don't work correctly with varying inputs in batched execution.

I've now got a pretty simple repro, and I've tested with the recently released 1.12.7.0, so I think this is probably a legit issue:

badBatch.osl
```
shader badBatch()
{
    Ci = diffuse( N, "label", format( "%f", u ) );
    printf( " Ci = %s", Ci );
}
```

Testing
```
> oslc badBatch.osl && testshade -g 2 1 badBatch
Ci = (1, 1, 1) * diffuse ((0, 0, 1), "label", "0.000000") Ci = (1, 1, 1) * diffuse ((0, 0, 1), "label", "1.000000")
```

```
> oslc badBatch.osl && testshade -batched -g 2 1 badBatch
Ci = (1, 1, 1) * diffuse ((0, 0, 1), "label", "0.000000") Ci = (1, 1, 1) * diffuse ((0, 0, 1), "label", "0.000000")
```

Passing the `-batched` flag causing the second shading point to incorrectly get the same value for "label" as the first shading point.

### Versions

* OSL branch/version:  1.12.7.0 ( first noticed in 1.12.6.2 )
* OS: CentOS 7.9
* C++ compiler: gcc-9.3.1
* LLVM version: 13.0.0
* OIIO version: 2.3.11.0

lgritz:
I believe this was closed by #1620

PR #1620, opened 2022-11-16
Fix Issue With Evaluating Varying Closure Keyword Parameters During Batched Evaluation

danieldresser-ie:
When evaluating varying closure keyword parameters in batched execution, the value of the first lane in the batch would be used for the whole batch.  More details in an issue I made here: #1619

I don't really understand this code at all, this fix was just made by following the suggestion of Stephen on the dev list to copy the code from llvm_gen_closure to llvm_gen_keyword_fill.

It fixes the one simple case which I've added a test for, but the more complex functionality of this code I copied would only be tested if you used array keyword parameters on a closure.  This would be more annoying to test, since I don't think testshade declares any closures with array keyword parameters.

Also, while testing, I found a segfault with adding a null closure in batched evaluation - I'll post a repro for that on the dev list, and hopefully we can clean that up soon too.


src/liboslexec/batched_llvm_gen.cpp
```diff
@@ -7088,10 +7088,43 @@ llvm_gen_keyword_fill(BatchedBackendLLVM& rop, Opcode& op,
...
+
+                llvm::Value* dest_base = rop.ll.offset_ptr(mem_void_ptr, p.offset);
+                dest_base              = rop.llvm_ptr_cast(dest_base, p.type);
+
+                for (int a = 0; a < num_elements; ++a) {
+                    llvm::Value* arrind = simpletype.arraylen ? rop.ll.constant(a)
+                                                               : NULL;
+                    for (int c = 0; c < num_components; c++) {
+                        llvm::Value* dest_elem;
+                        if (num_components > 1)
```
AlexMWells, line 7106: What happens for an array of of objects with 1 compoenent?
Looks like maybe the else case is taken which disregards the "a" which is the element index.
Does the else case need to incorporate the arrayind or "a"

danieldresser-ie: Hmm, that does look like a bug to me too ... which quite possibly means it is also a bug on line 7243 where I copied the code from?

Not that I want to make it harder to get this quick fix I need over the line, but I do wonder if we should add a closure with more types of parameters to testshade so that it's possible to actually write testcases for this code.

AlexMWells: This same comment might apply to the copy loop in llvm_gen_closure of batched_llvm_gen.cpp

danieldresser-ie: This started out as a quick fix to a bug that was holding me up ... modifying testshade seems like a somewhat larger change - adding new closure parameters to testshade feels a bit more involved, since they would potentially be visible to other users of OSL.  Does someone want to decide what this should look like?  Should we add some keyword parameters to `debug()`?  Oh, wait, looking at the list of macros in genclosure.h, it looks like keyword array parameters aren't currently supported anyway?  So potentially I could drop the iteration code and use a simple memcpy - though the bug would still potentially exist in the non-keyword array parameter case that already exists.  Should we add a special `test` closure with a parameter of every type in genclosure.h?

sfriedmapixar: I agree, that looks like a bug.  I think the two main purposes of testshade are to provide an example of how to integrate the library and to allow for testing the important features of the library.  A special 'test' closure which exercises the types in genclosure.h would be something that accomplishes both, so sounds like a good idea to me -- it'll keep me from accidentally checking things like this in that work for our renderer which doesn't make heavy use of closures but break those that depend on them heavily.

AlexMWells: I think you can still get this fix over the fence without everything else, maybe remove the array index stuff if you know its not needed, replace it with an assert?  You have added a test case and modified source to allow it to pass without breaking any other testcases.  We just noticed some other issues along the way.  Maybe drop a "todo" on the line 7243 to add testcase and fix bug.

danieldresser-ie:
As per Alex's comment, I've removed support for array indexing, avoiding the bug he observed with the faulty array handling I copied.  I wasn't able to get it working with one big memcpy ... unsure if this is due to not understanding the context well enough, or whether a per-component approach is required in order to get masking and such working.  Anyway, I started from the other end, with the more complex code that worked, and removed the array handling but kept the component handling - this seems to have resulted in working code, and I've removed the array part, and added an assert that the arraylen is 0.

As for the bug at line 7243, I've added #1621 with a test case that demonstrates that bug - does it make sense to merge this and then continue that part of the conversation over there?

AlexMWells (approved):
Looks Great To Me

PR #1621, opened 2022-11-18
Add new testshade closure to allow testing all parameter types

danieldresser-ie:
While working on #1620, it came up that it is currently difficult to test some closure parameter types.  Folks seemed to agree that it would be a good idea to have a special closure supported by testshade which exercises all the parameter types defined in genclosure.h.

This is a first pass at adding that test closure.  It fails without #1620, but even with #1620, it fails on the array closure types in batched mode.  This is expected based on the bug that @AlexMWells spotted ( in llvm_gen_closure in batched_llvm_gen, while looping through elements and components, it incorrectly ignores the element index `a` unless `num_components > 1` )  - but I tried an obvious fix based on using a zero offset to the first element, but still using the component offset, and that didn't fix the issue - there's probably something obvious here that I'm missing because I'm not that familiar with LLVM.  Anyway, perhaps this sketch of a test might be useful to someone who is able to fix this.

## Checklist:


sfriedmapixar (approved):
Looks good to me -- thanks for setting this up!
However -- it does point out the existing bug in the batched side, so until that's fixed, perhaps we should leave BATCHED out, and once that issue is addressed, we can re-enable it -- or if PR #1620 addresses this, just merge that one first

lgritz:
I think this is a great idea, thanks.

Unfortunately, looks like many of the CI tests are failing with llvm errors. Can you take a look and see if you can figure out what's broken?

lgritz:
I didn't read carefully. I see now that you said this fails without the other PR merged.

src/testshade/simplerend.cpp
```diff
@@ -84,6 +85,23 @@ struct MicrofacetParams {
 struct DebugParams {
     ustring tag;
 };
+struct ParameterTestParams {
+    int int_param;
+    float float_param;
+    Color3 color_param;
+    Vec3 vector_param;
+    ustring string_param;
+    int int_array[3];
```
AlexMWells, line 94: Wondering if a size other than 3 can be used to avoid any confusion with triples.  By confusion I mean anyone debugging llvm IR might confuse triple values with an array length 3.  And a number other than 3 might stick out better.  maybe [5]

AlexMWells:
I don't see any file changes to fix the array issue in batched llvm_gen_closure .  Which is fine.  I suggest removing the file testsuite/closure-parameters/BATCHED from this PR and having this just add the new test closure type and test.  Then in a different PR add BATCHED and the fix for batched llvm_gen_closure, or adding the fix (or attempted fix) here so we can review/comment/iterate the change until the tests pass.

danieldresser-ie:
As suggested by Alex, I have removed the BATCHED flag file, and have switched all the array lengths from 3 to 5 ( hopefully 5 is a length that is pretty distinct from other types ).

AlexMWells (approved):
Looks Great To Me!

lgritz:
The CI failures don't appear to be related to these changes, and are happening on the nightly builds as well. Merging.

lgritz:
Hey, @danieldresser-ie, just curious about the status of this... it wasn't completely clear if you were planning on submitting a follow-up PR for the issue here on the batch side? Or if you were only submitting the test and assuming someone else would pick up the task of making it all work for the batch case?

danieldresser-ie:
I'm honestly not quite sure - the process of adding this test allowed me to have some confidence that #1620 was implemented correctly - it does work for the non-array keyword argument case, which was difficult to test before.  That was my main concern.

The next thing I'll probably take a look at is a separate issue - the crash when adding a default initialized closure to another closure in batch mode - I isolated an issue in an email with Stephen, and Stephen suggested a solution - that one seems pretty straightforward, and hopefully I'll be able to tackle it ( and that one does affect a Gaffer use case ).

I'm fairly out of my depth in this codebase, having never touched LLVM before.  In the case of this issue, my first basic attempt at fixing the possible cause that Alex identified didn't fix the crash.  If I can successfully fix that other crash, I might come back to this for one more try, but I certainly not guaranteeing I'll be able to fix it.  Hopefully if I don't find the fix, this test will be useful to someone who can fix it.

danieldresser-ie:
To follow up on this - I took another look at the issues with batch mode.  Alex had noted that there seems to be an obvious bug in batched llvm_gen_closure when num_components == 1, since it fails to take into account the array index `a`.  So my hope was I could take a look at how it was working when num_components > 1, and make it work the same.

However, testing reveals that this code currently doesn't work for any sort of arrays ... after staring at it for a bit, I am somewhat confused as to whether we're storing contiguous arrays for each lane, or interleaving elements between lanes ... and I think maybe the code is confused as well, but I can't quite put my finger on it.

This might require someone who is actually familiar with this code base to sort it out.

PR #1624, opened 2022-11-30
Fix Batched Mode Crash When Adding A Default Initialized Closure

danieldresser-ie:
## Description

This is another fix I'm attempting without really understanding all this LLVM stuff.

I started by adding a simple test case that attempts to add two closure parameters, where one has not been set.  This crashes or returns random values from uninitialized memory, matching a problem I ran into while trying to get batched mode working in Gaffer.

I investigated some, and found that add_closure_closure is assuming closure pointers are stored with the width of the current batch, but build_llvm_init is creating uniform closures.  I checked with @sfriedmapixar, who confirmed that all the batched closure code is assuming wide closures, so it seems build_llvm_init is what needs fixing.

Adding a `sym.make_varying();` to build_llvm_init appears to fix the problem ... I really don't know enough about the overall structure to know if this is the right place to make this fix, or if it could have other consequences, but it's enough to get this test passing, so hopefully someone who does know the background can review it.

## Checklist:

### DIFF
--- a/src/liboslexec/batched_llvm_gen.cpp
+++ b/src/liboslexec/batched_llvm_gen.cpp
@@ -1263,6 +1263,7 @@ LLVMGEN(llvm_gen_add)
     OSL_ASSERT(!A.typespec().is_array() && !B.typespec().is_array());
     if (Result.typespec().is_closure()) {
         OSL_ASSERT(A.typespec().is_closure() && B.typespec().is_closure());
+        OSL_ASSERT(!A.is_uniform() && !B.is_uniform() && !result_is_uniform);
         llvm::Value* valargs[] = { rop.sg_void_ptr(), rop.llvm_void_ptr(Result),
                                    rop.llvm_void_ptr(A), rop.llvm_void_ptr(B),
                                    rop.ll.mask_as_int(rop.ll.current_mask()) };
--- a/src/liboslexec/batched_llvm_instance.cpp
+++ b/src/liboslexec/batched_llvm_instance.cpp
@@ -1813,6 +1813,7 @@ BatchedBackendLLVM::build_llvm_init()
         FOREACH_PARAM(Symbol & sym, gi)
         {
             if (sym.typespec().is_closure_based()) {
+                OSL_ASSERT(sym.is_varying());
                 int arraylen     = std::max(1, sym.typespec().arraylength());
                 llvm::Value* val = ll.constant_ptr(NULL, ll.type_void_ptr());
                 if (!sym.is_uniform())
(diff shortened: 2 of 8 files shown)


src/liboslexec/batched_llvm_gen.cpp
```diff
@@ -1263,6 +1263,7 @@ LLVMGEN(llvm_gen_add)
     OSL_ASSERT(!A.typespec().is_array() && !B.typespec().is_array());
     if (Result.typespec().is_closure()) {
         OSL_ASSERT(A.typespec().is_closure() && B.typespec().is_closure());
+        OSL_ASSERT( !A.is_uniform() && !B.is_uniform() && !result_is_uniform );
```
AlexMWells, line 1266: Good assert to verify our assumption that any closures will implicitly be varying

src/liboslexec/batched_llvm_instance.cpp
```diff
@@ -1813,10 +1813,10 @@ BatchedBackendLLVM::build_llvm_init()
         FOREACH_PARAM(Symbol & sym, gi)
         {
             if (sym.typespec().is_closure_based()) {
+                sym.make_varying();
+
                 int arraylen     = std::max(1, sym.typespec().arraylength());
-                llvm::Value* val = ll.constant_ptr(NULL, ll.type_void_ptr());
```
AlexMWells, line 1817: So the original code here should have sufficed.  maybe could just put 
OSL_ASSERT(sym.is_varying());

Which will probably fire? (let me know if it does).   If the assert does fire that means there is a bug in BatchedAnalysis, which gets run as part of the runtimeoptimizer.  BatchedAnalysis is responsible for calling symbol::make_varying() not sure anyone else should be calling it (maybe it should be made private and friend BatchedAnalysis).

Even though this allows the unit test to pass, calling mark_varying() outside of BatchedAnalysis means that any operations that use that symbol downstream will notice that their input was varying and therefore so should their result be varying.

If you add at the top of batched_analysis.cpp
#define OSL_DEV 1
it will dump the results of batched_analysis and can see more of what's going on.

danieldresser-ie: I've switched the build_llvm_init changes to just this assert - it is indeed triggering.

I again, don't know anything about BatchedAnalysis, but as a placeholder, I added a quick commit where I just force it to tag all closures as varying - this seems to resolve the assert in build_llvm_init and get these test passing again, so that supports the idea that BatchedAnalysis is the right place to fix this.

AlexMWells (approved):
That is exactly how to fix this!

Approved!

sfriedmapixar:
Nice fix -- Looks good to me!

PR #1625, opened 2022-12-02
Fix Issues With Derivatives of getattribute

danieldresser-ie:
Adding a test for getting the derivatives of custom userdata accessed through getattribute reveals a couple of problems.

The first is very simple:  in testshade's simplerend.cpp, in get_userdata, there is a branch for dealing with derivatives, but it is completely wrong.  The values stored in custom userdata aren't stored with derivatives, so we  should just set the derivatives to 0 rather than reading from uninitialized memory.

In batched_simplerend, it calls assign_all_from_scalar, which requires the source pointer to point to something stored with derivatives when the target has_derivs() is true.  This is a pretty inconvenient API - I have made it call assign_all_from_scalar correctly here by allocating a buffer large enough to hold both the data and derivatives, and zeroing out the derivatives - but obviously this pointless allocation is sub-optimal.  A better solution would be a version of assign_all_from_scalar that doesn't require the derivatives ... either allowing the derivatives to be set separately, or just assuming they are zero.  Actually, the more I think about this, the more I think assign_all_from_scalar should never require derivatives, and always set them to zero ( if derivatives are present ).  What use case could there be for setting all lanes to the same value ( implying the value doesn't vary at all ) but setting non-zero derivatives ( implying the value does vary spatially )?

Usually, I would think we couldn't just switch an API like this to make assign_all_from_scalar not use derivatives, because it could affect existing uses of assign_all_from_scalar, but in this case it might be pretty safe, because of the bug that originally caught my attention:  if you do use assign_all_from_scalar with derivatives, even if you have correctly allocated them, it applies an offset to the address for source derivatives in both assign_all_from_scalar  and assign_all_from_scalar_type, resulting in a double offset that again hits uninitialized memory - since this has never been working, maybe it would make sense to change how it behaves?

## Checklist:


danieldresser-ie:
Just pinging this PR since it's been a while.  As far as I know, this is the final thing that needs sorting out to get all Gaffer functionality working with batched OSL execution, which would be great to get into our January release.

Anything I can do to help get this worked out?

AlexMWells:
Sorry, been swamped.  I'll take a look 12/15 or 12/16, regards.  Thanks for your efforts!

AlexMWells:
Please refer to
https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1630
If that is acceptable suggest we close this PR.

danieldresser-ie:
As far as I can tell, #1630 looks great!  Thanks a lot, and yeah, you can go ahead and close this.

lgritz:
I merged #1630, so closing this one. Thanks for the help.

PR #1627, opened 2022-12-08
Remove dependency on OptiX SDK for building core OSL with OptiX support

pmoursnv:
OSL currently includes the OptiX header in several places, but without actually making any use of its contents. OptiX support is implemented soly using LLVM and its NVPTX backend, without any types or functions used from the OptiX SDK, making these includes superfluous. The only part of the codebase that does need them are the tests, so this patch makes finding the SDK dependent on whether tests are enabled or not.
Avoiding including the OptiX headers in the core makes it possible still to build OSL with OptiX support (`USE_OPTIX=ON;OSL_BUILD_TESTS=OFF;`), but without needing the OptiX SDK in the process, simplifying the build process on build systems where having the OptiX SDK installed is a challenge.

Additionally this patch also makes it possible to use OSL in OptiX without pre-built LLVM bitcode containing OSL intrinsics (`USE_LLVM_BITCODE=OFF;`). Building that bitcode module requires the CUDA toolkit, which too can be a challenge to have available on build systems, but it's not technically needed if the renderer implementing OSL does provide an implementation for all the OSL intrinsics declared in `builtindecl.h` itself (and it already had to do so for some of them right now anyway). In that case only have to initialize the now empty LLVM module OSL creates with a proper target and data layout (which would normally be inherited from the pre-built bitcode), otherwise compiling shader groups will later fail, as the NVPTX backend is looked up based on the target triple set on the module (see implementation of `LLVM_Util::ptx_compile_group()`).

CMake previously errored when attempting to build with OptiX support and without the LLVM bitcode, but due to the previous change it only does that when tests are enabled too now, making the following a valid combination of build options when no CUDA/OptiX SDK is available, but OptiX support is still desired: `USE_OPTIX=ON;USE_LLVM_BITCODE=OFF;OSL_BUILD_TESTS=OFF;`. This is e.g. the case with Blender (more discussion regarding this on https://developer.blender.org/D15902).

Note: The change to externalpackages.cmake looks larger than it is due whitespace changes caused by the new if clause.

## Checklist:

- [x] I have ensured that the change is tested somewhere in the testsuite (adding new test cases if necessary). [Existing tests build and run]

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -167,44 +167,46 @@ endif ()
 
 # CUDA setup
 if (USE_CUDA OR USE_OPTIX)
-    if (NOT CUDA_TOOLKIT_ROOT_DIR AND NOT $ENV{CUDA_TOOLKIT_ROOT_DIR} STREQUAL "")
-        set (CUDA_TOOLKIT_ROOT_DIR $ENV{CUDA_TOOLKIT_ROOT_DIR})
-    endif ()
+    if (USE_LLVM_BITCODE)
+        if (NOT CUDA_TOOLKIT_ROOT_DIR AND NOT $ENV{CUDA_TOOLKIT_ROOT_DIR} STREQUAL "")
+            set (CUDA_TOOLKIT_ROOT_DIR $ENV{CUDA_TOOLKIT_ROOT_DIR})
+        endif ()
 
-    if (NOT CUDA_FIND_QUIETLY OR NOT OptiX_FIND_QUIETLY)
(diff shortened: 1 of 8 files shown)


brechtvl:
The changes look good to me.

The failing `VFX2022 clang-format` check can be fixed by reordering the includes in `src/liboslexec/oslexec_pvt.h`.

The other 3 failures look like preexisting issues not caused by this pull request.

lgritz:
I know why the mac tests are failing, it's on me to fix. (I've been on vacation.)

I'll look into that other failure, too, which happens when setting up dependencies, so I don't think it's at all related to this PR, just ignore it.

Please fix the formatting for the clang-format failure, and then I'll merge this. Thanks for the fix!

PR #1635, opened 2023-01-06
platform.h: fully define bitcast template

lgritz:
Ported from OIIO. This lets us not have to depend on the specific definition in OIIO's fmath.h, and in particular not worry about whether it's a new enough version of OIIO that the old bit_cast is now marked as deprecated.


aconty (approved):

AlexMWells (approved):
LGTM

PR #1636, opened 2023-01-12
Remove unnecessary setting of locale::global

danieldresser-ie:
We used to need to set the global locale before calling something that used OIIO parsing functions.  This is no longer the case ( as of OIIO-1.9, and we now require 2.3 ), so we can now remove these calls.

It came up for me because this code seems to interact strangely with locale in Python - I think this is because something synchs the C++ locale across to Python, but if locale::global hasn't been called in C++ yet, and the locale has only been set in Python, oldlocale doesn't get set to the Python locale ... I could try to figure out more of what's going on, but the simple answer is just to remove this code that is no longer needed.  LG though it was reasonable when I asked on the dev list: https://lists.aswf.io/g/osl-dev/message/5081

A potential follow-up would be to move the lock_guard inside parse, since there's a comment indicating that the locale was the reason this hadn't been done.

## Checklist:

### DIFF
--- a/src/liboslcomp/osllex.l
+++ b/src/liboslcomp/osllex.l
@@ -397,15 +397,6 @@ OSLCompilerImpl::osl_parse_buffer (const std::string &preprocessed_buffer)
     // Thread safety with the lexer/parser
     std::lock_guard<std::mutex> lock(oslcompiler_mutex);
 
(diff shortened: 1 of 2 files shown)


lgritz (approved):
LGTM, thanks!

PR #1638, opened 2023-01-19
Fix some windows GPU build errors

DeclanRussell:
## Description

This PR fixes a couple of build failures when building `USE_OPTIX=true` on windows. Firstly, I added some search paths to correctly find the CUDA toolkit libs on windows. Secondly, this align on the array, doesn't seem to compile on windows, which is weird.
```
OSL_CONSTANT_DATA const float cie_colour_match[81 * 3] OSL_ALIGNAS(64) =
```
I get the following error when generating `opcolor_cuda.bc`
```
src/liboslexec/opcolor_impl.h(51,54): error G9BCB7FFB: 'alignas' attribute cannot be applied to types
```
Therefore, I have just ifdef'ed out the align when compiling for GPU. Correct me if I am wrong, but as I understand it the align is not really important for GPU anyway and is mainly for SIMD access on the CPU?

## Checklist:

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -190,14 +190,14 @@ if (USE_CUDA OR USE_OPTIX)
         set (CUDA_LIB_FLAGS "--cuda-path=${CUDA_TOOLKIT_ROOT_DIR}")
 
         find_library(cuda_lib NAMES cudart
-                    PATHS "${CUDA_TOOLKIT_ROOT_DIR}/lib64" "${CUDA_TOOLKIT_ROOT_DIR}/x64"
+                    PATHS "${CUDA_TOOLKIT_ROOT_DIR}/lib64" "${CUDA_TOOLKIT_ROOT_DIR}/x64" "${CUDA_TOOLKIT_ROOT_DIR}/lib/x64"
                     REQUIRED)
         set(CUDA_LIBRARIES ${cuda_lib})
 
         # testrender & testshade need libnvrtc
         if ("${CUDA_VERSION}" VERSION_GREATER_EQUAL "10.0")
             find_library(nvrtc_lib NAMES nvrtc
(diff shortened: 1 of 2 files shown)


AlexMWells:
For the alignas it looks like it might be considering it a type declaration vs. an instantiation.
I've seen lots of "interesting" behavior around arrays.  Typical solution involves declaring a typedef for the array before its instantiation:
https://www.godbolt.org/z/arMKTofxh

Something like:
typedef const float CieColourMatchLookup[81 * 3];
CieColourMatchLookup cie_colour_match alignas(64) = { 0.0076,0.0002,0.0362, ..., 0.0232,0.0006,0.1102};

It that works for you, it would be prettier than conditional compilation

lgritz:
@DeclanRussell Can you please fix the clang-format and DCO?  For the clang-format issue, I think is should be sufficient to move the `// clang-format on` / `off` to fully surround the declaration. The DCO fix just involves amending the PR so that all of your commits are edited to include the "Signed-off-by:" line.

(Don't worry about the Mac CI failure, that appears unrelated to this PR.)

lgritz:
Ping @DeclanRussell -- If you are able to fix the DCO and clang-format in the next few days, I can include it in the Feb 1 release.

DeclanRussell:
> Ping @DeclanRussell -- If you are able to fix the DCO and clang-format in the next few days, I can include it in the Feb 1 release.

I'll try and get it done today :)

DeclanRussell:
I've fixed the clang-format and the DCO. Unfortunately, I haven't been able to get around to trying Alex's suggestion yet. With the fast approaching release, perhaps I can just do that cleanup in a future PR.

lgritz (approved):
Looks fine for now. If you want to pursue Alex's suggestion, it's fine to do that in a subsequent PR.

PR #1640, opened 2023-01-25
Changed snprintf formatting to satisfy some compilers.

pellerington:
Signed-off-by: Peter Ellerington <elleringtonp@gmail.com>

## Description

When building OSL my compiler (GCC 11.3 on Ubuntu) was complaining about these snprintfs format strings not being a string literal.
Writing them to a "%s" seems to satisfy it and should produce the same result.

## Tests

no changes in tests

## Checklist:

### DIFF
--- a/src/testrender/optixraytracer.cpp
+++ b/src/testrender/optixraytracer.cpp
@@ -1072,7 +1072,7 @@ OptixRaytracer::processPrintfBuffer(void* buffer_data, size_t buffer_size)
                     switch (format[j]) {
                     case '%':
                         // seems like a silly to print a '%', but it keeps the logic parallel with the other cases
-                        dst += snprintf(&buffer[dst], BufferSize - dst,
+                        dst += snprintf(&buffer[dst], BufferSize - dst, "%s",
                                         fmt_string.c_str());
                         format_end_found = true;
                         break;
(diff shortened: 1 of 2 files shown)


lgritz (commented):
Oh wow, yes. Thanks for the fix.

Can you please get the CLA squared away, and then I'll merge this.

pellerington:
Okay, I think I've done everything now.

I've made some other modifications to make integrating my renderer easier. Let me know if your interested!

lgritz (approved):

lgritz:
Sure, what other modifications did you have in mind?

pellerington:
The main change I added was a optional register closure which lets that closure allocate with a unique function.
Eg. for microfacet instead of "osl_allocate_closure_component(...)" being run a function of the form "osl_microfacet(sg, weight, distribution, normal, roughnes ... )" gets run.

It's slightly more work because the renderer needs to implement a function for each closure and copy the arguments more manually, but there's much more freedom.

It let me do processing/interfacing with my renderer memory directly in that function (rather than storing args then looping back over everything after OSL had run).
It also meant I didn't have to deal with ClosureComponent data and offset calculations/alignment issues on the gpu. I could just define a class (eg. ClosureMicrofacet) and copy the data there.
Also a simpler register closure function because you just have to list the args (not their alignment/size in the struct).

PR #1641, opened 2023-01-31
RendererServices API for letting get_texture_handle consider colorspace

lgritz:
This makes some minor changes to the RS APIs to add an additional optional parameter to get_texture_handle so it takes a `TextureOpt*` to accommodate a similar change on the OIIO TextureSystem side from https://github.com/OpenImageIO/oiio/pull/3761

The purpose here is to add color management options to texture calls. However, please note that this PR only reserves a public API change we know we'll want, but it does not fully implement the color management. Further work will be needed, as follows:

* Current calls to RS::get_texture_handle should be updated to pass the TextureOpt*. (This is a formality for future-proofing in case future TS or alternate TS implementations want to bake the color space choice into the texture handle itself. But the OIIO change doesn't currently rely on that at the moment.)

* When the texture.h symbol `OIIO_TEXTURESYSTEM_SUPPORTS_COLORSPACE` exists, we are building against an OIIO that has the new int field `colortransformid` in the TextureOpt and TextureOptBatch structures. We should fill this in with a nonzero value when we want the texture call to be doing a color transformation as it reads texture tiles.

* Use the new TextureSystem:get_colortransform_id() to get that color transform ID given two named from/to color spaces. We agreed that it's a reasonable restriction that the color space names should be known by the end of runtime optimization, so that should be constant foldable.

* Recognize new optional argument(s) to the texture() call family that lets the caller specify the presumed color space of the texture, and use that value (with constant folded lookup of the ID, above) to set the value in the TextureOpt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 
 cmake_minimum_required (VERSION 3.12)
 
-set (OSL_VERSION "1.13.2.1")
+set (OSL_VERSION "1.13.3.0")
 set (OSL_VERSION_OVERRIDE "" CACHE STRING
      "Version override (use with caution)!")
 mark_as_advanced (OSL_VERSION_OVERRIDE)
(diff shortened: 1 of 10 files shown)


lgritz:
I will merge this at end of day tomorrow unless somebody has concerns.

PR #1649, opened 2023-03-01
testrender: Don't use the cached background map on the first bounce

fpsunflower:
This lets users customize the behavior of the background for camera rays (requested by the MaterialX team). Without this, you always get the cached copy which is only valid for one raytype (and has limited resolution) or you don't get any importance sampling of the environment.


fpsunflower:
As discussed on slack, I also bumped the default resolution in the event that someone forgets to set it in the xml so that we get better sampling out of the box.

lgritz:
I think that this changes a couple of the test outputs slightly so they need new ref images. If you don't have time to do that tonight, I believe I can just retrieve the updated images from the CI and push another commit on top of yours that fixes it.

jstone-lucasfilm (commented):
Looks good to me, thanks @fpsunflower!

PR #1653, opened 2023-03-13
oslc: simple constant folding of binary expressions on the oslc side

lgritz:
The runtime optimization does a great job of constant folding, and we'd pushed it there because, with instance parameter values and shader network connections known, it can find so much more opportunity to optimize than we could in oslc.

However, there is one pesky problem, which is that shader parameters that are initialized to even simple expressions such as 3+1 end up with "init ops", which although later constant folded by the time the shader is JITed, make it impossible to know the value via OSLQuery.

So this patch just takes the simplest cases -- certain `int OP int` and `float OP float` expressions involving literal constants, and performs the operation as it's parsing the code. So

    int val = 3 + 1;

actually just immediately is turned into `val = 4` instead of making an "add".

To reiterate, the add would never have happened while executing the shader -- at runtime in the renderer, when it's time to optimize and JIT the shader, it would know it's a constant 4 value. This is strictly about making oslc directly output an .oso file that knows that parameter `val` has default value 4 instead of throwing up its hands and saying "it's math code that will be evaluated later."


lgritz:
Anybody have comments or concerns about this?

fpsunflower (approved):
LGTM!

I am curious what real world idiom led to this being necessary. Is it common to init shader parameter with `foo = 3 + 1` ?

lgritz:
The idiom is actually `int myparam = enumA | enumB`, and we wanted oslquery to be able to query the value instead of saying "init ops, you dope!"

PR #1657, opened 2023-03-20
Enable userdata derivatives for interpolated params on GPU

chellmuth:
## Description

Enable userdata derivatives for interpolated parameters on the GPU.

When generating calls to `osl_bind_interpolated_param` in llvm_instance.cpp, the `userdata_has_derivs` field is set to true if the parameter's derivatives are needed, but the code does not take into account that on gpu there is no space allocated for these derivatives.

This leads to an incorrect derivatives pointer getting passed into the renderer, which the renderer will use to write into unrelated groupdata fields.

For our renders, getting the correct image was worth the 1-2% inflation of groupdata struct size that we observed by simply re-enabling gpu userdata derivatives. If desired, adding an option to turn them on/off is straightforward, but I didn't want to preemptively bloat the code if it's not necessary.

## Checklist:

### DIFF
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -279,14 +279,10 @@ BackendLLVM::llvm_type_groupdata()
         ++order;
         for (int i = 0; i < nuserdata; ++i) {
             TypeDesc type = types[i];
-            // NB: Userdata derivs are not currently supported in OptiX, since
-            //     making room for them in the GroupData struct can result in a
-            //     large per-thread memory allocation, which could negatively
-            //     impact performance.
-            int n         = (!use_optix()) ? type.numelements()
-                                         * 3  // make room for derivs by default
-                                           : type.numelements();
-            type.arraylen = n;
+            // make room for float derivs only
+            type.arraylen = type.basetype == TypeDesc::FLOAT
+                                ? type.numelements() * 3
+                                : type.numelements();
             fields.push_back(llvm_type(type));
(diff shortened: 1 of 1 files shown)


src/liboslexec/llvm_instance.cpp
```diff
@@ -279,14 +279,7 @@ BackendLLVM::llvm_type_groupdata()
...
             TypeDesc type = types[i];
-            // NB: Userdata derivs are not currently supported in OptiX, since
-            //     making room for them in the GroupData struct can result in a
-            //     large per-thread memory allocation, which could negatively
-            //     impact performance.
-            int n         = (!use_optix()) ? type.numelements()
-                                         * 3  // make room for derivs by default
-                                           : type.numelements();
-            type.arraylen = n;
+            type.arraylen = type.numelements() * 3;  // make room for derivs by default
```
lgritz, line 282: Now that I'm looking at this, I think that we should only multiply by 3 if `type.basetype == TypeDesc::FLOAT`. Because anything based on int or string can't possibly have derivatives, so perhaps you can reclaim a little space that way (if indeed anything is ever declared userdata with those types, but it is possible).

Also, I'm seeing a CI complaint about clang-format on this line.

lgritz (changes requested):

src/liboslexec/llvm_instance.cpp
chellmuth: Sounds good, I pushed an update addressing both points.

lgritz (approved):
LGTM

PR #1658, opened 2023-03-20
build: proper warnings about LLVM 16

lgritz:
LLVM 16 was just released. I'm pretty sure we won't build properly against it because of the phase-out of typed pointers in LLVM IR, so warn prominently so that anybody building against it who has problems will know why.

Also, LLVM itself has bumped their toolchain requirements with the release of 16. Technically, I'm not sure if that extends to projects using the LLVM APIs, but warn anyway and hope for the best. If it fails, at least the warning will supply an explanation.

At a later time, we'll need to upgrade the code on our side to have the option of using opaque pointers so we can run with LLVM 16+.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -12,7 +12,7 @@ Dependencies
 OSL requires the following dependencies or tools.
 NEW or CHANGED dependencies since the last major release are **bold**.
 
-* Build system: [CMake](https://cmake.org/) 3.12 or newer (tested through 3.25)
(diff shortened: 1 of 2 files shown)


oscarfv:
LLVM 16 still works with typed pointers, at least per the documentation [here](https://llvm.org/docs/OpaquePointers.html)

```
LLVM 16: Opaque pointers are enabled by default. Typed pointers are supported on a best-effort basis only and not tested.
```

The "not tested" part sounds scary, but in reality the LLVM test suite still is in transition from typed to opaque pointers, AFAIK.

PR #1659, opened 2023-03-21
Build: Fix cuda compilation flags for LLVM 15

lgritz:
For LLVM 15+, we need to disable opaque pointers. We did this for regular bitcode generation in llvm_macros.cmake, but neglected to make the corresponding change for compiling to ptx in cuda_macros.cmake.


lgritz:
We missed this spot in #1592

PR #1662, opened 2023-03-29
Switch lockgeom to interpolated and interactive

lgritz:
Establish two new metadata hints about shader parameters:

  - "interpolated" means that the parameter comes from the geometry somehow and may be bound with a different value at different points on the surfaces, or different values for different primitives that share the material. This corresponds to concepts known variously as "geometric primitive data/variables", "user data", or "vertex variables", depending on the renderer.

  - "interactive" means that the user desires the parameter to be allowed to change when the scene is interactively rerendered, without incurring a full recompile of the shader. (Variables not so marked might still be capable of being changed, but not necessarily without a re-JIT of the shaders.)

Both of these require that the parameter not be fully constant-folded or optimized away during runtime optimizeation or JIT, and thus supplant the old "lockgeom=0" hint. But their semantics may differ in other ways, so we thought it was worth introducing a separation, and at the same time, finally fixing the mis-design of "lockgeom" (which in retrospect is both a confusing name and also has the 0|1 sense backwards, thwarting even my own intuition every time I use it).

We introduce new variants of `ShadingSystem::Parameter()` that changes the optional `bool lockgeom` parameter for a new enum `ParamHints` that allows for future expansion with more hinted properties. For now, the old calls still work (lockgeom=false is the same as passing ParamHints::interpolated) and this does not introduce any break to backwards ABI compatibility at this time.

In this PR, "interpolated" is basically the replacement for the old lockgeom (though the sense is reversed). The "interactive" hint is put in place but doesn't do anything especially useful at the moment that is different from interpolated.  A few more things are coming down the pike in subsequent PRs: (1) Make interactive work and fully hook it up to ReParameter(), including on GPU. (2) Make a new global SS option that says if you're building shaders for an interactive render or an offline render. (If offline, the even "interactive" hinted params should be constant-folded.)

### DIFF
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -111,7 +111,8 @@ public:
 
 
 
-/// Parameter property hint bitfield values
+/// Parameter property hint bitflag values. The enum values must be powers of
+/// two so they can be combined with bitwise operators.
 enum class ParamHints : uint32_t {
     /// `none`: No special properties
     none = 0,
--- a/src/include/OSL/oslexec.h
+++ b/src/include/OSL/oslexec.h
@@ -111,6 +111,66 @@ public:
 
 
 
+/// Parameter property hint bitfield values
+enum class ParamHints : uint32_t {
+    /// `none`: No special properties
+    none = 0,
+    /// `interpolated`: This parameter may be an interpolated "user data" or
+    /// "geometric primitive" variable, and so may take on different values at
+    /// different points on the surface or on different objects that share the
+    /// same material. Any optimization of the shader should take this into
+    /// account.
+    interpolated = 1,
+    /// `interactive`: Parameter may have its value interactively modified by
+    /// subsequent ReParameter calls. Any optimization of the shader should
+    /// take this into account.
+    interactive = 2
+};
+
+inline constexpr ParamHints
+operator|(ParamHints a, ParamHints b)
+{
+    return static_cast<ParamHints>(static_cast<int>(a) | static_cast<int>(b));
+}
+
+inline constexpr ParamHints
+operator&(ParamHints a, ParamHints b)
+{
+    return static_cast<ParamHints>(static_cast<int>(a) & static_cast<int>(b));
+}
+
+inline ParamHints&
+operator|=(ParamHints& a, ParamHints b)
+{
+    a = a | b;
+    return a;
+}
+
+inline ParamHints&
+operator&=(ParamHints& a, ParamHints b)
+{
+    a = a & b;
+    return a;
+}
+
+inline constexpr ParamHints
+operator~(ParamHints a)
+{
+    return static_cast<ParamHints>(~static_cast<int>(a));
+}
+
+inline ParamHints&
+set(ParamHints& a, ParamHints b, bool on = true)
+{
+    if (on)
+        a |= b;
+    else
+        a &= ~b;
+    return a;
+}
+
+
+
 class OSLEXECPUBLIC ShadingSystem {
 public:
     ShadingSystem(RendererServices* renderer   = NULL,
@@ -173,7 +233,7 @@ public:
     ///    int lazyerror          Run layers lazily even if they have error
     ///                              ops after optimization (1).
     ///    int lazy_userdata      Retrieve userdata lazily (0).
-    ///    int userdata_isconnected  Should lockgeom=0 params (that may
+    ///    int userdata_isconnected  Should interpolated=1 params (that may
     ///                              receive userdata) return true from
     ///                              isconnected()? (0)
     ///    int greedyjit          Optimize and compile all shaders up front,
@@ -200,6 +260,7 @@ public:
     ///                              that don't specify it (1).  Lockgeom
     ///                              means a param CANNOT be overridden by
     ///                              interpolated geometric parameters.
+    ///                              This option is slated for deprecation.
     ///    int countlayerexecs    Add extra code to count total layers run.
     ///    int allow_shader_replacement Allow shader to be specified more than
     ///                              once, replacing former definition.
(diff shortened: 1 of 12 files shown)


src/include/osl_pvt.h
```diff
@@ -1029,10 +1056,12 @@ class Symbol {
...
     unsigned m_const_initializer : 1;  ///< initializer is a constant expression
-    unsigned m_connected_down : 1;   ///< Connected to a later/downstream layer
-    unsigned m_initialized : 1;      ///< If a param, has it been initialized?
-    unsigned m_lockgeom : 1;         ///< Is the param not overridden by geom?
-    unsigned m_allowconnect : 1;     ///< Is the param not overridden by geom?
+    unsigned m_connected_down : 1;  ///< Connected to a later/downstream layer
+    unsigned m_initialized : 1;     ///< If a param, has it been initialized?
+    unsigned m_interpolated : 1;    ///< Is the param overridden by geom?
+    unsigned m_interactive : 1;     ///< May the param change interactively?
+    unsigned m_noninteractive : 1;  ///< The param def won't modify interactively
```
aconty, line 1063: Not sure I understand why we need interactive and noninteractive flags.

lgritz: In discussion with Lee, we wanted leave room for the source code hints to be able to explicitly say `[[ int interactive = 0 ]]` to indicting being quite sure that this parameter is not one that lookdev artists will need to adjust interactively, whereas lacking any hint one way or the other has the possibility for a runtime control to allow everything not marked to be flipped to interactive.

It's still a half-baked idea that won't be fleshed out until we have a UI and understand more how it's used in practice, but I'm leaving as a placeholder so that the hint is really 3-state on, off, or not specified.

aconty (approved):

ee33:
Also curious, is there a case where you’d want both “interpolated” and “interactive” on the same attr?


From: Larry Gritz ***@***.***>
Date: Thursday, March 30, 2023 at 10:25 AM
To: AcademySoftwareFoundation/OpenShadingLanguage ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [AcademySoftwareFoundation/OpenShadingLanguage] Switch lockgeom to interpolated and interactive (PR #1662)

By the way, if people feel like "interpolated" is the wrong name for this concept, it can be debated. We can discuss at today's TSC meeeting.

ee33:
Never mind, easier to leave them separate bits as you have them.

From: Eric Enderton ***@***.***>
Date: Thursday, March 30, 2023 at 12:56 PM
To: AcademySoftwareFoundation/OpenShadingLanguage ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [AcademySoftwareFoundation/OpenShadingLanguage] Switch lockgeom to interpolated and interactive (PR #1662)
Also curious, is there a case where you’d want both “interpolated” and “interactive” on the same attr?


From: Larry Gritz ***@***.***>
Date: Thursday, March 30, 2023 at 10:25 AM
To: AcademySoftwareFoundation/OpenShadingLanguage ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [AcademySoftwareFoundation/OpenShadingLanguage] Switch lockgeom to interpolated and interactive (PR #1662)

By the way, if people feel like "interpolated" is the wrong name for this concept, it can be debated. We can discuss at today's TSC meeeting.

src/include/OSL/oslexec.h
```diff
@@ -111,6 +111,66 @@ struct SymLocationDesc {
...
+    /// `interpolated`: This parameter may be an interpolated "user data" or
+    /// "geometric primitive" variable, and so may take on different values at
+    /// different points on the surface or on different objects that share the
+    /// same material. Any optimization of the shader should take this into
+    /// account.
+    interpolated = 1,
+    /// `interactive`: Parameter may have its value interactively modified by
+    /// subsequent ReParameter calls. Any optimization of the shader should
+    /// take this into account.
+    interactive = 2
```
AlexMWells, line 127: Should these enum values be specified in hex so its more obvious they are bitflags vs. a vanilla enum?

lgritz:
Sorry for the delay in following up with this. At last week's TSC meeting, we discussed this and there was a lot of back-and-forth about the naming. To summarize from my notes: Some people liked "live" instead of "interactive." And also some people didn't like "interpolated", resulting in a variety of not-quite-better suggestions like generated, provided, queried, bound, overrideable, inconstant (ugh).

After agonizing over this for a while, none of these alternatives have really won me over as being any more clear than my original choices, so I'm inclined to claim "implementor's prerogative" and just merge this PR as-is and move on to get the next set of changes ready to show. My guess is that once it's in, it will seem fine ("what color should the bike shed be?" is a different dynamic from "I painted it green, let me know you think that's a problem after you've used it for a week"). There's nothing stopping us from changing it, up until we branch for a 1.13 release. Everything in main is understood to be subject to revision, and this will be a lot less intrusive than other changes we know are coming down the road.

src/include/OSL/oslexec.h
lgritz: Considering that the only values now are 1 and 2, I'm not sure it's any clearer in hex. :-)

But I take your point, and I will clarify in the comments about the class itself.

src/include/OSL/oslexec.h
```diff
@@ -111,6 +111,66 @@ struct SymLocationDesc {
 
 
 
+/// Parameter property hint bitfield values
```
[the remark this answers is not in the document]
lgritz: yes, good point

src/include/OSL/oslexec.h
fpsunflower: C++ has binary literals since C++14 which I think is our floor, right?

fpsunflower (approved):
LGTM

src/include/OSL/oslexec.h
lgritz: Yes, 14 is the floor. I've never used binary literals, but they make sense here. Another idiom I've used is

    pepperoni = (1 << 13),
    mushroom  = (1 << 14),
    pineapple = (1 << 15),
    ...

But when the only choices are 1 and 2, I'm not sure any of these language flourishes really improves the clarity. If and when there are enough bits to be confusing, we can choose a different notation.

PR #1665, opened 2023-04-06
Don't insert redundant run layer calls inside a basic block

chellmuth:
## Description

This patch is a very simple part one of work to optimize the number of conditional run layer calls that OSL makes, which take the form:
```
if (!groupdata->run[parentlayer])
    parent_layer([...]);
```

After examining our slowest JIT’ing shaders on gpu, @tgrant-nv showed us that these run layer calls were significantly bloating our ptx and slowing down their compilation. Many of them are unnecessary because static analysis can show we are guaranteed to have already run the layer in question.

For example, reading different fields from a connected struct like below will often generate multiple conditional run layer calls:
```
struct Pair {
    float x;
    float y;
};

shader example_shader(Pair p = {})
{
    if ([…]) {
        float xx = p.x;
        float yy = p.y;
        […]
    }
}
```

Since this is not really a performance issue on cpu (runtime overhead is just checking the conditional bit), OSL doesn’t make a huge effort to optimize away these redundant cases. Each useparam op will be careful not to insert more than one per layer, and code that is called unconditionally in the main section of the shader will also not have redundant checks, but there are plenty of examples like above that are not currently removed.

----

This patch takes a basic first step, adding metrics to track run layer call counts and removing duplicate run layer calls inside a basic block, covering cases like the example above.

On my set of production test scenes, this alone removes anywhere from 20-50% of total run layer calls, with our biggest shaders seeing the most improvement. Total ptx line reduction varies between 3-20%, and optix module compilation (jit time) is reduced between 1-15%.

## Tests

I don't know how to write a test that asserts an intermediate result of an optimization (fewer run layer calls) instead of asserting that an image looks correct.

Internally, I replaced omitted run layer calls with asserts that `groupdata->run[parentlayer] == true`, and ran against several test scenes totaling around a hundred or so shaders.

## Checklist:

### DIFF
--- a/src/liboslexec/llvm_gen.cpp
+++ b/src/liboslexec/llvm_gen.cpp
@@ -142,13 +142,14 @@ BackendLLVM::llvm_call_layer(int layer, bool unconditional)
 
     if (!unconditional)
         ll.op_branch(after_block);  // also moves insert point
+
+    shadingsys().m_stat_call_layers_inserted++;
 }
 
 
 
 void
-BackendLLVM::llvm_run_connected_layers(Symbol& sym, int symindex, int opnum,
-                                       std::set<int>* already_run)
+BackendLLVM::llvm_run_connected_layers(Symbol& sym, int symindex, int opnum)
 {
     if (sym.valuesource() != Symbol::ConnectedVal)
         return;  // Nothing to do
@@ -160,16 +161,6 @@ BackendLLVM::llvm_run_connected_layers(Symbol& sym, int symindex, int opnum,
         const Connection& con(inst()->connection(c));
         // If the connection gives a value to this param
         if (con.dst.param == symindex) {
-            // already_run is a set of layers run for this particular op.
-            // Just so we don't stupidly do several consecutive checks on
-            // whether we ran this same layer. It's JUST for this op.
-            if (already_run) {
-                if (already_run->count(con.srclayer))
-                    continue;  // already ran that one on this op
-                else
-                    already_run->insert(con.srclayer);  // mark it
-            }
-
             if (inmain) {
                 // There is an instance-wide m_layers_already_run that tries
                 // to remember which earlier layers have unconditionally
@@ -187,6 +178,17 @@ BackendLLVM::llvm_run_connected_layers(Symbol& sym, int symindex, int opnum,
                     m_layers_already_run.insert(con.srclayer);
                 }
             }
+
+            // m_call_layers_inserted tracks if we've already run this layer either:
+            //   * inside the current basic block (opnum >= 0)
+            //   * while writing output parameters (opnum == -1)
+            const auto key = std::make_pair(opnum >= 0 ? m_bblockids[opnum]
+                                                       : opnum,
+                                            con.srclayer);
+            if (m_call_layers_inserted.count(key)) {
+                continue;
+            }
+            m_call_layers_inserted.insert(key);
 
             // If the earlier layer it comes from has not yet been
             // executed, do so now.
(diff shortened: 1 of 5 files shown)


tgrant-nv:
Looks great. Is there any change to the output of the shader? Does this change the number of actual layer-to-layer calls when a shader is executed?

src/liboslexec/llvm_gen.cpp
```diff
@@ -160,16 +161,6 @@ BackendLLVM::llvm_run_connected_layers(Symbol& sym, int symindex, int opnum,
...
-            // already_run is a set of layers run for this particular op.
-            // Just so we don't stupidly do several consecutive checks on
-            // whether we ran this same layer. It's JUST for this op.
-            if (already_run) {
-                if (already_run->count(con.srclayer))
-                    continue;  // already ran that one on this op
-                else
-                    already_run->insert(con.srclayer);  // mark it
-            }
-
```
AlexMWells, line 172: Can you explain a bit on what the difference between the removed code block with the "local" already_run set and the m_layers_already run?  And how removing the block improves things?

chellmuth: Definitely!

The "local" `already_run` is restricted in scope to preventing redundant checks originating from a single useparam op. The new `m_call_layers_inserted` set will prevent all redundant checks in the scope of an entire basic block.

Consider this shader:
```
shader example_shader(ConnectionData my_input = {})
{
    if (u > 0.0) {
        float r = my_input.x * my_input.y;
        Ci = color(r, 0, 0) * emission();
    } else {
        float r = my_input.x * u;
        float g = my_input.y * v;
        Ci = color(r, g, 0) * emission();
    }
}
```

Its control flow graph looks like this:

plot-cfg

The existing `already_run` check would catch the "then" redundancy (block-3), but not the "else" redundancy (block-4). `m_call_layers_inserted` catches both- hence the removal of `already_run`.

`m_layers_already_run` is scoped to the entire main section of the instance, but only catches cases outside of conditionals. That's why all my examples are wrapped in  if-statements.

chellmuth:
> Is there any change to the output of the shader?

There shouldn't be any semantic difference. While testing, I changed all omitted layer checks to assertions that the layer had already been run, and didn't trigger any exceptions.

>Does this change the number of actual layer-to-layer calls when a shader is executed?
It shouldn't, and would be a bug if it did, but I didn't log that and verify.

src/liboslexec/llvm_gen.cpp
lgritz: This may be the most interesting reply to a review comment in this project's history. The visualization of the oso IR is extremely helpful.

Is the script or whatever you used to generate that graph something that makes sense to check into the project? I can imagine it being really helpful for a variety of debugging tasks, not to mention documenting internals of the shader optimization.

lgritz (commented):
LGTM. Nice improvement!

AlexMWells:
Could you add the same changes in the batched version: batched_llvm_gen.cpp and batched_llvm_instance.cpp?

src/liboslexec/oslexec_pvt.h
```diff
@@ -2606,6 +2608,8 @@ class OSOProcessorBase {
     std::vector<char> m_in_conditional;  ///< Whether each op is in a cond
     std::vector<char> m_in_loop;         ///< Whether each op is in a loop
     int m_first_return;                  ///< Op number of first return or exit
+    std::set<std::pair<int, int>>
+        m_call_layers_inserted;  ///< Lookup for (bblockid, layerid)
```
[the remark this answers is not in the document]
lgritz: This is a good suggestion

lgritz:
Alex is right. I was thinking that this was all oso manipulation, and it slipped my mind that because this happens as we're generating LLVM IR from he oso IR, it will need a corresponding change for batched shading, so I merged without remembering to make that request. It should be straightforward a straightforward change for you, Chris, since this doesn't in any way need any deep knowledge of how the SIMD code works.

src/liboslexec/oslexec_pvt.h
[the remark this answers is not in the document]
chellmuth: Cool, I'll improve this in a follow-up PR that also ports the changes to the batched code path.

PR #1668, opened 2023-04-11
build: Change cmake option USE_OPTIX -> OSL_USE_OPTIX

lgritz:
This avoids some possible confusion with other projects, or if this is built as a subproject.

Also, change from the C++ side having an OSL_USE_OPTIX defined -- only when building OSL itself -- to having it always defined in oslconfig.h and set to 0 or 1 depending on whether OptiX support was turned on at build time. This allows app-side use of OSL headers to know if it's dealing with an OptiX-capable OSL build.


lgritz:
Does this look ok to everyone? It seems fine with OSL internals, but I want to make sure people building externally are aware of (and ok with) the name change so it doesn't break your OSL builds if you need the OptiX support enabled.

PR #1672, opened 2023-04-20
Make recent run-layer optimizations optional, and fix init ops-related false positive

chellmuth:
## Description

My initial implementation of the run layer check optimizations didn’t account for the fact that init ops for symbols have their llvm code generated before we build basic blocks from the OSL IR.

This caused the code to be wrong in two ways: 1) the set of known run layer calls was being reset between instances too late in the pipeline, so a useparam inside an init op would actually be checking the previously compiled layer’s set, and 2) even when reset earlier, it assumed valid basic block ids existed for the init ops when they did not.

So, if a layer had a useparam inside one of its init ops that generally matched the same code location as a useparam in the previously compiled layer, the run layer check could be incorrectly omitted.

I’ve updated the code so that the default optimization behavior is reverted to the way it was before #1665, and added a new option `opt_useparam` that can be turned on to enable a fixed, restricted version of the optimization that only applies to code in the main section.

## Tests

I'm working on a test that properly triggers the issue.

## Checklist:

### DIFF
--- a/src/liboslexec/shadingsys.cpp
+++ b/src/liboslexec/shadingsys.cpp
@@ -1024,6 +1024,7 @@ ShadingSystemImpl::ShadingSystemImpl(RendererServices* renderer,
     , m_opt_middleman(true)
     , m_opt_texture_handle(true)
     , m_opt_seed_bblock_aliases(true)
+    , m_opt_useparam(false)
     ,
 #if OSL_USE_BATCHED
     m_opt_batched_analysis((renderer->batched(WidthOf<16>()) != nullptr)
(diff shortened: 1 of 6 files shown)


lgritz (approved):
LGTM, good catch, I also missed noticing that the init_ops might interact poorly with your prior change

AlexMWells:
Is the new `opt_useparam` just to allow a way to turn this off in case there is a bug?  Seems like, if its correct (the big question) you would always want it.

src/liboslexec/shadingsys.cpp
```diff
@@ -1024,6 +1024,7 @@ ShadingSystemImpl::ShadingSystemImpl(RendererServices* renderer,
     , m_opt_middleman(true)
     , m_opt_texture_handle(true)
     , m_opt_seed_bblock_aliases(true)
+    , m_opt_useparam(false)
```
AlexMWells, line 1027: Will then be enabled by default once confidence in its correctness is established?

PR #1675, opened 2023-04-28
Fix builds when USE_PYTHON=OFF

DeclanRussell:
## Description
This issue was caused by https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/commit/5efe15a79eb2d29a0f526f4c7e43c6f483b39d15. Essentially, we now use `Python_EXECUTABLE` to execute the `serialize-bc.py` script. Therefore, we have to run `find_package(python)` so that the `Python_EXECUTABLE` variable gets setup.

## Checklist:

### DIFF
--- a/src/cmake/pythonutils.cmake
+++ b/src/cmake/pythonutils.cmake
@@ -20,12 +20,9 @@ macro (find_python)
 
     # Attempt to find the desired version, but fall back to other
     # additional versions.
-    unset (_req)
-    if (USE_PYTHON)
-        set (_req REQUIRED)
-        if (PYTHON_VERSION)
-            list (APPEND _req EXACT)
-        endif ()
+    set (_req REQUIRED)
+    if (PYTHON_VERSION)
+        list (APPEND _req EXACT)
     endif ()
     checked_find_package (Python ${PYTHON_VERSION}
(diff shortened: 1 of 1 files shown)


lgritz:
Is there anybody for whom it might cause trouble for python to be a *required* dependency, as I believe this change effectively makes it?
I don't necessarily have any objections, but just making sure we are aware of that changes we are implicitly proposing.

DeclanRussell:
If there are people whom having python as a dependency is an issue, then they would already be hitting the problem before this change. Using the `Python_EXECUTABLE` variable currently means that you have to build with `USE_PYTHON=ON`, otherwise the build will fail. To remove that limitation, we would need to revert the `Python_EXECUTABLE` changes introduced by https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/commit/5efe15a79eb2d29a0f526f4c7e43c6f483b39d15 , but even then, the build would instead just fail trying to execute the python scripts instead if you don't have python installed on your machine.

lgritz:
No, I think that's reasonable. I guess we're saying that you need Python in order to build OSL or to run the testsuite, but not just to run it.

lgritz (approved):

PR #1677, opened 2023-05-04
refactor: Get rid of needless error check

lgritz:
This was apparently something I put in to track down a possible llvm-side problem many years ago. I haven't seen the error message in recent memory (and it was only enabled at SPI anway), so I guess it was a red herring or has long since been fixed. Get this cruft out of here.


lgritz:
Does this seem acceptable to everybody?

AlexMWells (approved):
LGTM

PR #1678, opened 2023-05-05
Expose API to add __optix_enabled__ prefix to shadeops that could call OptiX functions

DeclanRussell:
## Description

I'm creating this PR to start a discussion on how to best solve an issue that we hit while using OptiX 7 + OSL. With the OptiX 7 changes, the way we deal with renderer specific shadops is a little different from the OptiX 6 days. Essentially, renderer specific shadeops are now just marked extern, and then when we do the OptiX JIT compile, we link the definitions in from some other OptiX compiled module. The problem comes when these shadeop definitions call `optix*` device functions (e.g. `optixTrace`, `optixDirectCall`, etc...). If you didn't already know, if you have an `optix*` call in some function, the OptiX JIT compiler will inline the whole function call stack and remove all the definitions. This is a problem with OSL, because we need these definitions to be linked into our OSL program, not removed. A problematic example would be implementing `oslTrace`. Presumably, to implement this, you would need the `osl_trace` shadeop to call `optixTrace`. Currently, if you do this, the definition will be inlined and discarded by the OptiX JIT compiler and your OSL program that calls trace will just be left with a bunch of unresolved extern function calls.

The solution that we came up with, and is demonstrated by this PR, is to use the new OptiX 7.7 `__optix_enabled__` feature. Essentially, if you add the `__optix_enabled__` prefix to your function name, it will tell the OptiX JIT compiler that although this function contains `optix*` calls, we do not want it to be inlined. The modifications in the PR exposes an API (`register_optix_enabled_functions`) for the renderer to specify to the runtime which shadops will be calling `optix*` functions and will need this prefix added. There is then logic that will add this prefix to the function call. Unfortunately, `__optix_enabled__` functions can only be called by other `__optix_enabled__` functions. Therefore, there are some places, such as some matrix shadeops, that I have had to hardcode with the `__optix_enabled__` prefix. The downside to this, is that if a renderer is not calling `optix*` functions here, they may have some worse performance due to less inlining.

I am interested to know what you guys think of this. Its a fairly specific problem, but one I'm sure other renderers will hit if creating a full OSL implementation in an OptiX renderer.

## Checklist:

### DIFF
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -7,6 +7,7 @@
 #include <OSL/oslconfig.h>
 
 #include <unordered_set>
+#include <unordered_map>
 #include <vector>
 
 #ifdef LLVM_NAMESPACE
@@ -87,6 +88,9 @@ public:
     LLVM_Util(const PerThreadInfo& per_thread_info, int debuglevel = 0,
               int vector_width = 4);
     ~LLVM_Util();
+
+    // map of function names and prefixes that we would like prepended to these functions
+    std::unordered_map<ustring, ustring, ustringHash> m_function_prefixes;
 
     // JIT'd code needs to exist with a longer lifetime than the LLVM_Util object.
     // To enable better cleanup at shutdown, the lifetime of all JIT'd code is
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -2448,8 +2448,16 @@ LLVM_Util::make_function(const std::string& name, bool fastcall,
                          llvm::Type* rettype, cspan<llvm::Type*> params,
                          bool varargs)
 {
+    ustring function_prefix;
+    if (m_function_prefixes.size())
+    {
+        auto it = m_function_prefixes.find(ustring(name.c_str()));
+        if (it != m_function_prefixes.end())
+            function_prefix = it->second;
+    }
     llvm::FunctionType* functype = type_function(rettype, params, varargs);
-    auto maybe_func = module()->getOrInsertFunction(name, functype).getCallee();
+    auto maybe_func = module()->getOrInsertFunction(function_prefix.c_str() ? std::string(function_prefix.c_str()) + name : name,
+                                                    functype).getCallee();
     OSL_ASSERT(maybe_func && "getOrInsertFunction returned NULL");
     // if (!llvm::isa<llvm::Function>(maybe_func)) {
     //     print("make_function: getOrInsertFunction returned non-function for {}\n", name);
@@ -2457,8 +2465,9 @@ LLVM_Util::make_function(const std::string& name, bool fastcall,
     //         print("   param type: {}\n", llvm_typename(p));
     // }
     OSL_ASSERT_MSG(llvm::isa<llvm::Function>(maybe_func),
-                   "Declaration for %s is wrong, LLVM had to make a cast",
-                   name.c_str());
+                   "Declaration for %s%s is wrong, LLVM had to make a cast",
+                    function_prefix.c_str() ? function_prefix.c_str() : "",
+                    name);
     llvm::Function* func = llvm::cast<llvm::Function>(maybe_func);
 
     int vectorRegisterBitWidth = 8 * sizeof(float) * m_vector_width;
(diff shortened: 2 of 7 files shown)


src/include/OSL/llvm_util.h
```diff
@@ -88,6 +89,9 @@ class OSLEXECPUBLIC LLVM_Util {
               int vector_width = 4);
     ~LLVM_Util();
 
+    // map of function names and prefixes that we would like prepended to these functions
+    std::unordered_map<ustring, ustring, ustringHash> m_function_prefixes;
```
lgritz, line 93: Recent changes in the code base have made this declaration change, the ustringHash is no longer necessary.

```suggestion
    std::unordered_map<ustring, ustring> m_function_prefixes;
```

src/liboslexec/llvm_util.cpp
```diff
@@ -2448,17 +2448,26 @@ LLVM_Util::make_function(const std::string& name, bool fastcall,
                          llvm::Type* rettype, cspan<llvm::Type*> params,
                          bool varargs)
 {
+    ustring function_prefix;
+    if (m_function_prefixes.size())
```
AlexMWells, line 2452: Maybe add comment here as to "why" we are doing this whole prefix thing.  Fresh reader might be confused.

src/liboslexec/llvm_util.cpp
```diff
@@ -3637,12 +3646,18 @@ LLVM_Util::call_function(llvm::Value* func, cspan<llvm::Value*> args)
 llvm::Value*
 LLVM_Util::call_function(const char* name, cspan<llvm::Value*> args)
 {
-    llvm::Function* func = module()->getFunction(name);
+    ustring function_prefix;
```
AlexMWells, line 3649: Seeing this pattern multiple times, suggest add a helper build_prefixed_functionname(...) to better encapsulate it

AlexMWells: Perhaps "prefix" is a bit too specific, basically the target wants to translated function names, perhaps a more generic translation layer than just "prefix".

src/liboslexec/opmatrix.cpp
```diff
@@ -198,20 +198,24 @@ osl_get_inverse_matrix(void* sg_, void* r, const char* to)
 // Implemented by the renderer
```
AlexMWells, line 198: For opmatrix, would some of this optix specific stuff go away, if osl_get_matrix and osl_get_inverse_matrix were NOT implemented by the renderer and just use the bitcode based free function renderer services.  Seems like alot of this might be because of the call out to optix then wanting optix to call back in.  Now the problem in general may still exist, but not sure we need all of these functions to be callable from the renderer side.  In general there might be a set of osl functions meant to be called by free renderer services, and perhaps they should be explicitly defined someplace with a different prefix building scheme (compatible with Optix).

src/liboslexec/opmatrix.cpp
```diff
@@ -198,20 +198,24 @@ osl_get_inverse_matrix(void* sg_, void* r, const char* to)
...
-
+#ifdef __CUDACC__
+#define OPTIX_ENABLED_CALL(call) __optix_enabled__##call
+#else
+#define OPTIX_ENABLED_CALL(call) call
+#endif
 
 OSL_SHADEOP OSL_HOSTDEVICE int
-osl_prepend_matrix_from(void* sg_, void* r, const char* from)
+OPTIX_ENABLED_CALL(osl_prepend_matrix_from(void* sg_, void* r, const char* from))
```
AlexMWells, line 214: Suggest not having OPTIX mentioned directly in the macro and just have it on the function name.
Perhaps similar to what's in
https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/blob/dc6ce6c82b58e5638bf2798d9650445bbdac50f9/src/liboslexec/wide/define_opname_macros.h

Such that its clearly defined its meant to be called back into from render and from OSL JIT:
`__OSL_RS_OP(prepend_matrix_from)(void* sg_, void* r, const char* from)`

where 

```
#ifdef __CUDACC__
#define __OSL_RS_OP(call) __OSL_CONCAT3(__optix_enabled__,osl_,call)
#else
#define __OSL_RS_OP(call) __OSL_CONCAT2(osl_,call)
#endif
```

AlexMWells:
The more I look at this, I feel that a single llvm optimization pass could be used to just iterate over functions in module and change their name.  Maybe, if memory serves correct changing and existing function name might be more difficult than it sounds but is doable.  The point is perhaps we can not modify any of the code gen at all and just handle "decorating" (renaming) these optix callable functions directly in the generated module based on a mapping handed to llvm_util.  Would certainly avoid all these top level changes.

AlexMWells:
Another thought is not to change the names of osl_get_matrix at all and instead provide a set of OPTIX enabled shim functions that turn around and call the inlined regular osl_get_matrix calls.

DeclanRussell:
> The more I look at this, I feel that a single llvm optimization pass could be used to just iterate over functions in module and change their name. Maybe, if memory serves correct changing and existing function name might be more difficult than it sounds but is doable. The point is perhaps we can not modify any of the code gen at all and just handle "decorating" (renaming) these optix callable functions directly in the generated module based on a mapping handed to llvm_util. Would certainly avoid all these top level changes.

This sounds ideal, as these top level changes are certainly pretty nasty. I'm not sure I really have the expertise to implement such a thing. I could certainly take a look though if someone could help point me in the right direction.

Its worth noting that Tim's new PR https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1680 may also solve the the same issue, as it replaces the linking of shadops via OptiX JIT with instead compiling the shadops to LLVM IR and then giving that directly to the runtime to compile in directly.

aconty:
Yes, I would also vote for Tim's patch instead of the prefixes.

tgrant-nv:
Declan, have you had a chance to consider whether or not https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1680 covers the functionality you need? We'd definitely like to close any functional gaps in that PR.

If it's helpful to add the `__optix_enabled__` prefix to certain function names, that could be done before the `do_optmize()` call for each ShaderGroup. Alex's suggestion of using shim functions also sounds promising.

lgritz:
This is abandoned, right? Can we close?

DeclanRussell:
> This is abandoned, right? Can we close?

Apologies, I haven't had time to try Tim's changes from https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1680. I think we can close this, as it sounds like Tim's work solves the problem. I'll be sure to circle back around once I get to try them and notify if there are any issues.

PR #1679, opened 2023-05-06
cleanup: switch deprecated oiio simd names to modern

lgritz:
Oops, I noticed that OSL was still using deprecated OIIO::simd type aliases (float4, int4, bool4) instead of the modern vfloat4, vint4, vbool4.


lgritz:
Does this seem acceptable?

aconty (approved):

PR #1681, opened 2023-05-09
build(cmake): Fix incorrect variable name to control symbol visibility

lgritz:
As Fabrice Macagno pointed out on the oiio-dev mail list, we weren't setting the correct variable name CMAKE_CXX_VISIBILITY_PRESET.

And, oops, this also revealed that there were some things that needed to be declared public that were not (because, specifically, they needed to be called across library boundaries by the per-ISA dynamic libraries used by batch shading!).


AlexMWells (approved):
LGTM

PR #1683, opened 2023-05-12
Introduce OptiX direct callable API that owns groupdata buffer

chellmuth:
## Description

On gpu, the renderer currently owns a shader's groupdata params buffer so that it can pass its pointer to both the init and entry functions.

The downside of this approach is that in order to avoid dynamic memory allocation, the renderer must commit to a buffer size at build time. Using a conservative size to accommodate potentially large shaders can lead to gigabytes of unnecessary memory footprint.

This patch adds a new OptiX direct callable alternative to the existing init and entry callables. This new function allocates a perfectly-sized groupdata params buffer, so that the gpu only pays for the memory it needs, and then it calls init and entry.

Particularly large shaders can require larger buffers than there is space on the cuda stack. To handle this case, there is a new option, "max_optix_groupdata_alloc". Any shader requiring a groupdata buffer larger than this value will not allocate its own buffer, and instead use the pointer passed in by the renderer (presumably coming from a global memory pool).

## Tests

All testshade tests run with the existing api, and again with the new fused api.

## Checklist:

### DIFF
--- a/src/cmake/testing.cmake
+++ b/src/cmake/testing.cmake
@@ -160,6 +160,11 @@ macro ( TESTSUITE )
             # and optimized
             add_one_testsuite ("${_testname}.optix.opt" "${_testsrcdir}"
                                ENV TESTSHADE_OPT=2 TESTSHADE_OPTIX=1 )
+            if (NOT EXISTS "${_testsrcdir}/NOFUSED")
+              # and fused
+              add_one_testsuite ("${_testname}.optix.fused" "${_testsrcdir}"
+                                 ENV TESTSHADE_OPT=2 TESTSHADE_OPTIX=1 TESTSHADE_FUSED=1 )
+            endif()
         endif ()
 
         if (OSL_BUILD_BATCHED)
--- a/src/liboslexec/llvm_instance.cpp
+++ b/src/liboslexec/llvm_instance.cpp
@@ -1014,11 +1014,12 @@ BackendLLVM::build_llvm_optix_callables()
         ll.current_function(
             ll.make_function(dc_entry_name, false,
                              ll.type_void(),  // return type
-                             { llvm_type_sg_ptr(), llvm_type_groupdata_ptr(),
-                               ll.type_void_ptr(),  // userdata_base_ptr
-                               ll.type_void_ptr(),  // output_base_ptr
-                               ll.type_int(),
-                               ll.type_void_ptr(),  // interactive params
+                             {
+                                 llvm_type_sg_ptr(), llvm_type_groupdata_ptr(),
+                                 ll.type_void_ptr(),  // userdata_base_ptr
+                                 ll.type_void_ptr(),  // output_base_ptr
+                                 ll.type_int(),
+                                 ll.type_void_ptr(),  // interactive params
                              }));
 
         llvm::BasicBlock* entry_bb = ll.new_basic_block(dc_entry_name);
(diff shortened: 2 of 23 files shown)


src/cmake/testing.cmake
```diff
@@ -160,6 +160,11 @@ macro ( TESTSUITE )
             # and optimized
             add_one_testsuite ("${_testname}.optix.opt" "${_testsrcdir}"
                                ENV TESTSHADE_OPT=2 TESTSHADE_OPTIX=1 )
+            if (NOT EXISTS "${_testsrcdir}/NOFUSED")
```
aconty, line 163: All the patch LGTM! I'm wondering about this /NOFUSED dir, where and how you generate it?

chellmuth: I haven't added support for the new api to testrender yet, so the NOFUSED file is put in directories with optix tests that we want to skip testing with the new api. They're added by hand.

tgrant-nv:
This looks very cool, Chris. I'm still kicking the tires, but it seems like a great step forward.

src/liboslexec/oslexec_pvt.h
```diff
@@ -2502,6 +2503,37 @@ struct NoiseParams {
 };
 
 
+// Mangle the group and layer into a unique function name
+static std::string
```
tgrant-nv, line 2507: I get "unused function" errors for these three static functions, presumably because they're not called in all translation units that include oslexec_pvt.h.

tgrant-nv:
I see that the fused group function and the separate init/group functions all make it into the generated PTX. If you know in advance that you will or won't be using the fused function, it might be a good idea to drop the unused entry points. It would certainly make the PTX a lot smaller, and could save time in codegen and optimization.

src/liboslexec/oslexec_pvt.h
chellmuth: Thanks, should be fixed now.

chellmuth:
> I see that the fused group function and the separate init/group functions all make it into the generated PTX. If you know in advance that you will or won't be using the fused function, it might be a good idea to drop the unused entry points. It would certainly make the PTX a lot smaller, and could save time in codegen and optimization.

I'm happy to add an option that lets the user specify which callable(s) they're using, so we can remove the other. For what it's worth, in my tests the wrapper callables are just a couple dozen lines and keeping them all doesn't appear to affect codegen times.

chellmuth:
Latest push rebases on the recent reparameter work, and switches optix-testrender from the current api to the single-callable api.

src/liboslexec/llvm_instance.cpp
```diff
@@ -188,6 +188,35 @@ helper_function_lookup(const std::string& name)
 }
 
 
+std::string
+layer_function_name(const ShaderGroup& group, const ShaderInstance& inst,
+                    bool api)
+{
+    bool use_optix     = inst.shadingsys().use_optix();
+    std::string prefix = use_optix && api ? "__direct_callable__" : "";
```
lgritz, line 196: I feel like this could be `const char*` and not need a string allocation?

lgritz (changes requested):
LGTM.

I noted two places where you don't really need to construct a std::string, that's just a minor preference.

Could you please edit oslexec.h to add `max_optix_groupdata_alloc` to the documentation of attribute() that lists all of the attributes that one is allowed to set or query?


Is this ready to go, everybody ok with it?

lgritz:
Code looks fine, but can you please document the new attribute name you added, in oslexec.h where all the others are documented? Thanks.

chellmuth:
Ah yeah, you caught me pushing commits only partially addressing your feedback :) Latest commit adds the attribute documentation.

lgritz (approved):

PR #1687, opened 2023-05-24
Test interactive parameter initialization in optix testrender

chellmuth:
## Description

This change updates testrender's optix implementation of reparameter, and adds a new test.

Previously we were grabbing the pointers from incorrect SBT records, so this moves them into a global lookup table.

## Tests

I don't know how to test that parameter updates work correctly, but this at least validates that initial values are correct.

## Checklist:


aconty:
LGTM!

lgritz (approved):

lgritz:
I don't want you to bother changing this PR, but just noting that *technically*, although this is testing that the values you specify for parameters marked "interactive" make it to the shader, I believe it's never calling ReParameter and verifying that you can change the value after the shader has run, and then execute it again to see use of the new value. (That would be a somewhat bigger change to testrender, and probably not worth our time at the moment, since we do test it via testshade.)

PR #1700, opened 2023-06-19
Zero derivs for interactive params when needed

aconty:
Sometimes an interactive param is used in a way that gets flagged as needing derivs. But we were neither allocating space for that nor zeroing them. This resulted on memory corruption.

## Description

This fixes incorrect renders in a production context where params are often tagged as needing derivs.

## Checklist:

### DIFF
--- a/src/liboslexec/runtimeoptimize.cpp
+++ b/src/liboslexec/runtimeoptimize.cpp
@@ -3283,11 +3283,15 @@ RuntimeOptimizer::run()
                 // interactive parameter, including correct alignment.
                 size_t offset    = interactive_data.size();
                 size_t typesize  = s.typespec().simpletype().size();
+                size_t totalsize = typesize * (s.has_derivs() ? 3 : 1);
                 size_t alignment = typesize > 4 ? 8 : 4;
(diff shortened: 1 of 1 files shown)


lgritz (approved):

lgritz:
Hmmm, I think this will work, and I think it's fine to accept. But I wonder if a better fix long-term is that we should be turning derivatives off for a param that is marked as interactive. It won't have non-zero derivs, so any computations done using its derivative are pointless anyway.

aconty:
I thought about that, but I wasn't familiar with all the moving parts and didn't try.

lgritz:
I can take a look at that angle tomorrow when I'm back in the office.

If you need a workaround today for builds, it's fine to merge this, it doesn't hurt, I just think we can do better... if we can figure out the right spot to disable the derivs.

aconty:
We can wait for your take on it tomorrow

lgritz:
I'm going to merge this now so we can do an internal release, then I will come back to the problem afterwards and see if I can make it disable derivs entirely for these params that are marked interactive. (I don't want to hold up the fix waiting for me to get that right.)

PR #1708, opened 2023-08-03
admin: Add missing copyright notice to many files

lgritz:
Our latest license scan from LF reveailed that the usual copyright and license declaration was missing on many shader source files in testsuite. This was never intentional, it was simply overlooked for certain test files. Just add the standard notice to those files.


AlexMWells (approved):
LGTM

PR #1710, opened 2023-08-09
Move qualifying GroupData params onto stack

chellmuth:
## Description

On SPI Arnold, we are seeing GroupData as a bottleneck for GPU performance. Regardless of whether the buffer is backed by local or global memory, it's off-chip and slow to access.

This change moves qualifying parameters out of the GroupData struct and onto the stack. On GPU this makes those parameters more likely to be stored in registers, leading to better performance.

#### Qualifying parameter details

The qualifying parameters are all the output parameters minus three exceptions.

A typical output parameter's lifetime is scoped to the execution of the instance it is associated with. The parameter is initialized at the beginning of instance execution, accessed during the body of the instance, and after the instance body its final state is broadcast to all its downstream connections. This lifetime makes output parameters a good fit for stack storage.

The three exceptions:
* output parameters can be connected upstream and written to before instance execution
* output closures are initialized at the beginning of shader execution
* output parameters can be renderer outputs and read after instance execution

These are all identifiable at code generation time and can be flagged as disqualifying.

#### Improvements
On internal scenes, this change improves performance by about 15% on GPU. On CPU there aren't any performance regressions, but we don't see significant gains either. It also improves ptx generation and compilation by about 10%, and slightly lowers our memory footprint due to the GroupData size reduction.

## Tests

Exceptions 2 & 3 are covered by existing tests. A new test is added for exception 1, and for the case of successful optimization. This patch was also run through our internal shading testsuite.

## Checklist:

### DIFF
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -240,8 +240,9 @@ BackendLLVM::getLLVMSymbolBase(const Symbol& sym)
                              llvm_ptr_type(sym.typespec().elementtype()));
     }
 
-    if (sym.symtype() == SymTypeParam || sym.symtype() == SymTypeOutputParam) {
-        // Special case for params -- they live in the group data
+    if (sym.symtype() == SymTypeParam
+        || (sym.symtype() == SymTypeOutputParam && !is_stack_parameter(sym))) {
+        // Special case for most params -- they live in the group data
         int fieldnum = m_param_order_map[&sym];
         return groupdata_field_ptr(fieldnum,
                                    sym.typespec().elementtype().simpletype());
@@ -271,13 +272,26 @@ BackendLLVM::llvm_alloca(const TypeSpec& type, bool derivs,
 }
 
 
+bool
+BackendLLVM::is_stack_parameter(const Symbol& sym)
+{
+    if (!shadingsys().m_opt_groupdata)
+        return false;
+
+    // Some output parameters that are never needed before or
+    // after layer execution can be relocated from GroupData
+    // onto the stack.
+    return sym.symtype() == SymTypeOutputParam && !sym.renderer_output()
+           && !sym.typespec().is_closure_based()
+           && !sym.connected();
+}
 
 llvm::Value*
 BackendLLVM::getOrAllocateLLVMSymbol(const Symbol& sym)
 {
     OSL_DASSERT(
         (sym.symtype() == SymTypeLocal || sym.symtype() == SymTypeTemp
-         || sym.symtype() == SymTypeConst)
+         || sym.symtype() == SymTypeConst || is_stack_parameter(sym))
         && "getOrAllocateLLVMSymbol should only be for local, tmp, const");
     Symbol* dealiased                = sym.dealias();
     std::string mangled_name         = dealiased->mangled();
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -241,7 +241,8 @@ BackendLLVM::getLLVMSymbolBase(const Symbol& sym)
     }
 
     if (sym.symtype() == SymTypeParam
-        || (sym.symtype() == SymTypeOutputParam && !is_stack_parameter(sym))) {
+        || (sym.symtype() == SymTypeOutputParam
+            && !can_treat_param_as_local(sym))) {
         // Special case for most params -- they live in the group data
         int fieldnum = m_param_order_map[&sym];
         return groupdata_field_ptr(fieldnum,
@@ -273,7 +274,7 @@ BackendLLVM::llvm_alloca(const TypeSpec& type, bool derivs,
 
 
 bool
-BackendLLVM::is_stack_parameter(const Symbol& sym)
+BackendLLVM::can_treat_param_as_local(const Symbol& sym)
 {
     if (!shadingsys().m_opt_groupdata)
         return false;
@@ -290,7 +291,7 @@ BackendLLVM::getOrAllocateLLVMSymbol(const Symbol& sym)
 {
     OSL_DASSERT(
         (sym.symtype() == SymTypeLocal || sym.symtype() == SymTypeTemp
-         || sym.symtype() == SymTypeConst || is_stack_parameter(sym))
+         || sym.symtype() == SymTypeConst || can_treat_param_as_local(sym))
         && "getOrAllocateLLVMSymbol should only be for local, tmp, const");
     Symbol* dealiased                = sym.dealias();
     std::string mangled_name         = dealiased->mangled();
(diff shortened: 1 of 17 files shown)


aconty:
This looks very good to me!

src/liboslexec/backendllvm.cpp
```diff
@@ -267,13 +268,25 @@ BackendLLVM::llvm_alloca(const TypeSpec& type, bool derivs,
...
+BackendLLVM::is_stack_parameter(const Symbol& sym)
+{
+    if (!shadingsys().m_opt_groupdata)
+        return false;
+
+    // Some output parameters that are never needed before or
+    // after layer execution can be relocated from GroupData
+    // onto the stack.
+    return sym.symtype() == SymTypeOutputParam && !sym.renderer_output()
+           && !sym.typespec().is_closure_based() && !sym.connected();
```
aconty, line 281: Question for my understanding: You don't allow any type of connection? I thought downstream was allowed.

chellmuth: Thanks for taking a look! Symbol::connected() is upstream only, there's a separate method connected_down() for downstream connections.

aconty: I see, thanks!

AlexMWells:
@chellmuth I haven't had time to review this deeply, but had a higher level question.  Are there advantages to this PR's approach vs. code generate a shim that instantiates a GroupData on the stack, then passes it into the existing shader function?

src/liboslexec/backendllvm.cpp
AlexMWells: Does is_stack_parameter take into account "lockgeom()==0" or "interpolated" or reparameter into account?  Does it need to (maybe not)?

src/liboslexec/backendllvm.cpp
```diff
@@ -267,13 +268,25 @@ BackendLLVM::llvm_alloca(const TypeSpec& type, bool derivs,
 }
 
 
+bool
+BackendLLVM::is_stack_parameter(const Symbol& sym)
```
AlexMWells, line 272: I am wondering if it ending up on the stack is the results, and the query here is more like "param_could_be_local" or something indicating that the parameter to could be pro(de)moted from a parameter to a local variable because it has no upwards connections.

AlexMWells:
As this seems useful change in general, would be good to update batched version to make the same type of change (actually should looks pretty much identical)

src/liboslexec/backendllvm.cpp
chellmuth: Sure, makes sense to me.

chellmuth:
> As this seems useful change in general, would be good to update batched version to make the same type of change (actually should looks pretty much identical)

Yep, happy to port this over.

src/liboslexec/backendllvm.cpp
AlexMWells:
```
bool
BackendLLVM::can_treat_param_as_local(const Symbol& sym)
```

src/liboslexec/backendllvm.cpp
chellmuth: That's a good question about input parameters. I took a look at how OSL is currently optimizing 4 cases of unconnected parameters:

```
float u_func() { return u; }

shader opt_candidates(
    float w = 0.,
    float x = u,
    float y = u_func(),
    float z = 0. [[ int lockgeom = 0 ]]
)
{...}
```
Currently two of these cases are optimized out of the GroupData:
* `w`: constant-folded 
* `x`: I'm not sure if constant-folding is still the correct term here, but x is optimized out

The other two aren't:
* `y`: I hoped y would optimize like x, but it doesn't
* `z`: triggers a user-data lookup, so not constant-folded

I updated my patch to catch `z` (I believe unconnected + interpolated is safe for stack placement). It passed the unit tests but led to some image differences internally. I'll stick to output parameters for now, hopefully I can figure out that bug and follow up with input parameters.

src/liboslexec/backendllvm.cpp
chellmuth: This is pushed.

chellmuth:
Updated with batched implementation. I'm relying on the Github CI for testing.

src/liboslexec/backendllvm.cpp
AlexMWells: You might also want to examine behavior with `-O0` or `setenv TESTSHADE_OPT 0` that might cause that x to be left around as well as an input.

lgritz:
Chris, you can build and test batched mode at work. Ask me how tomorrow.

src/liboslexec/batched_backendllvm.cpp
```diff
@@ -458,8 +458,10 @@ BatchedBackendLLVM::getLLVMSymbolBase(const Symbol& sym)
                              llvm_ptr_type(sym.typespec().elementtype()));
     }
 
-    if (sym.symtype() == SymTypeParam || sym.symtype() == SymTypeOutputParam) {
-        // Special case for params -- they live in the group data
+    if (sym.symtype() == SymTypeParam
+        || sym.symtype() == SymTypeOutputParam
```
AlexMWells, line 462: Per build warning/error add parenthesis
```
    if ((sym.symtype() == SymTypeParam
        || sym.symtype() == SymTypeOutputParam)
               && !can_treat_param_as_local(sym)) {
```

src/testshade/testshade.cpp
```diff
@@ -2254,7 +2254,7 @@ test_shade(int argc, const char* argv[])
             std::cout << texturesys->getstats(5) << "\n";
         std::cout << ustring::getstats() << "\n";
     }
-    if (debug1 || print_groupdata) {
+    if ((debug1 || print_groupdata) && !batched) {
         int groupdata_size;
```
AlexMWells, line 2258: Does this mean that llvm_groupdata_size was not being set properly on the batched side?  Is that fixable (although might be outside the scope of your PR?.  If so can we leave a TODO comment in the code to fix it?

chellmuth: Step one I just wanted to pass these CI tests :)

I can remove that if-clause, and then when batched == true I can output llvm_groupdata_wide_size. But am I correct in thinking that number will vary with architecture? I will need a ref/out.txt for each possible output if so.

AlexMWells: @chellmuth correct on both counts.  Typically 8wide and 16wide will have the same sizes/outputs.  But perhaps for AVX512 some integers only used as booleans might only take 2 bytes because they use native bit masks.

chellmuth: I think I'll follow whatever you recommend then. I can enable the groupdata size output, run the CI tests, and add new ref.txt's from the different architectures' unique failures. Or I can omit the line. Happy either way :)

AlexMWells: To make life easier maybe just the TODO, but could the comparison can just verify the "optimized" group data is smaller than non-opt vs. specific sizes.

chellmuth: Sounds good. Unfortunately with how the testsuite is setup, I don't know any precedent for that kind of assert.

AlexMWells: ok, lets just do the TODO and move on.  
Thanks for your work on this.

AlexMWells (approved):
LGTM

lgritz (commented):
LGTM

PR #1711, opened 2023-08-18
Restore stubs for printf (fixes OptiX problems with 1702)

lgritz:
The recently merged PR 1702 fails all the OptiX tests. I tracked down the problem and it's that the patch was a bit too aggressive about removing things no longer needed. The intent was that even if you don't use the new journaling, nothing should break. But it does break for OptiX, and this patch restores some of the functions 1702 removed (like the implementation of osl_printf) and some other necessary stubs.

This takes us from failing all ~150 OptiX oriented tests to failing only 4, which I believe are unrelated to 1702 and are merely revealed by them. I will fix those in a separate PR.


AlexMWells (approved):
LGTM

PR #1713, opened 2023-08-18
Draft: Remove more unnecessary conditional run layer calls

chellmuth:
This PR is a work-in-progress that I've hit a (hopefully) minor stumbling block on.

#### Overview
@tgrant-nv was looking at some of our shader ptx awhile ago and pointed out that a surprisingly high amount of the total instructions were just conditional layer call checks (due to all the stores and loads surrounding calls). The concern was that this was slowing down OptiX module compilation.

OSL has some basic checks in place to remove unnecessary conditional layer calls, but this patch takes things further. The high level strategy is:

* Mark all our conditional layer calls during llvm generation
* Run a custom llvm optimization pass that associates calls with their basic blocks
* For each call, walk up llvm's dominator tree to find basic blocks guaranteed to have already run
* If a guaranteed basic block already contains the same call as the current one, remove the current one

It doesn't help the compilation times for every shader, but we've seen improvements as speedups as 5x on some of our most expensive shaders.

#### Work in progress
The problem I've run into is that the optimization pass introduces quite a bit of overhead to the ptx generation. Most assets take 10-20% longer to generate ptx with this patch. For us, it's probably still a good trade to turn on for GPU, but I was hoping some more eyes could help improve things. Especially since I've never written a LLVM pass before, maybe I'm doing things particularly inefficiently.

Any feedback or ideas would be appreciated!

## Checklist:

### DIFF
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -79,6 +79,7 @@
 #include <llvm/Transforms/Scalar/GVN.h>
 #include <llvm/Transforms/Utils.h>
 #include <llvm/Transforms/Utils/UnifyFunctionExitNodes.h>
+#include <llvm/Transforms/Utils/BasicBlockUtils.h>
 
 #include <llvm/Support/DynamicLibrary.h>
 
@@ -91,6 +92,7 @@
 #include <llvm/Analysis/TargetTransformInfo.h>
 #include <llvm/Transforms/Utils/Cloning.h>
 #include <llvm/Transforms/Utils/SymbolRewriter.h>
+
 
 OSL_NAMESPACE_ENTER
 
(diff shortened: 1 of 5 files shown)


src/liboslexec/llvm_util.cpp
```diff
@@ -1751,7 +1753,98 @@ LLVM_Util::InstallLazyFunctionCreator(void* (*P)(const std::string&))
...
+        // Delete all the unnecessary stubs identified above
+
+        // WIP: This appears to be the cause of the performance regression
+        // (Not the dominator tree analysis above)
+        llvm::Value* fake = llvm::ConstantInt::get(F.getContext(), llvm::APInt(32, 1));
+        for (llvm::CallInst* inst : delete_queue) {
+            llvm::BasicBlock::iterator iterator(inst);
+            llvm::ReplaceInstWithValue(inst->getParent()->getInstList(), iterator, fake);
+            m_calls_removed++;
+        }
```
chellmuth, line 1835: This appears to be where the performance hit originates from. It seems like the replacement and subsequent code-folding is pretty simple, so I'm confused why it is taking so much time compared to everything else that happens while codegen'ing a shader (+10-20%).

tgrant-nv:
This sounds very promising. A 10-20% increase in JIT time isn't _so_ bad, if it increases the shader performance or even reduces the PTX compile time. We'll take a closer look and see what recommendations we can make.

src/liboslexec/llvm_util.cpp
```diff
@@ -1767,6 +1860,8 @@ LLVM_Util::setup_optimization_passes(int optlevel, bool target_host)
 
     m_llvm_module_passes           = new llvm::legacy::PassManager;
     llvm::legacy::PassManager& mpm = (*m_llvm_module_passes);
+    // TODO: Add based on optlevel
+    mpm.add(new CheckLayerRemovalPass());
```
tgrant-nv, line 1864: Since the new pass is being run on the entire module, I think that means that it's running on all of the shadeops functions. Would it be possible to detect that you are processing a library function (by function name, or perhaps a function attribute), and early-exit from `CheckLayerRemovalPass::runOnFunction()` in that case?

That might not buy much since module pruning should remove unused library functions, but maybe it will help in some cases.

chellmuth: That seems like a responsible safe-guard to put in, I'll add that.

chellmuth:
> This sounds very promising. A 10-20% increase in JIT time isn't _so_ bad, if it increases the shader performance or even reduces the PTX compile time. We'll take a closer look and see what recommendations we can make.

Thanks! Looking forward to your findings.

chellmuth:
> Is this ready to be a non-draft?

It's close. It turned out the speed hit we've been seeing is because removing so many run-layer checks made the remaining ones more palatable for inlining. So LLVM's inlining pass times are growing quite a bit.

Now that it's been merged, I want to test this patch internally on #1680, since that also has inlining ramifications and I want to make sure everything interacts okay.

PR #1715, opened 2023-08-21
fix: Recent journaling changes break with some fmtlib versions

lgritz:
Certain fmt library versions don't automatically know how to format atomics. Explicitly load them to turn into regular ints to avoid new build errors introduced by the recent journaling changes.


AlexMWells (approved):
LGTM

PR #1717, opened 2023-08-22
PR/BatchedFixForce_LLVM_Boolean

### LINKED ISSUE #1716: Crash assigning boolean expression to `int` in batched mode
### Problem

OSL crashes when running expressions of the form `int = float == float` in batched mode, but not in non-batched mode. This was originally reported as a Gaffer bug in https://github.com/GafferHQ/gaffer/issues/5430, but is reproducible using `testshade` alone (see below). The original bug report suggests the crash may occur only on certain CPUs. We've seen crashes on `Intel(R) Xeon(R) W-2145` and `Intel(R) Xeon(R) Silver 4216 CPU`.

### Steps to Reproduce

1. Compile `test.osl` (see below) with `oslc test.osl`
2. Run `testshade --batched --res 128 128 -o result testout.txt test`
3. Observe stack trace (see below)

<details><summary>test.osl</summary>

```
shader test(
        float a1 = u,
        float a2 = v,
        output int result = 0
)
{
        result = a1 == a2;
}
```

</details>

<details><summary>Stack trace</summary>

 ```
new_val type=<16 x i1> dest_ptr type=<16 x i32>*
/home/john/dev/gafferDependencies/OpenShadingLanguage/working/OpenShadingLanguage-1.12.9.0/src/liboslexec/batched_backendllvm.cpp:1234: llvm_store_value: Assertion 'll.type_ptr(ll.llvm_typeof(new_val)) == ll.llvm_typeof(dst_ptr)' failed.
/home/john/dev/gafferDependencies/OpenShadingLanguage/working/OpenShadingLanguage-1.12.9.0/src/liboslexec/llvm_util.cpp:2958: native_to_llvm_mask: Assertion 'native_mask->getType() == type_native_mask()' failed.
Invalid operands for select instruction!
  %31 = select <16 x i1> %10, <16 x i1> %26, <16 x i32> %30, !dbg !12
Stored value type does not match pointer operand type!
  store <16 x i1> %31, <16 x i32>* %29, align 16, !dbg !12
 <16 x i32>LLVM ERROR: Broken module found, compilation aborted!
 0# OpenImageIO_v2_4::Sysutil::stacktrace() in /home/john/dev/build/gaffer-1.3/lib/libOpenImageIO_Util.so.2.4
 1# 0x00007F9C23E6A79B in /home/john/dev/build/gaffer-1.3/lib/libOpenImageIO_Util.so.2.4
 2# 0x00007F9C1BA2E400 in /lib64/libc.so.6
 3# gsignal in /lib64/libc.so.6
 4# abort in /lib64/libc.so.6
 5# 0x00007F9C216680E5 in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
 6# 0x00007F9C21668208 in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
 7# 0x00007F9C2130F54F in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
 8# 0x00007F9C212C1E97 in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
 9# 0x00007F9C2141D26B in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
10# OSL_v1_12::pvt::LLVM_Util::prune_and_internalize_module(std::unordered_set<llvm::Function*, std::hash<llvm::Function*>, std::equal_to<llvm::Function*>, std::allocator<llvm::Function*> >, OSL_v1_12::pvt::LLVM_Util::Linkage, std::string*) in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
11# OSL_v1_12::pvt::BatchedBackendLLVM::run() in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
12# OSL_v1_12::pvt::ShadingSystemImpl::Batched<16>::jit_group(OSL_v1_12::ShaderGroup&, OSL_v1_12::ShadingContext*) in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
13# OSL_v1_12::ShadingContext::Batched<16>::execute_init(OSL_v1_12::ShaderGroup&, int, OSL_v1_12::Wide<int const, 16>, OSL_v1_12::BatchedShaderGlobals<16>&, void*, void*, bool) in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
14# OSL_v1_12::ShadingContext::Batched<16>::execute(OSL_v1_12::ShaderGroup&, int, OSL_v1_12::Wide<int const, 16>, OSL_v1_12::BatchedShaderGlobals<16>&, void*, void*, bool) in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
15# OSL_v1_12::ShadingSystem::BatchedExecutor<16>::execute(OSL_v1_12::ShadingContext&, OSL_v1_12::ShaderGroup&, int, OSL_v1_12::Wide<int const, 16>, OSL_v1_12::BatchedShaderGlobals<16>&, void*, void*, bool) in /home/john/dev/build/gaffer-1.3/lib/liboslexec.so.1.12
```

</details>

### Versions

* OSL branch/version: 1.12.9
* OS: CentOS 7
* C++ compiler: GCC 9.3.1
* LLVM version: 11.1.0
* OIIO version: 2.4.11

### PR DESCRIPTION
**Pull request by AlexMWells**
## Description

Fixes #1716:  Crash assigning boolean expression to int in batched…mode

Modifying BatchedAnalysis::establish_symbols_forced_llvm_bool to excluded renderer outputs to fix the issue. 

## Tests

Added new test compassign-bool to recreate the issue and verify it is fixed.

## Checklist:

### CODE REVIEW

**Review by sfriedmapixar — APPROVED** [review]
LGTM

**Comment by lgritz** [comment]
Is this ready to merge?

**Comment by AlexMWells** [comment]
No, I will be updating it to handle the follow on issue reported in the issue 1716

**Comment by AlexMWells** [comment]
Just committed the comprehensive fix(es): 

Rearchitected BatchedAnalysis::establish_symbols_forced_llvm_bool to track a BoolStatus per Symbol evaluating all operations that write to the symbol. Only if all ops that write to it always produce a logically boolean result will the BoolStatus be set to Yes.
There is another set of ops that could produce a logically boolean result if their inputs were also boolean.  In that case the BoolStatus is set to Maybe and its input symbols tracked (as they may not have a known BoolStatus yet) and deferred for a follow on pass.
After the initial pass is made, if there were any deferred Symbols, we can check their dependencies to see if they have a known BoolStatus and we can set the Symbol's BoolStatus to Yes or No depending on those input dependencies.  If the BoolStatus of any dependencies is still Unknown, we just defer it again.  We continue making passes over deferred Symbols until we have established a known BoolStatus for them all.  Afterwards we use forced_llvm_bool(true) on the Symbols where BoolStatus is Yes.
    Improved support for storing forced_llvm_bool Symbol values by automatically converting from int->bool and then a vector of bools (mask) to the platforms native mask type (might be vector of integers on platforms without native mask support).
    Fixed bug in BatchedAnalysis where arrays of closures were not being marked as varying, only closures where.
    Allow GroupData to hold actual boolean or native mask data type vs. integers when a Symbol is forced_llvm_bool.
    Improved batched GroupData layout calculation to use llvm offset and alignment calculations based on the actual DataLayout (might consider upgrading the scalar code gen to do the same).
    Fixed bug where results of partial results from RendererServices::get_userdata allowed initops to execute unmasked effectively overwriting the results from RendererServices::get_userdata.  Issue stemmed from op_useparam not being modelled properly by BatchedAnalysis.

Also for scalar and batched enabled ability to have interpolated output parameters to match oslc behavior of allowing it.

    Added new test compassign-bool to recreate the issue and verify it is fixed.
    Added new test closure-layered to recreate the issue and verify it is fixed.
    Added new test userdata-partial recreate the issue and verify it is fixed.

**Comment by lgritz** [comment]
Is this still in progress?

Alex, why did a bunch of methods get marked as NOINLINE here?

**Comment by johnhaddon** [comment]
I tested this PR in Gaffer today, and can confirm that it fixes the original problem involving the boolean assignment. But it also appears to have introduced two new test failures for us. In both cases we're reading an integer via `getattribute()` and then doing something based on its value :

```
int shadingIndex;
if( getattribute( "shading:index", shadingIndex ) )
{
    printf( "SHADING INDEX %d", shadingIndex );
    if (shadingIndex % 2 == 0)
    {
         ...
```

This is the meat of the `BatchedRendererServices::get_userdata()` call that invokes :

```
if( wval.type() == OIIO::TypeDesc( OIIO::TypeDesc::INT32 ) )
{
    maskedDataInitWithZeroDerivs( wval );
    wval.mask().foreach ([&wval, pointIndex](ActiveLane lane) -> void {
        int i = pointIndex + lane.value();
        wval.assign_val_lane_from_scalar( lane, &i );
    });
}
```

Can you see anything wrong with that? If I instrument the C++ then it appears to be running for all the lanes, and assigning the value I expect, but that's not coming through on the OSL side.

What seems to work around the problem is to make sure `shadingIndex` is initialised before calling `getattribute()` :

```
int shadingIndex = -1;
if( getattribute( ...
```

Then all of a sudden the tests pass. I don't suppose that gives you any clues as to something we might be doing wrong, or as to something that might have been inadvertently broken in this PR?

**Comment by AlexMWells** [comment]
> Is this still in progress?
> 
> Alex, why did a bunch of methods get marked as NOINLINE here?

Debugging, when inlined its quite difficult to follow through.  And for most of these not sure they would benefit from inlining, especially the ones recursive in nature.  They can be removed later if a slow down in analysis is observed.  But really something with a larger working set of complex shaders to measure the difference between inline and NOINLINE here.

**Comment by AlexMWells** [comment]
@johnhaddon thanks for testing.

For the broken one, what output are you seeing from your
 printf( "SHADING INDEX %d", shadingIndex );
 vs. what you see when you set 
 int shadingIndex = -1;
 ?
Are they all 0 or 1's or garbage?
 And you can set attribute in the ShadyingSystem "dump_forced_llvm_bool_symbols" to 1, and run the broken and workaround version and report back on which symbols.
 Perhaps getattribute is not participating is in the forced bool analysis correctly.

**Comment by AlexMWells** [comment]
@johnhaddon , I think I was able to reproduce the issue, the getattribute doesn't appear to disqualify the local variable from being forced_llvm_bool.  
```
testshade --options opt_batched_analysis=1,dump_forced_llvm_bool_symbols=1 -t 1 test
shadingIndex is forced llvm bool.
$tmp1 is forced llvm bool.
$tmp3 is forced llvm bool.
```

But when you initialized shadingIndex to -1, it could tell the initial value couldn't be bool.
```
testshade --options opt_batched_analysis=1,dump_forced_llvm_bool_symbols=1 -t 1 test
$tmp1 is forced llvm bool.
$tmp3 is forced llvm bool.
```

I'll see if I can get getattribute to play nicely here.

johnhaddon:
> I think I was able to reproduce the issue, the getattribute doesn't appear to disqualify the local variable from being forced_llvm_bool.

Yep, I think that's it. I see the same output in my original test case here, if I run it with `dump_forced_llvm_bool_symbols=1`. And for what it's worth, in the broken case, `shadingIndex` was always `0` for all lanes.

AlexMWells:
@johnhaddon the pr has been updated with a fix for getattribute, whose return value is always bool, but it was treating its result parameter as always being bool as well.  Fix was to restrict consideration to write's from return values, vs any argument.

Let us know if it works, otherwise I think the PR is good to go.

johnhaddon:
Thanks @AlexMWells - I'm happy to confirm that the whole GafferOSL test suite is running successfully with this PR. Many thanks once again!

PR #1718, opened 2023-08-23
fix: Handle mixed string casting that plagues optix codegen

lgritz:
The string situation is still a bit tricky on the OptiX side.  There's a full overhaul of that coming separately as we continue to shift to an all-ustringhash world.

But in the mean time, we were getting some testsuite failures that were because of a mismatch between times we think of a gpu-side ustring as a char* and when it's a uint64. IN FACT, they are always the hashes, but we backed ourselves into a corner where sometimes that 64 bit pattern is passed around as a pointer, ugh.

This small patch attempts to catch this case and just do the appropriate cast to compensate. It's a band-aid; a more comprehensive fix is coming, as I said.

This gets us down to just ONE failing optix test case, which I will tackle separately.

The change necessitated adding a new `ptr_to_int64_cast()` method to LLVM_Util.

### DIFF
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -5224,7 +5224,7 @@ LLVM_Util::op_store(llvm::Value* val, llvm::Value* ptr)
         std::cerr << "op_store val->getType()=" << std::flush;
         val->getType()->print(llvm::errs());
         std::cerr << std::endl;
-        // OSL_ASSERT(0);
+        OSL_DASSERT(0 && "We should not have a pointer type mismatch here");
     }
     if (m_mask_stack.empty() || val->getType()->isVectorTy() == false
         || (!is_masking_required())) {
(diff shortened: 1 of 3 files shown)


src/liboslexec/llvm_util.cpp
```diff
@@ -5216,6 +5224,7 @@ LLVM_Util::op_store(llvm::Value* val, llvm::Value* ptr)
         std::cerr << "op_store val->getType()=" << std::flush;
         val->getType()->print(llvm::errs());
         std::cerr << std::endl;
+        // OSL_ASSERT(0);
```
AlexMWells, line 5227: Remove commented out line(?)

lgritz: Wait, maybe I'll change it to DASSERT.

AlexMWells: I think my point was that it was currently commented out, so doing nothing.
Having it OSL_ASSERT or OSL_ASSERT is useful when in debugger because you get the builtin_trap and break right there.

lgritz: Updated to not-commented-out OSL_DASSERT, so it will hit the assertion in debug mode, but not for production builds.

PR #1719, opened 2023-08-23
ci: Fix broken CI from new OIIO that needs newer cmake and openexr

lgritz:
This test was pulling OIIO master, which had recently been upgraded to require newer cmake and openexr versions than were provided in the 2019 era container we were using for the GPU test. Upgrade the test to modern 2023 standards.


lgritz:
Look ok to everyone? This fixes a minor CI break.

PR #1721, opened 2023-08-25
Add missing members to ShaderGlobals in rend_lib.h

tgrant-nv:
## Description

There is a separate definition of `ShaderGlobals` in `testrender/cuda/rend_lib.h` that has drifted away from the definition in `include/OSL/shaderglobals.h`. Two new members were added in #1702 which changed the byte offset of the output closure `Ci`, which leads to incorrect shader output since the offset is baked into the PTX.

Adding the new members fixes the shading issue. But this should be regarded as a short-term fix, since the two `ShaderGlobals` definitions have diverged in other ways that might be significant.

## Tests

All OptiX tests passing with the exception of `testoptix.optix.opt`, which is a preexisting failure.

## Checklist:

### DIFF
--- a/src/testrender/cuda/rend_lib.h
+++ b/src/testrender/cuda/rend_lib.h
@@ -99,6 +99,8 @@ struct ShaderGlobals {
     float3 Ps, dPsdx, dPsdy;
     void* renderstate;
     void* shadingStateUniform;
+    int thread_index;
+    int shade_index;
     void* tracedata;
     void* objdata;
     void* context;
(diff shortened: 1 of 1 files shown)


lgritz:
should we have a static_assert somewhere to ensure that these structs don't drift apart inadvertently again?

tgrant-nv:
There are some other differences that I'm not quite sure how to sort out. Such as `shaderID` being tacked on to the end of the CUDA version. The order of some members is different as well. It would definitely be nice to synchronize them and keep them in sync, though.

tgrant-nv:
I'll reorder the CUDA version to match the "real" version.

tgrant-nv:
It was just `shadingStateUniform` that was a bit out of place. I've rearranged the members so the layout matches `shaderglobals.h`, with the exception of `shaderID`. `shaderID` could probably be tucked into one of the existing pointees, but that would be a bit of a refactor.

lgritz (approved):
LGTM, good catch!

PR #1722, opened 2023-08-27
build(deps): Raise minimum OpenImageIO version to 2.4

lgritz:
This is just for future-looking OSL 1.13 releases, it won't be backported to the 1.12 release branch.

Remember that by the time we release OSL 1.13 as a supported release family, OIIO will also have done its annual release, of 2.5. So our minimum of OIIO 2.4 means that we will be supporting the current release, as well as the prior year's release (and, of course, we always stay compatible with OIIO's trunk as it evolves). So we're really keeping compatibility with 3 major releases, which seems like plenty to juggle.

We discussed doing this a few TSC meetings ago, and nobody raised any objections to our raising the OIIO floor to 2.4 for the upcoming OSL 1.13 release. But if anyone is bothered by this, please speak up on this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
             container: aswftesting/ci-osl:2019-clang9
             vfxyear: 2019
             cxx_std: 14
-            openimageio_ver: v2.3.17.0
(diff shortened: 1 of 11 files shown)


lgritz:
This seem ok to everybody?

PR #1723, opened 2023-09-02
The recently added SS::find_symloc can have const args

lgritz:


lgritz:
This seem ok?

sfriedmapixar (approved):
I'm not thinking of any obvious ways finding a symbol location would need to alter a shader group unless we were to introduce some sort of lazily constructed accelerator cache...but that could easily be marked mutable or we change back at that point. So, LGTM.

PR #1724, opened 2023-09-04
build(deps): Raise minimum CMake dependency from 3.12 to 3.15

lgritz:
This bump brings access to the following interesting features (not a comprehensive list):

- 3.13  Nov 21, 2018

    - -B, -S let you specify source and build directories without the "cd" dance
    - target_link_options()
    - ctest --progress / CTEST_PROGRESS_OUTPUT

- 3.14  Mar 14, 2019

    - cmake --build now supports -v/--verbose
    - file(CREATE_LINK )

- 3.15  Jul 17, 2019

    - env CMAKE_GENERATOR can be used instead of -G
    - when using --build, the --target can now list multiple targets (and also can be `-t`)
    - --install can be used separately without running through the whole build
    - message() now supports NOTICE, VERBOSE, DEBUG and TRACE.
    - CMAKE_FIND_PACKAGE_PREFER_CONFIG variable
    - Enable C++20 for Apple Clang

Make sure all CI cases have a new enough cmake

Use message(VERBOSE ...) feature to simplify lots of `if` blocks.

Clean up code for older cmake versions

Freshen some of our CI build_<dependency>.bash scripts.

Note that this will NOT be backported to the current release branch, but it will be part of the upcoming 1.13 family of releases.


lgritz:
This has sat for a while without any comments. Last chance to object...

lgritz:
No objections, so I'm merging this.

PR #1726, opened 2023-09-10
Change RendererServices::get_texture_handle back to ustring

lgritz:
We were overzealous in switching everything to ustringhash.  The filename argument to get_texture_handle should allow taking a ustring.


AlexMWells (approved):
LGTM

PR #1727, opened 2023-09-11
LLVM 16 and opaque pointers support

brechtvl:
## Description

LLVM 16 was released a while ago but OSL does not support it yet. The main challenge is that this version requires opaque pointers, that is you can no longer tell from a pointer value which data type it points to.

Changes made:

* Bump minimum supported LLVM version to 16.
* Enable opaque pointers on LLVM 15 and 16.
* Remove deprecated `GEP` and `op_load` methods without type, and refactor code to always pass the type.
* Add element type argument to `op_scatter` and `op_gather`.
* Eliminate some pointer type comparison which no longer work with opaque pointers. Some asserts remain which pass but are not so useful with opaque pointers.
* Add `clangASTMatchers` library for LLVM 16, will be skipped for older LLVM version.
* Rename `makeArrayRef` to `toArrayRef` to `llvm::makeArrayRef` deprecation warnings.
* Minor required changes related to `llvm::Align`.
* Add support for new pass manager, as the legacy pass manager no longer supports default optimization levels.
* The legacy pass manager is still used for custom optimization levels 10-13.

Remaining work:

* OptiX support was not tested yet. It is likely broken due to the sneaky conversion in `BackendLLVM::llvm_store_value` between `int64` and `char*`, but we can no longer compare the pointer type. This function is called from many places and it's unclear to me when this happens. According to a4e779c4 this code is only temporary.

## Tests

No new tests as the whole testsuite covers this.

Some tests are failing currently as this is work in progress.

## Checklist:

### DIFF
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -888,7 +888,7 @@ public:
     llvm::Value* op_load(llvm::Type* type, llvm::Value* ptr,
                          const std::string& llname = {});
 
-    llvm::Value* op_gather(llvm::Type* ptr_element_type, llvm::Value* ptr,
+    llvm::Value* op_gather(llvm::Type* ptr_type, llvm::Value* ptr,
                            llvm::Value* wide_index);
 
     /// Store to a dereferenced pointer
@@ -906,7 +906,7 @@ public:
     /// converting it to the native mask type for storage:   *ptr = llvm_mask_to_native(val);
     void op_store_mask(llvm::Value* llvm_mask, llvm::Value* native_mask_ptr);
 
-    void op_scatter(llvm::Type* ptr_element_type, llvm::Value* wide_val,
+    void op_scatter(llvm::Value* wide_val, llvm::Type* ptr_type,
                     llvm::Value* ptr, llvm::Value* wide_index);
 
     // N.B. "GEP" -- GetElementPointer -- is a particular LLVM-ism that is
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -624,6 +624,9 @@ public:
                             const std::string& name = "",
                             bool is_packed          = false);
 
+    // Return type of field at given index in a struct type.
+    llvm::Type* type_struct_field_at_index(llvm::Type* type, int index);
+
     /// Return the llvm::Type that is a pointer to the given llvm type.
     llvm::PointerType* type_ptr(llvm::Type* type);
 
@@ -884,10 +887,9 @@ public:
     /// type is the type of the thing being pointed to.
     llvm::Value* op_load(llvm::Type* type, llvm::Value* ptr,
                          const std::string& llname = {});
-    // Blind pointer version that's deprecated as of LLVM13:
-    llvm::Value* op_load(llvm::Value* ptr, const std::string& llname = {});
 
-    llvm::Value* op_gather(llvm::Value* ptr, llvm::Value* index);
+    llvm::Value* op_gather(llvm::Type* ptr_element_type, llvm::Value* ptr,
+                           llvm::Value* wide_index);
 
     /// Store to a dereferenced pointer
     /// respecting the current mask & masking_enabled flag:   *ptr = val
@@ -904,8 +906,8 @@ public:
     /// converting it to the native mask type for storage:   *ptr = llvm_mask_to_native(val);
     void op_store_mask(llvm::Value* llvm_mask, llvm::Value* native_mask_ptr);
 
-    void op_scatter(llvm::Value* wide_val, llvm::Value* ptr,
-                    llvm::Value* wide_index);
+    void op_scatter(llvm::Type* ptr_element_type, llvm::Value* wide_val,
+                    llvm::Value* ptr, llvm::Value* wide_index);
 
     // N.B. "GEP" -- GetElementPointer -- is a particular LLVM-ism that is
     // the means for retrieving elements from some kind of aggregate: the
@@ -918,16 +920,10 @@ public:
     /// we're retrieving.
     llvm::Value* GEP(llvm::Type* type, llvm::Value* ptr, llvm::Value* elem,
                      const std::string& llname = {});
-    // Blind pointer version that's deprecated as of LLVM13:
-    llvm::Value* GEP(llvm::Value* ptr, llvm::Value* elem,
-                     const std::string& llname = {});
 
     /// Generate a GEP (get element pointer) with an integer element
     /// offset. `type` is the type of the data we're retrieving.
     llvm::Value* GEP(llvm::Type* type, llvm::Value* ptr, int elem,
-                     const std::string& llname = {});
-    // Blind pointer version that's deprecated as of LLVM13:
-    llvm::Value* GEP(llvm::Value* ptr, int elem,
                      const std::string& llname = {});
 
     /// Generate a GEP (get element pointer) with two integer element
@@ -937,9 +933,14 @@ public:
     /// retrieving.
     llvm::Value* GEP(llvm::Type* type, llvm::Value* ptr, int elem1, int elem2,
                      const std::string& llname = {});
-    // Blind pointer version that's deprecated as of LLVM13:
-    llvm::Value* GEP(llvm::Value* ptr, int elem1, int elem2,
-                     const std::string& llname = {});
+
+    /// Generate a GEP (get element pointer) with three integer element
+    /// offsets.  This is just a special (and common) case of GEP where
+    /// we have a 3-level hierarchy and we have fixed element indices
+    /// that are known at compile time.  `type` is the type of the data we're
+    /// retrieving.
+    llvm::Value* GEP(llvm::Type* type, llvm::Value* ptr, int elem1, int elem2,
+                     int elem3, const std::string& llname = {});
 
     // Arithmetic ops.  It auto-detects the type (int vs float).
     // ...
--- a/src/include/OSL/llvm_util.h
+++ b/src/include/OSL/llvm_util.h
@@ -1029,6 +1029,7 @@ public:
 private:
     class MemoryManager;
     class IRBuilder;
+    struct NewPassManager;
 
     void SetupLLVM();
     IRBuilder& builder();
@@ -1046,6 +1047,7 @@ private:
     llvm::Function* m_current_function;
     llvm::legacy::PassManager* m_llvm_module_passes;
     llvm::legacy::FunctionPassManager* m_llvm_func_passes;
+    NewPassManager* m_new_pass_manager;
     llvm::ExecutionEngine* m_llvm_exec;
     TargetISA m_target_isa = TargetISA::UNKNOWN;
 
@@ -1208,6 +1210,9 @@ private:
                                        llvm::Value* half_vec_2);
     llvm::Value* op_combine_4x_vectors(llvm::Value* half_vec_1,
                                        llvm::Value* half_vec_2);
+
+    void setup_legacy_optimization_passes(int optlevel, bool target_host);
+    void setup_new_optimization_passes(int optlevel, bool target_host);
 };
 
 
--- a/src/liboslexec/batched_analysis.cpp
+++ b/src/liboslexec/batched_analysis.cpp
@@ -1117,15 +1117,9 @@ public:
     }
 
 #ifdef OSL_DEV
-    OSL_FORCEINLINE int op_num() const
-    {
-        return m_op_num;
-    }
+    OSL_FORCEINLINE int op_num() const { return m_op_num; }
 
-    OSL_FORCEINLINE int loop_op_index() const
-    {
-        return m_loop_op_index;
-    }
+    OSL_FORCEINLINE int loop_op_index() const { return m_loop_op_index; }
 #endif
 };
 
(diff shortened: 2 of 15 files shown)


brechtvl:
Maybe a better solution would be to introduce an OSL non-opaque pointer type that wraps an `llvm::Value*` and `llvm::Type*` for the data it points to. That would sidestep the need to figure out the right types for batch mode types and `llvm_store_value`.

The full consequences of adding that additional abstraction to `LLVM_Util` are not clear to me, if perhaps using higher level abstractions than the LLVM API will get us into trouble later.

AlexMWells:
> Maybe a better solution would be to introduce an OSL non-opaque pointer type that wraps an `llvm::Value*` and `llvm::Type*` for the data it points to. 

If llvm's point of view are that type's don't matter for pointers, well something higher needs to actually track them.  So an having a wrapper of those 2 seems like a very reasonable approach to be able to ease into the transition.

lgritz:
What I was thinking is that LLVM_Util::op_load would be augmented with another parameter giving the type that we expect, and everyplace we call op_load, we would pass the expected type. For old LLVM versions, it would just check that they matched, for new LLVM that would be what the opaque pointer would need to know what it is.

So the first step -- add the type paramter to op_load, pass it everywhere it's needed, check that it's right -- can be done first, independently, to get the whole thing plumbed to throw the switch for opaque pointers.

brechtvl:
@lgritz there is already a version of `op_load` (and  `GEP`) with a type, and for non-batched code it was relatively straightforward to use them everywhere.

For the batched code I'm having difficulty finding what the right types are just from the context of each function. When tracing further backwards I can't see clearly when a type is wide or not, or when it's a wide bool or integer mask. It may just be my unfamiliarity with the code and @AlexMWells can tell quickly.

I left TODO comments in the code where I was uncertain.

src/liboslexec/batched_backendllvm.cpp
```diff
@@ -811,26 +811,29 @@ BatchedBackendLLVM::llvm_load_value(llvm::Value* ptr, const TypeSpec& type,
...
             OSL_DEV_ONLY(std::cout << "step to the right field " << component
                                    << std::endl);
-            ptr = ll.GEP(ptr, 0, component);
+            ptr = ll.GEP(element_type, ptr, 0, component);
         }
 
         // Now grab the value
-        llvm::Value* result;
-        result = ll.op_load(ptr);
+        // TODO: fails with assert later in ll.native_to_llvm_mask
```
AlexMWells, line 834: Well, here we are relying on the underlying llvm type to identify if the value we just loaded is uniform or varying.
This differs from the parameter "bool op_is_uniform" which indicates if the op the participating in is varying or not.  I.E.:
`[[varying]] int C = [[uniform]] A != [[varying]] B;`

In that case the op itself is varying, and when `A` is loaded, it will need to be widened before being passed to LLVM_Util::op_eq(...) so it matches types with the varying value loaded for `B`.

So for batched, not only do we need to know the TypeSpec of the symbol being loaded, we will also need an explicit "bool symbol_is_uniform".  In this case if that parameter was added:
```
llvm::Value*
BatchedBackendLLVM::llvm_load_value(llvm::Value* ptr, const TypeSpec& type,
                                    int deriv, llvm::Value* arrayindex,
                                    int component, TypeDesc cast,
                                    bool symbol_is_uniform, bool op_is_uniform, bool index_is_uniform,
                                    bool symbol_forced_boolean)
```
Then the code that was causing problems (asserting):
```
        if (type.is_int() && symbol_forced_boolean) {
            // We only need to convert wide native masks
            // and op_is_uniform doesn't guarantee that the symbol it self
            // in non-unform, it could just be a single bool (vs. mask).
            if (!op_is_uniform && (ll.llvm_typeof(result) != ll.type_bool())) {
                // We just loaded a native mask need to convert it
                // to a vector of bools
                result = ll.native_to_llvm_mask(result);
            }
```
could be written as:
```
        if (type.is_int() && symbol_forced_boolean) {
            // We only need to convert wide native masks
            // and op_is_uniform doesn't guarantee that the symbol it self
            // in non-unform, it could just be a single bool (vs. mask).
            if (!op_is_uniform && symbols_is_uniform) {
                // We just loaded a native mask need to convert it
                // to a vector of bools
                result = ll.native_to_llvm_mask(result);
            }
```            
You see how the previous use of the llvm::Value *result's type,  `(ll.llvm_typeof(result) != ll.type_bool())`, was just to figure out if the boolean value loaded was uniform or varying.

Of course this means having all callers pass a valid symbols_is_uniform parameter in, luckily the `Symbol::is_uniform()` has the answer.

So would suggest starting with that change and use the extensive testsuite to route out other useses.
Your might even comment out LLVM_Util::llvm_typeof to route out all its uses.

brechtvl: Thanks! That really helped me understand things better.

brechtvl:
I got batched mode tests to pass as well now on my system. OptiX likely still broken due to the casting issue mentioned in the description.

Other changes:
* Add temporary CMake variable to easily test opaque pointers on/off
* Add asserts in op_load and GEP to validate pointer types (when opaque pointers are off)
* Keep interactive param symbols always marked as uniform, to help finding correct pointer type
* Fix build with LLVM 15

brechtvl:
Which version of `clang-format` should I be using? I tried 12, 15 and 16 and none matched exactly.

lgritz:
Our CI job uses clang-format 14

brechtvl:
It turns out there is another significant change in LLVM 16, where support for default optimization levels (O0, O1, O2, O3) was removed from the legacy pass manager.

I added support for the new pass manager to address that. For the custom optimization levels 10-13 it still uses the legacy pass manager since I did not implement those. And of course the performance impact of this should be checked.

It's probably time to break up this PR into multiple smaller ones.

lgritz:
You're a big hero for all of this, Brecht.

Yeah, the more bite sized chunks, the better. Maybe the order should be:

1. Switch to new pass manager when possible.
2. The changes that allow us to build against LLVM16 but still using typed pointers (I think 16 still allows it, optionally, but in 17 it's gone for good).
3. Switch to being able to use opaque pointers.

lgritz:
@brechtvl, We currently claim to support as far back as llvm 9, but such a wide range is really not necessary (and perhaps not desirable, because it's too much trouble to always test against them all). I would prefer (if possible) to preserve the ability to build against several old versions (i.e., not force us to move our minimum all the way up to llvm 16), but at the same time, if it helps you to bring the floor up somewhat, you should definitely feel free to do so.

brechtvl:
I don't expect any of these changes to require raising the floor.

brechtvl:
> The changes that allow us to build against LLVM16 but still using typed pointers (I think 16 still allows it, optionally, but in 17 it's gone for good).

I was hoping the opaque pointers would still work, but I got this error loading the bitcode. Even though `-Xclang -no-opaque-pointers` is passed when compiling the bitcode.
```
Opaque pointers are only supported in -opaque-pointers mode (Producer: 'LLVM16.0.0' Reader: 'LLVM 16.0.0')
```

I guess it's not too surprising given the LLVM docs:
> LLVM 15: Opaque pointers are enabled by default. Typed pointers are still supported.
> LLVM 16: Opaque pointers are enabled by default. Typed pointers are supported on a best-effort basis only and not tested.
https://llvm.org/docs/OpaquePointers.html

brechtvl:
This is now replaced by #1728, #1729 and #1730.

lgritz:
Is there a preferred order in which we review and integrate them?

brechtvl:
For reviewing the order does not matter. For merging, #1730 should be done last.

src/include/OSL/llvm_util.h
```diff
@@ -906,7 +906,7 @@ class OSLEXECPUBLIC LLVM_Util {
     /// converting it to the native mask type for storage:   *ptr = llvm_mask_to_native(val);
     void op_store_mask(llvm::Value* llvm_mask, llvm::Value* native_mask_ptr);
 
-    void op_scatter(llvm::Type* ptr_element_type, llvm::Value* wide_val,
+    void op_scatter(llvm::Value* wide_val, llvm::Type* ptr_type,
```
AlexMWells, line 909: Was there a reason to swap type and value parameters?

src/liboslexec/batched_analysis.cpp
```diff
@@ -2008,7 +2002,13 @@ struct Analyzer {
     {
         bool previously_was_uniform = symbol_to_be_varying->is_uniform();
         if (previously_was_uniform | force) {
-            symbol_to_be_varying->make_varying();
+            // Interactive params are never varying, following getLLVMSymbolBase()
```
AlexMWells, line 2011: While it is technically correct that interactive params are always uniform, we shouldn't have to force it.
The need for this change show's that there is bug elsewhere.



Previously there was just Symbol::lockgeom() and that changed to Symbol::interpolated() and Symbol::interactive() and the lockgeom covered both:

    bool Symbol::lockgeom() const
    {
        // We can lock a value to a constant (at all places on all pieces of
        // geometry) if it is neither interpolated nor interactively modified.
        return !m_interpolated && !m_interactive;
    }


However not all code that used lockgeom got updated, in particular the method:    


    void Analyzer::make_interpolated_parameters_varying()
    {
        // Mark all interpolated parameters as varying,
        // As we expect interpolated data,  get_userdata will
        // only provide varying values, so we must mark
        // our symbols appropriately!
        FOREACH_PARAM(Symbol & s, inst())
        {
            if (s.everread() && !s.lockgeom() && !s.typespec().is_closure()) {
                recursively_mark_varying(&s);
            }
        }
    }


Could you try changing it from !lockgeom() to interactive() there:

    void Analyzer::make_interpolated_parameters_varying()
    {
        // Mark all interpolated parameters as varying,
        // As we expect interpolated data,  get_userdata will
        // only provide varying values, so we must mark
        // our symbols appropriately!
        FOREACH_PARAM(Symbol & s, inst())
        {
            if (s.everread() && s.interpolated() && !s.typespec().is_closure()) {
                recursively_mark_varying(&s);
            }
        }
    }

  
Then replace your force_uniform change with an
    `OSL_ASSERT(!symbol_to_be_varying->interactive());`

src/liboslexec/batched_backendllvm.cpp
```diff
@@ -800,40 +804,56 @@ BatchedBackendLLVM::llvm_load_mask(const Symbol& cond)
...
+    TypeDesc t = type.simpletype();
+    llvm::Type *ptr_type, *ptr_scalar_type;
+
+    if (symbol_forced_boolean) {
+        if (ptr_is_uniform) {
+            ptr_type        = ll.type_bool();
+            ptr_scalar_type = ll.type_bool();
+        } else {
+            ptr_type        = ll.type_native_mask();
+            ptr_scalar_type = ll.type_native_mask();
```
AlexMWells, line 826: ptr_scalar_type = ll.type_native_mask();
I'm not sure if that is quite right or not, should it just be 
      ptr_scalar_type = ll.type_bool();
Perhaps its not used later when symbol_forced_bool && !ptr_is_uniform

src/liboslexec/batched_llvm_instance.cpp
```diff
@@ -1024,7 +1024,6 @@ BatchedBackendLLVM::llvm_assign_initial_value(
             // Because we allow temporaries and local results of comparison operations
             // to use the native bool type of i1, we can just skip initializing these
             // as they should always be assigned a value.
-            // TODO: is this equivalent to the old pointer type check? probably not
```
AlexMWells: Can remove this comment, looks ok, no pointer types, just Symbol stuff.

src/liboslexec/batched_llvm_instance.cpp
```diff
@@ -1974,7 +1972,7 @@ BatchedBackendLLVM::build_llvm_instance(bool groupentry)
 
     llvm::Value* previously_executed_value = nullptr;
     if (!group().is_last_layer(layer())) {
-        previously_executed_value = ll.op_load(ll.type_wide_bool(), layerfield);
+        previously_executed_value = ll.op_load(ll.type_int(), layerfield);
```
AlexMWells, line 1975: Did this need to change?

brechtvl:
This PR is closed and outdated now, please see #1728 instead.

src/liboslexec/batched_backendllvm.cpp
```diff
@@ -800,40 +804,56 @@ BatchedBackendLLVM::llvm_load_mask(const Symbol& cond)
 llvm::Value*
 BatchedBackendLLVM::llvm_load_value(llvm::Value* ptr, const TypeSpec& type,
                                     int deriv, llvm::Value* arrayindex,
-                                    int component, TypeDesc cast,
-                                    bool op_is_uniform, bool index_is_uniform,
+                                    int component, bool ptr_is_uniform,
```
AlexMWells, line 809: `bool ptr_is_uniform`

I found the name to be confusing, because varying pointers exist and this is really saying what the ptr points to is uniform.

In the store I saw dst_is_uniform and it made total sense, so maybe for the loads we could just use 
`bool src_in_uniform`
or something instead

src/liboslexec/batched_llvm_instance.cpp
brechtvl: Already removed in the new PR.

src/liboslexec/batched_backendllvm.cpp
brechtvl: Resolved in the new PR.

src/liboslexec/batched_backendllvm.cpp
brechtvl: Discussion moved to the new PR.

PR #1728, opened 2023-09-16
LLVM opaque pointers support

brechtvl:
## Description

New LLVM versions require opaque pointers, that is you can no longer tell from a pointer value which data type it points to.

* Remove deprecated `GEP` and `op_load` methods without type, and refactor code to always pass the type.
* Add element type argument to `op_scatter` and `op_gather`.
* Eliminate pointer type comparisons which no longer work with opaque pointers.
* Some useful asserts to verify correct pointer types must unfortunately be disabled with opaque pointers.
* Keep interactive params marked as uniform in batched analysis, so we can deduce uniform/varying correctly from just the symbol.
* Enable opaque pointers on LLVM 16+.

OptiX opaque pointers might not be working yet due to the conversion in`BackendLLVM::llvm_store_value` between `int64` and `char*`, where we can no longer compare the pointer type. This function is called from many places and it's unclear to me when this happens.

According to a4e779c4 this code is only temporary, so this may resolve itself as the ustring refactoring gets completed.

## Tests

No new tests added, entire testsuite tests this.

## Checklist:

### DIFF
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -749,7 +749,8 @@ llvm::Value*
 BackendLLVM::layer_run_ref(int layer)
 {
     int fieldnum = 0;  // field 0 is the layer_run array
-    return ll.GEP(llvm_type_groupdata(), groupdata_ptr(), 0, fieldnum, layer);
+    return ll.GEP(llvm_type_groupdata(), groupdata_ptr(), 0, fieldnum, layer,
+                  llnamefmt("layer_runflags_ref"));
 }
 
 
@@ -759,7 +760,7 @@ BackendLLVM::userdata_initialized_ref(int userdata_index)
 {
     int fieldnum = 1;  // field 1 is the userdata_initialized array
     return ll.GEP(llvm_type_groupdata(), groupdata_ptr(), 0, fieldnum,
-                  userdata_index);
+                  userdata_index, llnamefmt("userdata_init_flags_ref"));
 }
 
 
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -509,8 +509,8 @@ BackendLLVM::llvm_load_value(llvm::Value* ptr, const TypeSpec& type, int deriv,
         ptr = ll.GEP(element_type, ptr, 0, component);
 
     // Now grab the value
-    llvm::Type* scalar_type = llvm_type(t.scalartype());
-    llvm::Value* result     = ll.op_load(scalar_type, ptr, llname);
+    llvm::Type* component_type = llvm_type(t.scalartype());
+    llvm::Value* result        = ll.op_load(component_type, ptr, llname);
 
     if (type.is_closure_based())
         return result;
--- a/src/liboslexec/backendllvm.cpp
+++ b/src/liboslexec/backendllvm.cpp
@@ -212,7 +212,7 @@ BackendLLVM::llvm_global_symbol_ptr(ustring name)
     // the ShaderGlobals struct.
     int sg_index = ShaderGlobalNameToIndex(name);
     OSL_ASSERT(sg_index >= 0);
-    return ll.void_ptr(ll.GEP(sg_ptr(), 0, sg_index),
+    return ll.void_ptr(ll.GEP(llvm_type_sg(), sg_ptr(), 0, sg_index),
                        llnamefmt("glob_{}_voidptr", name));
 }
 
@@ -429,7 +429,7 @@ BackendLLVM::llvm_get_pointer(const Symbol& sym, int deriv,
             arrayindex = ll.op_add(arrayindex, ll.constant(d));
         else
             arrayindex = ll.constant(d);
-        result = ll.GEP(result, arrayindex);
+        result = ll.GEP(llvm_type(t.elementtype()), result, arrayindex);
     }
 
     return result;
@@ -493,22 +493,24 @@ BackendLLVM::llvm_load_value(llvm::Value* ptr, const TypeSpec& type, int deriv,
 
     // If it's an array or we're dealing with derivatives, step to the
     // right element.
-    TypeDesc t = type.simpletype();
+    TypeDesc t               = type.simpletype();
+    llvm::Type* element_type = llvm_type(t.elementtype());
     if (t.arraylen || deriv) {
         int d = deriv * std::max(1, t.arraylen);
         if (arrayindex)
             arrayindex = ll.op_add(arrayindex, ll.constant(d));
         else
             arrayindex = ll.constant(d);
-        ptr = ll.GEP(ptr, arrayindex);
+        ptr = ll.GEP(element_type, ptr, arrayindex);
     }
 
     // If it's multi-component (triple or matrix), step to the right field
     if (!type.is_closure_based() && t.aggregate > 1)
-        ptr = ll.GEP(ptr, 0, component);
+        ptr = ll.GEP(element_type, ptr, 0, component);
 
     // Now grab the value
-    llvm::Value* result = ll.op_load(ptr, llname);
+    llvm::Type* scalar_type = llvm_type(t.scalartype());
+    llvm::Value* result     = ll.op_load(scalar_type, ptr, llname);
 
     if (type.is_closure_based())
         return result;
@@ -597,10 +599,10 @@ BackendLLVM::llvm_load_component_value(const Symbol& sym, int deriv,
     OSL_DASSERT(sym.typespec().simpletype().aggregate != TypeDesc::SCALAR);
     // cast the Vec* to a float*
     result = ll.ptr_cast(result, ll.type_float_ptr());
-    result = ll.GEP(result, component);  // get the component
+    result = ll.GEP(ll.type_float(), result, component);  // get the component
 
     // Now grab the value
-    return ll.op_load(result);
+    return ll.op_load(ll.type_float(), result);
 }
 
 
@@ -664,21 +666,23 @@ BackendLLVM::llvm_store_value(llvm::Value* new_val, llvm::Value* dst_ptr,
 
     // If it's an array or we're dealing with derivatives, step to the
     // right element.
-    TypeDesc t = type.simpletype();
+    TypeDesc t               = type.simpletype();
+    llvm::Type* element_type = llvm_type(t.elementtype());
     if (t.arraylen || deriv) {
         int d = deriv * std::max(1, t.arraylen);
         if (arrayindex)
             arrayindex = ll.op_add(arrayindex, ll.constant(d));
         else
             arrayindex = ll.constant(d);
-        dst_ptr = ll.GEP(dst_ptr, arrayindex);
+        dst_ptr = ll.GEP(element_type, dst_ptr, arrayindex);
     }
 
     // If it's multi-component (triple or matrix), step to the right field
     if (!type.is_closure_based() && t.aggregate > 1)
-        dst_ptr = ll.GEP(dst_ptr, 0, component);
+        dst_ptr = ll.GEP(element_type, dst_ptr, 0, component);
 
     // Finally, store the value.
+    // TODO: this breaks OptiX, pointer type comparison no longer works with opaque pointers.
     if (t == TypeString && dst_ptr->getType() == ll.type_int64_ptr()
         && new_val->getType() == ll.type_char_ptr()) {
         // Special case: we are still ickily storing strings sometimes as a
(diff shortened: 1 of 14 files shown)


brechtvl:
The `pointcloud.batched` and `debug-uninit.batched` CI failures I guess are a real issue to be solved.

src/liboslexec/batched_backendllvm.h
```diff
@@ -114,7 +114,7 @@ class BatchedBackendLLVM : public OSOProcessorBase {
     /// (no conversion is performed if cast is the default of UNKNOWN).
     llvm::Value* llvm_load_value(llvm::Value* ptr, const TypeSpec& type,
                                  int deriv, llvm::Value* arrayindex,
-                                 int component,
+                                 int component, bool ptr_is_uniform,
```
AlexMWells, line 117: `bool ptr_is_uniform`

I found the name to be confusing, because varying pointers exist and this is really saying what the ptr points to is uniform.

In the store I saw `dst_is_uniform` and it made total sense, so maybe for the loads we could just use
`bool src_in_uniform`
or something instead

src/liboslexec/batched_analysis.cpp
```diff
@@ -2008,7 +2008,13 @@ struct Analyzer {
     {
         bool previously_was_uniform = symbol_to_be_varying->is_uniform();
         if (previously_was_uniform | force) {
-            symbol_to_be_varying->make_varying();
+            // Interactive params are never varying, following getLLVMSymbolBase()
+            const bool force_uniform = symbol_to_be_varying->symtype()
+                                           == SymTypeParam
+                                       && symbol_to_be_varying->interactive();
+            if (!force_uniform) {
+                symbol_to_be_varying->make_varying();
```
AlexMWells, line 2016: While it is technically correct that interactive params are always uniform, we shouldn't have to force it.
The need for this change show's that there is bug elsewhere.

Previously there was just `Symbol::lockgeom()` and that changed to `Symbol::interpolated()` and `Symbol::interactive()` and the lockgeom covered both:

```
bool Symbol::lockgeom() const
{
    // We can lock a value to a constant (at all places on all pieces of
    // geometry) if it is neither interpolated nor interactively modified.
    return !m_interpolated && !m_interactive;
}
```
However not all code that used lockgeom got updated, in particular the method:

```
void Analyzer::make_interpolated_parameters_varying()
{
    // Mark all interpolated parameters as varying,
    // As we expect interpolated data,  get_userdata will
    // only provide varying values, so we must mark
    // our symbols appropriately!
    FOREACH_PARAM(Symbol & s, inst())
    {
        if (s.everread() && !s.lockgeom() && !s.typespec().is_closure()) {
            recursively_mark_varying(&s);
        }
    }
}
```
Could you try changing it from !lockgeom() to interactive() there:

```
void Analyzer::make_interpolated_parameters_varying()
{
    // Mark all interpolated parameters as varying,
    // As we expect interpolated data,  get_userdata will
    // only provide varying values, so we must mark
    // our symbols appropriately!
    FOREACH_PARAM(Symbol & s, inst())
    {
        if (s.everread() && s.interpolated() && !s.typespec().is_closure()) {
            recursively_mark_varying(&s);
        }
    }
}
```
Then replace your force_uniform change with an
`OSL_ASSERT(!symbol_to_be_varying->interactive());`

src/liboslexec/batched_backendllvm.cpp
```diff
@@ -813,37 +819,56 @@ BatchedBackendLLVM::llvm_load_mask(const Symbol& cond)
...
+    TypeDesc t = type.simpletype();
+    llvm::Type *ptr_type, *ptr_scalar_type;
+
+    if (symbol_forced_boolean) {
+        if (ptr_is_uniform) {
+            ptr_type        = ll.type_bool();
+            ptr_scalar_type = ll.type_bool();
+        } else {
+            ptr_type        = ll.type_native_mask();
+            ptr_scalar_type = ll.type_native_mask();
```
AlexMWells, line 839: `  ptr_scalar_type = ll.type_native_mask();`
I'm not sure if that is quite right or not.  It just looked off to me.
Should it just be
`ptr_scalar_type = ll.type_bool();`
Perhaps its not used later when symbol_forced_bool && !ptr_is_uniform

brechtvl: My understanding is that `ptr_type` and `ptr_scalar_type` are meant to have the same value for booleans, ints and floats. The "scalar" here refers to taking a single component of a triple or matrix.

src/liboslexec/batched_backendllvm.h
brechtvl: Ok, I will rename `ptr` to `src_ptr` and `ptr_is_uniform` to `src_is_uniform`.

src/liboslexec/backendllvm.cpp
```diff
@@ -755,8 +758,8 @@ llvm::Value*
 BackendLLVM::userdata_initialized_ref(int userdata_index)
 {
     int fieldnum = 1;  // field 1 is the userdata_initialized array
```
AlexMWells, line 760: Once difference here is that groupdata_field_ref created some additional name/debug info associated with the GEP instruction.  

```
llvm::Value*
BackendLLVM::groupdata_field_ref(int fieldnum)
{
    return ll.GEP(llvm_type_groupdata(), groupdata_ptr(), 0, fieldnum,
                  llnamefmt("{}_ref", m_groupdata_field_names[fieldnum]));
}
```

As pointers become typeless, the names/debug info we generate might become more and more important.  Perhaps a follow on task after this transition.  But for those of us who end up reading llvm-ir, any metadata left around at all can be very useful.

src/liboslexec/batched_backendllvm.cpp
```diff
@@ -667,7 +667,12 @@ BatchedBackendLLVM::llvm_get_pointer(const Symbol& sym, int deriv,
             arrayindex = ll.op_add(arrayindex, ll.constant(d));
         else
             arrayindex = ll.constant(d);
-        result = ll.GEP(result, arrayindex);
+
```
AlexMWells, line 670: Add comment here:
`// Array's will not be forced_llvm_bool, so no need to check`

src/liboslexec/batched_analysis.cpp
AlexMWells: Might be more broken there, let me look a bit wider in scope.

src/liboslexec/batched_backendllvm.cpp
AlexMWells: Hmm, I think scalar vs vector type when I see scalar.  Vs. element or component type.  So would `ptr_comp_type` be more accurate?

AlexMWells:
```
OptiX opaque pointers were not tested yet and left disabled. It is likely broken due to the sneaky conversion in
BackendLLVM::llvm_store_value between int64 and char*, but we can no longer compare the pointer type. This function is called from many places and it's unclear to me when this happens.
```
In an upcoming PR, StringPhase3, we remove all type mutability for strings, its either a ustring_pod (char *) or a ustringhash_pod (int64).  So you might wait for that PR to flow in before this one as to be able to try and handle the situation above.

src/liboslexec/batched_analysis.cpp
AlexMWells: This might not be it, but was one other spot using lockgeom, in Analyzer::discover_init_symbols
```
            // Skip if it's an interpolated (userdata) parameter and we're
            // initializing them lazily.
            if ((s.symtype() == SymTypeParam
                 || s.symtype() == SymTypeOutputParam)
                && !s.lockgeom() && !s.typespec().is_closure() && !s.connected()
                && !s.connected_down() && m_ba.shadingsys().lazy_userdata())
                continue;

```                
Try changing it to more closely match code from BatchedBackendLLVM::build_llvm_instance
```
        // Skip if it's an interpolated (userdata) parameter and we're
        // initializing them lazily, or if it's an interactively-adjusted
        // parameter.
        if ((s.symtype() == SymTypeParam
                 || s.symtype() == SymTypeOutputParam)
            && !s.typespec().is_closure()
            && !s.connected() && !s.connected_down()
            && (s.interactive()
                || (s.interpolated() && shadingsys().lazy_userdata())))
            continue;

```
Note there is still a difference between the build_llvm_instance logic and the Analyzer.  As written the Analyzer allow OutputParams to be interactive or interpolated, while build_llvm_instance doesn't.  In a recent discussion, since oslc allows interpolated attribute on output params, it must be legal.  So in a different PR, I think we are changing the build_llvm_instance and llvm_instance to match.

And based on all that, I suppose the requested
`OSL_ASSERT(!symbol_to_be_varying->interactive());`
was way to simple.

It should be more like:
`OSL_ASSERT(!symbol_to_be_varying->interactive() || s.typespec().is_closure() || s.connected() || s.connected_down());`

I guess if that assert fires, I'd be interested in knowing all the symbol's data member values.

brechtvl: This didn't work either.

What happens is that the `reparam` test has a parameter `float user = 2 [[ int interpolated = 1 ]]`, that is made interactive. `Analyzer::make_interpolated_parameters_varying` encounters this symbol, which is marked both interpolated and interactive, and marks it as varying.

The follwing makes all the batched tests pass. It's basically just moving the check to force interactive parameters as uniform one level up in the callstack.
```
diff --git a/src/liboslexec/batched_analysis.cpp b/src/liboslexec/batched_analysis.cpp
index 18f0edd..e46fa74 100644
--- a/src/liboslexec/batched_analysis.cpp
+++ b/src/liboslexec/batched_analysis.cpp
@@ -2303,7 +2298,8 @@ struct Analyzer {
         // our symbols appropriately!
         FOREACH_PARAM(Symbol & s, inst())
         {
-            if (s.everread() && !s.lockgeom() && !s.typespec().is_closure()) {
+            if (s.everread() && s.interpolated() && !s.interactive()
+                && !s.typespec().is_closure()) {
                 recursively_mark_varying(&s);
             }
         }
```

My understanding from reading the code is that interactive always override interpolated values. If a symbol is interactive it is not interpolated. Because the interpolated values are bound in `BackendLLVM::llvm_assign_initial_value`, but `BackendLLVM::build_llvm_instance` skips calling that for interactive parameters.

src/liboslexec/batched_backendllvm.cpp
brechtvl: Renamed now.

src/liboslexec/backendllvm.cpp
brechtvl: I added (back) names to those groupdata GEP instructions now, both in batched and non-batched code.

src/liboslexec/batched_analysis.cpp
brechtvl: I added a fix along these lines now.

AlexMWells:
@brechtvl the changes to batched_analysis look good.  But you will want to rebase/resolve on main, as
https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1717
has been merged in which had changes in batched_analysis.cpp.

There is also another large pending PR
https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1732
which might resolve (or be closer to) some of the Optix issues, as all the stores should be explicit ustring (const char *) or ustringhash (uint64_t)

lgritz:
I checked out this branch, rebased it myself and added minor fixups to address the merge conflict, and it all seems to work. In addition to passing CI (except for the MacOS 13 test which is broken for unrelated reasons), I confirmed that it works fine on my laptop with LLVM 15 (and thus with the opaque pointers enabled), as well as at work -- which used LLVM 14, so no opaque pointers, but at least that confirms that it doesn't break anything that worked before, including testing both batched and OptiX modes.

I wanted to update your tree for you to save you the work, and then approve the PR. But for some reason, GitHub won't let me push back to your branch, even though you have marked the PR as "Maintainers are allowed to edit this pull request."

I pushed it to my fork: https://github.com/lgritz/OpenShadingLanguage/tree/llvm-opaque-pointers

Brecht, if you want to grab it from there and fold it into this PR, I think we are ready to merge this, unless you think there is more work needed.

brechtvl:
I made this PR match the branch in your fork now.

But, LLVM 15 + OptiX might be broken because of that remaining TODO. So I only enabled opaque pointers on LLVM 16+ now to be on the safe side.

lgritz:
I'm in favor of taking steps forward, even if not perfect. As long as we don't break anybody using LLVM 15 or older, anything that makes it work on 16+, even for some modes is forward movement. We can fix what's still missing in subsequent patches.

lgritz (approved):

lgritz:
LGTM. Shall I merge?

PR #1729, opened 2023-09-16
LLVM new pass manager support

brechtvl:
## Description

LLVM 16 removes support for default optimizations levels (O1, O2, O3) from the legacy pass manager, which are used by OSL. This adds supports for the new pass manager while keeping the legacy pass manager working for older LLVM versions.

* Enable new pass manager for LLVM 16+
* Both default and custom optimization levels have been implemented
* But the performance and correctness beyond passing tests has not been verified
* `ptx_compile_group` still uses the legacy pass manager for linking

## Tests

No new tests, the entire testsuite tests this.

## Checklist:


brechtvl:
The conservative option here would to only enable this for LLVM 16+, so that anyone already building with LLVM 15 will see no difference. But for CI in this PR it's useful to test at least one LLVM version with the new pass manager.

lgritz:
I agree, let's not enable for LLVM 15 to avoid any possible breakage for now. 

Let's get these reviewed and merged, and then we'll adjust the CI to have a test that builds against 16.

src/liboslexec/llvm_passes.h
```diff
@@ -37,9 +38,8 @@ namespace {
 // how instruction lowering happens then this pass may not be necessary.
 // Also if future LLVM version takes on the work of this optimization pass,
 // then it may be removed.
-template<int WidthT>
-class PreventBitMasksFromBeingLiveinsToBasicBlocks final
-    : public llvm::FunctionPass {
+
+template<int WidthT> class PreventBitMasksFromBeingLiveinsToBasicBlocks {
```
lgritz, line 42: I'm not sure I understand what's going on here. Why does this no longer inherit from llvm::FunctionPass?

brechtvl: Only for the legacy pass manager do we need to inherit from `llvm::FunctionPass`. This class contains the implementation that is shared by `NewPreventBitMasksFromBeingLiveinsToBasicBlocks` and `LegacyPreventBitMasksFromBeingLiveinsToBasicBlocks`.

lgritz:
I think what I said before may not have been clear. Let me try again:

For now, enable this for LLVM 15 so it's well tested in CI.
Once all of these PRs get merged, we'll be sure to add LLVM 16 to the CI, and then if we want, we can re-disable LLVM so that it uses the old pass manager and therefore can't break what was working before for people using LLVM <= 15.

lgritz:
I think this is the next one we should merge. I believe it needs a rebase to double check everything.

brechtvl:
This one I think we also want to switch it to LLVM 16+ only before merging.

lgritz:
It seems to pass all the tests. Are you just worried about the risk, not wanting to disrupt the versions that worked before?

brechtvl:
The main risk I see is that OptiX might break due to the problem mentioned in https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1730#issuecomment-1738971908, or that performance might be bad since I don't have a good way to test it.

I guess I don't see a good reason to take any risk if we can just add LLVM 16 to CI soon and test the new code that way.

lgritz:
Your plan is fine with me.

Do you have more to add here, then?

brechtvl:
I think this is ready to merge now.

lgritz:
OK, will merge when the CI run completes.

AlexMWells (commented):
LGTM!

PR #1731, opened 2023-09-19
RS::get_texture_info make robust to empty shaderglobals param

lgritz:
The unconditional dereferencing of the sg pointer is only necessary if texture_thread_info is not supplied. And for that matter, the texture_thread_info is only necessary if the texture_handle is not supplied.

So rearranging the nesting skips a bunch of the work, and the need for the caller to supply the sg if they have the texture_thread_info, nor for them to need texture_thread_info if they already have the handle.


AlexMWells (approved):
LGTM

PR #1732, opened 2023-09-20
Ustringhash phase 3

steenax86:
## Description

Move OSL library functions and remaining execution time interface from ustring to ustringhash.

Intent is all code for cpu or device just use ustringhash only.  This allows us to remove
STRING_PARAM, ustringrep, and OSL_USTRINGREP_IS_HASH.  All interfaces and library code
should explicitly use ustringhash or ustring,
or for library functions llvm can call ustringhash_pod or ustring_pod

Compile time strings used in the OSL Library are represented in strdecls.h
and will be available as ustring variables in the Strings namespace as well as
available as unstringhash constexpr variables in the Hashes namespace.

Compile time strings used by the Renderer should adopt a similiar strategy as
exemplified by TestShade which keeps its strings in rs_strdecls.h
and made available as ustring variables in the RS::Strings namespace as well as
available as unstringhash constexpr variables in the RS::Hashes namespace.

Added helper utility functions to generate ustring from ustringhash_pods

Constant string arrays: Previously, array copys required copying data via address
of symbol on host. To enable working on device, replaced memcpy with array allocation
and loading of values. Constant arrays are now present as
global arrays in final llvm module.

On the Batched side compatability/integration:
Batched OSL Library functions continue to use ustrings (not ustringhashes to reduce scope/risk).
Batched OSL library functions that call their scalar counterparts were adjusted
to pass arguments as ustringhash_pods and store results as ustrings

STRING_PARAMS, device_string.h, and string_hash.h: Built-in strings on the CUDA side are now accessed via Hashes namespace, instead of STRING_PARAMS and are
evaluated as ustringhashes. Similarly, DeviceString, StringParams, HDSTR are obsolete, and device_string.h is eliminated.
Hashes.h, which only contains OIIO's strhash and pvtstrlen functions (both formerly in string_hash.h), replaces all inclusions of device_string.h. Namespace Hashes in hashes.h
contains pre-compiled ustringhashes representing strings from strdecls.h.

Texture, dictionary, closure: Helper implementation functions are used by both scalar and batched versions. However, since scalar uses ustringhashes and
the batched version uses ustrings added a flag to inform functions on treating TypeDesc::STRING has either a ustring or a ustringhash

## Tests

Added testing of bitcode rendereservices for unoptimized OSL shader.
array-copy tests regression: Added new test to verify copying of arrays of unequal lengths

## Checklist:


src/liboslexec/batched_llvm_gen.cpp
```diff
@@ -8267,7 +8318,7 @@ LLVMGEN(llvm_gen_split)
             rop.llvm_load_arg(*optSep, false /*derivs*/, op_is_uniform));
     } else {
         if (op_is_uniform) {
-            args.push_back(rop.ll.constant(""));
+            args.push_back(rop.ll.constant(ustring("").c_str()));
```
AlexMWells, line 8321: We thought this was a bug passing a char * vs. a true ustring, 
  `args.push_back(rop.ll.constant(""));`
but Turns out 
  `llvm::Value* LLVM_Util::constant(string_view s) { return constant(ustring(s)); }`
did the conversion for us already. 
Although as we move towards being more explicit, this change is probably welcome so its clear to the reader.

So 2 lines later we see a similar call:
` llvm::Value* wide_sep      = rop.ll.wide_constant("");`
 which again is handled properly by LLVM_Util, by implicitly creating a ustring.
 But perhaps we should update it to match the 2 lines before
`  llvm::Value* wide_sep      = rop.ll.wide_constant(ustring("").c_str());`
 ?

src/liboslexec/llvm_instance.cpp
```diff
@@ -660,10 +708,21 @@ BackendLLVM::llvm_assign_initial_value(const Symbol& sym, bool force)
                 llvm_zero_derivs(sym);
 #endif
         } else if (sym.interpolated() && !sym.typespec().is_closure()) {
+#if 1
```
AlexMWells, line 711: Remove #if 1 
and #endif

src/liboslexec/opclosure.cpp
```diff
@@ -76,7 +76,7 @@ osl_closure_to_string(ShaderGlobals* sg, ClosureColor* c)
     // Special case for printing closures
     std::ostringstream stream;
     stream.imbue(std::locale::classic());  // force C locale
-    print_closure(stream, c, &sg->context->shadingsys());
+    print_closure(stream, c, &sg->context->shadingsys(), false);
```
AlexMWells, line 79: add comment to the `false `parameter identifying the parameter name "treat_string_as_ustring_hash"?

src/liboslexec/opclosure.cpp
```diff
@@ -86,7 +86,7 @@ osl_closure_to_ustringhash(ShaderGlobals* sg, ClosureColor* c)
     // Special case for printing closures
     std::ostringstream stream;
     stream.imbue(std::locale::classic());  // force C locale
-    print_closure(stream, c, &sg->context->shadingsys());
+    print_closure(stream, c, &sg->context->shadingsys(), true);
```
AlexMWells, line 89: add comment to the `true` parameter identifying the parameter name "treat_string_as_ustring_hash"?

src/liboslexec/wide/wide_opdictionary.cpp
```diff
@@ -145,12 +145,12 @@ __OSL_MASKED_OP(dict_next)(void* bsg_, void* wout, void* wNodeID,
...
 OSL_BATCHOP int
-__OSL_OP(dict_value)(void* bsg_, int nodeID, void* attribname, long long type,
-                     void* data)
+__OSL_OP(dict_value)(void* bsg_, int nodeID, ustring_pod attribname,
+                     long long type, void* data)
 {
     auto* bsg = reinterpret_cast<BatchedShaderGlobals*>(bsg_);
     return bsg->uniform.context->dict_value(nodeID, USTR(attribname),
-                                            TYPEDESC(type), data);
+                                            TYPEDESC(type), data, false);
```
AlexMWells, line 153: Add comment on `false `argument identifying the name of parameter `/* treat_ustrings_as_hash */`

src/liboslexec/wide/wide_opdictionary.cpp
```diff
@@ -169,7 +169,7 @@ template<typename ValueT> struct DictValueGetter<ValueT, false> {
 
             ValueT value;
             int result = context->dict_value(nodeID, attribname, wdest.type(),
-                                             &value);
+                                             &value, false);
```
AlexMWells, line 172: Add comment on `false `argument identifying the name of parameter `/* treat_ustrings_as_hash */`

src/liboslexec/wide/wide_opdictionary.cpp
```diff
@@ -191,7 +191,7 @@ template<typename ValueT> struct DictValueGetter<ValueT, true> {
             ustring attribname = wAttribName[lane];
             ElementType value[dest_array.length()];
             int result = context->dict_value(nodeID, attribname, wdest.type(),
-                                             &value[0]);
+                                             &value[0], false);
```
AlexMWells, line 194: Add comment on `false `argument identifying the name of parameter `/* treat_ustrings_as_hash */`

AlexMWells (approved):
Left a couple cleanup comments, but looks good to go (as is).

src/liboslexec/batched_llvm_gen.cpp
steenax86: llvm_gen_split  in 

> llvm_gen.cpp

 still uses
`   else {
        args[2] = rop.ll.constant("");
    }`
and I will change it to `rop.ll.wide_constant(ustring("").c_str());`

It might make sense to remove  
`llvm::Value* LLVM_Util::constant(string_view s) { return constant(ustring(s)); }` 
after the batched side is through with ustringhash phase3 conversion.

src/testrender/cuda/wrapper.cu
```diff
@@ -131,7 +130,10 @@ process_closure(const OSL::ClosureColor* closure_tree)
             const char* mem = (const char*)((OSL::ClosureComponent*)cur)->data();
             const char* dist_str = *(const char**)&mem[0];
 
-            if (HDSTR(dist_str) == STRING_PARAMS(default))
+            OSL::ustring dist_u(dist_str);
```
brechtvl, line 133: Build error with OptiX enabled:
```
cuda/wrapper.cu(133): error: calling a __host__ function("OpenImageIO_v2_6_0::ustring::ustring(const char *)") from a __device__ function("process_closure") is not allowed
```

src/testshade/optixgridrender.cpp
```diff
@@ -272,15 +272,15 @@ OptixGridRenderer::synch_attributes()
         const ustring* cpuString
             = (const ustring*)(colorSys
                                + (cpuDataSize
-                                  - sizeof(StringParam) * numStrings));
+                                  - sizeof(ustringhash) * numStrings));
         CUdeviceptr gpuStrings = d_color_system + podDataSize;
         for (const ustring* end = cpuString + numStrings; cpuString < end;
              ++cpuString) {
             // convert the ustring to a device string
             uint64_t devStr = cpuString->hash();
```
brechtvl, line 280: OptiX tests are crashing here with an invalid `cpuString` pointer.

src/liboslexec/string_hash.h
```diff
@@ -1,33 +0,0 @@
-// Copyright Contributors to the Open Shading Language project.
```
brechtvl, line 1: This removed file is still included from `oslexec_pvt.h`:
```
#ifdef __CUDACC__
#    include "string_hash.h"
#endif
```

brechtvl (changes requested):

src/testrender/cuda/wrapper.cu
AlexMWells: @steenax86, I think that dist_str should already be a ustringhash already because it was previously just being cast to HDSTR:
`if (HDSTR(dist_str) == STRING_PARAMS(default))`
So you should just be able to change line 131 from
`const char* dist_str = *(const char**)&mem[0];`
to
 `OSL::ustringhash dist_uh = *(OSL::ustringhash*)&mem[0];`
and then get ride of lines 133 and 134.

src/testshade/optixgridrender.cpp
AlexMWells: @steenax86 , as this is the "cpuString" side I things, the
`sizeof(StringParam)`
would map to a ustring vs. ustringhash.
We see later on line 280 the hashcode is extracted from it:
`uint64_t devStr = cpuString->hash();`

So try changing it to 
`sizeof(ustring)`

lgritz:
This needs a rebase and fix any merge conflicts with the several Brecht patches that just went in.

brechtvl:
OptiX tests are now crashing with this:
```
src/liboslexec/llvm_util.cpp:3032: make_function: Assertion 'llvm::isa<llvm::Function>(maybe_func)' failed: Declaration for osl_texture is wrong, LLVM had to make a cast
```

AlexMWells:
@brechtvl , PR has been updated with fixes.  If you could please retest, we think its read to go.

brechtvl:
This passes the OptiX tests now for me (ignoring some tests that fail the same way in main).

lgritz:
Yes, passes for me, too. 

I'm working on some additional changes that will make this less of a stark source code break for renderers, though.

AlexMWells:
@lgritz this PR includes a lot of cleanup.  Makes things explicit (as far as ustrings and hashes go).  And yes we did make lots of changes to remove confusion, why have multiple ways of doing things.  Although that means more changes to user code to update to it.  But minimizing the delta at least in the OSL code base wasn't really a goal per say, but trying to be explicit what was passed to who was.  And removing as many casts between char * and int64_t (hash code) as possible including llvm JIT.

Now, if adding back a deprecated DeviceStrings or equivalent makes Renderer transition easier, then maybe that should happen.  Although I think we imagined a renderer might just define those itself vs. adding them back in OSL proper.  But either way would work.

Specifics:
llvm_load_string used to at one point be a customization point for CUDA so it could go off to some global variable and actually load a value, after switching to hash_codes, nothing was loaded anymore.  So the cleanup was to rename the remaining variations to what they actually do which is to emit a const hash.  Much more clear when reading/debugging the code (we didn't make the change by accident).  This shouldn't have impact on renderers though.  And yes it could be renamed back to llvm_load_string and there would be fewer diffs.

So if you still want to be able to switch back and forth between ustrings and ustringhash'es then this isn't that PR.  We went out of our way to rip out the old way and what would be come unused/confusing as the interfaces.

And yes you could add back into OSL
```
typedef ustringhash StringParam;
typedef ustringhash DeviceString;
typedef ustringhash ustringrep;
#define STRING_PARAMS(unquoted_string) OSL_HASHIFY(unquoted_string)
```
and reduce the textual differences in this PR.  Which might make back patching easier.  Although we started with a goal ripping off the band aid and just removing ustringrep and being explicit everywhere.

AlexMWells:
NOTE:  if a renderer chooses to use STRING_PARAMS(mystring), then it is responsible for instantiating a ustring of "mystring" on the host someplace so the reverse lookup of the hashed mystring exists in the ustring table.  This is partly why we recommend a rs_strdecls.h approach so the text of the string exists in one place that is #included to produce hashcodes on device and again to populate ustrings on the host.

lgritz:
My current thinking is to 

1. Branch a dev-1.13 prior to this patch and use that as the binary-compatible basis for the 1.13 release family.
2. Merge this as-is into the new main (aka future 1.14), but instruct renderer authors not to sync with it quite yet, because it's currently rough sailing to switch a renderer.
3. Submit a separate PR afterwards that contains a fairly minimal set of additional changes to make the renderer side easier and more stageable, as well as to establish a 'ustringrep' type alias that more clearly distinguishes between the things that are irrevocably ustring hashes, and those that are "the representation for strings in shader execution" which we all know today, for most of us, is the same as a ustringhash, but makes it easier for a renderer to diverge and use a different representation.

lgritz (approved):

PR #1733, opened 2023-09-28
Fix tests passing when test commands fail

brechtvl:
## Description

37cbab5b accidentally changed tests to accept failing commands by default.

This can cause tests to pass even though the program crashed, when a previous run created the correct output file.

This reveals existing asan and lsan errors. The asan error was fixed. The memory leaks remain ignored, as the exact cause is unclear.

## Tests

N/A

## Checklist:

### DIFF
--- a/src/build-scripts/ci-startup.bash
+++ b/src/build-scripts/ci-startup.bash
@@ -22,7 +22,7 @@ export OSL_ROOT=$PWD/dist
 export DYLD_LIBRARY_PATH=$OSL_ROOT/lib:$DYLD_LIBRARY_PATH
 export LD_LIBRARY_PATH=$OSL_ROOT/lib:$LD_LIBRARY_PATH
 export OIIO_LIBRARY_PATH=$OSL_ROOT/lib:${OIIO_LIBRARY_PATH}
-export LSAN_OPTIONS=suppressions=$PWD/src/build-scripts/nosanitize.txt:exitcode=0
+export LSAN_OPTIONS=suppressions=$PWD/src/build-scripts/nosanitize.txt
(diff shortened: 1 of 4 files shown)


brechtvl:
I made it pass tests even when there are memory leaks. Better would be to fix these (if possible), but I don't really have the time for that. There is not a lot to go on.
```
=================================================================
==17951==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 140 byte(s) in 1 object(s) allocated from:
    #0 0x510ab7 in operator new(unsigned long) /llvm-project-llvmorg-13.0.1/compiler-rt/lib/asan/asan_new_delete.cpp:95:3
    #1 0x7f2828947cd8 in std::string::_Rep::_S_create(unsigned long, unsigned long, std::allocator<char> const&) (/lib64/libstdc++.so.6+0xbdcd8)

Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #0 0x510d67 in operator new(unsigned long, std::nothrow_t const&) /llvm-project-llvmorg-13.0.1/compiler-rt/lib/asan/asan_new_delete.cpp:101:3
    #1 0x7f28288e6f1d in __cxa_thread_atexit (/lib64/libstdc++.so.6+0x5cf1d)

SUMMARY: AddressSanitizer: 164 byte(s) leaked in 2 allocation(s).

```

brechtvl:
It was masking all memory leaks. I made it more specific now, though it's still pretty broad.

lgritz (approved):
LGTM

Apologies on the "failureok = 1" mistake -- that was me, probably something I had temporarily done while testing, and then forgot to restore.

PR #1734, opened 2023-09-29
build(deps): Raise OpenEXR min to 2.4

lgritz:
Only for OSL 1.13. Will not be backported to the current release.


lgritz:
Seem ok to everybody?

fpsunflower (approved):

PR #1736, opened 2023-10-02
docs: First stab at ReadTheDocs

lgritz:
Convert the LaTeX docs to Sphinx + Myst Markdown extension

I think this is in decent shape. It's going to need a lot of thorough proofreading to ensure that thing got lost or mangled on the way from the original docs. I suspect we will have several follow-up PRs to fix things.

After we're reasonably happy with it, we'll eventually remove the .tex and the pdf.


lgritz:
Preview of the staged docs: https://open-shading-language.readthedocs.io/en/latest/

fpsunflower:
This is looking great already!

lgritz:
Is there anything additional I should do before a first merge?  (Knowing that there will probably be lots of little fixes and enhancements I find later and will follow up with additional PRs as needed.)

lgritz:
If there are no objections surfacing, I'm going to merge this today, then we can move on to more careful proofreading and smaller patches to fix anything we see that needs to be fixed.

fpsunflower:
I didn't read through the text in detail, but the overall structure looks good to me.

We should definitely handle the next steps in follow up commits, this is already a great start.

fpsunflower (approved):

PR #1737, opened 2023-10-02
build: Bump Cuda minimum to 9.0

lgritz:
Based on Brecht's tests, we realize that OSL doesn't work at all with Cuda 8.0 (which didn't support the C++14 that we use), so we're raising the minimum to 9.0. And there are some flaky tests with 9 and 10, so we're making the recommended minimum be 11.0.

It's not our policy to change the minimum versions of dependencies on release branches, but I think we should backport this to the release branch. I believe what this change is doing is not removing actual support for old versions, but rather, correctly documenting what has been working (or not) for quite some time.

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -212,9 +212,9 @@ if (OSL_USE_OPTIX)
 
         checked_find_package (CUDA REQUIRED
                              VERSION_MIN 9.0
-                             RECOMMENDED_MIN 11.0
-                             RECOMMENDED_MIN_REASON
-                                "We don't actively test Cuda older than 11"
+                             RECOMMEND_MIN 11.0
+                             RECOMMEND_MIN_REASON
+                                "We don't actively test CUDA older than 11"
                              PRINT CUDA_INCLUDES)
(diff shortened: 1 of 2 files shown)


src/cmake/externalpackages.cmake
```diff
@@ -211,8 +211,11 @@ if (OSL_USE_OPTIX)
         endif ()
 
         checked_find_package (CUDA REQUIRED
-                            VERSION_MIN 8.0
-                            PRINT CUDA_INCLUDES)
+                             VERSION_MIN 9.0
+                             RECOMMENDED_MIN 11.0
+                             RECOMMENDED_MIN_REASON
+                                "We don't actively test Cuda older than 11"
```
brechtvl, line 217:
```suggestion
                             RECOMMEND_MIN 11.0
                             RECOMMEND_MIN_REASON
                                "We don't actively test CUDA older than 11"
```

Wrong argument name.

brechtvl (changes requested):

src/cmake/externalpackages.cmake
lgritz: D'oh! you're right. Fixed.

brechtvl (approved):

PR #1738, opened 2023-10-02
ci: Fix GPU job not actually enabling OptiX

brechtvl:
## Description

In #1668 USE_OPTIX changed to OSL_USE_OPTIX, but CI was not updated.

And fix CPU SIMD flags being used for CUDA compilation, making the build fail.

## Tests

N/A

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,7 +158,7 @@ jobs:
             pybind11_ver: v2.10.0
             simd: avx2,f16c
             skip_tests: 1
-            setenvs: export OSL_CMAKE_FLAGS="-DUSE_OPTIX=1" OPTIX_VERSION=7.0
+            setenvs: export OSL_CMAKE_FLAGS="-DOSL_USE_OPTIX=1" OPTIX_VERSION=7.0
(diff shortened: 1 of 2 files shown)


lgritz (approved):

lgritz (changes requested):
Oof, something is broken here

AlexMWells:
```
 Found CUDA 11.8 
 ...
clang-15: warning: CUDA version is newer than the latest supported version 11.5 [-Wunknown-cuda-version]

```
So does the clang version need to be paired up with CUDA version?  Meaning CUDA 11.5 is newest that could be used when using clang-15?
And similar pairings might be true of different clang versions?

brechtvl:
The GPU job is failing because CPU flags `-mavx2 -mf16c` are part of the command for building CUDA bitcode, which in turns makes the Imath `half.h` header try to use CPU intrinsics.

brechtvl:
That seems to have fixed it. Other CI failures are unrelated.

> So does the clang version need to be paired up with CUDA version? Meaning CUDA 11.5 is newest that could be used when using clang-15?
> And similar pairings might be true of different clang versions?

CUDA 11.8 + LLVM 15 works fine for me despite the warning. But in general, I guess so.

lgritz (approved):
LGTM

PR #1743, opened 2023-10-20
ci: For clarity, use the term 'runner' instead of 'os'

lgritz:


lgritz:
this seem ok?

lgritz:
No objections were voiced to this innocuous CI-only change, so I'm going to merge.

PR #1745, opened 2023-10-26
fix: SymOverrideInfo bitfields should be the same type

lgritz:
Stephen Friedman alerted me to the fact that on Windows, bit fields within a struct will only be combined if they are the same type.  We were a bit sloppy in the defintion of SymOverrideInfo, and that was causing the struct to end up a different size in Windows.  I think the solution is that all those bit fields should be `unsigned int`.


AlexMWells (approved):
Good catch!  Maybe leave comment above that to the effect "all bit field data members need to be the same type to share the same space"

AlexMWells (approved):

sfriedmapixar:
Thank you!

PR #1746, opened 2023-10-26
ci: Disable MacOS-11 CI

lgritz:
The MacOS 11 CI seems to be broken on GitHub's runners. I'm not sure why, but since it's old and getting various deprecation warnings from Homebrew and GH, I'm inclined to just disable it rather than work hard to keep this old version of the OS running.


lgritz:
Preemptively merging because this undeniably unbreaks our CI.

PR #1747, opened 2023-10-27
docs: Regenerate PDF docs before 1.13 release

lgritz:
This is probably the last time we'll regen the PDF docs, since we are mid-transition to ReadTheDocs. We will probably not directly distribute a PDF version with 1.14 or subsequent releases.

Bump the version and regularize a few of the copyright notices to the current recommended presentation format.


lgritz:
Merging without review -- this changes no code or content, just rebuilds the PDF document for release to incorporate changes we've already made to the docs over the prior months.

PR #1748, opened 2023-10-28
ci: Test ABI compliance

lgritz:
This patch adds a new test to the CI matrix: verifying that we do not break the ABI during releases. This is based on similar work done a couple months ago in OpenImageIO.

Basically, in addition to building the current checkout, it also builds a checkout of the designated "ABI standard" (generally, the first non-beta release of a major version), then generates ABI dumps of both versions and runs the ABI compliance checker using the tools from https://github.com/lvc

Making two builds sounds expensive, but both builds are rigged to be minimal, since only the libraries are needed for ABI checks (we can turn off the binary tools, unit tests, python bindings, etc.), and of course the test suite need not run. So actually, it's less expensive than most of the test matrix entries, and it helps ensure that we don't accidentally break ABI when we don't mean to.

I'm not sure how many false positives we'll have. We might need to adjust the methodology slightly as we gain experience with this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -236,6 +236,30 @@ jobs:
             setenvs: export BUILDTARGET=clang-format
                             OPENIMAGEIO_CMAKE_FLAGS=-DUSE_PYTHON=0
 
+          # Test ABI stability. `abi_check` is the version or commit that we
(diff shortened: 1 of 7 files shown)


lgritz:
No objections -> merging

PR #1760, opened 2023-12-27
ci: test against LLVM 17

lgritz:
Bump the "bleeding edge" test to LLVM 17, the "stable latest releases" test to 16.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -380,7 +380,7 @@ jobs:
             setenvs: export LLVM_VERSION=10.0.0
                             OPENIMAGEIO_CMAKE_FLAGS="-DBUILD_FMT_VERSION=7.0.1"
                             PUGIXML_VERSION=v1.10
-          - desc: latest releases gcc11/C++17 llvm15 boost1.71 exr3.2 py3.9 avx2 batch-b16avx512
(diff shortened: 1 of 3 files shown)


lgritz:
Does this seem ok?

AlexMWells:
So which clang versions will CI be exercising now?

lgritz:
Clang or LLVM?

With this test, I believe our CI will test against llvm (for JIT) with every major version from 9 to 17, inclusive. (That actually feels like unnecessarily many, maybe it's time to raise the minimum for future OSL 1.14+ releases?)

But our CI only builds with a few versions of clang: 9, 13, and 14 (on Linux), and 15 (on Mac). Most of our build matrix uses gcc, plus the icc & icx tests. It would be unwieldy to expand the CI tests to make every combination test against every compiler, so we kind of sample. Historically, we have rarely or never found much sensitivity to building against one or another version of clang, as long as they both support the same level of C++ standard.

The one exception is that sometimes newer versions monotonically add new warnings, so we should always be doing one build with the very latest clang to make sure we compile cleanly with no warnings. I'll amend this PR to make sure one build uses the latest clang 17 to ensure we have full coverage there.

lgritz:
Amended: Test clang 17 compilation with MacOS-13 build

brechtvl (approved):

PR #1763, opened 2024-01-11
Fix NVPTX TargetMachine leak

tgrant-nv:
## Description

This PR includes fixes for three issues:

1. A leak of the `llvm::TargetMachine` used for NVPTX.
2. ~~A segfault that can occur in `testrender` when the renderer is destroyed before the  `ShadingSystem`.~~
3. A compilation failure when using CUDA Toolkit 12.0 or newer due to references to the long-deprecated (and now removed) `textureReference` API still being present in the LLVM headers.

These changes aren't related, but I've bundled them up for convenience. I can break these into individual PRs if that would be preferable.

## Tests

No new tests, but all previously passing tests still pass.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -585,6 +585,9 @@ LLVM_Util::~LLVM_Util()
     delete m_new_pass_manager;
     delete m_builder;
     delete m_llvm_debug_builder;
+    if (m_nvptx_target_machine != nullptr) {
+        delete m_nvptx_target_machine;
+    }
     module(NULL);
     // DO NOT delete m_llvm_jitmm;  // just the dummy wrapper around the real MM
 }
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -428,6 +428,7 @@ LLVM_Util::LLVM_Util(const PerThreadInfo& per_thread_info, int debuglevel,
     , m_llvm_func_passes(NULL)
     , m_new_pass_manager(NULL)
     , m_llvm_exec(NULL)
+    , m_nvptx_target_machine(nullptr)
     , m_vector_width(vector_width)
     , m_llvm_type_native_mask(nullptr)
     , mVTuneNotifier(nullptr)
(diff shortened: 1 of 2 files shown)


tgrant-nv:
@lgritz, is this (in particular, the leak fix) something that we should get in for the 1.13 release?

src/liboslexec/llvm_util.cpp
```diff
@@ -585,6 +586,9 @@ LLVM_Util::~LLVM_Util()
     delete m_new_pass_manager;
     delete m_builder;
     delete m_llvm_debug_builder;
+    if (m_nvptx_target_machine != nullptr) {
+        delete m_nvptx_target_machine;
+    }
```
lgritz, line 591: I believe that this can be

```
    delete m_nvptx_target_machine;
```

C++ semantics say that delete itself checks that the pointer is non-null so you don't need to enclose it in yet another nullptr check.

src/testrender/testrender.cpp
```diff
@@ -368,7 +368,7 @@ main(int argc, const char* argv[])
 
     // We're done with the shading system now, destroy it
     rend->clear();
-    delete rend;
     delete shadingsys;
+    delete rend;
```
lgritz, line 372: The sanitizer CI tests are failing for this PR, and I wonder if it's related to this deletion reversal here?

tgrant-nv: When the renderer is being destroyed it calls back into the shading system, deleting the shading system before the renderer is definitely erroneous. I don't see anything in the CI output that sheds much light on the failure.

src/liboslexec/llvm_util.cpp
tgrant-nv: Thanks, I deleted the null check.

src/testrender/testrender.cpp
lgritz: Hmmm, I'll try to repro on my end.

tgrant-nv: Ah, I just managed to get the full log and I can see the errors reported in the `ShadingSystem` destructor. I can drop this change from the PR and then I can investigate these issues in isolation.

lgritz (approved):
LGTM, thanks.

PR #1764, opened 2024-01-18
fix(build): Remove unused static ustring

lgritz:
A recent change in OIIO master made some of the ustring methods "trivial", and somehow that makes this variable warn about being unused, which it did not do before.


AlexMWells (approved):
LGTM

PR #1765, opened 2024-01-26
Add API for building interpolated getter free functions.

Uedaki:
## Description

Adds a new `RendererServices` API `build_interpolated_getter` which allows for building custom free functions that provide values for interpolated datas.

The new API is involved at material compilation time and allow the developers to provide specialized functions leveraging the information known at compile time. It will allow for many optimization opportunities for the compiler by replacing runtime branch with direct memory reads.

This PR is a followup to the API build_attribute_getter proposed in this PR: https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1704

## Tests

TestShade provides an example implementation showing how this compile time information can be used to select an appropriate function to use as the attribute provider, and how to configure the function signature.
Since the feature is enable by default when `rs_bitcode` is used, the existing testsuite ensure the feature is generating the same result as the previous API.

## Checklist:

  being nonconforming.


src/liboslexec/llvm_instance.cpp
```diff
@@ -320,9 +320,9 @@ BackendLLVM::llvm_type_groupdata()
         TypeDesc* types = &group().m_userdata_types[0];
         int* offsets    = &group().m_userdata_offsets[0];
         int sz          = (nuserdata + 3) & (~3);
-        fields.push_back(ll.type_array(ll.type_bool(), sz));
+        fields.push_back(ll.type_array(ll.type_int8(), sz));
```
AlexMWells, line 323: Can you talk to changing the type of the 
`bool GroupData::userdata_init_flags;`
to
`int8 GroupData::userdata_init_flags;`
?
Is it to ensure the size of the GroupData structure is consistent across platforms?

src/liboslexec/llvm_instance.cpp
```diff
@@ -638,6 +638,99 @@ BackendLLVM::llvm_assign_initial_value(const Symbol& sym, bool force)
...
                 ll.op_memset(ll.offset_ptr(dstptr, size), 0, 2 * size);
+        } else if (renderer()->supports("build_interpolated_getter")) {
+            InterpolatedGetterSpec spec;
+            renderer()->build_interpolated_getter(group(), symname, type,
+                                                  sym.has_derivs(), spec);
+            if (!spec.function_name().empty()) {
+                std::vector<llvm::Value*> args;
+                args.reserve(spec.arg_count() + 1);
+                for (size_t index = 0; index < spec.arg_count(); index++) {
+                    const auto& arg = spec.arg(index);
```
AlexMWells, line 649: Even though the code flow says it, maybe add a comment here explaining that the argument could be dynamic value identified by an enum or constant value that can directly be emitted.

src/liboslexec/llvm_instance.cpp
```diff
@@ -638,6 +638,99 @@ BackendLLVM::llvm_assign_initial_value(const Symbol& sym, bool force)
...
+                        case InterpolatedSpecBuiltinArg::ParamName:
+                            args.push_back(llvm_const_hash(symname));
+                            break;
+                        }
+                    } else {
+                        append_constant_arg(*this, arg, args);
+                    }
+                }
+                args.push_back(
+                    ll.void_ptr(groupdata_field_ptr(2 + userdata_index)));
```
AlexMWells, line 674: Add comment identifying which field your are extracting

src/testshade/simplerend.cpp
```diff
@@ -673,6 +675,65 @@ SimpleRenderer::build_attribute_getter(
...
+               && type == OIIO::TypeDesc::TypeFloat) {
+        spec.set(rs_get_interpolated_blue,
+                 InterpolatedSpecBuiltinArg::OpaqueExecutionContext,
+                 InterpolatedSpecBuiltinArg::Derivatives);
+    } else if (param_name == RS::Hashes::test
+               && type == OIIO::TypeDesc::TypeFloat) {
+        spec.set(rs_get_interpolated_test);
+    } else {
+        spec.set(rs_get_interpolated_generic,
+                 InterpolatedSpecBuiltinArg::OpaqueExecutionContext,
```
AlexMWells, line 728: Since we always know param_name, type, and derivatives at compile time, do we need a fallback at all?
Either we should be able to emit a specific call or not?

src/testshade/rs_simplerend.cpp
```diff
@@ -366,6 +366,93 @@ rs_get_attribute(OSL::OpaqueExecContextPtr oec, OSL::ustringhash_pod object_,
...
+OSL_RSOP bool
+rs_get_interpolated_generic(void* sg_, OSL::ustringhash_pod _name,
+                            OSL::TypeDesc_pod _type, bool derivatives,
+                            void* val)
+{
+    auto name                = OSL::ustringhash_from(_name);
+    const OSL::TypeDesc type = OSL::TypeDesc_from(_type);
+
+    OSL::ShaderGlobals* sg = (OSL::ShaderGlobals*)sg_;
+    return sg->renderer->get_userdata(derivatives, name, type, sg, val);
```
AlexMWells, line 453: To be device compatible, we should break encapsulation of the first parameter` OSL::OpaqueExecContextPtr` by casting it to `ShaderGlobals`.  In particular we can't make calls to a `RendererServices` virtual function on the device side.
So what ever functionality you want from `sg->renderer->get_userdata` should be placed right here with any required supporting data placed inside struct `RenderState `located `testshade/render_state.h`

The idea being if you are using free functions version, you wouldn't even need to implement RendererServices::get_userdata, because rs_get_interpolated* has replaced it.

src/testshade/rs_simplerend.cpp
```diff
@@ -366,6 +366,93 @@ rs_get_attribute(OSL::OpaqueExecContextPtr oec, OSL::ustringhash_pod object_,
...
+OSL_RSOP bool
+rs_get_interpolated_test(void* val)
+{
+    printf("rs_get_interpolated_test\n");
+    ((float*)val)[0] = 0.42f;
+    return true;
+}
+
+OSL_RSOP bool
+rs_get_interpolated_generic(void* sg_, OSL::ustringhash_pod _name,
```
AlexMWells, line 445: `void* sg_`
should be 
`OSL::OpaqueExecContextPtr ec`

AlexMWells:
@Uedaki , overall this looks great!  There are just some encapsulation/device compatibility issues in the testshade rs_get_interpolated_* functions to clear up and a few more comments.

AlexMWells:
@lgritz , can you comment on the CI /VFX2023 abi check failing?  Is that because a new virtual function was added to RendererServices?  Do we need to bump a version number or something to indicate the ABI might be changing?

AlexMWells:
From the CI build artifacts:
<html>
<body>
<span class="h_name" style="color: rgb(204, 51, 0); font-size: 0.875em; font-weight: bold; font-family: Arial, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">rendererservices.h</span><br style="color: rgb(0, 0, 0); font-family: Arial, sans-serif; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="ns" style="color: rgb(64, 128, 128); font-size: 0.94em; font-family: Arial, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">namespace<span> </span><b>OSL_v1_14_0</b></span><br style="color: rgb(0, 0, 0); font-family: Arial, sans-serif; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="section" onclick="javascript:showContent(this, 'c_4')" style="font-weight: bold; cursor: pointer; color: rgb(0, 62, 105); white-space: nowrap; margin-left: 0.3125em; font-family: Arial, sans-serif; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="ext" style="font-weight: normal;">[−]</span><span> </span><span class="ttype" style="font-weight: normal;">class</span><span> </span>RendererServices<span> </span><span class="failed" style="background-color: rgb(255, 204, 204); font-weight: normal;"> 1 </span></span><span style="color: rgb(0, 0, 0); font-family: Arial, sans-serif; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;"></span><br style="color: rgb(0, 0, 0); font-family: Arial, sans-serif; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><div id="c_4" style="color: rgb(0, 0, 0); font-family: Arial, sans-serif; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: block; visibility: visible;">
  | Change | Effect
-- | -- | --
Virtual method build_interpolated_getter ( ShaderGroup const&, OpenImageIO_v2_4::ustring const&, struct OpenImageIO_v2_4::TypeDesc, bool, InterpolatedGetterSpec& ) has been added to this class. | The layout of v-table has been changed. Call of any virtual method at higher position in this class or its subclasses may result in crash or incorrect behavior of applications.

<span class="sect_info" onclick="javascript:showContent(this, 'c_5')" style="cursor: pointer; padding-left: 1.55em; font-size: 0.875em; color: black;">[+] show v-table (old and new)</span><br><span class="sect_aff" onclick="javascript:showContent(this, 'c_6')" style="cursor: pointer; padding-left: 1.55em; font-size: 0.875em; color: rgb(204, 51, 0);">[+] affected symbols: 21 (3.7%)</span><br></div>
</body>
</html>

So perhaps the new virtual method needs to be moved after all the other virtual methods, try that.  Then on ABI breaking version changes, we could rearrange.  Add comment why its at the end.

src/liboslexec/llvm_instance.cpp
Uedaki: `userdata_init_flags` is a variable used as a `char` in `osl_bind_interpolated_param` which can take 3 different values: not initialized (0), not found (1), and initialized (2).
Boolean values are treated as 2-state variables with llvm operators. Due to this limitation, it was impossible to implement the `osl_bind_interpolated_param` functionality in `build_interpolated_getter` with llvm operators. Cases 1 and 2 were treated the same way with the bool array.
To solve this problem, I changed the array type from bool to int8. This solution doesn't change the logic, since `userdata_init_flags` was already used as a `char` array.
That way I could implement `osl_bind_interpolated_param` with llvm operators for the `build_interpolated_getter` functionality.

AlexMWells: @Uedaki , excellent explanation.  Can you add a comment that it is representing 3 states and list them.
Follow on, should it just go to a vanilla int vs int8?  Do we gain much by using 1 byte vs 4 bytes, not sure if alignment rules would cause the following variable to start on 4 byte boundary anyway.

src/liboslexec/llvm_instance.cpp
lgritz: I'd rather keep the virtual functions in a good logical order when reading the header. Appending to the end is a bit of a suspect way to avoiding ABI break, I don't think it's guaranteed.

We're allowed to break ABI compatibility in main. We just need to adjust the version numbers to reflect it. Which we can do in a subsequent commit (we already need a subsequent commit in order to update the reference commit for ABI compatibility, so this is not a problem).

lgritz: I think I put this comment in the wrong place!

lgritz:
> So perhaps the new virtual method needs to be moved after all the other virtual methods, try that. Then on ABI breaking version changes, we could rearrange. Add comment why it's at the end.



I'd rather keep the virtual functions in a good logical order when reading the header. Appending to the end is a bit of a suspect way to avoiding ABI break, I don't think it's guaranteed.

We're allowed to break ABI compatibility in main. We just need to adjust the version numbers to reflect it. Which we can do in a subsequent commit (we already need a subsequent commit in order to update the reference commit for ABI compatibility, so this is not a problem).

Unless you need this backported to 1.13, in which case we should do this (even if not guaranteed) to try to preserve ABI compatibility in practice as best we can.

AlexMWells:
> We're allowed to break ABI compatibility in main. We just need to adjust the version numbers to reflect it. Which we can do in a subsequent commit (we already need a subsequent commit in order to update the reference commit for ABI compatibility, so this is not a problem).
> 
@lgritz (asking as I think I might be missing something here)  Why wouldn't we just require this PR update the version number, as it will need to to pass CI before we merge right?

lgritz:
> @lgritz (asking as I think I might be missing something here) Why wouldn't we just require this PR update the version number, as it will need to to pass CI before we merge right?

Bumping the version is what we do when we have to break the ABI, yes!

But that alone won't make it pass the automated test that compares the ABI to a designated prior commit (named by its hash) that is the "reference" ABI we are trying to match. So the sequence we follow is:

1. Merge the fix that breaks the ABI. It is at this point expected to fail the ABI CI check, and that's ok because we know why it's failing and are going to immediately fix it.
2. Change the ABI check to make the "reference ABI" name the commit from step 1, and merge that as a separate fix. At this point, the automated ABI check will pass again.

You can't combine these into just one step, because the actual merge of PR (1) will result in a different commit hash than when the PR was submitted. There's no way to know what the new ABI reference commit's hash will be post-merge, so you simply cannot include it in step 1.

If you can think of a better solution to this, I'm all ears. I haven't been able to think of an easy way to say "make THIS commit, whatever its hash may end up being, be the new ABI reference going forward."

AlexMWells (approved):
LGTM, real test will be enabling in production renderer to see if all scenarios where covered.

PR #1773, opened 2024-02-22
Initial support for LLVM-18

MehdiChinoune:
## Description

Adds primary support for LLVM-18
I couldn't fix all issues because it needs a better understanding of the code.

## Checklist:

- [ ] My code follows the prevailing code style of this project. If I haven't
  being nonconforming.


lgritz (approved):
LGTM (We don't have a test for this yet because it's not a released LLVM version. I'm assuming that you tested it against this version on your end.)

MehdiChinoune:
> LGTM (We don't have a test for this yet because it's not a released LLVM version. I'm assuming that you tested it against this version on your end.)

As I had mentioned above It doesn't fix it completely as It needs more knowledge of code.

PR #1774, opened 2024-02-28
Mute partio error prints

olegul:
## Description

This patch mutes the error printed to std::err from inside Partio when it fails to read a pointcloud.  OSL's pointcloud_search and pointcloud_find already reports errors "the proper way" via errorfmt, so a failed read will still thrown an error. Plain prints like this can sneak past log-limiting filtering  and result in very large log files.

Currently, running something like

shader partioerror(){
    float test[1];
    if(pointcloud_search("non-existing",P,1.0,1,"test",test))
        printf("Success");
}

in testshade, will print

Partio: No extension detected in filename
ERROR: pointcloud_search: could not open "non-existing"

with the patch, it will just print
ERROR: pointcloud_search: could not open "non-existing"

## Tests

I haven't added any tests yet

## Checklist:

- [X ] I have ensured that the change is tested somewhere in the testsuite (adding new test cases if necessary).
- [X ] My code follows the prevailing code style of this project. If I haven't
  being nonconforming.

### DIFF
--- a/src/liboslexec/pointcloud.cpp
+++ b/src/liboslexec/pointcloud.cpp
@@ -3,6 +3,7 @@
 // https://github.com/AcademySoftwareFoundation/OpenShadingLanguage
 
 #include <cstdarg>
+#include <sstream>
 
 #include "pointcloud.h"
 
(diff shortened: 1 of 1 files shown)


AlexMWells (approved):
LGTM

lgritz:
@olegul You need to click the right things to add yourself to the CI. You're already approved on the SPI end, but you need to accept it.

PR #1775, opened 2024-02-28
ci: Increase timeout for slow MacOS-13 runner

lgritz:
We didn't change anything on our end, but in the last few days, I believe the GHA MacOS-13 runners have slowed done somehow, and so now a few test are failing only because they don't complete within the ctest timeout period for "small tests". So this just bumps up the limit on this runner platform to allow the tests to pass.


lgritz:
Fixes CI, merging

PR #1776, opened 2024-03-02
PR/FixCompRefWithVaryingIndex

AlexMWells:
## Description
Fix bug in batched codegen of compref when the index is varying.
The llvm_load_value was left to defaulted to uniform load, which we can't store into a varying/wide result. Fixed similar bug in getmessage, and added check for spline knotcount to be uniform (although varying support could be added).

This bug exposed a missed optimization in BatchedAnalysis where a early out (return) needs to change the loop conditional to be varying to support masking.  However no check was being done to see if the loop control existed higher in the callstack.  So now the execution scope tracks loop nesting depth to see if the current function is really in a loop or not.  This allows the loop control flow to remain uniform (avoiding masking and more complex control flow).

## Tests
Added nestloop-reg to testsuite to reproduce bug and verify its fixed.

## Checklist:

  being nonconforming.

### DIFF
--- a/testsuite/nestedloop-reg/test.osl
+++ b/testsuite/nestedloop-reg/test.osl
@@ -1,3 +1,6 @@
+// Copyright Contributors to the Open Shading Language project.
+// SPDX-License-Identifier: BSD-3-Clause
+// https://github.com/AcademySoftwareFoundation/OpenShadingLanguage
 
 float limit_early_return(float channel) {
     if (channel > 1.0) {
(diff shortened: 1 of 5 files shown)


testsuite/nestedloop-reg/test.osl
```diff
@@ -0,0 +1,36 @@
+
```
lgritz, line 1:
```suggestion
// Copyright Contributors to the Open Shading Language project.
// SPDX-License-Identifier: BSD-3-Clause
// https://github.com/AcademySoftwareFoundation/OpenShadingLanguage

```

lgritz: Every year LF runs an automated license scan and sends me a report of all the files that don't clearly reveal licenses. Better to add them from the start than for me to have to go find them months later when I get the report.

lgritz (approved):
LGTM, thanks for the incredibly fast fix!

PR #1778, opened 2024-03-05
build: Remove a lot of old SPI build cruft

lgritz:
Sweep away some SPI-specific build customization that hasn't been relevant for a long time.


sfriedmapixar (approved):
LGTM

PR #1780, opened 2024-03-07
ci: Fix broken Macos-13 CI

lgritz:
Some noodling was necessary to fix recently broken Mac CI.


lgritz:
Merging, as it undeniably fixes the problem.

PR #1782, opened 2024-03-07
Make isconnected() work with downstream renderer "connections."

sfriedmapixar:
## Description

The isconnected() returns 2 when used on an output that is connected to another downstream shader node -- this can be used to limit computation to only the outputs used.  However, that misses the important case of where the renderer has declared a shader node output as a renderer output, which happens often when using OSL purely for pattern generation purposes.  This change makes that case also return 2.

## Tests

The isconnected test has been updated to include testing this functionality.

## Checklist:

  being nonconforming.


lgritz (approved):
Sorry for the delay. LGTM.

PR #1787, opened 2024-03-22
oslquery: Simplify include and link needs

lgritz:
The (should be) tiny oslquery library was over-including and over-linking beyond its actual needs.

* Remove`#include <OSL/oslconfig.h>`, which is a lot of setup, type definitions, and many transitive header includes.  Instead, just include the very small number of headers and types actually needed by the OSLQuery class definition.
* In particular, this removes includes of OpenEXR/Imath headers and references to Imath types, the OIIO TextureSystem (!), and other big things, from what oslquery.h needs to see.
* liboslquery can link against libOpenImageIO_Util (which it uses primarily for ustring and TypeDesc), it doesn't need the full libOpenImageIO, since it doesn't read or write image files.
* Remove some link references to Boost which isn't actually needed for this library.
* This also lightens the load for the Python module for oslquery.

### DIFF
--- a/src/include/OSL/oslquery.h
+++ b/src/include/OSL/oslquery.h
@@ -19,9 +19,38 @@ Sony Pictures Imageworks terms, above.
 #include <string>
 #include <vector>
 
-#include <OSL/oslconfig.h>
+#include <OSL/export.h>
+#include <OSL/oslversion.h>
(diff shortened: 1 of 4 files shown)


chellmuth (approved):
LGTM!

PR #1803, opened 2024-04-28
Add full LLVM 18 support

lgritz:
Some methods and enums got renamed from LLVM 17 -> 18, etc.

Also some adjustment to warning messages about version mixing between clang and llvm.


lgritz:
There are some CI failures that are unrelated to this patch, I'm addressing those separately.

It's the Macos-14 CI that is building against LLVM 18.1 to test this.

Macos-13 would be too, but it is failing for entirely different reasons -- because building OIIO is failing with clang 18. That's already fixed and this failure will go away when we push out a new "release" branch on May 1.

fpsunflower (approved):

PR #1808, opened 2024-04-30
Remove remaining traces of boost

fpsunflower:
## Description

Replace the last few uses of boost with robin-map (which is already a dependency for OIIO, although it is new for OSL itself) and functions in OIIO.

## Tests

Testsuite passes on my mac laptop. Will keep an eye on how the CI steps go.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/doc/languagespec.tex
+++ b/src/doc/languagespec.tex
@@ -384,7 +384,6 @@ other open source packages:
 \url{http://openimageio.org}
 \item Imath \copyright\ Contributors to Imath.
 \url{http://www.openexr.com}
-\item Boost \copyright\ various authors.  \url{http://www.boost.org}
 \item LLVM \copyright\ 2003-2010 University of Illinois at
   Urbana-Champaign. \url{http://llvm.org}
 \end{itemize}
--- a/src/doc/languagespec.tex
+++ b/src/doc/languagespec.tex
@@ -182,7 +182,7 @@ Unported License. \\
 \smallskip
 \spc \includegraphics[width=0.85in]{Figures/CC-BY.png}
 \spc http://creativecommons.org/licenses/by/4.0/
- \bigskip 
+ \bigskip
 
 
 
@@ -318,7 +318,7 @@ find are different in OSL compared to other languages:
 
   OSL shaders are evaluated in SIMD fashion, executing shaders on many
   points at once, but there is no need to burden shader writers with
-  declaring which variables need to be uniform or varying.  
+  declaring which variables need to be uniform or varying.
 
   In the open source OSL implementation, this is done both automatically
   and dynamically, meaning that a variable can switch back and forth
@@ -331,7 +331,7 @@ find are different in OSL compared to other languages:
   and use arbitrary quantities as texture coordinates and expect correct
   filtering.  This does not require that shaded points be arranged in a
   rectangular grid, or have any particular connectivity, or that any
-  "extra points" be shaded.  
+  "extra points" be shaded.
 
   In the open source OSL implementation, this is possible because
   derivatives are not computed by finite differences with neighboring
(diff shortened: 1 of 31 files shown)


src/doc/languagespec.tex
```diff
@@ -182,7 +182,7 @@
 \smallskip
 \spc \includegraphics[width=0.85in]{Figures/CC-BY.png}
 \spc http://creativecommons.org/licenses/by/4.0/
- \bigskip
```
lgritz, line 185: Can you please revert languagespec.tex? It's just making whitespace changes all throughout.

src/liboslexec/CMakeLists.txt
```diff
@@ -62,13 +62,13 @@ set ( liboslexec_target_srcs
    ../liboslnoise/wide/wide_gabor_hybrid_disabled
    ../liboslnoise/wide/wide_gabor_hybrid_enabled
    ../liboslnoise/wide/wide_gabor_isotropic_disabled
-   ../liboslnoise/wide/wide_gabor_isotropic_enabled
```
lgritz, line 65: Can you revert all the whitespace changes here? I don't mind fixing that eventually, but I'd rather do it as a separate PR and do the git blame trick since nothing really changed.

lgritz:
I would recommend copying OIIO's FindRobinmap.cmake and the parts of externalpackages.cmake that deal with robinmap, just exactly copy all of that because it works.

src/doc/languagespec.tex
fpsunflower: Will do - it looks like my editor is trimming whitespace on the whole file instead of on edited lines.

src/liboslexec/CMakeLists.txt
fpsunflower: Fixed.

fpsunflower:
> I would recommend copying OIIO's FindRobinmap.cmake and the parts of externalpackages.cmake that deal with robinmap, just exactly copy all of that because it works.

I gave a quick try to `FetchContent` since I was reading about it earlier and this is a particularly "easy" case with a header only dependency that has a modern cmake build system with exported targets. It seems to work as advertised, and will automatically do a `find_package` first to check if you already have the library on your system before executing `FetchContent`. 

It doesn't seem to work properly on the CI machines though -- and the error message is not helpful at all. Is there a trick to accessing the internet from the CI machines perhaps?

Of course I'm happy to clone the existing setup from OIIO if this fails, but the amount of boilerplate is definitely reduced with this approach.

lgritz:
There is no trick, the CI runners have direct internet access.

lgritz:
I would recommend just copying the old way OIIO did it, it has worked perfectly for a long time.

When I finish perfecting the new OIIO build system, we'll port it over here, too.

At that point, it will be easy to experiment with FetchContent if we want, and if that works any better or simpler instead of my build_dependency_with_cmake approach, I'm happy to switch.

fpsunflower:
Will do.

fpsunflower:
I believe this is working now. There are still some failures on OSX but they appear to be unrelated.

fpsunflower:
I think partio was broken because it was previously pulling in zlib implictly via boost. It seems a bit weird that the cmake setup for partio doesn't handle this for us, but perhaps it isn't fully correct on their end?

Adding zlib to the target_link_libraries fixes it locally for me (was building without partio previously).

lgritz:
> I believe this is working now. There are still some failures on OSX but they appear to be unrelated.

That's unrelated. I expected it to be fixed after the OIIO release a couple days ago. I'll need to look into it separately, but it's definitely not at all related to your PR.

lgritz (approved):
LGTM!

PR #1811, opened 2024-05-05
docs: fix outdated or wrong repo URLs

lgritz:
* Fix references to old OIIO home
* Fix embarrassing release notes generating script botch that accidentally used OpenImageIO URLs where OSL URLs were needed.


lgritz:
No objections + docs-only change = merge

PR #1815, opened 2024-05-09
Add lazytrace option to avoid combinatoric optixTrace inlining

chellmuth:
OptiX inlines all calls to optixTrace at module compilation time. Due to OSL's lazy layer evaluation, if layer A calls layer B M times, and layer B calls layer C N times, that can lead to M*N inlines of optixTrace if layer C contains a trace operation.

In practice we've observed single trace ops being inlined hundreds of times, leading to minutes-long shader compilations.

This patch adds a new option, lazytrace, to run layers with trace ops unconditionally at the start of shader evaluation. This costs some potential performance in cases where the trace layer would never be evaluated, but removes all the compilation penalties that the inlining was presenting.

## Tests

Added a new test, lazytrace, that checks via printf to make sure the non-lazy execution ordering rules were correctly followed when lazytrace=0.

## Checklist:

  being nonconforming.


src/liboslexec/llvm_gen.cpp
```diff
@@ -187,6 +187,13 @@ BackendLLVM::llvm_run_connected_layers(Symbol& sym, int symindex, int opnum,
         const Connection& con(inst()->connection(c));
         // If the connection gives a value to this param
         if (con.dst.param == symindex) {
+            // Non-lazy layers are run upfront directly via llvm_call_layer.
+            // Eliding them here doesn't change the semantics of execution,
+            // but it will prevent optixTrace calls from being repeatedly
+            // inlined when lazytrace=0.
+            if (!group()[con.srclayer]->run_lazily())
+                continue;
```
chellmuth, line 195: This is the only part of the PR that affects the status quo lazytrace=1 path. I could narrow it to only execute when lazytrace=0, but it seems better to always skip the run-layer check on unconditionally-executed layers.

lgritz (approved):
LGTM

PR #1816, opened 2024-05-10
testing: Update ref output for OptiX execution order variance

lgritz:
Was spuriously failing tests only because on GPU, the points didn't always execute in the same order.

(This just adds an additional reference output with same results, but different point order.)


lgritz:
@chellmuth I think this is the last of the OptiX test failures I was getting that was holding up my release for you.

chellmuth (approved):

PR #1818, opened 2024-05-26
build: Default to C++17 mode

lgritz:
This does not change the minimum C++ standard, which is still C++14, even in main. But it does change the default in main to build in C++17 mode, and to get 14 you have to say so explicitly.

We do anticipate bumping the minimum before this fall's release, and this is just a half step to get us ready. Maybe it will help shake loose any compatibility problems and remind people on main that they are using a branch that in a few months will have a new minimum C++.

This will not be backported to the release.

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -25,7 +25,7 @@ message (STATUS  "CMAKE_CXX_COMPILER_ID  = ${CMAKE_CXX_COMPILER_ID}")
 ###########################################################################
 # C++ language standard
 #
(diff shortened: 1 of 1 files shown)


lgritz:
No objections -> merging.

Remember, this doesn't change the minimum (yet), it just changes the default C++ standard to 17. 

You can still build in C++14 mode with -DCMAKE_CXX_STANDARD=14, but beware of dependencies that we use that themselves require 17 at this point. You may need to also use old dependencies to use a compiler that old.

PR #1819, opened 2024-06-01
Implement OpenPBR's sheen BRDF

fpsunflower:
## Description

This PR adds an implementation of the Zeltner-Burley sheen closure which is the new default in OpenPBR. I've exposed it via a new (optional) mode keyword arg in the existing closure, though it would be easy to switch the interface if needed.

The implementation is based on a fit done by Stephen Hill here: https://blog.selfshadow.com/sheen/ltc_sheen.html

## Tests

I've modified the existing sheen render unit tests to use the new closure on half of the sphere so we can keep testing old and new modes.

## Checklist:

  being nonconforming.


src/testrender/shading.cpp
```diff
@@ -1446,6 +1452,114 @@ struct MxSheen final : public BSDF, MxSheenParams {
...
+        const Vec3 T1 = (V - N * NdotV).normalize();
+        const Vec3 T2 = N.cross(T1);
+        const Vec3 L  = Vec3(wi.x * T1 + wi.y * T2 + wi.z * N);
+
+        // TODO: is there a cheaper way to get the jacobian here?
+        const Vec3 wiOriginal(aInv * wi.x + bInv * wi.z, aInv * wi.y, wi.z);
+        const float len2     = dot(wiOriginal, wiOriginal);
+        const float det      = aInv * aInv;
+        const float jacobian = det / (len2 * len2);
+        pdf = jacobian * std::max(wiOriginal.z, 0.0f) * float(M_1_PI);
```
aconty, line 1524: Is there a reason not to call eval(wo, wi) here?

fpsunflower: It just seemed redundant to transform/untransform the vector and recompute the ltc factors several times. Since its just a few lines of code, it seemed easier to just do the small amount of work again.

selfshadow (commented):
I think it would be better to reference the MaterialX implementation for now*, since the WebGL sandbox was just a test and may go away or move.

* I'm planning to publish something about the fitting methodology later once I've settled on a final implementation, at which point this can be the canonical reference.

src/testrender/shading.cpp
```diff
@@ -1446,6 +1452,113 @@ struct MxSheen final : public BSDF, MxSheenParams {
...
+#if USE_LTC_SAMPLING == 1
+        const Vec3 V = wo;
+        const float NdotV = clamp(N.dot(V), 0.0f, 1.0f);
+        const Vec3 ltc = fetch_ltc(NdotV);
+        const float aInv = ltc.x, bInv = ltc.y, R = ltc.z;
+        Vec3 wi;
+        float pdf;
+        Sampling::sample_cosine_hemisphere(Vec3(0, 0, 1), rx, ry, wi, pdf);
+
+        wi = Vec3(wi.x - wi.z * bInv, wi.y, wi.z * aInv).normalize();
```
selfshadow, line 1513: Nice idea of scaling up by aInv here. I'll see if I can borrow that for the MaterialX implementation.

src/testrender/shading.cpp
```diff
@@ -1446,6 +1452,114 @@ struct MxSheen final : public BSDF, MxSheenParams {
...
+
+        const Vec3 T1 = (V - N * NdotV).normalize();
+        const Vec3 T2 = N.cross(T1);
+        const Vec3 L  = Vec3(wi.x * T1 + wi.y * T2 + wi.z * N);
+
+        // TODO: is there a cheaper way to get the jacobian here?
+        const Vec3 wiOriginal(aInv * wi.x + bInv * wi.z, aInv * wi.y, wi.z);
+        const float len2     = dot(wiOriginal, wiOriginal);
+        const float det      = aInv * aInv;
```
selfshadow, line 1522: Here's what I'm doing, using the length of L / wi instead (before normalisation): https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_sheen.glsl#L169-L183

fpsunflower: Nice - I'll go through this and try to consolidate the logic.  Didn't realize yours was added to the MaterialX side already.

selfshadow: I think if you retain the `aInv` multiplication of `wi` but otherwise follow what I'm doing then you'll end up with `square(len2 / aInv)` for the jacobian, which should still be a little bit cheaper than what you have here.

NB: I may end up changing to fitting to work with the non-inverse matrix coefficients (TBD, but it's possible that they could be more linear), but that shouldn't change the implementation all that much.

src/testrender/shading.cpp
```diff
@@ -1446,6 +1452,113 @@ struct MxSheen final : public BSDF, MxSheenParams {
...
+        return Color3(fetch_ltc(NdotV).z);
+    }
+
+    Sample eval(const Vec3& wo, const Vec3& wi) const override
+    {
+        const Vec3 L = wi, V = wo;
+        const float NdotV = clamp(N.dot(V), 0.0f, 1.0f);
+        const Vec3 ltc = fetch_ltc(NdotV);
+
+        const Vec3 T1 = (V - N * NdotV).normalize();
```
selfshadow: There's a danger of generating NaNs here if V = N (I've certainly hit this in other LTC-related code in production). I didn't bother to handle that in the WebGL sandbox, but I'm doing so in the MaterialX version:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_sheen.glsl#L117-L133

Things get a little simpler if vectors are already in local space:
https://github.com/shill-lucasfilm/OpenPBR-viewer/blob/main/glsl/fuzz_brdf.glsl#L30-L37

fpsunflower:
I've made the suggested changes. I also fixed the TangentFrame class so it could be used here and replace the ad-hoc one (which wasn't robust as you pointed out).

This did invalidate all the reference images with small noise differences, but I think its worth the cleanup.

fpsunflower:
@jstone-lucasfilm Do you have any preference for the keyword argument used to pick the new sheen variant?

As I've implemented it so far, using the new sheen would look like this:

```
Ci = sheen_bsdf (N, SheenColor, Roughness "mode", 1);
```

testsuite/render-mx-sheen/sheen.osl
```diff
@@ -16,7 +16,7 @@ sheen
   )
 {
     if (N.x < 0)
-        Ci = sheen_bsdf (N, Cs, roughness * 2, "mode", 1);
+        Ci = sheen_bsdf (N, Cs, sqrt(roughness), "mode", 1);
```
selfshadow, line 19: Are you sure this is correct? I thought the external parameterisation was roughness (alpha in the Zeltner implementation) and that only the underlying SGGX model uses sigma. The table itself (and my analytic refit) should be parameterised by alpha, not sigma.

src/testrender/shading.cpp
```diff
@@ -1537,23 +1537,16 @@ struct ZeltnerBurleySheen final : public BSDF, MxSheenParams {
...
-                  / (0.046391f + y);
-        float o = y * (0.000654023f + (-0.0207818f + 0.119681f * y) * y)
-                  / (1.26264f + y * (-1.92021f + y));
-        float q = (x - m) / s;
-        float R = expf(-0.5f * q * q) / (s * float(sqrt(2.0 * M_PI))) + o;
+        const float x = NdotV;
+        const float y = roughness;
+        const float A = ((2.58126f * x + 0.813703f * y) * y) / (1.0f + 0.310327f * x * x + 2.60994f * x * y);
+        const float B = sqrtf(1.0f - x) * (y - 1.0f) * y * y * y / (0.0000254053f + 1.71228f * x - 1.71506f * x * y + 1.34174f * y * y);
+        const float invs = (0.0379424f + y * (1.32227f + y)) / (y * (0.0206607f + 1.58491f * y));
```
selfshadow: Good idea! I had originally tried to fit 1/s directly but it made the optimisation unstable. I don't know why I didn't follow through on doing it as a post step but I'll bring this over to a future  MaterialX GLSL version, assuming I don't replace the Gaussian fitting (I might try a logistic).

src/testrender/sampling.h
```diff
@@ -14,18 +14,28 @@ OSL_NAMESPACE_ENTER
...
     }
 
     // build frame from unit normal and unit tangent
-    TangentFrame(const Vec3& n, const Vec3& t) : w(n)
-    {
-        v = w.cross(t);
-        u = v.cross(w);
+    // fallsback to an arbitrary basis if the tangent is 0 or colinear with n
+    static TangentFrame from_normal_and_tangent(const Vec3& n, const Vec3& t) {
+        Vec3 y = n.cross(t);
```
selfshadow: Out of curiosity, was there a reason you went with computing y here rather than x (via t - n*dot(n, t))? Numerical stability or performance?

fpsunflower: No real reason actually. Now that I look at both in the light of day, they should be equivalent, and the dot product way might be a few less operations (especially considering the dot product is needed for the LTC lookup anyway).

testsuite/render-mx-sheen/sheen.osl
fpsunflower: When looking at the wedges of both sheen models, the sqrt seemed to bring them closer together. This part is just for the testsuite where its handy to see both models at once on the same sphere.

The materialx closure just exposes the "raw" parameter in any case (just like GGX).

src/testrender/sampling.h
selfshadow: Okay, fair enough. I just wanted to check that I wasn't missing anything.

testsuite/render-mx-sheen/sheen.osl
selfshadow: Okay, makes sense.

jstone-lucasfilm:
@fpsunflower @shill-lucasfilm I've had a chance to write a pull request for the `mode` input of `sheen_bsdf` in MaterialX, and I wanted to post a link here to compare notes between these two implementations:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/1880

Following additional discussion with Stephen, our current thinking is that the two modes should be named `conty_kulla` and `zeltner`.  Since the Zeltner paper has three authors (Zeltner, Burley, and Chiang), it seemed incorrect to mention just two of them in the sheen option, so we're instead leaning towards `zeltner` as a parallel construction to "Zeltner, et al".

Let me know how this sounds to you, and I'm very open to making adjustments based on feedback from this group.

fpsunflower:
Sounds good to me.

fpsunflower:
Finally got the testsuite passing. Just needed a handful of extra reference images and had to clamp the roughness to a small value to avoid generating NaNs for the R term.

I will squash and merge since I don't believe there are any outstanding comments.

The two remaining failures also happen in main and are udim related -- I am guessing that is coming from the latest OIIO.

src/testrender/shading.cpp
```diff
@@ -1529,7 +1529,7 @@ struct ZeltnerBurleySheen final : public BSDF, MxSheenParams {
         // for the implementation in MaterialX:
         // https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_sheen.glsl
         const float x = NdotV;
-        const float y = roughness;
+        const float y = std::max(roughness, 1e-3f);
```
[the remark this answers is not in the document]
selfshadow: Oops, I left this comment as pending earlier. Oh well!

PR #1820, opened 2024-06-01
docs: Fixes to documentation of new oren_nayar_diffuse_bsdf parameters

lgritz:
* The new parameter was explained, but not added to the function prototype in the docs.
* The semi-deprecated languagespec.tex was fixed, but it slipped our minds that we've shifted the docs to markdown for ReadTheDocs (the real place we should have changed was stdlib.md).


fpsunflower (approved):
LGTM!

PR #1822, opened 2024-06-05
build: Make finding bison work better on Apple Silicon Macs + homebrew

lgritz:
Homebrew changes location by default on the new ARM-based Macs. Check the new paths, too.


lgritz:
Lee gave a thumb, I'm considering that an approval given that nobody else objected.

PR #1823, opened 2024-06-06
Fix the subsurface_bssrdf parameters

fpsunflower:
## Description

We intended this closure to take the effective color and radius so that the SSS color remapping can be controlled by the renderer.

## Tests

The implementation in testrender uses a simple diffuse response (which was effectively assuming it got a multiple-scattered albedo. So all that was needed was renaming the parameters and updating docs.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/doc/languagespec.tex
+++ b/src/doc/languagespec.tex
@@ -5049,27 +5049,19 @@ Constructs a BSSRDF for subsurface scattering within a homogeneous medium.
   
   \apiitem{albedo}
   \vspace{12pt}
-  Single-scattering albedo of the medium.
+  Effective albedo of the medium (after multiple scattering). The renderer is expected to
+  invert this color to derive the appropriate single-scattering albedo that will produce
+  this color for the average random walk.
   \apiend
   \vspace{-16pt}
   
-  \apiitem{transmission_depth}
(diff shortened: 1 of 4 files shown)


fpsunflower:
@jstone-lucasfilm @niklasharrysson this is for addressing the issue you brought up on slack. Let me know if the documentation looks ok.

@lgritz The documentation exists in 3 places. I'm guessing the latex docs are deprecated, but is there anyway for the new md docs to use the header file somehow?

lgritz:
The src/doc/*.md (particularly stdlib.md) is the primary documentation. The LaTeX is obsolete, and exists now only for us to double-check that the new docs have faithfully reproduced what we have.

We don't yet have a way to generate docs from the headers, but it sure would be nice. (Open question: whether Doxygen can be coaxed into parsing osl headers so that we can use Sphinx + Breath + Doxygen like we do for the OIIO docs.)

lgritz (approved):
LGTM. Is this ready to go?

jstone-lucasfilm (approved):
Looks good to me!

niklasharrysson (approved):
Looks good to me too. Thanks @fpsunflower!

PR #1792, opened 2024-03-28
Use new OIIO ustringhash decode methods when available

chellmuth:
## Description

We noticed a few ustring-related performance regressions while converting Arnold to the new ustringhash pipeline.

These changes update relevant shadeops to avoid a ustringhash -> ustring conversion. The decode_wrapmode call into OIIO requires the latest master: https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4207/commits/6759d4d85dc6f41f16c56379d18daf4833663913

## Checklist:

- [x] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [x] I have updated the documentation, if applicable.
- [x] I have ensured that the change is tested somewhere in the testsuite (adding new test cases if necessary).
- [x] My code follows the prevailing code style of this project. If I haven't
  already run clang-format v17 before submitting, I definitely will look at
  the CI test that runs clang-format and fix anything that it highlights as
  being nonconforming.

PR #1824, opened 2024-06-07
Remove unnecessary ustring lookup

chellmuth:
## Description

I missed an unnecessary ustringhash to ustring conversion back in #1792. Looking up the interp value directly from the ustringhash fixes a performance regression caused by the conversion.

## Checklist:

  being nonconforming.


aconty (approved):
Good catch!

lgritz (approved):

PR #1826, opened 2024-06-07
Feature: Ability to sparsely gather scalar values and connect them to elements of a dynamic or statically sized array parameter

sfriedmapixar:
## Description

    This requires you to first declare a setting for the dynamic array input that
    will be used to size and provide values for any unconnected elements of the
    dynamically sized array.  After that, individual scalar->arrayelem connections
    can be specified from any number of different upstream shading nodes.  This
    facilitates gathering inputs from across your shading network, for example
    several procedurally generated color signals and using them to drive inputs
    to shading nodes that are best left dynamically sized, such as the inputs
    to spline operations.

## Tests

    This includes a new test in vararray-scalar-connect to demonstrate and
    test the new connection method.

## Checklist:

  being nonconforming.


johnhaddon:
This seems very useful. We currently achieve the same thing with shaders which take N scalar inputs and output arrays of length N, which isn't very elegant. Seems like the PR has been stale for a while - is there any chance of it being revived?

lgritz:
I'm not sure how this slipped through the cracks; apologies about that. I'm in favor of this generally. @sfriedmapixar, can you please rebase on the current main, and then I promise I will quickly review and merge it.

PR #1828, opened 2024-06-08
fix(batched): Assume BatchedRendererServices texture derivatives are in st space.

sfriedmapixar:
The convention in the single-point RendererServices is that the texture call returns derivatives in st space, and they are transformed to xy space before returning from the wrapper to RenderServices.  This change makes BatchedRendererServices follow the same convention.

This is still tested by testrender/testshade, but the code motion means the assumptions visible at the renderer/OSL boundary are now consistent between batched/non-batched renderservices.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/liboslexec/wide/wide_optexture.cpp
+++ b/src/liboslexec/wide/wide_optexture.cpp
@@ -117,14 +117,6 @@ default_texture(BatchedRendererServices* bsr, ustring filename,
             has_derivs ? (float*)&dresultds_simd : NULL,
             has_derivs ? (float*)&dresultdt_simd : NULL);
 
-        OIIO::simd::vfloat4 dresultdx_simd;
-        OIIO::simd::vfloat4 dresultdy_simd;
-        if (has_derivs) {
-            // Correct our st texture space gradients into xy-space gradients
-            dresultdx_simd = dresultds_simd * dsdx + dresultdt_simd * dtdx;
-            dresultdy_simd = dresultds_simd * dsdy + dresultdt_simd * dtdy;
-        }
-
         // NOTE: regardless of the value of "retVal" we will always copy over the texture system's results.
         // We are relying on the texture system properly filling in missing or fill colors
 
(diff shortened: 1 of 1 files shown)


src/liboslexec/wide/wide_optexture.cpp
```diff
@@ -561,6 +553,165 @@ dispatch_environment(BatchedRendererServices* bsr, ustring filename,
...
+static OSL_NOINLINE void
+transformWideTextureGradientsTexture3d(BatchedTextureOutputs& outputs,
+                                       Wide<const Vec3> Pdx,
+                                       Wide<const Vec3> Pdy,
+                                       Wide<const Vec3> Pdz)
+{
+    MaskedData resultRef = outputs.result();
+    bool has_derivs      = resultRef.has_derivs();
+    // Matching scalar code path behavior to check only result derivs.
+    ASSERT(has_derivs);
```
lgritz, line 626: I think this should be `OSL_ASSERT` if you want to check all the time, `OSL_DASSERT` if you only want to check for debug builds. I believe the `ASSERT` is just an old alias.

lgritz:
Yay for fixing to make the batch match the single point shading for texture derivs.

Seems to be failing a bunch of the CI. Do reference outputs need to be updated as part of this?

sfriedmapixar:
I thought this went through CI cleanly for me, so I'll investigate the diffs.

sfriedmapixar:
*sigh* This is why we have CI, and why you should always let your local CI finish before making a pull request.  I just had gotten interrupted and hadn't finished getting the changes for alpha and texture3d().  Now it's a complete change and passing CI.

src/liboslexec/wide/wide_optexture.cpp
sfriedmapixar: I went with OSL_DASSERT for the asserts I left -- in production if the types don't match because we start returning new types, those types will get mis-scaled derivatives, but shouldn't crash, and so I think the debug assert for the developer to notice is appropriate.

lgritz:
Sorry for the delay on this.

This look ok to everybody? @AlexMWells @fpsunflower ?

lgritz:
Is there a simple test we can add that would validate that this is (a) correct, and (b) matches between batch and single point?

fpsunflower:
This looks ok to me.  I assume the testsuite already has something that covers this since there were some CI failures before the patch was complete?

lgritz:
I think the reason this was wrong all along (for batch, but correct for single point) is that we *don't* have a test that ensures that the derivatives of the texture results are correct.

I've been fighting a lot of CI failures lately -- github changed some things, OIIO changed some things. Only recently that I've got everything passing again, so it's not at all surprising that CI was broken (for reasons unrelated to this PR) when it was first submitted but is ok now.

PR #1833, opened 2024-06-23
ci: Mods to CI to deal with OIIO master raising dependency mins

lgritz:


lgritz:
Merging proactively because this definitely fixes the broken CI.

PR #1839, opened 2024-07-07
ci: deal with CentOS 7 EOL and disappearance of yum mirrors

lgritz:
This was breaking CI for us. Hard break for icc/icx tests, since those REQUIRED a yum install of the intel compilers. Softer undetected break for all the ASWF <= 2022 containers based on CentOS 7, which were not failing outright but were failing to install certain optional packages.

The solution (for now) is to configure yum to exclude the missing repo.

### DIFF
--- a/src/build-scripts/gh-installdeps.bash
+++ b/src/build-scripts/gh-installdeps.bash
@@ -12,11 +12,18 @@ set -ex
 # Install system packages when those are acceptable for dependencies.
 #
 if [[ "$ASWF_ORG" != ""  ]] ; then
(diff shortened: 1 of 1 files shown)


steenax86:
Hi Larry, I looked at the CI logs from the failing CI actions, and I do not see a failed yum installation message with icx. Is the failure you observe to do with how Centos is now managing repos or is that a missing icx license/repo?

lgritz:
We were getting failed yum actions, but I have it fixed now.

lgritz:
There's a totally different failure I'm chasing now...

PR #1841, opened 2024-07-09
Fix reparameter string representation

chellmuth:
## Description

This PR fixes a regression that went unnoticed during the ustringhash conversion.

Right now OSL expects strings to be represented at runtime as ustringhashes, but interactive parameters' values are still being stored as ustrings. This requires fixing in two locations, first the code that sets the initial value of an interactive parameter, and second the code that copies updated parameters into the interactive buffer.

## Tests

New reparameter-strings test.

## Checklist:

  being nonconforming.


aconty (approved):

lgritz (approved):
LGTM2!

PR #1845, opened 2024-07-25
build: port set_utils and dependency_utils from oiio cmake scripts

lgritz:
For those who didn't see this get added in OIIO land a few months back, the highlights are:

* Variable setting utilities, including replacements/wrappers for set(CACHE) and option() that have a few extra bells and whistles, including that they allow environment variables of the same name to override the default values.

* An end-of-config dependency report that prints which dependencies were found and their versions, which were found but were too old, which were missing.

* The skeleton of an automatic local dependency build facility.  The short story is that if a dependency is not found or is too old AND it can find a file called src/cmake/build_PACKAGE.cmake, it will run that file and expect it to download and build the dependency on the fly. I haven't added any of those scripts here yet, but on the OIIO side we already have a few to build certain critical dependencies. Future PRs to OSL will also add the missing pieces to build the dependencies.


lgritz:
Here's what the "dependency report" looks like in the CI logs, which is immensely helpful:

```
-- =========================================================================
-- = Dependency report                                                     =
-- =========================================================================
-- 
-- The following dependencies were found externally:
--     BISON 3.0.4
--     FLEX 2.6.1
--     LLVM 15.0.7
--     OpenEXR 3.1.11
--     OpenImageIO 2.5.13.1
--     pugixml 1.11
--     Python 3.10.11
--     Qt5 5.15.9
--     ZLIB 1.2.11
-- The following dependencies were not found:
--     Qt6  
-- 
-- =========================================================================
-- 
```

lgritz:
As soon as this little organizational change is reviewed and merged, right on top of that I will have PRs that at long last bump our C++, Python, and Imath requirements.

lgritz:
No objections -> merging

PR #1850, opened 2024-08-18
Adjust to OIIO change to IC/TS API

lgritz:
Didn't catch them all in #1848 -- missed the BatchedRendererServices side.


lgritz:
Merging immediately as it demonstrably un-breaks the CI.

PR #1853, opened 2024-08-23
build(deps): Raise Imath minimum to 3.1

lgritz:
Cull a whole lot of things that were necessary only for older Imath/OpenEXR.


fpsunflower (approved):
Nice simplification!

PR #1855, opened 2024-08-25
build: Raise Python minimum to 3.7, pybind11 to 2.7

lgritz:
Drop Python 2.7 support in main.

Upgrade from very old cmake FindPython to FindPython3, update idioms.

Simplify a lot of cruft that was supporting python2, as well as supporting build features we don't use (long ago, I needed to support some special library naming conventions at SPI that we no longer use).

Change the module filename from "OSL" to "oslquery" to reflect that it's only covering oslquery and make room for other python bindings in the future.


lgritz:
No objections -> merge

PR #1869, opened 2024-09-21
Fix testrender OptiX build

tgrant-nv:
## Description
This PR restores build and run-time functionality for the OptiX path in `testrender` following #1865. It includes much of the host code refactoring from #1829. Having only one geometry type is a nice simplification.

## Tests
No new tests have been added, but I did update the reference images to reflect slight changes from the tessellated geometry. I also updated the expected output for `test_str_ops`, which has been broken since the change to string hashes.

## Checklist:

  being nonconforming.


fpsunflower (approved):
LGTM!

lgritz (approved):

lgritz:
Since our CI only builds OptiX but doesn't test it, I'll need to test it on my work machine (will try later today), and if that works, I will merge.

lgritz:
Passes my tests. But on the other hand, it looks like the testsuite isn't actually running the "render" tests in optix mode because those files don't have the "OPTIX" file tag within them. So this is certainly building, but it's not really being tested.

What's the relationship between this and #1829? Should this be merged first and then the other will be rebased and looked at again? Are there parts of the (much more extensive) description message body for 1829 that should be included here to reflect the refactoring that's present in both?

lgritz:
This doesn't actually work. Try:

    cd testsuite/render-microfacet
    testrender -r 320 240 -aa 8 -O2 --optix scene.xml out.exr

I just see solid cyan for the spheres

Is this expected to work with this PR? Or is this only expected to build and it's #1829 that will make it work?

tgrant-nv:
If you run `ctest -R testoptix`, it should run a handful of tests that exercise the OptiX path in `testrender`.

Yes, I'll need to more or less re-implement #1829 before I can re-open that PR. But that's probably for the best, since this PR takes care of a fair chunk of the host-side refactoring, so the rendering change will be a bit more comprehensible. There's nothing from the other description that needs to be restated here.

tgrant-nv:
None of the `render-*` test cases will currently work in OptiX mode. As things stand now, shading in OptiX is limited to returning the albedo color (at best). #1829 adds full path tracing and BSDF support, but that will probably have to wait until after Dev Days. It's a sprawling change.

lgritz:
OK, great, just wanted to be sure this PR had the functionality we expected. I'll merge!

PR #1873, opened 2024-09-24
build: Support for LLVM 19

### LINKED ISSUE #1872: LLVM 19 support
LLVM 19.1 was just released, and we don't build properly against it. Somebody is going to need to figure out what changed in LLVM's APIs and fix our side.

lgritz:
Fixes #1872


lgritz:
Pre-emptively merging, since it fixes CI on Mac, which due to a recent bump of homebrew's designated llvm version to 19, had broken prior to this fix.

PR #1875, opened 2024-09-27
Replace the tab key with 4 spaces in olstoy editor

rydrman:
## Description

Updates the osltoy application to use 4 spaces in the editor window instead of tab characters.

## Tests

I'm not sure that this change warrants a test case, but I did launch the application and ensure that the tab key does indeed insert 4 spaces into the exitor.

## Checklist:

  being nonconforming.


rydrman:
Sounds like ILM is going to handle the dev days CLAs all together in the coming days, but I'm aware that it needs to be done here still

src/osltoy/codeeditor.cpp
```diff
@@ -68,6 +68,25 @@ CodeEditor::CodeEditor(QWidget* parent, const std::string& filename)
...
+void
+CodeEditor::keyPressEvent(QKeyEvent* event)
+{
+    if (event->key() == Qt::Key_Tab) {
+        // replace the tab key with 4 spaces
+        // TODO: make this a configurable preference?
+        auto replacement
+            = QKeyEvent(QEvent::KeyPress, Qt::Key_Space,
+                        Qt::KeyboardModifiers(event->nativeModifiers()),
+                        "    ");
```
lgritz, line 80: Hmmm, in most editors, does TAB do the equivalent of hitting SPACE 4 times? Or does it do the number of spaces to take you to the next 4-space boundary?

rydrman: Good point, I'm not sure why I didn't think of that. I'll take another look when I have a moment

lgritz: Ping just in case you're interested in fixing this up.

Issue #1862, opened 2024-09-10
Meta-issue: osltoy features and fixes


lgritz:
# Master list of `osltoy` feature requests

`osltoy` is a small example program loosely inspired by [shadertoy](https://www.shadertoy.com/) -- a Qt application that lets you load or type in an OSL shader, which is then used to evaluate a color for each pixel in the display window of the app, and to interactively change the controls. Unlike the real Shadertoy, it's just meant to be a local app, not a web service. But it's a great way to tinker with OSL shaders, but it hasn't gotten a lot of development love lately.

We will use this issue to collect individual suggestions or feature requests for `osltoy`. Please comment/reply on this ticket to request that items be added (or if you're on the TSC or have permissions to edit this description, feel free to do so.

Comment/reply to say "hey, I want to do task X" so that multiple people don't independently attempt the same tasks (if it's something where only one can really be accepted); we don't want anybody's work time to be wasted.

## Good First Issues

These are small tasks ideal for newcomers to the project, suitable for Dev Days. They don't really require knowledge of how OSL itself is implemented under the covers, but some experience making C++ Qt GUI applications may be helpful, and it probably wouldn't hurt if you've fooled around a bit with the real Shadertoy to get inspiration for some things to add.

```[tasklist]
- [ ] way to set search path for headers?
- [ ] auto-detect output variable -- doesn't need to be Cout if there's only one
- [ ] editor improvement: can we make the tab stops 4 spaces each?
- [ ] watch a shader file on disk, auto reload/recompile when updated
```

## Bigger projects

These might require more extensive experience, planning complex features, or making big design decisions. Or maybe they will be fast? I'm only guessing.

```[tasklist]
- [ ] allow mouse and keyboard input -- somehow they could be fed into inputs to the shader on the next frame or something? Probably it would help to see how ShaderToy did it, for inspiration.
- [ ] allow "memory" -- can we turn any output into an input for the next frame?
```

agreen-creator:
Hello! I am going to be participating in dev days this thursday and friday. I would like to take on this task!
watch a shader file on disk, auto reload/recompile when updated

maxwelliverson:
I too am participating in dev days, and I'm going to be tackling the following task:

> way to set search path for headers?

rydrman:
hello, I'm going to be looking at this one for dev days!

> editor improvement: can we make the tab stops 4 spaces each?

silvialpz:
[...]

PR #1876, opened 2024-09-28
osltoy - Feature Request: custom #include search paths

maxwelliverson:
## Description

For Dev Days

#1862 

Addressing a osltoy feature request, I've added the ability to specify custom search paths. 
The changes consist of the following:

- Added a command line option `-I <path>` to specify search paths (`OIIO::ArgParse` unfortunately can't mirror `OSLCompiler`'s `-Ipath` syntax)
- OSLToyMainWindow now stores an array of `OSLCompiler` options alongside a flag indicating whether or not they should be regenerated before compilation. The only option currently able to be specified is `-Ipath`, but this should ease any future extension.
- Added a GUI component consiting of a window with a mutable list of the current set of search paths. This list is initially populated with any paths passed via the command line, but any modifications to the list will be reflected when the shader is next recompiled. Modifying this list does not currently force recompilation.
- Added an entry to the Tools menu to open the new GUI component.
- Added a line of output to the error window indicating what options the compiler was invoked with. I've defaulted to having it print for every run, but this may want to only print on failure.

## Tests

I did not add any automated tests for this feature, as is consistent with the rest of osltoy. I did, however, do some manual testing to make sure the functionality works as expected.

Shaders that failed to `#include` some file instead succeeded when the parent directory of that file was added to the list of search paths via both the command line and the GUI component. Respectively, when said search path is then removed via
the GUI component, any subsequent compilation attempts fail again. The GUI component otherwise behaves as expected.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/osltoy/osltoyapp.cpp
+++ b/src/osltoy/osltoyapp.cpp
@@ -711,6 +711,22 @@ OSLToyMainWindow::open_file(const std::string& filename)
 }
 
 
+void 
+OSLToyMainWindow::add_include_search_path(const std::string& path) 
+{
+    auto begin = m_include_search_paths.begin();
+    auto end = m_include_search_paths.end();
+    if (std::find(begin, end, path) == end) { 
+        m_include_search_paths.push_back(path);
+        m_should_regenerate_compile_options = true;
+
+        // Open question: Do we want to force a recompile whenever the list is updated?
+        //                For now this doesn't matter cause the list is only modified 
+        //                via command line arguments, but if we want to have a GUI component,
+        //                this is worth considering. 
+    }
+}
+
 
 void
 OSLToyMainWindow::action_saveas()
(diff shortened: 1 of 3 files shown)


src/osltoy/osltoyapp.cpp
```diff
@@ -351,6 +352,276 @@ class OSLToyRenderView final : public QLabel {
...
+
+        m_maxIndexWithContent
+            = (int)paths.size()
+              - 1;  // ok that this is -1 if the paths are empty
+
+
+
+        auto initialLineCount = required_lines();
+
+        m_lines.reserve(initialLineCount);
```
lgritz, line 471: As a style nit, we tend to only put blank lines within a function body where the extra whitespace is clarifying -- for example, if you have several code statements that are doing a well defined task A, followed immediately by several statements that do a separate task B, then a single blank line between them can help understanding of the conceptual grouping of the lines into tasks. But if the space doesn't especially clarify the algorithm, you may as well omit it and benefit from the function being compact. Also, even when whitespace makes it more clear, a single line suffices.

maxwelliverson: Gotcha! That makes perfect sense; I definitely have a bad habit of inserting extra line breaks while working on something for my own legibility and subsequently forgetting to remove them later on. I'll fix that now.

maxwelliverson:
Noting that when the automated test suite ran on Friday evening for this pull request, the icx build got caught in an infinite loop at some point while building the BVH. Not that it's particularly relevant to this, but I wanted to point it out as it appears to be another case of #1874 cropping up

lgritz:
The icx thing is a known problem, we're looking into it

PR #1880, opened 2024-10-04
Update osltoyrenderer.cpp: auto-detect output variable -- doesn't need to be Cout if there's only on

lunakhan:
auto-detect output variable -- doesn't need to be Cout if there's only one

## Checklist:

- [ ] My code follows the prevailing code style of this project. If I haven't
  being nonconforming.

### DIFF
--- a/src/osltoy/osltoyrenderer.cpp
+++ b/src/osltoy/osltoyrenderer.cpp
@@ -129,12 +129,31 @@ OSLToyRenderer::render_image()
         m_framebuffer.reset(
             OIIO::ImageSpec(m_xres, m_yres, 3, TypeDesc::FLOAT));
 
-    static ustring outputs[] = { ustring("Cout") };
+    std::vector<ustring> output_vars;
+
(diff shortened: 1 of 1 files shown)


src/osltoy/osltoyrenderer.cpp
```diff
@@ -129,15 +129,34 @@ OSLToyRenderer::render_image()
...
+
+    // If there's only one output variable, automatically use it
+    const ustring* outputs = &output_vars[0];
+    if (output_vars.size() == 1) {
+        outputs = &output_vars[0]; // auto-detect and use the single output
+    } else {
+        // If there are multiple outputs, you could implement additional logic here
+        // to choose between them or provide the user with a choice.
+        outputs = &output_vars[0]; // default to the first output
+    }
```
lgritz, line 150: I don't understand what's going on here. You set outputs to point to the first element of output_vars. OK. Then if that's the only item in output_vars, you... do it again? But if there are multiple items in output_vars, you... also do the same thing?

PR #1888, opened 2024-10-14
build(deps): Adjust to OIIO changes to TextureOpt structure

lgritz:
OIIO 3.0 changes the TextureOpt structure a bit, and we need to adjust on the OSL end, too.

To have context for this, you need to be aware of https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4485 and https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4490.

PR #1890, opened 2024-10-22
Reduce shared dependencies between liboslcomp and liboslexec

fpsunflower:
## Description

This PR reduces some of the code sharing going on between `liboslcomp` and `liboslexec`. In particular, we no longer need access to `OSLCompilerImpl` from `liboslexec` which lets us avoid pulling in `libclang` for this library, as well as avoids needing a copy of the `.osl` lexer/parser in `liboslexec`.

Most of the shared code centered around a few utility functions related to `TypeDesc`. I've extracted these as standalone methods in that class instead.

The only really substantial function is `track_variable_lifetimes` which is used both by `liboslcomp` and `liboslexec`. Luckily it is totally standalone, so I've now extracted it from its parent class and left it as a free function. Further refactors may enable us to move it to only one of the two libraries.

I discovered this while working on swapping out the preprocessor.

## Tests

Unit tests are passing on my machine.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/include/osl_pvt.h
+++ b/src/include/osl_pvt.h
@@ -98,6 +98,10 @@ public:
     /// is not responsible for freeing the characters.
     const char* c_str() const;
 
+    /// Return the c_str giving a human-readable name of a type, fully
+    /// accounting for exotic types like structs, etc.
+    const char* type_c_str() const;
+
     /// Stream output
     friend std::ostream& operator<<(std::ostream& o, const TypeSpec& t)
     {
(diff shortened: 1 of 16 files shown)


fpsunflower:
Build appears to be broken from the TextureOption changes -- but otherwise this is ready to review. I can rebase once #1888 is reviewed and merged.

lgritz:
I merged #1888, please try a rebase on current main and see how CI does.

fpsunflower:
It looks like everything is passing but the ABI check -- I'm not sure how to interpret the output on that one. Might just be coming from trimming the libraries?

fpsunflower:
I found the artifact download (keep forgetting how to find that one -- its rather well hidden).

The report on the ABI just shows the expected diff now that `OSLCompiler` is no longer included in `liboslexec`. What is the workflow to update the reference for this check?

lgritz:
> I found the artifact download (keep forgetting how to find that one -- its rather well hidden).

Yeah, only the "summary" lists the artifacts for some reason, it doesn't show you their download links if you click one of the test runs to look at the logs. 

> The report on the ABI just shows the expected diff now that `OSLCompiler` is no longer included in `liboslexec`. What is the workflow to update the reference for this check?

We merge this, that secures the new reference commit for the ABI we're trying to match, then I (or you) paste that commit hash into ci.yml where it lists a hash for "abi_check", as the only change for a subsequent commit (commit message, e.g., "Update ABI reference commit"). Push to your branch to ensure that the ABI check passes, then you can force push that to main (or if you really want, go through a PR).

lgritz (approved):
LGTM.

I'm not 100% sure I understand how there's not a linkage problem for apps that need both liboslcomp and liboslexec. Is it not a problem that both contain the code for TypeSpec, for example?

lgritz:
Feel free to merge it, you have the power. And then to subsequently update the ABI reference when you see what the post-merged hash of this is.

fpsunflower:
I think you are right that there is still a potential linkage issue here. The "right way" would be to make sure all the private symbols are not exported so they can't conflict even though they are duplicated across both libraries.

In any case, this is at least no worse than it was before.

PR #1891, opened 2024-10-23
Add hair_chiang_bsdf closure definition from MaterialX

fpsunflower:
## Description

This just adds a closure definition so the MaterialX project can work on their OSL backend bindings for this new BSDF.

## Tests

We don't have an implementation of this closure (nor would be it be easy to provide since testrender doesn't support hair fibers and this model won't work well on triangulated hair. We'll have to rely on MaterialX and renderers implementing this closure for testing.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/shaders/stdosl.h
+++ b/src/shaders/stdosl.h
@@ -607,6 +607,39 @@ closure color subsurface_bssrdf(normal N, color albedo, color radius, float anis
 closure color sheen_bsdf(normal N, color albedo, float roughness) BUILTIN;
 
 
+// Constructs a hair BSDF based on the Chiang hair shading model. This node does not support vertical layering.
+//  \param N                            Normal vector of the surface.
+//  \param curve_direction              Direction of the hair geometry.
+//  \param tint_R                       Color multiplier for the R-lobe.
+//  \param tint_TT                      Color multiplier for the TT-lobe.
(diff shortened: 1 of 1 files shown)


fpsunflower:
@jstone-lucasfilm for vis.

fpsunflower:
Related MaterialX issue for posterity: https://github.com/AcademySoftwareFoundation/MaterialX/pull/1968

src/shaders/stdosl.h
```diff
@@ -607,6 +607,39 @@ closure color subsurface_bssrdf(normal N, color albedo, color radius, float anis
...
+    float   ior,
+    float   longitudual_roughness_R,
+    float   longitudual_roughness_TT,
+    float   longitudual_roughness_TRT,
+    float   azimuthal_roughness_R,
+    float   azimuthal_roughness_TT,
+    float   azimuthal_roughness_TRT,
+    float   cuticle_angle,
+    color   absorption_coefficient
+);
```
lgritz, line 640: Should this have a `BUILTIN` tag here like the others?

fpsunflower: Yes indeed. Updated.

lgritz (approved):
LGTM

PR #1892, opened 2024-10-25
fix: Bugs in vector2.h, vector4.h, color2.h, color4.h, docs

### LINKED ISSUE #1889: min() is bugged in vector2.h
### Problem
The following two functions in the "vector2.h" header are bugged.

1. vector2 add(vector2 a, vector2 b)
2. vector2 add(vector2 a, float b)

The first has the bug, the second uses the first so gets bugged as well.
Looks like a copy/pate bug that just got left in there, trivial to fix (not bothering with a PR)

**Expected behavior:**  
The minimum value between the 2 arguments should be in the result, for both components

**Actual behavior:**  
The result contains just the first component of the first argument, and the second component of the second argument - regardless.

### Steps to Reproduce
Write a simple shader like that uses this function from the `vector2.h` header:
```
#include "vector2.h"

shader Texture(
    output color value = 0,
    vector2 a = {1, 0},
    vector2 b = {0, 1})
{
    vector2 c = min(a, b);
    value = color(c.x, c.y, 0.0);
}
```
Result should be black, not yellow.

OSL_min_bug_Blender_test

### Versions
all (I suppose)

lgritz:
Fix a variety of typos in the 2- and 4-component color and vector helper structs. Very embarrassing!

Also embarrassing, a typo in the docs had said that the way to overload the `!=` operator was with a function called `__operator__ne__` while actually, all along it had been `__operator__neq__`.

Comprehensive test of color2, color4, vector2, vector4.

New testing header: testsuite/common/shaders/osl-unittest.h This has utilities to make it easier to make unit tests for shader functions in our testsuite.

Various other minor fixes to these headers.

Fixes #1889


jstone-lucasfilm:
This looks like a great improvement, thanks @lgritz, though it seems worthwhile to consider whether the `color2` data type is still needed in the OSL codebase.

The `color2` data type originated in the MaterialX project (sorry!), but we were successfully convinced by the CG community that it was not a useful feature, and it was removed from the MaterialX codebase for the v1.38 release in 2021:

https://materialx.org/assets/MaterialX.v1.38.Changelist.pdf

> The "color2" and "color2array" types have been removed, since there hasn't proven to be any practical use for these types in 3D rendering applications which have become the primary focus of MaterialX.

I'm not aware of any teams that are still building projects with pre-2021 versions of MaterialX, and legacy MaterialX documents are automatically upgraded in modern versions, so we may want to consider whether the `color2` data type can now be safely removed from OSL as well.

lgritz:
Yes, I had to do it anyway because (a) I wanted to backport the fixes to the 1.12 branch, where we can't remove functionality within a supported release family, and also (b) because of the high symmetry between color2 and vector2, it wouldn't really have saved me any work to fix only one.

I think we should integrate this now, for the reasons stated above, but then remove the color2 from main before we beta 1.13.

I feel like my procrastination on the types I didn't like has been 25% successful so far. I guess I can't convince you of the error of color4, if not vector2/4? :-)

But seriously, I feel like OSL has to support these types to properly live in an ecosystem where MaterialX is important (which it is), but while my mild annoyance at vector2 and vector4 being built-in types is merely that they are unnecessary, I fear we are all going to regret color4 if spectral rendering ever catches on more widely.

jstone-lucasfilm:
Got it, and that approach makes sense to me.

On the value of the `color4` data type, it wouldn't be needed in MaterialX if we were only focused on processing data in linear color spaces, but we need to support applications that bind `color3` and `color4` values to arbitrary spaces, and manage the conversions of those values to the linear color space of the renderer.

That would be challenging without the ability to bind `color4` values to color spaces, and we'd be forcing applications to split all colors from their associated alpha values before passing them along to MaterialX and OpenUSD.

Although we'll learn more about the implications of spectral rendering as it becomes more popular in open-source renderers, it's my early sense that we won't associate spectral color values with alpha channels at all, since there's no history of combining these two concepts in rendering applications.  That would allow us to have a future `spectrum` data type that would be independent of both color spaces and associated alpha channels, and could coexist with all of the current data types in OSL, MaterialX, and OpenUSD.

It's worthwhile to consider how this would be integrated in OSL, and whether `spectrum` would simply be a supported behavior of `color`, or whether it would have its own data type.  In MaterialX, since spectral colors would not be associated with color spaces, it seems more logical for `spectrum` to be an independent data type, but we're very open to discussion on this.

lgritz:
Ha, that's where I was going with my thinking as well.

I agree that the way it will eventually shake out is that we'll need to introduce a `spectrum` type. And then `color` will finally mean only what the color scientists think it means -- a perceptual locus in a known color space, not a "pretend-spectrum". If `color` is definitively a color and not doubling as a spectrum, then I'm not nearly as annoyed by a type that combines a color and an alpha. Now I'm only annoyed by the fact that we don't yet have a spectrum type. :-)

testsuite/common/shaders/osl-unittest.h
```diff
@@ -0,0 +1,95 @@
...
+            printf("\nFAIL: %s:%d: %s\n\n", __FILE__, __LINE__, #x);    \
+            if (OSL_UNITTEST_EXIT_ON_FAILURE)                           \
+                exit();                                                 \
+        }                                                               \
+    } while (0)
+
+
+// Check that two expressions are equal. For non-built-in types, this
+// requires that a tostring() function is defined for that type.
+#define OSL_CHECK_EQUAL(x,y)                                                \
```
AlexMWells, line 84: @lgritz do we need some fuzzy comparison here, allowing for some amount of epsilon or other floating point differences?  Exact compare sounds like a setup for issues with different hardware/compilers especially with trigonometric functions

lgritz:
Anybody have remaining concerns about this PR?

fpsunflower:
LGTM!

PR #1898, opened 2024-10-30
testrender: Implement basic displacement shader support

fpsunflower:
## Description

Associate an (optional) displacement shader with each material. On startup, we execute the displacement for each face-vertex that has a valid displacement shader assigned.

## Tests

Added a new test with a displaced sphere to show how this works.

## Checklist:

  being nonconforming.


fpsunflower:
Looks like I need to grab updated reference images from the CI machine.

In any case, this PR should be reviewed after we get #1897 merged.

lgritz:
Update: I've merged #1897 now, so you ought to be able to rebase on top of it in main.

lgritz:
I took the liberty of rebasing this on top of the current main, fixing the formatting problems, and adding one more reference image, and now we are passing all CI but the icx (I think that will need another reference image, checking...) and the gpu test.

The GPU one is tricky -- this PR breaks the build when OptiX is enabled. It's because of the change of SimpleRaytracer::m_shaders from a ShaderGroupRef to a `vector<Material>` (and a Material is now a ShaderGroupRef for each of surf and disp). I think you made changes in simpleraytracer.cpp related to this, but didn't make the corresponding changes to optixraytracer.cpp, and so now it's confused.

lgritz:
I believe I've updated now with a fixed icx reference image. So the only remaining problem with this PR is the build break when enabling OptiX. Is this something you're able to look at, @fpsunflower ?

fpsunflower:
Yes, sorry I had started to look into this over the break and then got distracted. I'll pick up where you left off.

lgritz:
Thanks, Chris. I think I at least got a bunch of other distractions out of your way.

fpsunflower:
Alright this should be good to go now. I believe the OptiX side is hooked up correctly, though I don't have a setup to run it here.

lgritz:
It's really unfortunate that our CI can't (yet) do OptiX tests. Turns out that we had two somewhat recent changes that broke the CI tests on OptiX and needed minor updates. I have fixes in #1926 (posted yesterday and merged this morning) and #1927 (just posted and awaiting review). I can confirm that with the addition of those, your patch here fully build and passes all tests with OptiX, so it's ready to merge!

lgritz (approved):
Looks great!

There were still some remaining failed tests when I tried it, but they were unrelated to your changes and I have separate patches that address them.

PR #1901, opened 2024-11-10
ci: Various adjustments for OIIO 3.0 becoming the release

lgritz:
Some of the changes are specifically related to OIIO 3.0 requiring OCIO and raising the minimum version it needs.

Also, we can eliminate the extra test we added to specifically test the 3.0 betas (now "release" picks up 3.0, so getting dev-3.0 explicitly isn't really needed).


lgritz:
Could somebody please review? This one is necessary to solve some broken CI now that OIIO has moved its release branch marker to 3.0.

fpsunflower (approved):

PR #1902, opened 2024-11-10
docs: Update some URLs to our new vanity URLs

lgritz:
Also a little be of typo fixing and docs updating, mostly from comparing to the same docs in OIIO and picking the more thorough description.


lgritz:
Any comments on this docs-only change?

PR #1908, opened 2024-12-09
admin: Add code reviewing guide

lgritz:
Ported from OIIO, with minor changes in some areas that differ between the two projects.


lgritz:
No objections, docs only -> merging

PR #1909, opened 2024-12-09
build: add option for build profiling with clang -ftime-trace

lgritz:
(Ported from OIIO)


lgritz:
No objections, optional build mode only, ported from OIIO -> merging

PR #1913, opened 2025-01-02
admin: Sign release artifacts

### LINKED ISSUE #1857: Consider signing release artifacts
The OpenSSF Best Practices Badge suggests [signing release artifacts](https://www.bestpractices.dev/en/criteria). Consider using OpenEXR's [release-sign.yml](https://github.com/AcademySoftwareFoundation/openexr/blob/main/.github/workflows/release-sign.yml) workflow as a template. It's triggered on release creation and does these steps:
1. Runs `get archive` to generate a `<release>.tar.gz` artifact
2. Signs the `<release>.tar.gz` via [sigstore](https://www.sigstore.dev/) 
3. Uploads the resulting sigstore signature file along with the tarball.

lgritz:
Closes #1857

Same approach that was taken by OpenEXR and OpenImageIO.


lgritz:
Merging without review because it's purely an administrative GH thing, can't affect user builds. And this is identical to the approach successfully used for OpenEXR and OpenImageIO.

PR #1914, opened 2025-01-03
deps: Raise dependencies: cmake 3.19, OIIO 2.5, LLVM 11

lgritz:
Some last minute dependency adjustments before I cut a beta for OSL 1.14.

* Raise CMake minimum to 3.19 (from 3.15). This is still very conservative -- 4 years old, even more lax than our usual "3 years back" rule.

  Raising to 3.18 matches OIIO 3.0, and compared to 3.15, it adds Cuda support as well as the `--debug-find` flag.  Further raising to 3.19 -- which in 2025 seems prudent, even though it's beyond the version chosen by OIIO 6 months ago -- adds more helpful features, including presets, and Apple Silicon support.

* Raise OpenImageIO to 2.5 (from 2.4). The current release is 3.0, FYI.

* Bump LLVM minimum to 11.0 (from 9.0). This is a pretty conservative move, still gives several years of back support, but brings a little code simplification. I expect that next year's OSL will raise the LLVM minimum by several more versions.

### DIFF
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -1896,6 +1896,10 @@ void
 LLVM_Util::setup_new_optimization_passes(int optlevel, bool target_host)
 {
 #ifdef OSL_LLVM_NEW_PASS_MANAGER
+#    if OSL_LLVM_VERSION <= 110
+#        error "New pass manager not supported in LLVM 11 and earlier"
+#    endif
+
     OSL_DEV_ONLY(std::cout << "setup_new_optimization_passes " << optlevel);
     OSL_ASSERT(m_new_pass_manager == nullptr);
 
--- a/src/liboslexec/llvm_util.cpp
+++ b/src/liboslexec/llvm_util.cpp
@@ -13,8 +13,8 @@
 #include <OSL/oslconfig.h>
 #include <OSL/wide.h>
 
-#if OSL_LLVM_VERSION < 90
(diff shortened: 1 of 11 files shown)


src/liboslexec/llvm_util.cpp
```diff
@@ -1902,10 +1896,6 @@ void
 LLVM_Util::setup_new_optimization_passes(int optlevel, bool target_host)
 {
 #ifdef OSL_LLVM_NEW_PASS_MANAGER
-#    if OSL_LLVM_VERSION <= 110
-#        error "New pass manager not supported in LLVM 11 and earlier"
-#    endif
```
brechtvl, line 1907: This can't be removed yet, since it's <= 11 and not < 11.

lgritz: I'll revert that, but curiously, the "oldest everything" CI test passed using LLVM 11. That's because this doesn't technically matter -- we only turn on the new pass manager when LLVM version >= 16, and this check merely ensures that if it's on, we're using at least 11. So in a sense, it's redundant (other than to ensure that we don't change the logic about when we use the new pass manager to a version too early).

lgritz: Good catch, by the way, thanks for looking carefully at this.

lgritz: Pushed an update

brechtvl (approved):

PR #1915, opened 2025-01-04
ci: bump to latest version of upload-artifact action

lgritz:


lgritz:
Any objections? This only affects CI and demonstrably works.

lgritz:
No objections, CI only change -> merge

PR #1916, opened 2025-01-06
Generate symbol derivatives for outputs when requested

lukasstockner:
## Description

Currently, symbols are only flagged for derivative computation if
- They're a dependency of an operation which requires derivatives
- They're connected to a downstream layer parameter which requires derivatives
- They're writable globals (except for N)

However, there's a fourth relevant case - when the renderer requests the symbol to be written as an output with derivatives. This is possible via `SymLocationDesc`, but in the current implementation, this only ends up writing derivatives if one of the three cases above also happens to be true.

Therefore, detect this case and explicitly mark the symbol as requiring dependencies.

## Tests

No tests at the moment since it would require extensive changes to testshade. Let me know if this is something that you'd consider necessary here.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/liboslexec/batched_analysis.cpp
+++ b/src/liboslexec/batched_analysis.cpp
@@ -2175,11 +2175,8 @@ struct Analyzer {
         const SymLocationDesc* symloc = nullptr;
         if (interpolate_param) {
             // See if userdata input placement has been used for this symbol
-            ustring layersym = ustring::fmtformat("{}.{}", inst()->layername(),
-                                                  s.name());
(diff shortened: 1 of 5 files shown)


brechtvl (approved):

lgritz:
CI doesn't (yet) test Windows -- I am working on it now, actually -- but I trust that you two have verified that this is the right fix, so I will merge.

lgritz:
Sorry, that last comment was bogus, I had a different PR in mind. :-)

lgritz (approved):
LGTM2

PR #1919, opened 2025-01-08
ci: break out reusable build steps

lgritz:
Modeled after what we recently did for OIIO.

Now the new build-steps.yml contains all the logic for the sequence of how to build and test for any configuration, and the main ci.yml has the specifications of all the different testing configurations.

The different platforms share the same build steps, so there's a lot less needless repetition (and opportunty for divergence and mistakes).


lgritz:
Comments? I'd like to merge this and then I will have follow-ons that build upon this. I'm hot on the trail of getting Windows CI working.

lgritz:
I will merge this if I don't hear any concerns by tomorrow.

PR #1920, opened 2025-01-08
Build fixes for OSL on Windows

jstone-lucasfilm:
## Description

This changelist contains two fixes for OSL builds on the Windows platform, both of which were authored by Kai Rohmer at NVIDIA.

## Tests

With these fixes in place, OSL builds on Windows now include `testshade` and `testrender`, where these two applications would previously fail to build.

For visual validation, I'm attaching the latest MaterialX render comparisons between GLSL and OSL, where the OSL renders are using the codebase in OSL main plus these two fixes.

[MaterialXRenderTests_01_07_2025_GitHub.pdf](https://github.com/user-attachments/files/18351045/MaterialXRenderTests_01_07_2025_GitHub.pdf)

## Checklist:

  being nonconforming.

### DIFF
--- a/src/testshade/CMakeLists.txt
+++ b/src/testshade/CMakeLists.txt
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # https://github.com/AcademySoftwareFoundation/OpenShadingLanguage
 
+set (LLVM_COMPILE_FLAGS ${LLVM_COMPILE_FLAGS} -std=c++${CMAKE_CXX_STANDARD})
+
 # The 'testshade' executable
 set ( testshade_srcs
       testshade.cpp
--- a/src/testrender/CMakeLists.txt
+++ b/src/testrender/CMakeLists.txt
@@ -79,7 +79,7 @@ add_executable (testrender ${testrender_srcs})
 
 target_link_libraries (testrender
     PRIVATE
-        oslexec oslquery
+        oslexec oslquery oslcomp
         pugixml::pugixml
         Threads::Threads)
 


src/testshade/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # https://github.com/AcademySoftwareFoundation/OpenShadingLanguage
 
+set (LLVM_COMPILE_FLAGS ${LLVM_COMPILE_FLAGS} "-std=c++17")
```
lgritz, line 5: This one doesn't feel right. Shouldn't it use whatever standard is used by the rest of the codebase, instead of hard-coding it to 17?

jstone-lucasfilm: That sounds like a good request for an improvement to this change, and I'm CC'ing @krohmerNV for his thoughts.

jstone-lucasfilm: As an initial proposal for this improvement, perhaps it could be written as follows?

```
set (LLVM_COMPILE_FLAGS ${LLVM_COMPILE_FLAGS} -std=c++${CMAKE_CXX_STANDARD})
```

jstone-lucasfilm: I've made the change described above, and have verified that the OSL build for Windows works as before.

jstone-lucasfilm:
As an update on the Contributor License Agreement, it sounds like Lucasfilm may not have contributed to OSL before, so it's going through the approval process now!

src/testshade/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # https://github.com/AcademySoftwareFoundation/OpenShadingLanguage
 
+set (LLVM_COMPILE_FLAGS ${LLVM_COMPILE_FLAGS} -std=c++${CMAKE_CXX_STANDARD})
+
```
lgritz, line 6: I ran into this as well, but there is actually a better solution!

lgritz:
I think my Windows CI PR is going to subsume this. I still think you should push the CLA through the bureaucracy so you've got the ducks in a row for the next time somebody from ILM needs to make an OSL patch. But I suspect this PR is ultimately not going to be merged.

jstone-lucasfilm:
No problem, Larry, and a set of Windows fixes directly from the OSL team is even better from my perspective.  I'll close out this PR, but will continue with the Lucasfilm approval process so that we can contribute in the future.

PR #1922, opened 2025-01-15
api: Switch to a 2-level namespace

lgritz:
Related OIIO PR: https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4567

Like in the corresponding OIIO PR, I'm proposing switching to a 2-level namespace scheme (basically, `OSL::v1_14` instead of the current `OSL_v1_14) because that opens the door to possible future schemes to preserve better ABI compatibility between minor releases.

I would like to squeeze this in before I finalize a 1.14 beta, so that it doesn't change again during the beta period.

Also, I changed some macro names.


lgritz:
Please excuse the fact that this touches every source file. Almost all the individual changes are the result of changing `OSL_NAMESPACE_ENTER/EXIT` to `OSL_NAMESPACE_BEGIN/END`, which I did to match other projects including OIIO. The substantive changes are only to CMakeLists.txt and oslversion.in.h.

fpsunflower (approved):

PR #1927, opened 2025-01-21
test: Update ref output for render-microfacet with OptiX

lgritz:
A while back, PR #1897 slightly changed the output of this test under OptiX, so we're belatedly updating the reference output.


lgritz:
Merging, this is only a testsuite reference output update and can't break any code.

PR #1930, opened 2025-01-24
ci: Windows CI (take 1)

lgritz:
First step at getting full Windows CI: This PR gets us as far as doing a Windows build of OSL for the "Windows 2019" flavor.  This is an important first step that at least makes it much harder for us to inadvertently break the build on Windows. There are a number of shortcomings that will be addressed over time in follow-up PRs:

* We currently only build, but we don't run the testsuite. This is in progress, but we don't pass all tests quite yet, but I didn't want to wait any longer to get this much in our CI.
* We build the Windows 2019 flavor, but not the Windows 2022 flavor. I'm having trouble there, a little confused about why it's not working, also will finish that as a side quest and/or plea for help from others.
* This is just about getting our own CI runs for Windows and relies on helper scripts being run as part of the CI to build or install the dependencies before the main build starts. We want eventually to get to the point where OSL's build system can itself auto-build all the important dependencies if not found and it's not something most users need to worry about.

A few notes about specific changes in this patch that came along for the ride:

* Port many build_blah.bash scripts (or recent changes thereto) from OIIO to OSL to synchronize them between the two projects and move improvements made on the OIIO side to OSL's versions.
* Augment the build_llvm.bash script with logic to work on Windows, too, using pre-built binaries associated with LLVM releases. Note that not all of their releases have Windows binaries pre-built, so you have to pick one of the right releases for it to work.
* Fix a problem with the way we propagate C++ std level flags from the main build to the bitcode-generating pass. On Linux and Mac, `-std=c++17` works identically on gcc/clang that we use for OSL itself and the clang we use to generate precompiled bitcode. But on Windows, the standard-setting flags for MSVC we use to build the cpp is different from the flags needed for clang when compiling to bicode.
* Modify the ci.yml to allow special branch names to exclude platforms, to simplify iterating over many CI runs when only caring about one platform. If the branch name contains the substring "windows-only", the Linux and Mac CI test variants will be skipped, and analogously for "macos-only" and "linux-only".


src/build-scripts/build_openexr.bash
```diff
@@ -64,9 +62,8 @@ popd
 
 # Set up paths. These will only affect the caller if this script is
 # run with 'source' rather than in a separate shell.
-export ILMBASE_ROOT=$OPENEXR_INSTALL_DIR
+export OpenEXR_ROOT=$OPENEXR_INSTALL_DIR
 export OPENEXR_ROOT=$OPENEXR_INSTALL_DIR
```
brechtvl, line 66: For Windows 2022, it's pretty strange but this seems to make it work.
https://github.com/brechtvl/OpenShadingLanguage/actions/runs/12959852395/job/36152802147

```suggestion
+# Environment variables are case insensitive on Windows, don't set
+# it a second time with different case. This seems to trigger a bug
+# where the next cmake command fails to find the compiler.
+if [[ $(uname -s) != "Windows" && "${RUNNER_OS}" != "Windows" ]]; then
+  export OPENEXR_ROOT=$OPENEXR_INSTALL_DIR
+fi
```

src/cmake/modules/FindLLVM.cmake
```diff
@@ -76,6 +77,9 @@ else ()
 endif ()
 string (REPLACE " " ";" LLVM_SYSTEM_LIBRARIES "${LLVM_SYSTEM_LIBRARIES}")
 
+# Seems necessary on our Windows CI?
+string (REPLACE "libxml2s.lib" "" LLVM_SYSTEM_LIBRARIES "${LLVM_SYSTEM_LIBRARIES}")
```
brechtvl, line 81: libxml2 is used for llvm-mt, which is a tool for Windows manifest files. Not something that OSL would depend on.

brechtvl (commented):
I was just trying to build with CUDA bitcode on Windows, and ran into the same std flag issue. A further fix was needed to avoid some flags being treated as a single argument due to quotes.
```diff
diff --git a/src/cmake/cuda_macros.cmake b/src/cmake/cuda_macros.cmake
index 2f15ca65..cf38234e 100644
--- a/src/cmake/cuda_macros.cmake
+++ b/src/cmake/cuda_macros.cmake
@@ -122,7 +122,7 @@ function ( MAKE_CUDA_BITCODE src suffix generated_bc extra_clang_args )
         # these are warnings triggered by the dllimport/export attributes not being supported when
         # compiling for Cuda. When all 3rd parties have their export macro fixed these warnings
         # can be restored.
-        set (CLANG_MSVC_FIX "${CLANG_MSVC_FIX} -Wno-ignored-attributes -Wno-unknown-attributes")
+        set (CLANG_MSVC_FIX -Wno-ignored-attributes -Wno-unknown-attributes)
     endif ()

     if (NOT LLVM_OPAQUE_POINTERS AND ${LLVM_VERSION} VERSION_GREATER_EQUAL 15.0)
```

src/cmake/modules/FindLLVM.cmake
lgritz: Yeah, it seems to end up in the list of libraries that llvm-config reports should be used, but I guess not for our use case.

src/build-scripts/build_openexr.bash
lgritz: Holy cow, you are a life saver for this one! I've spent a lot of hours trying to understand what's going on here, and never in a million years would I have suspected this innocent line to be the culprit.

lgritz: I feel like I've been around computers for a very long time now, and yet I don't have even the vaguest guess how setting both OPENEXR_ROOT and OpenEXR_ROOT would start a cascade of problems that leads to the OIIO cmake build not being able to figure out which C++ compiler it should be using.

brechtvl: Bash manipulates the `extern char **environ;` array.
https://github.com/bminor/bash/blob/master/variables.c#L4997

I suspect that somewhere in the Bash -> MinGW -> MSVC runtime -> OS stack, it ends up using the wrong index and overwriting an unrelated environment variable, or considers the array invalid due to duplicate entries and discards it.

brechtvl (approved):

lgritz:
With @brechtvl's detective work, I've amended this PR to enable the build for both Windows 2022 and the new GitHub experimental Windows 2025 runners.

PR #1934, opened 2025-02-04
testing: env variable OSL_TESTSUITE_THRESH_SCALE scales test thresholds

lgritz:
For any individual test running, the OSL_TESTSUITE_THRESH_SCALE environment variable, if set, will scale the image comparison thresholds used to decide if a test passes.

Furthermore, a special OSL_TESTSUITE_OPTIX_THRESH_SCALE environment variable, if set, will cause *all* OptiX related tests to be set up with scaled thresholds.


chellmuth (approved):
LGTM!

PR #1938, opened 2025-02-10
API for renderer to cache OptiX ptx

chellmuth:
## Description

This change adds a caching layer to the OptiX PTX pipeline that can skip LLVM generation and optimization. This helps improve scene build times for re-renders of scenes with large shader counts.

We generate a hash key from the optimized shadergroup, and depend on the renderer to provide a cache backend implementation. There is no overhead if the renderer doesn't explicitly opt-in to ptx caching.

A simple backend will be added to testshade/testrender in a follow-up PR.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/liboslexec/oslexec.cpp
+++ b/src/liboslexec/oslexec.cpp
@@ -50,8 +50,10 @@ shadertype_from_name(string_view name)
     return ShaderType::Unknown;
 }
 
+
+
 std::string
-optix_cache_wrap(const std::string& ptx, size_t groupdata_size)
+optix_cache_wrap(string_view ptx, size_t groupdata_size)
 {
     // Cache string is the ptx file with groupdata size on top as a comment.
     // This way the cache string is a valid ptx program, which can be useful
@@ -59,8 +61,10 @@ optix_cache_wrap(const std::string& ptx, size_t groupdata_size)
     return fmtformat("// {}\n{}", groupdata_size, ptx);
 }
 
+
+
 void
-optix_cache_unwrap(const std::string& cache_value, std::string& ptx,
+optix_cache_unwrap(string_view cache_value, std::string& ptx,
                    size_t& groupdata_size)
 {
     size_t groupdata_end_index = cache_value.find('\n');
--- a/src/liboslexec/oslexec.cpp
+++ b/src/liboslexec/oslexec.cpp
@@ -50,6 +50,29 @@ shadertype_from_name(string_view name)
     return ShaderType::Unknown;
 }
 
+std::string
+optix_cache_wrap(const std::string& ptx, size_t groupdata_size)
+{
+    // Cache string is the ptx file with groupdata size on top as a comment.
+    // This way the cache string is a valid ptx program, which can be useful
+    // for debugging.
+    return fmtformat("// {}\n{}", groupdata_size, ptx);
+}
+
+void
+optix_cache_unwrap(const std::string& cache_value, std::string& ptx,
+                   size_t& groupdata_size)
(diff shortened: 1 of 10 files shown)


lgritz:
So, if I may rephrase to make sure I fully understand this:

The sequence at runtime is group oso ->1-> runtime optimize by liboslexec ->2-> JIT to PTX via LLVM ->3-> driver converts PTX to actual executable GPU code on that hardware.

The "OptiX Cache" (part of OptiX & driver) speed up step (3) by not having the last step for optimized/JITed shaders it's encountered before.

This PR adds another cache to step (2), managed by OSL and/or the renderer internals, to allow you to skip the bulk of the work for that step for optimized shaders you've encountered already.

You still pay full price the very first time a shader is encountered, and that leads to terrible TTFP (time to first pixel). But this should take a big bite out of that in practice since it's very common to have encountered most shader configuration before. If the implementation the renderer provides is to store the cache persistently on disk or in a real database, it will be shared from run to run and possibly from user to user.

chellmuth:
> So, if I may rephrase to make sure I fully understand this [...]

Yes, precisely!

src/liboslexec/runtimeoptimize.cpp
```diff
@@ -3122,7 +3122,18 @@ RuntimeOptimizer::printinst(std::ostream& out) const
     return out;
 }
 
+std::string
```
lgritz, line 3125: Sorry, this is an extremely small nit, but our convention is to use 3 blank lines between separate standalone function implementations, in order to give a little more visual separation. (Though we do use 1 blank line between inline function declarations within a class definition.)

src/liboslexec/oslexec.cpp
```diff
@@ -50,6 +50,29 @@ shadertype_from_name(string_view name)
     return ShaderType::Unknown;
 }
 
+std::string
+optix_cache_wrap(const std::string& ptx, size_t groupdata_size)
```
lgritz, line 54: Up to you whether to do this here, but generally, a `const std::string&` parameter that you don't need to store anywhere is usually a good candidate to instead pass as a `string_view` parameter.

aconty:
Other than Larry's comment LGTM!

lgritz (approved):
LGTM, this is great!

Don't worry about the CI failure on the ABI test. That's expected because you added virtual functions. I'll fix up the ABI reference later, before I branch the beta.

I'll let you decide whether you want to fix any of the other minor suggestions I made, just let me know when you're ready to merge. (Or, if you have write perms -- I don't remember now -- feel free to merge it when you're ready.)

lgritz:
Another comment to clarify for the onlookers:

This PR establishes the interface so we can lock down the ABI/API, and also has the compilation side changes we need -- but the implementation is currently nothing, it's not backed by any cache.

A subsequent PR from Chris will add a working implementation of the cache as the default, that uses filesystem storage. But because it's implemented as an overridable part of RendererServices, a renderer will be free to replace it with their own custom implementation (for example, using their studio's database server of choice, or whatever).

src/liboslexec/runtimeoptimize.cpp
chellmuth: No problem, fixed!

src/liboslexec/oslexec.cpp
chellmuth: Sounds good, the latest changes swap all the `const std::string&` parameters for `string_view`.

lgritz (approved):
LGTM, thanks Chris!

PR #1944, opened 2025-02-18
Add rs_allocate_closure free function.

curtisblack:
## Description

Adds a new free function `rs_allocate_closure` which allows the renderer services to provide memory for closure storage, for both inbuilt (add/mul) and user defined closures.

The existing `osl_*` closure handling functions now use this new free function.

The CPU side fallback is to call through to the existing closure pool implementation. The GPU side fallback returns null. 

testshade/testrender provide an example implementation showing how to implement a stack based closure pool for the GPU. 

## Checklist:

  being nonconforming.


src/liboslexec/rs_fallback.cpp
```diff
@@ -316,6 +317,18 @@ rs_trace_get(OSL::OpaqueExecContextPtr exec_ctx, OSL::ustringhash name,
 #endif
 }
 
+OSL_RSOP OSL_HOSTDEVICE void*
+rs_allocate_closure(OSL::OpaqueExecContextPtr exec_ctx, size_t size,
+                    size_t alignment)
+{
+#ifndef __CUDA_ARCH__
```
AlexMWells, line 324: As this is rs_fallback.cpp, it should never be compiled by CUDA, can we remove the #ifndef __CUDA_ARCH__

src/testshade/rs_simplerend.cpp
```diff
@@ -368,6 +368,18 @@ rs_trace_get(OSL::OpaqueExecContextPtr ec, OSL::ustringhash name,
...
 }
 
+#ifdef __CUDA_ARCH__  // Host side uses rs_fallback implementation.
+OSL_RSOP OSL_HOSTDEVICE void*
+rs_allocate_closure(OSL::OpaqueExecContextPtr ec, size_t size, size_t alignment)
+{
+    auto sg       = (OSL::ShaderGlobals*)ec;
+    uintptr_t ptr = OIIO::round_to_multiple_of_pow2((uintptr_t)sg->renderstate,
+                                                    alignment);
+    sg->renderstate = (void*)(ptr + size);
```
AlexMWells, line 378: Suggest making use of the testshade/render_state.h RenderState object and add a stack_buffer data member.  Although I think RenderState is uniform accross all shades, so we might need a per thread renderstate in the ExecContext to properly handle this.

AlexMWells: Maybe something like

```
auto *pst = get_rs_per_shade<TestShadePerShadeState>(ec);
pst->stackbuf = (void*)(ptr + size);
```
Would require another pointer in ShaderGlobals (but hopefully we can remove some in the future).
But then a Renderer could have their own PerShadeState to implement some of this stuff.

src/liboslexec/rs_fallback.cpp
lgritz: Everything in rs_fallback is done this way. It is compiled by Cuda, is it not?

AlexMWells: Its is not, rs_fallback.cpp only exists in
`set (lib_src`
which is only compiled for the host, it turns around and just calls the virtual functions of renderer services (which are only on the host) and has no way to be customized for other targets.
More of a rs_legacy_adapter.cpp

src/testshade/render_state.h
```diff
@@ -24,6 +24,38 @@ struct RenderState {
...
+        uintptr_t p = OIIO::round_to_multiple_of_pow2((uintptr_t)ptr,
+                                                      alignment);
+        ptr         = (void*)(p + size);
+        if (ptr <= &buffer[256])
+            return (void*)p;
+        return nullptr;
+    }
+};
+
+struct RenderState {
```
AlexMWells, line 54: Ok, so with this you are defining "RenderState" to be PerShade, and introduce RenderContext to be for all threads.
And arguably ShaderGlobals always was PerShade.  So perhaps that is fine for terminology.

Stepping back a minute, if I were to want a kernel to execute 100,000 shades, when I launch that kernel I don't have a PerShade RenderState to pass, only a RenderContext.  That kernel would then use the RenderContext and other kernel arguments to build a RenderState to pass into the JIT'd shader.

Ultimately I think we were looking to get the JIT to produce that ShaderGroup specific kernel_adapter interacting with rs_* functions, like rs_population_shader_globals(...) that could be launched on the GPU (passing RenderContext, start index, end index), and we end up with a bit of chicken and egg between the RenderState and RendererContext.  As the OSL Jit doesn't know the concrete type/size of the PerThread/Shade RenderState object to create on the stack.  And the Renderer doesn't know the JIT'd shader function (just its prototype).  Ideally we want them all in the same llvm module so it can all get inlined / optimized together.

But perhaps this is simpler, the final Renderer would have to craft a kernel that sets up a RenderState and call the JIT'd Shader function (not sure how it got it).  Or for CPU side, renderer would have to do its own loop over shades configuring/reusing RenderState.  Although in this example, its really PerThread state vs. PerShade.

So for state/context we have:
1. All Threads
2. Per Thread
3. Per Shade (per shade could be subset/section of PerThread that is updated per shade).

I guess its a Render provided a custom ShadingContext with rs free functions to interact with (which makes sense as we are trying to remove/replace the CPU side ShadingContext by asking the renderer to take on it responsibilities).

As we have a rs_bitcode module to integrate, the OSL Jit could just directly use Renderer provided types and alloca a RenderState on the stack, then call a rs_init(OpaqueExecutionContext *oec, RenderState *) function to let the renderer populate it.

src/testshade/testshade.cpp
```diff
@@ -945,21 +945,23 @@ setup_transformations(SimpleRenderer& rend, OSL::Matrix44& Mshad,
...
+setup_shaderglobals(ShaderGlobals& sg, ShadingSystem* shadingsys,
+                    RenderState& renderState, int x, int y)
 {
     // Just zero the whole thing out to start
     memset((char*)&sg, 0, sizeof(ShaderGlobals));
 
     // Any state data needed by SimpleRenderer or its free function equivalent
     // will need to be passed here the ShaderGlobals.
-    sg.renderstate = &theRenderState;
+    renderState.context      = &theRenderState;
```
AlexMWells, line 962: rename theRenderState to theRenderContext

tgrant-nv:
The changes to the CUDA code look like a nice simplification. Perhaps it would make sense for the pool size to be a template parameter, but given that we've been hobbling along with the fixed-size allocations that's obviously not a requirement. LGTM.

lgritz:
What's the status here, @curtisblack ? This seems to still be broadly failing CI.

src/include/OSL/rs_free_function.h
```diff
@@ -316,6 +316,13 @@ OSL_RSOP OSL_HOSTDEVICE bool
 rs_trace_get(OSL::OpaqueExecContextPtr oec, OSL::ustringhash name,
              OSL::TypeDesc type, void* val, bool derivatives);
 
+/// Allocates memory for a closure color. May return null if no memory could
+/// be allocated. It is the renderers responsibility to clean up these
+/// allocations after a shader is run and the closures have been processed.
```
AlexMWells, line 321: Please mention that this can be called from multiple threads and needs to be thread safe, but same applies to all free functions so add to top.

AlexMWells (approved):
Looks really good to me!

lgritz:
I was holding this approved PR because of the CI failures, but I just rebased it myself and sent it through CI, and it was fine, so I will merge it now.

PR #1946, opened 2025-02-22
build: Fix libclang link order

lgritz:
Apparently, in LLVM 18, some things shifted between libraries and that requires a slightly different link order. But only on Linux, or maybe only with static libraries, so even though our Mac and Windows CI tests have used LLVM 18, it wasn't symptomatic there.

We didn't discover this earlier because we didn't actually test against LLVM 18 on any of the Linux CI runs. So we add that.

Also, to alleviate the pain in adjusting LLVM versions and often needing to modify our build_llvm.bash script, I'm trying out a GHA marketplace action KyleMayes/install-llvm-action, and it works well and saved me time, so we may switch to that for other cases over time as well.

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -44,6 +44,8 @@ on:
         type: string
       python_ver:
         type: string
+      llvm_action_ver:
+        type: string
       setenvs:
         type: string
       simd:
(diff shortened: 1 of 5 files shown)


lgritz:
Any comments or objections?

fpsunflower (approved):

fpsunflower:
Looking forward to being able to ditch this dependency :) I have poked things again on my end, the wheels are still in motion ...

PR #1947, opened 2025-02-24
Fix false positives when utilizing OSL's "debug_uninit" feature

AlexMWells:
## Description

It was checking a "while" loops condition variable before the condition had executed causing false positive.

It was checking all elements of arrays causing false positives when only a subset of the array is actually being read from in "pointcloud_get", "spline", "splineinverse" calls.  Implementation detail:  in batched mode, to handle a varying count or knot count parameters a new version of osl_uninit_check_values_offset was added Fix message formatting for batched debug_uninit messages which was printing out pointer to the string instead of the strings themselves.

## Tests

debug-uninit in testsuite was update to check for false positives for while(condition), spline, and splineinverse as well as correctly identifying uninitialized use of partially initialized arrays with spline and splineinverse.  
Adding test for pointcloud_get was omitted due to local build configuration.

## Checklist:

  being nonconforming.


lgritz (approved):
Good catches, LGTM!

Issue #1929, opened 2025-01-22
problem with integer geometry attributes and batched shader execution


cmstein:
I'm seeing a problem with batched execution that I've managed to distill to the repro case below with `testshade`.  Basically, if I have an integer geometry attribute, then batched OSL execution seems to disregard what the renderer writes to the destination depending on what the default value is specified in the shader.  For eample, suppose I have a shader with the following parameter declaration, `int face_idx = 0 [[int lockgeom=0]]`, the value returned by the renderer doesn't make it to the shader.  But in this case it does:  `face_idx = 2 [[int lockgeom=0]]`.  FWIW, a value of '1' also fails.

This is based on OSL-v1.14.2.0 which goes dates back to September so it's possible this has been fixed?

Here's my example shader:

```c
shader example(int face_idx = 0 [[int lockgeom = 0]],
               output color dst = 0)
{
    dst = color(face_idx * 0.25, 0, 0);
}
```

A patch to add the `face_idx` user-data to batched `testshade`:
```c++
diff --git a/src/testshade/batched_simplerend.cpp b/src/testshade/batched_simplerend.cpp
index ea2acbdf..6618fba7 100644
--- a/src/testshade/batched_simplerend.cpp
+++ b/src/testshade/batched_simplerend.cpp
@@ -21,6 +21,7 @@ struct UniqueStringCache {
         , red("red")
         , green("green")
         , blue("blue")
+        , face_idx("face_idx")
         , lookupTable("lookupTable")
         , blahblah("blahblah")
         , options("options")
@@ -51,6 +52,7 @@ struct UniqueStringCache {
     ustring red;
     ustring green;
     ustring blue;
+    ustring face_idx;
     ustring lookupTable;
     ustring blahblah;
     ustring options;
@@ -663,7 +665,17 @@ BatchedSimpleRenderer<WidthT>::get_userdata(ustringhash name,
 
     // For testing of interactions with default values
     // may not provide data for all lanes
-    if (name == ucache().s && Masked<float>::is(val)) {
+    if (name == ucache().face_idx && Masked<int>::is(val)) {
+        printf("found face_idx!\n");
+        Masked<int> out(val);
+        for (int i = 0; i < WidthT; ++i) {
+           if (out[i].is_on()) {
+               out[i] = int(4 * bsg->varying.u[i]);
+           }
+        }
+        return out.mask();
+    }
+    else if (name == ucache().s && Masked<float>::is(val)) {
[...]

PR #1948, opened 2025-02-25
Pr/batched geometry attributes

AlexMWells:
## Description

Fix issue #1929 with batched execution forcing interpolated integer arguments to be boolean when their default value is 0 or 1.
In actuality the renderer could assign any integer value, we update BatchedAnalysis to track the interpolated value as a write event which will disqualify forcing it to be a boolean, as well as ensure future writes are masked properly.
    
## Tests

Added new lockgeom test that reproduces the issue and verifies its fixed.

## Checklist:

  being nonconforming.


AlexMWells:
I need to rebase on main and try this again.

PR #1852, opened 2024-08-23
Free functions for texturing, point clouds, and trace

curtisblack:
## Description

Provides free function API for:
 - `rs_texture`
 - `rs_texture3d`
 - `rs_environment`
 - `rs_get_texture_info`
 - `rs_get_texture_info_st`
 - `rs_pointcloud_search`
 - `rs_pointcloud_get`
 - `rs_pointcloud_write`
 - `rs_trace`
 
Each of these match the existing signatures of the renderer services equivalents. A new function `rs_trace_get` is added to retrieve trace results and is intended to replace `getmessage("trace", ...)` which is currently not available on GPUs.

The fallback for each of these is to call the existing renderer services functions.

To achieve this, the helper structs `TextureOpt`, `TraceOpt`, `NoiseOpt` are removed from the host-only shading context, and are instead added into the LLVM generated code as stack allocations. These allocations only occur if the shader uses these features.

To allow these free functions to compile on GPUs, the texture per-thread context pointer is currently passed null. Future work will be needed to give this a proper home.

To make the pointcloud functions GPU friendly, they had to modified to not use variadic arguments and to avoid dynamic alloca.

## Checklist:

- [x] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [ ] I have updated the documentation, if applicable.
- [ ] I have ensured that the change is tested somewhere in the testsuite (adding new test cases if necessary).
- [x] My code follows the prevailing code style of this project. If I haven't
  already run clang-format v17 before submitting, I definitely will look at
  the CI test that runs clang-format and fix anything that it highlights as
  being nonconforming.

PR #1951, opened 2025-02-27
Re-remove device decoration from some osl_ functions

lgritz:
Recent PR #1852 (Free functions for texturing, point clouds, and trace) added host/device decorations to certain `osl_` functions that previously were replicated in OptiX-based renderers. This means that now the ones in those renderers will be seen as duplicate symbols.

The more clear intent that we'd like to work towards is that all the `osl_` functions are meant to be "internal" to the liboslexec or the bitcode it provides, and the `rs_` functions are the customization points that the renderer is expected to supply.

Except... a bunch of the `osl_` functions that set parameters for subsequent texture and trace calls do not yet have `rs_` customization points. So we're actually not quite ready to change things.

To unbreak, then, this PR removes the `__device__` decorations (so they are again safe/expected for GPU renderers to override them). To keep testshade/testrender working, we add copies of them to testshade/testrender specifically.

Like I said, eventually we'll add `rs_` customization points, and at that point, we'll truly expect the `osl_` functions to be only supplied by oslexec itself and not the renderers.

### DIFF
--- a/src/liboslexec/optexture.cpp
+++ b/src/liboslexec/optexture.cpp
@@ -66,13 +66,16 @@ osl_init_texture_options(OpaqueExecContextPtr oec, void* opt)
 }
 
 
-OSL_SHADEOP OSL_HOSTDEVICE void
+#define OSL_TEXTURE_SET_HOSTDEVICE /* just host */
+
+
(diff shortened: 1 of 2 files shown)


chellmuth:
LGTM!

PR #1952, opened 2025-03-03
Share Shading Context when optimizing a shader…

AlexMWells:
… vs. grabbing an additi onal context.

## Description

No need for a 2nd set of heaps/buffers when the existing context can be reused. Reduces memory footprint which increases cache efficiency, especially for threaded usage.

We can just swap the TexturePerThread info out for a null pointer and restore the previous one afterwards
For completeness we also process the errors in the context.

## Tests

Existing tests exercise the changes.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/liboslexec/context.cpp
+++ b/src/liboslexec/context.cpp
@@ -86,14 +86,21 @@ ShadingContext::execute_init(ShaderGroup& sgroup, int threadindex,
     if (sgroup.nlayers()) {
         sgroup.start_running();
         if (!sgroup.jitted()) {
-            auto ctx = shadingsys().get_context(thread_info());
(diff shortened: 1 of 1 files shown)


AlexMWells:
Hold up on this, going to update so that such that the work to "borrow" a shading context for opt/jit is internal to ShadingContext and not burden on its user to understand the implementation details.

lgritz:
#1953 fixes the broken CI, which is not the fault of this PR. I merged it, so if you rebase on the current main, you should get a passing CI.

lgritz (approved):
Sorry for the delay, this slipped through the cracks. LGTM.

PR #1953, opened 2025-03-04
ci: Update cache action to modern version

lgritz:
GitHub has deprecated the older versions we were still using.


lgritz:
Merging without review -- demonstrably fixes broken CI, doesn't change library code at all

PR #1954, opened 2025-03-07
Include entry layer name in PTX cache key

chellmuth:
## Description

The ptx cache key introduced in #1938 didn't include a shader's entry layer name in its cache key. The direct callable function generated during llvm construction includes this layer name, so we could get false positive cache hits when two identical networks with matching group names had different entry layer names. The end result is an optix compilation error. This PR extends the cache key to handle this scenario.

## Checklist:

  being nonconforming.


lgritz:
I don't think the CI failures here are related to this PR -- they appear to be (perhaps momentary?) failures to download the Intel compilers. I'm rerunning the failed tests now to see if it's fixed.

lgritz (approved):
LGTM

PR #1960, opened 2025-04-01
build: Support for CMake 4.0

lgritz:
The tricky part is that several dependencies break because of cmake 4.0 dropping support for strict compatbility with cmake < 3.5. Have to resort to some trickery with CMAKE_POLICY_VERSION_MINIMUM overriding what those packages declare via their own cmake_minimum_required(), and hope it works.


lgritz:
Merging as this is desperately needed to fix CI.

There are still two tests failing, but it's for a different reason (llvm 20 issues) and I will address those separately.

PR #1962, opened 2025-04-03
build: Unbreak CI on Mac by not letting it install llvm 20

lgritz:
We're not LLVM 20 compatible yet, we'll deal with that separately and not under the time pressure of a major releases that's overdue.


lgritz:
This has been preventing release, so I will merge it once the CI passes, so that we can get the release out of the way. Then I'll look at how to get ourselves LLVM 20 compatible (unless somebody else wants to give it a try -- that would sure make life easier for me).

PR #1966, opened 2025-04-05
feat: Optional texture argument for colorspace (part 1 -- don't reject)

lgritz:
A while back, we proposed a new optional token/value parameter for the texture functions: "colorspace", followed by a color space name, which indicates that the texture name referenced is understood to be in that colorspace (this is passed to the TextureSystem, which has known for some time how to ensure the right transformations are applied as tile data is read in).

We do not fully implement this yet! It's more involved than I have time to get right by the scheduled OSL 1.14 non-beta release (which was supposed to have been on Apr 1, oh no). Not wanting to delay things further, this change at least causes the use of the new parameter (which apparently, MaterialX has been using) to be silently accepted and ignored, rather than generating runtime errors. We'll come back right away for a more complete working implementation.


lgritz:
@lkerley @jstone-lucasfilm This doesn't make it "work" in the sense of being fully functional, but it makes it safe to pass the parameter.

As soon as this gets approved and merged, I will tag the non-beta and make 1.14 the new supported release family.

lgritz:
Would love a second pair of eyes on these, but I'm going to merge it by tomorrow either way if there are no specific objections, because the release is so hopelessly overdue.

fpsunflower (approved):
LGTM

PR #1981, opened 2025-04-19
admin: Set up new issue templates

lgritz:
It had escaped my notice that last month, GitHub ended support for a single ISSUE_TEMPLATE.md, and now only supports an ISSUE_TEMPLATE directory with individual templates. I can't imagine the engineering reason for not being able to continue to supporting just one template as always instead of that being in a subdirectory.

Anyway, so this PR conforms to what they support, by copying (and lightly modifying) the issue templates that OIIO has used for some time.


lgritz:
No objections for 2 weeks + this is a github issues administrative doesn't change code --> merge.

PR #1982, opened 2025-04-20
build: Fix various MacOS problems with GitHub and CMake

lgritz:
The combination of cmake 4.0, changing homebrew default package versions, and slight changes rolled out to the GitHub Actions runner images conspired to break our CI for 2 of 3 Mac variants. Fixes:

* Apple: only add the -isysroot to the bitcode-generating clang++ command if CMAKE_OSX_SYSROOT is defined. Plot twist -- this is no longer set by cmake  starting with cmake 4.0!

* Fix incorrect warning message for apple clang vs newish llvm in externalpackages.cmake.

* Remove homebrew unlinking of python@3.10 that warned in install_homebrew_deps.bash. On the GHA runners, it was giving us warnings because we were asking this package to be unlinked but it wasn't installed.  For future needs, add an env variable based way to unlink individual packges.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -520,17 +520,19 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - desc: MacOS-13-Intel llvm17
+          - desc: MacOS-13-Intel llvm17 oiio-release
             runner: macos-13
             nametag: macos13-p313
-            cc_compiler: /usr/local/opt/llvm/bin/clang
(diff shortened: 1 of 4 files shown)


lgritz:
I'll hold this until the end of the day because we try to avoid ever merging without a suitable time for meaningful review, but because it fixes broken CI, I will merge it at that point even if still not reviewed.

lgritz:
3 days + needed to fix broken CI --> merging

PR #1984, opened 2025-05-02
build: Remove unused symbol that was removed from OIIO main

lgritz:
This symbol was for an architecture that never shipped mainstream, we don't need to look for it.


lgritz:
Emergency merge without review, only because it's fixing broken CI and blocking the monthly release.

PR #1988, opened 2025-05-30
build: Simplify FindOptiX.cmake a bit now that OptiX 7 is the minimum

lgritz:
There was still some logic that was only needed during the period that we supported both OptiX 6 and 7+.


lgritz:
ping

PR #1989, opened 2025-05-30
build: Get OptiX headers from new improved location

lgritz:
NVIDIA recently set up a GitHub repo with the OptiX headers (with version tags!), which is much better than the unreliable semi-secret location we used to use.


lgritz:
ping

aconty (approved):

PR #1991, opened 2025-06-04
Implement MX conductor with multiple scattering

aconty:
## Description

This moves the BSDF to libbsdl so we use the albedo tables infrastructure. Adds roughness regularization and energy compensation (multiple scattering, Turquin style).

I removed the thinfilm options that were not implemented and don't exist in the MaterialX spec. Also I propose thinfilm to be implemented as a decoration via vertical layering so we don't repeat the parameters in every BSDF/closure. But that would be a separate issue.

Here's a roughness wedge. It is not testrender but it is the same BSDF code verbatim.

conductor_vd8

## Tests

Conductor tests pass and are now cleaner.

## Checklist:

  being nonconforming.


lgritz (approved):
LGTM.

Anybody else have comments?

PR #1996, opened 2025-06-07
testing: Add optix ref output for new test added in PR 1991

lgritz:
#1991 (adding conductor BSDF) crossed paths in flight with #1992 (adding full OptiX tests to CI). So the former was merged without the benefit of being confirmed to fully pass on GPU.


lgritz:
This is great because it highlights the kind of mistake that in theory will be very hard to make now that we have the optix testing in CI!

aconty:
Strange, because the optix side was passing for me locally. I guess it is an optix version driver thing? Does the reference look right?

Otherwise LGTM

aconty:
Just downloaded it, reference looks good.

aconty (approved):

lgritz:
> Strange, because the optix side was passing for me locally. I guess it is an optix version driver thing?

I assume that locally, your OptiX output matched the CPU output closely enough to pass, but that on the CI runners, differences in hardware, optix version, driver, and who knows what else compared to your workstation happened to make just enough differences to fail the test thresholds.

PR #1998, opened 2025-06-10
check for errors in ShadingSystem::ReParameter()

cmstein:
## Description

ShadingSystem() wasn't checking whether the offset returned by interactive_param_offset() was valid or not before performing the ReParameter() memcpy().  So, we could copy data to a location outside of a valid memory block and cause a crash.

## Tests

tests pass except for '987 - render-microfacet.opt (Failed)' but I get the same error for main.

## Checklist:

  being nonconforming.


lgritz (approved):
LGTM

PR #2000, opened 2025-06-12
Change MX conductor to sample bounded visible normals

aconty:
## Change MX conductor to sample bounded visible normals

Turquin style multiple scattering is very simple, but also way noiser
than using a diffuse lobe for MS. This is because sampling microfacets
for reflection, the way we do now, is far from ideal at high roughnesses.

Switching to the reflection optimized sampling technique from "Bounded
VNDF Sampling for Smith–GGX Reflections" fixes all the trouble. This
sampling avoids sampling micronormals that are going to reflect the ray
under the surface (and therefore throwing away samples).

Paper:
https://gpuopen.com/download/publications/Bounded_VNDF_Sampling_for_Smith-GGX_Reflections.pdf

## Tests

Here is a 64 rpp comparison between the two methods. This is main with Turquin style

old_vd8

and the SPI method with diffuse MS

new_vd8

And to compare the look difference here is a more converged couple of renders

Turquin

old2_vd8

SPI

new2_vd8

In the end the fix is using bounded viseble normals. See comments.

## Checklist:

  being nonconforming.


fpsunflower:
Its a good point that in many case you are sampling specular + diffuse anyway -- so the non-reciprocal version which adds a diffuse lobe can be obtained almost "for free" (although Turquin's approach is even more free in the sense that its simply a rescaling of the specular lobe).

However in the BSDL framework, the specular and diffuse lobes are not merged, so you can't easily take advantage of this and end up doing specular + diffuse_ms + diffuse which limits the benefit a little. Also, in practice you probably want the diffuse energy compensation lobe to be tagged as part of the specular lobe for the AOV tracking (again limiting the benefit of lobe sharing).

I haven't checked, but is the sampling of the GGX lobe taking advantage of the latest importance sampling scheme for GGX? https://gpuopen.com/download/publications/Bounded_VNDF_Sampling_for_Smith-GGX_Reflections.pdf

I had not noticed a big difference in sampling quality between the approaches in my tests, but I should repeat those experiments to confirm. Intuitively, most of the waste in Turquin's approach comes from the wasted samples after reflection of the visible micronormal. Adding a diffuse lobe will counter this a bit, but fixing this problem more directly as in the paper above should also help.

selfshadow:
@aconty: this may be faulty intuition on my part but I'm surprised by the noise difference for the leftmost shader balls in the 64 sample renders. What roughness value is that? I would've expected very little visual difference (and negligible multiple scattering) in that case.

aconty:
That would be the roughness regularization. Roughness is 0 on the left, but as soon as you have a diffuse bounce roughness goes to 1.0. I guess a depth=1 render be less confusing.

aconty:
Chris was 100% on point, improving the sampling with that paper fixes everything. It is now clean and the lost saturation was due to sample clamps. It is now back. 64rpp render

bnd_vd8

I rephrased and updated this pull reques.

selfshadow:
> That would be the roughness regularization.

Oh! That makes perfect sense then.

> Chris was 100% on point, improving the sampling with that paper fixes everything.

Awesome!

> I rephrased and updated this pull reques.

The PR title is still referring to the original change. Would it make sense to update that?

aconty:
Updated, is it better now?

selfshadow:
> Updated, is it better now?

Yup, all good now. Thanks!

Issue #2003, opened 2025-06-29
[BUILD] Windows CI vs LLVM


lgritz:
I'm a little stuck and could use some help/advice from somebody who is more Windows adept.

Several days back, osl's CI started failing its Windows jobs. [Here is an example log.](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/actions/runs/15952202367/job/44993740692#step:9:742) that has this lovely error:

```
  In file included from C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\include\algorithm:8:
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\include\yvals_core.h(908,1): error GC074E59A: static assertion failed: error STL1000: Unexpected compiler version, expected Clang 19.0.0 or newer. [D:\a\OpenShadingLanguage\OpenShadingLanguage\build\src\testshade\libtestshade.vcxproj]
    908 | _EMIT_STL_ERROR(STL1000, "Unexpected compiler version, expected Clang 19.0.0 or newer.");
        | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\include\yvals_core.h:534:58: note: expanded from macro '_EMIT_STL_ERROR'
    534 | #define _EMIT_STL_ERROR(NUMBER, MESSAGE)   static_assert(false, "error " #NUMBER ": " MESSAGE)
        |                                                          ^~~~~
  1 error generated.
```

Nothing changed on our end. But examining the logs, I see that on the day this started happening, the GitHub Windows runners upgraded to a new version of MSVS!

I am inferring that this newer MSVS comes with an STL that for some reason is not compatible with LLVM < 19.0 (we were, in fact, using 18.1.8). 

[...]

PR #2004, opened 2025-07-01
fix: pass the real number of channels to rs_texture and rs_texture3d on gpu

lgritz:
These were discovered when moving SPI's renderer to the the current release and switching to use the rs_ functions.

### DIFF
--- a/src/testshade/rs_simplerend.cpp
+++ b/src/testshade/rs_simplerend.cpp
@@ -211,8 +211,8 @@ rs_texture(OSL::OpaqueExecContextPtr ec, OSL::ustringhash filename,
         return false;
     const float4 fromTexture = osl_tex2DLookup(texture_handle, s, t, dsdx, dtdx,
                                                dsdy, dtdy);
-    *((float3*)result)       = make_float3(fromTexture.x, fromTexture.y,
-                                           fromTexture.z);
+    for (int c = 0; c < nchannels; ++c)
+        result[c] = ((const float*)&fromTexture)[c];
     return true;
 #endif
 }


lgritz:
The Windows tests are failing because of unrelated issue #2003.

I think I've made a mistake in the OptiX parts, actually. Don't review yet, I think this still needs a little work.

lgritz:
Eew, there is some confusion about how the rs_texture family of functions should work. The parameters are passed as `(int nchans, float* result)` (I'm simplifying), and in some parts of the code, we act as if result was an array of nchans, but in other places we seem to assume that it's always a pointer to `float[4]`, and it's up to the caller to later copy out individual floats if needed. 

It might be worth noting that for the old RendererServices, the texture methods also pass `int nchans, float* result` and expect that it's `result[nchans]`. But WITHIN the implementation, there is a lower level that assumes float[4] in order to use some SIMD acceleration. 

So do we imagine rs_texture to be a direct substitute (in a sense) for RenderServices::texture(), or is it more analogous to the hidden thing in side RS::texture that knows it has a `float[4]` to deal with? Which would you like it to be?

Now that I've written the above, I think that it makes most sense for rs_texture to mimic RendererServices::texture, in that the fact that we pass nchans at all makes me think that it's supposed to only touch `result[0..nchans-1]` and not just jam a float4 in there.

But if you disagree strongly, speak up before I finish fixing it all.

lgritz:
Never mind that last comment. I think I have it all fixed already, was just one spot making a bad assumption. Obviously, rs_texture ought to have the same meaning of its parameters as RendererServices:texture did.

lgritz:
@chellmuth @aconty This is a necessary step to get Spear fully caught up.

aconty (approved):
Good catch!

PR #2007, opened 2025-07-20
deps: test against pybind11 3.0 and minor fixes

lgritz:
Add pybind11 3.0 to the CI test matrix.

Found a place where we defined a python method twice, fixed that.

Ran into some trouble where the Mac runners which used a pre-built OIIO supplied by Homebrew, and apparently compiled against pybind11 2.x.  I think maybe there are some corner cases where if you have two python modules that share types (TypeDesc and a few more, in this case), they should not be built with pre-3.0 and post-3.0 versions of pybind11?

The quick fix was to switch to having OSL CI on Mac just build an OIIO from scratch to ensure that it was using the same pybind11. But that identified yet another problem (which I also think is a runner issue) that required some minor adjustment to how OIIO auto-builds libpng.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -532,13 +532,13 @@ jobs:
             cc_compiler: clang
             cxx_compiler: clang++
             cxx_std: 17
-            openimageio_ver: lg-buildpng
+            openimageio_ver: dev-3.0
             python_ver: "3.13"
             aclang: 14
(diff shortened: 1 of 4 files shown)


lgritz:
I will merge this change tomorrow unless there are objections -- it's needed to fix a CI break.

lgritz:
No objections -> merging to fix broken Mac CI.

PR #2009, opened 2025-07-25
MaterialX dielectric with Multiple Scattering

aconty:
## Description

This moves the dielectric implementation to BSDL and adds energy conservation via Turquin style multiple scattering. That is, we create a microfacet based reflection refraction model where we precumpute the energy loss to a lookup table. Then at render time we renormalize the lobe to remove the energy loss. This is a roughness wedge just using this technique.

mtx_clear_vd8

In addition we use Bounded Visible Normal sampling to avoid rejecting reflection samples. This improves noise significantly at the cost of losing valid refracted directions. This doesn't produce energy loss thanks to the renormalization, but biases the the refraction for high roughness and low IOR. The result is still plausible and very hard to see which one is correct.

mtx2_clear_vd8

Here is with higher IOR and some color:

mtx_plastic_vd8

If you have a spectral render or a hacky backport for dispersion you can get some color aberration, this BSDF can do wavelength dependent IOR.

mtx_cleardisp_vd8

Also keeping the ability to layer a reflection only coating on top of other BSDFs via layer().

mtx_coat_vd8

## Tests

Updated existing tests.

## Checklist:

  being nonconforming.


fpsunflower:
Great stuff! I hadn't thought of just ignoring the bias from bounded VNF in the refraction case, its true that its hard to tell which image is "correct". At the same time,  looking at the images, I would say the difference is noticeable enough that I am worried about compatibility with other renderers (since this is meant to be a reference for MaterialX and others).

Maybe it could be a compile time decision? Or maybe compatibility with other systems if more important overall?

Do you have the comparison of the new approach to the previous one with extra (reciprocal) diffuse lobes?

lgritz:
I took the liberty of pushing a 2-line update to your branch that adjusts the test parameters to allow a small number of pixels to fail in the render-mx-dielectric-glass test. Let's see how it does on the CI run now, it seemed to do the trick in my fork.

I would rather have made it as a code change suggestion as part of a review, which you could then look at and click whether or not to accept without my needing to actually alter your fork, but GH won't let me do that. I really wish that GH allowed one to make code edit suggestions to lines that were not directly part of the PR. Quite often the correct suggestion is to add something that was left out!

lgritz:
The Windows failures are "expected" -- all branches are currently failing the Windows tests for reasons unrelated to your PR.

lgritz (commented):
The code itself LGTM, but I'll leave it to @fpsunflower and @aconty (and any other experts with an opinion) to hash out acceptance of the implementation details.

aconty:
Thanks, Larry!

Chris, do you mean the SPI model? yes, I do have it here

spi2_clear_vd8

Remember this model incurs in energy gain with many bounces + TIR. Our math for the balance is not perfect, specially when using transparent shadows. Take into account both SPI and this pass a furnace test.

I do like the idea of making it a compile time option. I would make it runtime if it wasn't for the duplicated albedo tables + intuition that people will like cleaner best. Let me make it compile option.

fpsunflower:
That is quite different! I haven't done a comparison on my end to the "ground truth" Heitz model, but it would be interesting to know which is closest.

Given how far apart these two version of multiple scattering are, maybe the smaller difference from the Bounded VNF is not that actually that big ...

lgritz:
> There is still some CI stuff to sort out, but that should just be a matter of tweaking the reference thresholds or adding additional reference images.

No, the CI is totally passing now, except for the Windows builds which are failing for the llvm vs msvs version we discussed last week.

aconty:
Sounds good, ok to merge?

lgritz:
Alex, do you need a backport and internal build?

aconty:
No need, no. Thanks!

PR #2010, opened 2025-07-28
build: Test C++20 and gcc14 + fixes needed

lgritz:
We weren't testing against gcc-14 nor against C++20 in our CI test matrix.  Add them.

The fixes required were:

* Had to suppress some warnings gcc-14 complains about in LLVM's headers.

* There were some tricky interactions between C++20 and fmtlib, which is because when being compiled with C++ >= 20, fmt switches to a bunch of tricky runtime checking that has different needs for constexpr and consteval. I had to rearrange some things a bit to make it happy simultaneously with both old and new.


jfpanisset (approved):
I can confirm that with these changes I am able to build with gcc 14.2 in C++20 mode.

PR #2015, opened 2025-08-03
fix: batch shading compiler directives were wrong

lgritz:
I spotted two places where we used a *nonexistent* `OSL_GCCVERSION` macro. This is never defined.

In platform.h, I think it's a simple typo and should be `OSL_GNUC_VERSION`, which is the macro we actually use to hold the GCC version, or 0 if it's not being compiled by GCC.

In wide_opcolor.cpp, it's odder and seems like also a cut-and-paste error. At the very least it should be `OSL_CLANG_VERSION` since it's part of a clause that also tests that clang < 14.0. But upon staring at it, I can't help but wonder if this spot should *also* have a clause `OSL_GNUC_VERSION ||` to make this simd loop macro work properly for GCC, as it did in the other one in platform.h?

And should the one in platform.h include clang, or no?


lgritz:
@AlexMWells This look right to you?

lgritz:
This has been sitting around for a while. Any comments? @AlexMWells?

AlexMWells (approved):
LGTM, 
although we might want a todo to try different clang versions to see if at some version the WIDE_TRANSFORMC_OMP_SIMD_LOOP can be re-enabled.

PR #2016, opened 2025-08-03
cleanup: Simplify some clauses involving old compilers

lgritz:
Our minimum gcc 9 and clang 5 lets us simplify a few preprocessor conditionals.

Also give more clear cmake messages when too-old compilers are encountered, for gcc, clang, and Apple clang.


lgritz:
No comments for over a week? If there are no objections, I will merge in the relatively near future.

lgritz:
Any objections?

PR #2017, opened 2025-08-06
deps: Raise minimum supported Python version from 3.7 to 3.9

lgritz:
This also means ceasing testing against VFX Platform 2021, which was the last year that 3.7 was dictated.

This obviously is only for main (future 1.15), and does not affect the current supported release family.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,20 +51,20 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - desc: gcc9/C++17 llvm11 py3.7 oiio2.5 sse4 batch-b8avx2
-            nametag: linux-vfx2021
-            runner: ubuntu-latest
-            container: aswftesting/ci-osl:2021-clang11
-            vfxyear: 2021
-            old_node: 1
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -51,20 +51,20 @@ jobs:
...
-            old_node: 1
-            cxx_std: 17
-            openexr_ver: v3.1.3
-            openimageio_ver: v2.5.4.0
-            python_ver: 3.7
-            pybind11_ver: v2.7.0
-            simd: sse4.2
-            batched: b8_AVX2_noFMA
-            setenvs: export USE_OPENVDB=0
+          # - desc: gcc9/C++17 llvm11 py3.7 oiio2.5 sse4 batch-b8avx2
```
jessey-git, line 54: Are you leaving these commented out in case of some sort of catastrophe?

lgritz: I think I just didn't remove it after commenting it out. But as soon as this goes in, I have another PR that takes it out (and has a bunch of other changes related to raising the LLVM minimum).

PR #2019, opened 2025-08-06
deps: Raise llvm minimum to 14.0

### LINKED ISSUE #2013: Raise LLVM minimum to 14
Currently, OSL claims to support every LLVM release from 11 to 20 (with 21 coming at us soon). This is too wide a range, leads to awkward code on our side to support all of that, and it's hard to test them all individually in CI. Every couple years, we try to raise the floor.

The current proposal is to raise the minimum from LLVM 11 to LLVM 14. (Obviously this is only in main, a.k.a. future 1.15, and will not affect support for older LLVM in the 1.14 family.)

Please voice any objections you may have about this, now before we make the changes.

lgritz:
Drop support for llvm 11, 12, 13.

Simplify lots of conditional compilation.

With 14 as the minimum, the new pass manager and opaque pointers are both supported, so we are able to get rid of all code related to support of the legacy pass manager, as well as typed pointers. This is a lot of removed code!

When updating INSTALL.md, I noticed that we still had **bold** for "new minimum dependencies" left over from when we first released 1.14. So I'm unbolding them now, except for the ones we actually have changed for 1.15.

I had originally intended to pull the minimum to 15, but actually that makes it more difficult to continue CI testing vs VFX Platform 2022, since the ASWF-docker container for 2022 only support up to LLVM 14. And simultaneously, once I removed the code for legacy PM and typed pointers, there was, in comparison, only a small amount of additional simplification we could squeeze out of moving the minimum to llvm 15. So for OSL 1.15, we will have a minimum llvm 14, and then for OSL 1.16 when we would feel more comfortable droping all VFX Platform 2022 support and move to the 2023 containers as our oldest testing environment, we can bump to llvm 15.

Closes #2013

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,24 +51,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # - desc: gcc9/C++17 llvm11 py3.7 oiio2.5 sse4 batch-b8avx2
-          #   nametag: linux-vfx2021
-          #   runner: ubuntu-latest
-          #   container: aswftesting/ci-osl:2021-clang11
-          #   vfxyear: 2021
-          #   old_node: 1
-          #   cxx_std: 17
-          #   openexr_ver: v3.1.3
-          #   openimageio_ver: v2.5.4.0
-          #   python_ver: 3.7
-          #   pybind11_ver: v2.7.0
-          #   simd: sse4.2
-          #   batched: b8_AVX2_noFMA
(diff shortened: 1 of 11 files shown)


fpsunflower (approved):
Nice simplification!

lgritz:
I was hoping that this PR was discarding enough code to make Chris pleased. ;-)

fpsunflower:
You had me at "removed code" 😄

PR #2023, opened 2025-09-25
build: Add Conda environment setup and dependency installation script

bisqottii:
## Description

This script sets up the development environment by:
- Creating a Conda environment
- Installing all required dependencies

Enables contributors to quickly replicate a working environment.

## Checklist:

  being nonconforming.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -126,6 +126,31 @@ Here are the steps to check out, build, and test the OSL distribution:
 
         make test
 
+Conda Environment
+-----------------
+
+To simplify installation of Python and other dependencies, you can use
+the provided Conda environment setup script located at `src/build-scripts/` 
+by running:
+
+    source src/build-scripts/configure_conda_env.bash
+
+**This script will:**
+  * Check for Miniconda installation.
+  * Create a Conda environment named `osl-env` if it doesn't exist.
+  * Install all required dependencies into the environment.
+  * Activate the environment for the current shell session.
+
+After running this script, the `osl-env` environment will be created, and
+all you need to do when opening a new shell session is simply activate the 
+Conda environment.
+
+**When to use it:**  
+Run this script after cloning the repository and before building OSL. It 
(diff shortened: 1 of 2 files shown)


lgritz:
Maybe put this script in the src/build-scripts directory?

activate_conda_osl_env.sh
```diff
@@ -0,0 +1,51 @@
...
+        # Activate the environment
+        conda activate osl-env
+
+        # Install dependencies
+        echo "Installing dependencies in osl-env..."
+        conda install -y -c conda-forge \
+            cmake \
+            llvmdev=20.1.8 clangdev clangxx_linux-64 libcxx \
+            python=3.12 numpy pybind11 \
+            openimageio=2.5 imath flex bison pugixml zlib
```
lgritz, line 43: That's pretty old. Can we use at least OpenImageIO 3.0 by default?

lgritz:
I think this could be pretty useful for people. Can you add some documentation to "INSTALL.md" explaining how to use it?

activate_conda_osl_env.sh
Rafapp: Looks like conda-forge has up to `v2.5.18.0` sadly :/ 
https://anaconda.org/conda-forge/openimageio

Maybe `mamba `could be an alternative?
https://openmamba.org/en/rpms/base/python-OpenImageIO/aarch64/

bisqq: What we could do is, once we activate the environment, add to the script a step to clone and build from source.

lgritz: The OpenImageIO repo does not currently provide pre-built binaries.

However, this OSL repo already has a src/build-scripts/build_openimageio.bash that can build OIIO from scratch.

lgritz:
I think that to wrap up this PR, we're primarily waiting for some modification to INSTALL.md instructions to explain briefly how and when to use it.

PR #2027, opened 2025-10-03
ci: Adjust comparison threshold for render-microfacet test

lgritz:
Something must have changed on the GHA runners? Or a compiler version? On one test job, we started failing this one test, the noise pattern changed a few days ago, independently of any change to OSL repo.

Bumping up the comparison thresholds slightly appears to do the trick.


lgritz:
This is still going to fail one test until #2028 is also merged.

lgritz:
Merging to fix broken CI.

PR #2028, opened 2025-10-03
ci: Adjust test matrix for recent OIIO and OCIO releases

lgritz:
Most of these changes are actually just making the comments match what we really do. But some update or lock down the OIIO and OCIO versions we test against. Both packages had big releases this week, and for OIIO in particular, "release" now means 3.1, not 3.0 as it did last week.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


lgritz:
This is still going to fail one test until #2027 is merged. It needs both fixes, purely by coincidence, but I made them different PRs because they are addressing entirely unrelated things.

lgritz:
Merging to fix broken CI.

PR #2037, opened 2025-10-30
testrender: Fix uv derivatives for testrender [#1978]

### LINKED ISSUE #1978: Fix uv derivatives in testrender

AlexeySmolenchuk:
## Description

Fix dudx, dudy, dvdx, dvdy, as well as dpdx, dpdy in testrender. Now, bump mapping from both UV-based mapping and 3D-based patterns works well. Mip-Map selection also works fine now. Images in Issue: #1978
This PR is intended to validate math and discuss tests.

Fixes #1978

## Tests

This fix could potentially affect many other tests since dpdx, dpdy are changed.
I'm going to add tests representing bump and mipmapping in this PR.

### DIFF
--- a/src/testrender/raytracer.h
+++ b/src/testrender/raytracer.h
@@ -282,16 +282,15 @@ struct Scene {
     OSL_HOSTDEVICE
     void project(Dual2<Vec3>& p, const Vec3& N, const Vec3& I) const
     {
-        Vec3 nI = I.normalized();
+        Vec3 nI    = I.normalized();
         float cosI = dot(-nI, N);
 
-        if (fabsf(cosI)>1e-3f)
-        {
+        if (fabsf(cosI) > 1e-3f) {
             float deltaX = dot(p.dx(), N) / cosI;
             float deltaY = dot(p.dy(), N) / cosI;
 
-            p.dx() += nI * deltaX; 
-            p.dy() += nI * deltaY; 
+            p.dx() += nI * deltaX;
+            p.dy() += nI * deltaY;
         }
     }
 
@@ -332,13 +331,11 @@ struct Scene {
         Vec3 Lc = n.cross(vb - va);
         Lc /= dot(vc - va, Lc);
 
-        Vec2 dTdx = dot(La, p.dx()) * ta
-                  + dot(Lb, p.dx()) * tb
-                  + dot(Lc, p.dx()) * tc;
+        Vec2 dTdx = dot(La, p.dx()) * ta + dot(Lb, p.dx()) * tb
+                    + dot(Lc, p.dx()) * tc;
 
-        Vec2 dTdy = dot(La, p.dy()) * ta
-                  + dot(Lb, p.dy()) * tb
-                  + dot(Lc, p.dy()) * tc;
+        Vec2 dTdy = dot(La, p.dy()) * ta + dot(Lb, p.dy()) * tb
+                    + dot(Lc, p.dy()) * tc;
 
         return Dual2<Vec2>((1 - u - v) * ta + u * tb + v * tc, dTdx, dTdy);
     }


src/testrender/simpleraytracer.cpp
```diff
@@ -896,12 +896,17 @@ SimpleRaytracer::globals_from_hit(ShaderGlobalsType& sg, const Ray& r,
...
     Dual2<Vec3> P = r.point(t);
-    // We are missing the projection onto the surface here
     sg.P                  = P.val();
+    sg.N                  = scene.normal(P, sg.Ng, id, u, v);
+    // Projecting onto the surface here
+    scene.project(P, sg.N, sg.I);
     sg.dPdx               = P.dx();
     sg.dPdy               = P.dy();
-    sg.N                  = scene.normal(P, sg.Ng, id, u, v);
     Dual2<Vec2> uv        = scene.uv(P, sg.N, sg.dPdu, sg.dPdv, id, u, v);
```
AlexeySmolenchuk: I'm thinking of
Switching N also to Dual2<Vec3> to track changing differentials due to curvature.

And maybe replacing r.point(), scene.normal(), scene.uv() with one unified function so I can reuse some intermediate calculations.

src/testrender/raytracer.h
```diff
@@ -278,6 +278,24 @@ struct Scene {
...
 
+    // Project differentials onto surface defined by N
+    OSL_HOSTDEVICE
+    void project(Dual2<Vec3>& p, const Vec3& N, const Vec3& I) const
+    {
+        float cosx = dot(p.dx().normalized(), N);
+        float cosy = dot(p.dy().normalized(), N);
+        float cosI = dot(I.normalized(), N);
+
+        if (abs(cosI)>1e-3 )
```
fpsunflower: The overload specification of `std::abs` is a bit convoluted in C++ - it appears to be working here but I prefer to use `fabsf()` to be explicit that we want the float to float version.

Constant should be `1e-3f` to avoid extra casts to double precision.

fpsunflower:
Aside from those minor notes above, the logic seems like its correct. Results also look promising.

Someone with access to a working implementation might want to spot check it too for how to handle the grazing angle cases and other situations that can divide by 0.

A follow up might be to add a version that guarantees continuity of the derivatives between triangles - but that requires extra storage and is fine to leave as a second step.

src/testrender/simpleraytracer.cpp
fpsunflower: For testrender only - we could definitely try it out. In any case, your current change looks like a good improvement on its own, we can fine tune it in a second pass.

fpsunflower:
I think the only remaining step would be to update the reference images for the failing tests.

lgritz:
I took the liberty of quickly updating the ref image and formatting and pushed those to your branch. Let's see if those pass all the tests now.

lgritz:
I think we're almost there! The only one failing is the optix test. I think some of the changes you made to simpleraytracer need to also be applied to optixrantracer?

AlexeySmolenchuk:
> I took the liberty of quickly updating the ref image and formatting and pushed those to your branch. Let's see if those pass all the tests now.

Thank you @lgritz.
Unfortunately, I'm stuck on a Windows laptop now and can't move forward.
Opened this issue: https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/issues/2039

lgritz:
I have good news, @AlexeySmolenchuk! The OptiX test failures we're seeing here are also happening in main and other branches, too, so they appear unrelated to this PR. So I am going to merge what you have here, and we'll figure out how to fix the OptiX support separately. 

Since I never merged anything that didn't pass all tests including OptiX, I suspect that it's not a problem we introduced at all, but rather something that changed on the GH runners that we need to adapt to.

AlexeySmolenchuk:
I managed to make a very hacky OptiX build, and I can say that dPdx and dPdy work as expected. Derivatives of u and v also match the CPU version when visualised using emission().

What's not working is accessing the proper mip level from the texture call. Even if I read it like
`Ci = (color)texture(filename, u, v, 1, 1 ,1, 1) * emission();`
It reads the highest mip level anyway.

Reading derivatives of a texture signal is also incorrect.

lgritz:
Thanks, and sorry to distract you with this optix issue. The widespread failure of the optix jobs has been conclusively tied to an update of the docker container that occurred last week.

PR #2041, opened 2025-11-03
ci: unbreak optix jobs by locking down container

lgritz:
In all branches, the OptiX test job started failing several days ago, corresponding to an update of the aswf/ci-osl:2025-clang18.3 container. Reverting to aswf/ci-osl:2025-clang18.2 passes again!

Let's merge this fix so that our CI passes, and separately iterate to figure out what went wrong with the newer container.


jfpanisset (approved):
That should roll you back to the SIGGRAPH time release. Note that this is missing the compiler-rt LLVM bits.

lgritz:
>  Note that this is missing the compiler-rt LLVM bits.

I think that those are only important for the job where we run the sanitizers?

PR #2042, opened 2025-11-04
fix: Use more modern span-based set_pixels with OIIO 3.1.

lgritz:
Switch CI back to the aswf/ci-osl:2025-clang18.3 container that has OIIO 3.1, it will work fine with this change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -684,7 +684,7 @@ jobs:
             # select the right runner type and whether tests are run.
             nametag: linux-optix-vfx2025
             runner: ${{ (github.repository_owner == 'AcademySoftwareFoundation' && 'ubuntu-20.04-gpu-t4-4c-16g-176h') || 'ubuntu-latest' }}
-            container: aswf/ci-osl:2025-clang18.2
+            container: aswf/ci-osl:2025-clang18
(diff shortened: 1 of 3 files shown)


lgritz:
@jfpanisset This is the fix that makes the OptiX tests work with OIIO 3.1, it was my fault after all and your bumping the container helped me identify an important bug.

(To be pendantic, this change is not the fix, this is an OSL-wide workaround that avoids the OIIO bug by using a newer, and working, version of the call, which we should have switched to anyway as it is the preferred one, regardless. The actual bug fix on the OIIO side is https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4949)

jfpanisset:
Seems reasonable. If the GPU CI is going to be testing "bleeding edge", do you want to use the 2026 container? There shouldn't be much difference for OSL since even the 2025 container is using OIIO 3.1.x anyway.

lgritz:
I think when the 2026 container is updated, I'll maybe set up a second OptiX job -- one on 2025 using 2025 era releases, and one using 2026 closer to bleeding edge.

lgritz:
Any other comments, approvals, or objections?

fpsunflower (approved):

PR #2047, opened 2025-11-19
ci: Use bigger runner for icx job variation, needs more disk space

lgritz:
A week or two ago, the icx job variation stopped working, not enough space to finish installing the packages containing icx and related material. We didn't change, the Intel package (pegged to a specific version from a couple years ago) didn't change, the ASWF docker container (also old) didn't change, so we're assuming that the GHA runner itself changed, leaving us less disk space, which is now not enough.

The easiest way out of this for now is for the icx job variation to use a paid bigger GHA runner with more "disk space". This means it can only be enabled when running on the main repo, which uses the ASWF's account that has paid runners. Sorry, CI can't currently test with icx you people's individual forks until we figure out a way to require less space. Until then, developers who don't have icx locally will need to submit the PR to see how it does on icx.

A little bit of workflow magic was needed to ensure that this job variation doesn't run at all on a developer's fork -- we don't want it to run, and look like a failed test just because they don't have an account that doesn't allow use of the big paid runners.

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -12,6 +12,9 @@ on:
   workflow_call:
     # This inputs receive values via the "with:" section in ci_workflow.yml
     inputs:
+      disable:
+        type: boolean
+        default: false
       build:
         type: string
       runner:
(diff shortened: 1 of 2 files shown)


lgritz:
I am going to pre-emptively merge this because it is necessary to stop failing CI.

lgritz:
For the record: this gets is working again, but there are two problems we want to return to:
1. It would be better if we didn't need the big paid runner, both for financial reasons and also (primarily) for the sake of wanting people to be able to verify icx works correctly on their forks before submitting a PR.
2. This is still running the icx test in an old VFX year and on an old version of the icx compiler. It would be preferred to be testing the latest icx and on a modern platform.

PR #2053, opened 2025-11-30
Add testrender volumes

osomalley:
## Description

Alongside the rest of the initialization for `subpixel_radiance`, an empty `MediumStack` is initialized. Its behavior is conceptually similar to the existing `CompositeBSDF`. Media are stored in a statically allocated `char` pool, with pointers maintained in `mediums[MaxMediums]`. In processing medium closures, either `HomogeneousMedium` or `EmptyMedium` are inserted into the medium stack. There is a simple priority handling scheme that will shuffle the incumbent medium pointers to give high priority mediums lower indices. They both inherit from a shared CRTP interface, `Medium`. When processing BSDF closures, intersections with `MxDielectric` or `MxGeneralizedSchlick` may only be added when `priority < current_priority` or are both 0 (the precious priority). Attached to the `MediumStack`, an `integrate` method transmittance samples each medium and scatters using the phase function of the highest priority medium. As of now, only `HenyeyGreenstein` is implemented and inherits from `BSDF`.

With this being both a sizable update and my first commit to the project, I’m more than happy to iterate on this.

### Tests 
`render-mx-medium-glass`, `render-mx-medium-vdf`, `render-mx-medium-anisotropic` with OptiX references of the latter two. The images for `render-mx-generalized-schlick-glass` and `render-mx-dielectric-schlick-glass` have been updated as well.

<p align="center">
  
  
  
</p>

## Checklist:

  being nonconforming.


lgritz:
Hi, just checking up on this. It's still marked as draft, still needs a DCO sign-off on all commits, and it's failing some CI tests.

DCO -- the easiest fix is on your end to squash all the little commits into one, make sure that one has the "Signed-off-by: Your Name <your@email>", then do a force push to the same branch to replace what you have.

CI tests -- I suspect that some numerical differences between platforms simply require more than one reference output. I can help with this, if you can fix the other problems.

osomalley:
Thanks for the follow-up and guidance. I’ve squashed the commits into one and signed off. Are there any other changes I should address aside from the CI issues?

aconty:
Very nice addition. I would just ask for clarification on the stack, priorities and policy. In my mind, an ice cube with medium A floating in water with medium B pushes the water away. So in the stack inside the ice you have

A (highest prio)
B

But only A is active. When you exit the ice your stack is only B, making the water active. No?

osomalley:
Yes — This is true. I've made some changes. 

There were two problems with what I had. In the ice/water example, inside the ice the stack contains A (highest priority) and B, but only A is active, so B no longer incorrectly contributes. Also, what I had before didn't track the entry order and just decremented depth by 1. Which in the case of A (highest), B(middle), and C(lowest). If the ray enters B, C, then A. After sorting it will be A, B, C. If the ray exists B first C will be the one taken off the stack. Now when the ray exists B. the pointer to B is nullified and C takes its place. Does this sound right?

Mediums are now only combined when they share the same priority as the topmost medium. When that happens, a CDF is computed and sampled for scattering. This CDF is recomputed whenever `add_medium` or `pop_medium` is called.
Also, I had the false-intersection logic backwards. Priority 0 now disables false intersections rather than enabling them. 

A ring of volumes with the same priority.

image

The upper left spheres are the same and non precious. Upper right spheres are the same priority and are precious (priority == 0).

image

lgritz:
I'll let @fpsunflower and @aconty comment in detail, but glancing and the results, I think this is a lot more along the lines of what we expect for this feature in a renderer.

lgritz:
@ThyOwen Do you know what's up with the CI failures?

src/testrender/shading.h
```diff
@@ -454,22 +526,271 @@ struct CompositeBSDF {
...
+                }
+            }
+
+            current_params                 = new_params;
+            need_to_compute_current_params = false;
+        }
+
+        Vec3 rand_vol  = sampler.get();
+        float t_volume = -logf(1.0f - rand_vol.x)
+                         / current_params.avg_sigma_t();
```
fpsunflower: It would be better to perform MIS among the 3 color channels. Sampling using the average will not work well if there is a strong difference between the channels.

src/testrender/shading.h
```diff
@@ -454,22 +526,271 @@ struct CompositeBSDF {
...
+        Color3 weight;
+        Color3 tr;
+        bool scatter = t_volume < hit.t;
+        if (scatter) {
+            r.origin = r.point(t_volume);
+            tr       = transmittance(current_params.sigma_t, t_volume);
+            Color3 combined_albedo = current_params.sigma_s
+                                     / current_params.sigma_t;
+            weight = combined_albedo / tr;
```
fpsunflower: I think it would be simpler to merge this block and the one bellow so all the scatter logic is together.

Like below, I'm not quite following the logic on how the weight is computed here.

The weight should be transmittance / pdf, not albedo / transmittance.


To include MIS between the color channels, you want to consider sampling from r,g,b for `sigma_t` independently. The weighting between these should depend on the combined `sigma_s` (so that channels that don't have any scattering will not contribute). This way - if you don't have any scattering - you won't generate any sample for transmittance and can pass straight through without an additional random decision.

fpsunflower (commented):
The general setup seems like its going in the right direction - see comments below for some things I think should be checked though.


Also - I would make sure the code works with closures like glass than can both reflect and refract. At a high level it looks like the medium update happens when we hit the surface (depending on the backfacing flag). I would expect the medium stack to change only if we pick a direction that actually crosses from one side to another. I might have missed something though.

osomalley:
Thanks for the feedback. I’ve made several changes.

`MediumStack` stores `MediumParams` directly as a fixed-size array rather than using a byte pool. Since there is no inheritance pattern and the mediums are just params, the pool allocator was unnecessary. Entry and exit factor the outgoing direction of the BSDF sample. The dot product again the geometric normal (Ng · wi < 0) indicates transmission. Additionally, there was an address alignment issue that was causing problems in OptiX.

The `HenyeyGreenstein` phase function now wraps the BSDL implementation in `MediumParams::sample_phase_func`.

Volume integration uses single-channel free-flight estimation. A channel is selected with probability proportional to `sigma_s`, and the sampled channel's `sigma_t` drives the distance sampling.

Also, I initially changed the sorting in `MediumStack` to stored entry indices but changed it back to the shuffling pointers. 

`render-mx-medium-vdf-glass`, `render-mx-medium-vdf`, `render-mx-medium-anisotropic-vdf`:

image
image
image

lgritz:
Where are we on this?

We still have some test failing, and probably it needs to be rebased onto the current main and have any conflicts resolved.

ThyOwen:
I'll get this rebased and look into the CI failures. The failure that worries me most is a `testshade` timeout issue on one of the runners — it fails on all three of the volume closure tests. The remaining failures look to be build/dependency-related.

src/testrender/shading.h
```diff
@@ -13,11 +13,13 @@
 
 #include "bsdl_config.h"
 #include <BSDL/static_virtual.h>
+#include <queue>
```
fpsunflower: Is this header still used?

src/testrender/shading.h
```diff
@@ -434,22 +437,329 @@ struct CompositeBSDF {
...
-    Color3 sigma_s       = Color3(0.0f);
-    Color3 sigma_t       = Color3(0.0f);
-    float medium_g       = 0.0f;  // volumetric anisotropy
+struct MediumParams {
+    Color3 sigma_t       = Color3(0.0f);  // extinction coefficient
+    Color3 sigma_s       = Color3(0.0f);  // scattering
+    float medium_g       = 0.0f;          // volumetric anisotropy
     float refraction_ior = 1.0f;
     int priority         = 0;
+    bool is_empty        = true;
```
fpsunflower: Is this boolean really needed? I think it would be less error prone to just test `sigma_t` directly.

src/testrender/shading.h
```diff
@@ -434,22 +437,329 @@ struct CompositeBSDF {
...
+
+    OSL_HOSTDEVICE Color3 transmittance(const Color3& sigma_t,
+                                        float distance) const
+    {  // Beer-Lambert law
+        return Color3(expf(-sigma_t.x * distance), expf(-sigma_t.y * distance),
+                      expf(-sigma_t.z * distance));
+    }
+
+private:
+    /// Never try to copy this struct because it would invalidate the medium pointers
```
fpsunflower, line 707: Instead of making these private, they should be `= deleted` instead.

src/testrender/shading.h
```diff
@@ -434,22 +437,329 @@ struct CompositeBSDF {
...
+    OSL_HOSTDEVICE MediumStack(const MediumStack& c);
+    OSL_HOSTDEVICE MediumStack& operator=(const MediumStack& c);
+
+    enum { MaxEntries = 8 };
```
fpsunflower, line 711: If there are only 8 entries in the stack, why do you need 32 entries in the pool? In what case would the pool grow larger than the stack?

src/testrender/shading.h
```diff
@@ -434,22 +437,329 @@ struct CompositeBSDF {
...
+    float cdf[MaxEntries];
+    int overlapping_medium_indices[MaxEntries];
+
+    MediumParams current_params;
+
+    int depth;
+    int pool_size;
+    int num_overlapping;
+
+    bool need_to_compute_current_params;
```
fpsunflower: Instead of maintaining this boolean, could the current medium parameters just be computed as soon as the state is changed? Are there cases where multiple operations invalidate the stack before the current parameters are needed again?

src/testrender/shading.h
```diff
@@ -434,22 +437,329 @@ struct CompositeBSDF {
...
+
 };
 
+struct ShadingResult {
+    Color3 Le                = Color3(0.0f);
+    CompositeBSDF bsdf       = {};
+    MediumParams medium_data = {};
```
fpsunflower, line 734: Maybe add a comment here about what `medium_data` represents?

I assume it is the medium parameters that will take over if we "cross" the boundary of the surface hit for this shading result.

src/testshade/optixgridrender.cpp
```diff
@@ -1013,7 +1013,7 @@ OptixGridRenderer::processPrintfBuffer(void* buffer_data, size_t buffer_size)
             if (format[j] == '%') {
                 fmt_string            = "%";
                 bool format_end_found = false;
-                for (size_t i = 0; !format_end_found; i++) {
+                for (; !format_end_found; ) {
```
fpsunflower: This looks like a cleanup unrelated to this review? Could be a `while` loop instead to make it even more clear.

src/testrender/shading.h
```diff
@@ -434,22 +437,329 @@ struct CompositeBSDF {
...
+            Color3 tr = transmittance(current_params.sigma_t, hit.t);
+            path_weight *= tr;
+            return false;
+        }
+
+        Vec3 rand_vol     = sampler.get();
+        Vec3 rand_channel = sampler.get();
+
+        // Build per-channel selection weights from sigma_s
+        float channel_weights[3] = { current_params.sigma_s.x,
```
fpsunflower: In my experience `path_weight * sigma_s / sigma_t` gives even lower variance.

src/testrender/shading.h
```diff
@@ -434,22 +437,329 @@ struct CompositeBSDF {
...
+            channel = 0;
+        } else if (rand_channel.x < channel_weights[0] + channel_weights[1]) {
+            channel = 1;
+        } else {
+            channel = 2;
+        }
+
+        // sample distance along the ray using the selected channel's sigma_t
+        float sigma_t_channel = current_params.sigma_t[channel];
+        if (sigma_t_channel <= 0.0f) {
```
fpsunflower: Is this a case we want to support? It is not physically plausible to have sigma_s > sigma_t. I would argue we should probably enforce this by clamping the medium's sigma_s to sigma_t to avoid these corner cases.

src/testrender/shading.h
```diff
@@ -434,22 +437,329 @@ struct CompositeBSDF {
...
+                              + current_params.sigma_s.z;
+        
+        if (total_sigma_s <= 0.0f) {
+            // pure absorption: apply transmittance to the surface and continue
+            Color3 tr = transmittance(current_params.sigma_t, hit.t);
+            path_weight *= tr;
+            return false;
+        }
+
+        Vec3 rand_vol     = sampler.get();
```
fpsunflower: You are drawing 6 random numbers here, but only using two as far as I can  tell.

You can do the channel selection and then distance sampling with just one number by rescaling the random number after the discrete selection.

But it would also be fine to use `rand.x` and `rand.y` for channel selection and distance sampling respectively.

src/testrender/shading.h
```diff
@@ -434,22 +437,329 @@ struct CompositeBSDF {
...
+            Vec3 rand_index = sampler.get();
+
+            for (index = 0; index < num_overlapping; ++index) {
+                if (rand_index.x < cdf[index]) {
+                    break;
+                }
+            }
+        }
+
+        index = fmin(index, num_overlapping - 1);
```
fpsunflower: Should be `std::min` here, not `fmin` as this is an integer index, not a `double`.

The clamp only need to be inside the `num_overlapping > 1` case.

Or even better - the loop for the cdf only need to go up to `index < num_overlapping - 1` because we know that by construction the last entry of the cdf is 1.0. That would avoid the clamp entirely.

src/testrender/shading.h
```diff
@@ -434,22 +437,329 @@ struct CompositeBSDF {
...
+        }
+
+        // scattering pick a new direction 
+        r.origin = r.point(t_volume);
+
+        Vec3 rand_phase = sampler.get();
+
+        int index = 0;
+        if (num_overlapping > 1) {
+            Vec3 rand_index = sampler.get();
```
fpsunflower: Another case where 3 numbers are fetched but only 1 is used. One of the random values from an earlier `sampler.get()` could be used here.

fpsunflower (changes requested):
Sorry for the very delayed review here. Overall the code is looking pretty good, I just left a few comments below.

osomalley:
Thanks for the notes.

I've cleaned up the sampler calls, revised channel selection (which did reduce variance noticeably), and addressed the remaining suggestions.

fpsunflower (commented):
This is looking good!

I think clang-format is nit picking a few spots, otherwise I think this is ready to submit.

@lgritz to advise on the debug test timing out, not sure how we handle this for other `testrender` tests.

lgritz:
Look in ci.yaml, you can see that the sanitizer job has "-DOSL_TEST_BIG_TIMEOUT=1200" added to the `setenvs` option. You may need to do something similar for the debug job variant as well.

.github/workflows/ci.yml
```diff
@@ -44,47 +44,12 @@ jobs:
...
-      old_node: ${{ matrix.old_node || 0 }}
-      llvm_action_ver: ${{ matrix.llvm_action_ver }}
-      clang_format: ${{ matrix.clang_format }}
+    name: "VFX${{matrix.vfxyear}} ${{matrix.desc}}"
     strategy:
       fail-fast: false
       matrix:
         include:
-          - desc: VP2022 gcc9/C++17 llvm14 py3.9 exr3.1 oiio-2.5 avx2
+          - desc: gcc9/C++17 llvm14 py3.9 exr3.1 oiio-2.5 sse
```
lgritz: Why did this line change?

lgritz:
Something seems wrong, it says there are conflicts?

.github/workflows/ci.yml
osomalley: Thanks for catching that. That change was unintentional on my end. It should be fixed now.

lgritz:
I think the timeout needs to be longer. Rather than continually guessing, it may be easier to make it super long once, let it pass CI and see how long it actually takes, and then tighten up the value to be comfortably, but not hugely, longer than the longest test actually takes.

osomalley:
Previous CI run completed successfully with the following execution times:
```
886/894 Test #582: render-mx-medium-vdf.opt ............   Passed  1455.52 sec
887/894 Test #583: render-mx-medium-vdf-glass.opt ......   Passed  1348.65 sec
874/894 Test #570: render-mx-anisotropic-vdf.opt .......   Passed  1336.76 sec
```
I've decreased the timeout to 1800 seconds, which should comfortably accommodate these longer-running tests?

lgritz:
> I've decreased the timeout to 1800 seconds, which should comfortably accommodate these longer-running tests?

Yeah, that sounds good. Running CI now and if it passes, I will merge.

Because those tests take so very long, I do wonder --- not for now, but eventually -- if we may wish to simply exclude a couple of them from the debug-build CI tests, just to keep our CI runs from growing longer and longer. In other words, maybe our strategy should be to test result correctness on all tests when optimized, but on debug/sanitize CI runs, just test enough for code coverage and to make sure we aren't triggering any debug-only assertions? But don't change this now, we'll cross that bridge if/when we come to it.

PR #2055, opened 2025-12-06
ci: Address tight disk space on GHA runners

lgritz:
- Mount the host's root under /host/root so it's visible in the containers.
- Remove stuff we don't need to free lots of disk space.
- Tentatively go back to the free runner for the icx test.

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -21,6 +21,9 @@ on:
         type: string
       container:
         type: string
+      container_volumes:
+        type: string
+        default: '[]'
       cc_compiler:
         type: string
       cxx_compiler:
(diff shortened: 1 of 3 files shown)


lgritz:
The one failure we're still getting is where we're running out of disk space just loading the Docker image, before we ever get a chance to do the trick of this PR to clear out some space.

GitHub is really making this hard. It has to be that lots of other projects are suffering, right?

lgritz:
And the problem now is the 2026 container, it's not even the one that with the extra intel compiler installs, which with this PR, now happily runs on the free runner again.

lgritz:
I'm going to merge this because it fixes things somewhat. But still will need a re-issued ci-osl to take care of the 2026 problem.

jfpanisset:
All 2024 / 2025 / 2026 images have been re released with the Conan cache purge and should hopefully unpack again on a free runner.

PR #2056, opened 2025-12-07
Update CHANGES before branch

lgritz:


lgritz:
Merging immediately as this is release administrivia and does not change code.

PR #2057, opened 2025-12-07
admin: Prepare for 1.15.0.0 beta release

lgritz:


lgritz:
The ABI failure is expected, until I subsequently update the ABI reference commit.

The VP2026 failure is unrelated to this PR, more problems with disk space on the runners.

So gonna merge.

PR #2061, opened 2025-12-09
ci: switch all containers from aswftesting to aswf

lgritz:
aswftesting is no longer maintained.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
           - desc: gcc9/C++17 llvm14 py3.9 exr3.1 oiio-2.5 avx2
             nametag: linux-vfx2022
             runner: ubuntu-latest
-            container: aswftesting/ci-osl:2022-clang14
+            container: aswf/ci-osl:2022-clang14
             vfxyear: 2022
             old_node: 1
             cxx_std: 17
@@ -65,7 +65,7 @@ jobs:
           - desc: clang14/C++17 llvm14 oiio-3.0 py3.9 avx2 batch-avx512
             nametag: linux-clang14-llvm14-batch
             runner: ubuntu-latest
-            container: aswftesting/ci-osl:2022-clang14
+            container: aswf/ci-osl:2022-clang14
             vfxyear: 2022
             old_node: 1
             cxx_std: 17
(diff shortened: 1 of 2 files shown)


jfpanisset (approved):
That should get you the slimmer images without the baked in Conan download cache.

What's forcing you to remove the bundled OpenImageIO version in the 2024 images and having to rebuild your own with the `REMOVE_INSTALLED_OPENIMAGEIO` mechanism? I'd be happy to fix that so you don't have to spend CI time rebuilding OIIO.

lgritz:
> What's forcing you to remove the bundled OpenImageIO version in the 2024 images and having to rebuild your own with the `REMOVE_INSTALLED_OPENIMAGEIO` mechanism? I'd be happy to fix that so you don't have to spend CI time rebuilding OIIO.

It's often the case that I want to try combinations of a different OIIO than the one in the container, and sometimes it can be hard to guarantee that cmake only finds one and not any remnants of the others, creating an unholy mixture of headers from one version and libraries from another, or a Python module from one version but a library from another. Especially when one of the installs is in a "system area" that has to get included in various search paths. Sometimes the expedient way to solve this is just to remove the one I don't want from the container.

In this particular case, I wanted a newer OIIO for the sanitizer tests. I should really push the sanitizer tests to a more recent container, but I wanted to get this repair out quickly and not spend 3 days fighting some unrelated reason why it might not work in the 2025 or 2026 container. This was just the fastest way to an all-green lineup.

jfpanisset:
Makes perfect sense, just wanted to make sure there wasn't something specifically broken with the OIIO in the 2024 image which was forcing you to replace it.

lgritzecho:
No, no flaws in the container as far as I know.



On December 10, 2025 9:55:03 a.m. GMT+09:00, Jean-Francois Panisset ***@***.***> wrote:
>jfpanisset left a comment (AcademySoftwareFoundation/OpenShadingLanguage#2061)
>
>Makes perfect sense, just wanted to make sure there wasn't something specifically broken with the OIIO in the 2024 image which was forcing you to replace it.
>
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/2061#issuecomment-3634901569
>You are receiving this because you are subscribed to this thread.
>
>Message ID: ***@***.***>
--
Larry Gritz
***@***.***

PR #2064, opened 2026-01-13
Make linearstep definitely return 1 for x > edge1

### LINKED ISSUE #2063: [BUG] linearstep is not guaranteed to return exactly 0 or 1 outside the edges
For `x > e2` and `e1 < e2`, `linearstep(e1, e2, x)` should return 1.0. This however isn't guaranteed by the implementation in stdosl.h, due to possible rounding errors.

I think on platforms that exactly follow IEEE this is in fact guaranteed, but many platforms, especially with GPU targets, have optimisations with weaker error guarantees than what is required by IEEE floating point math. (see for example [`div.approx` in PTX](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#floating-point-instructions-div))

[This is the implementation in stdosl.h](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/blob/main/src/shaders/stdosl.h#L337):

```
float linearstep (float edge0, float edge1, float x) {
    float result;
    if (edge0 != edge1) {
        float xclamped = clamp (x, edge0, edge1);
        result = (xclamped - edge0) / (edge1 - edge0);
    } else {  // special case: edges coincide
        result = step (edge0, x);
    }
    return result;
}
```

A possible fix: since in the if branch we know edge0 is not equal to edge1, the following is mathematically equivalent, but it is logically guaranteed to return exactly 1 for `x > e2` and `e1 < e2`:

```
    if (edge0 != edge1) {
        result = clamp((x - edge0) / (edge1 - edge0), 0, 1);
    } else ....
```

roelandschoukens:
## Description

The division (xclamped - edge0) / (edge1 - edge0) is not guaranteed to be exactly 1 for xclamped == edge1 if the backend floating point math has higher than expected error bounds. Eg. it is common on GPU backends to use faster but less precise instructions for division.

The modified code is mathematically equivalent and it logically guarantees the returned value is exactly 1.0 for x > edge1 > edge0.

Fixes #2063

## Tests

This should not break the existing test for linearstep. A difference in behaviour will be observed only on back-ends that implement some form of fast math.

## Checklist:

  being nonconforming.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -115,7 +115,7 @@ Here are the steps to check out, build, and test the OSL distribution:
    dist/
 
 5. Add the "dist/bin" to your `$PATH`, and "dist/lib" to your
-   `$LD_LIBRAY_PATH` (or `$DYLD_LIBRARY_PATH` on MacOS), or copy the contents
+   `$LD_LIBRARY_PATH` (or `$DYLD_LIBRARY_PATH` on MacOS), or copy the contents
    of those files to appropriate directories.  Public include files
    (those needed when building applications that incorporate OSL)
    can be found in "dist/include", and documentation can be found


lgritz:
You seem to still need DCO and CLA fixed up

roelandschoukens:
Hi, the CLA should have been sent by e-mail.

lgritz:
@jmertic Is this something you can check on and poke something in the CLA system to make it approve @roelandschoukens if LF has received a signed CLA by email?

jmertic:
Hi @roelandschoukens you will need the CLA Manager for your organization to authorize you as a contributor under that CCLA ( docs at https://docs.linuxfoundation.org/lfx/easycla/v2-current/corporate-cla-managers/approve-and-manage-contributors ). If you run into issues, please open a ticket at support.linuxfoundation.org to have that team look more into it. Thanks!

lgritz:
@jmertic Are you saying that the system knows that his CLA has been signed (even though it was received by email and not done online), but that he just needs the CLA manager to add him to the authorized list?

Or are we not sure that the email-received CLA signature is understood or recorded?

roelandschoukens:
OK I think it should be authorized now.

PR #2066, opened 2026-01-16
ci: Bump python we test on Windows runners

lgritz:
They got rid of 3.9 on the runners this week and broke CI.

Also I took the opportunity to update INSTALL.md to refect the
correct dependency versions we have tested.


lgritz:
Pre-emptively merging because this fixes broken CI.

PR #2070, opened 2026-02-02
admin: Fix link to icon used for Slack release announcements

lgritz:
I think this used to work. Did the files change their path at some point, @jmertic?


jmertic:
@lgritz Yep - there was a cleanup of the filenames and so that seems to have broke. Sorry for not letting you know downstream :(

PR #2071, opened 2026-02-06
Fix "#if with no expression" errors in platform.h

jmather-sesi:
## Description

When building in debug mode, gcc rasied multiple "#if with no expression" errors with OSL_DEBUG.

## Tests

Not applicable. Code compiles.

## Checklist:

  being nonconforming.

### DIFF
--- a/src/include/OSL/platform.h
+++ b/src/include/OSL/platform.h
@@ -299,7 +299,7 @@
 #endif // OSL_DEBUG
 
 
-#if OSL_DEBUG
+#ifdef OSL_DEBUG
 	// Disable OPENMP_SIMD when debugging
 	#undef OSL_OPENMP_SIMD
 #endif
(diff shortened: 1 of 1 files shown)


lgritz:
I'm not sure why this is needed. The first mention of OSL_DEBUG in platform.h is at line 289:

```
#ifndef OSL_DEBUG
    #ifdef NDEBUG
        #define OSL_DEBUG 0
    #else
        #ifdef _DEBUG
            #define OSL_DEBUG _DEBUG
        #else
            #define OSL_DEBUG 1
        #endif
    #endif
#endif // OSL_DEBUG
```

That sure looks like it will always result in OSL_DEBUG being defined.

Additionally, in light of this, it's clear that "#ifdef OSL_DEBUG" is the wrong test, since sometimes it is defined and 0.

I feel like something else must be going on. What about that suspicious clause of `#define OSL_DEBUG _DEBUG`? Is it possible that `_DEBUG` is defined, but is empty?

lgritz:
@jmather-sesi Any thoughts on my question?

lgritz:
Hi, @jmather-sesi, checking in again. Please read my earlier comments, I would like your thoughts.

lgritz:
@jmather-sesi Checking in again. Can you explain more what's going on here?

jmather-sesi:
Hi @lgritz, apologies for my tardiness.

I'm not entirely sure how this issue occurred at the time, as I'm unable to reproduce it. I'll re-open if I run into it again.

Thanks!

PR #2097, opened 2026-03-21
ci: optimize openexr build time, don't run nightly on forks

lgritz:
* When building openexr as a dependency for OSL's CI, there is no need to build the openexr command line utilities, saving a bit of build time.

* Fix an error from PR #2074 where we tried to disable nightly CI runs on forks -- we didn't make the right change to the Windows jobs, so they were still running.


lgritz:
No objections for 2 weeks + only affects CI -> merge

PR #2105, opened 2026-04-04
admin: Rename the repo's top level doc/ subdir to docs/

lgritz:
Mostly for conformance with OpenImageIO, since we share so many scripts and management documentation. But looking across ASWF project, they almost all use "docs/" (OSL and OpenVDB were the only outliers).


fpsunflower (approved):

ThiagoIze (approved):
I like the change.

AlexeySmolenchuk:
Images in WIKI https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/wiki/OSL-Light-Path-Expressions are sourced from 'doc'. I remember just because fixed them last time.

lgritz:
I think they are from src/doc, which is different.

PR #2118, opened 2026-05-20
ci: No need to build OpenColorIO for the icx CI job (fixes broken CI)

lgritz:


lgritz:
No, this doesn't help -- it fixes one problem but seems to reveal another. I don't want to fight it for 2023 containers, so I'm going to close this one and merge the other that passes CI.

PR #2123, opened 2026-05-29
deps: Bump minimum OpenImageIO to 3.0, adjust some CI configs

lgritz:
In main only, i.e. for OSL 1.16 release late in 2026, raise the minimum OpenImageIO supported to 3.0.

That means that the mostly-2027 version of OSL will support building against OIIO 3.2 (2027), 3.1 (2026), or 3.0 (2025), which I think is enough range.

In the process, I shuffled some CI variants around. Removed a couple variants that used OIIO 2.5, and bumped the job we use for testing a debug build from being based on 2023 containers to 2026, and the job we use for checking ABI to 2025.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,34 +84,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - desc: VP2022 gcc9/C++17 llvm14 py3.9 exr3.1 oiio-2.5 avx2
-            nametag: linux-vfx2022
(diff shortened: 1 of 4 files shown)


lgritz:
Two weeks without comment. Any objections? I will merge tomorrow if nobody speaks up.

lgritz:
Three weeks without objection. Merging.
Henceforth, in main (future OSL 1.16), the OpenImageIO dependency is required to be 3.0 or above.

PR #2125, opened 2026-06-07
refactor: Reduce redundancy of argument code logic

lgritz:
Some of the translation of type codes to the abbreviations we use internally for argument type signatures and internal function naming, like 'f' for float, 'c' for color, etc., were repeated in various place. This tries to remove some of the duplication.


lgritz:
Any objections?

aconty (approved):

PR #2126, opened 2026-06-07
testing: Reduce output precision of geomath and trig, collapse variants

lgritz:
Analyzing the geomath and trig testsuite entries, each of them had multiple reference outputs due to small precision differences between platforms and optimization levels. By printing only 4 digits of precision (and in one case of an unstable value to pass to tan, just picking a different value), we are able to collapse those tests to one reference output each.

Assisted-by: Claude Code / Claude Opus 4.8


lgritz:
Any comments?

fpsunflower (approved):

PR #2128, opened 2026-06-15
admin: Clarify securty policy on what is not a vulnerability

lgritz:
This is a proposed clarification about our security stance.
Let's discuss at the next TSC meeting, and I will not merge before then.


aconty (approved):

lgritz:
Discussed in TSC also and nobody disagreed.

PR #2133, opened 2026-06-21
build: check for the bad combo of Cuda >= 13.2 but LLVM < 22.1.2

### LINKED ISSUE #2129: [BUILD] CUDA 13.3 requires LLVM 22.1.2 or newer
This is less of a build "problem", more of a FYI that might be added to the documentation: when trying to build current OSL 1.15.5.0 against latest CUDA 13.3.0, you need to use LLVM 22.1.2 or newer in order to get the following fix:

Artem Belevich: [[clang][CUDA] Define _NV_RSQRT_SPECIFIER for glibc-2.42/cuda-13.2 compatibility (#185701) (c1c833734cf0)](https://github.com/llvm/llvm-project/commit/c1c833734cf0ee0789985af32b0b3752d8bb469f)

as documented in the LLVM release notes:

https://discourse.llvm.org/t/llvm-22-1-2-released/90308

Otherwise OSL code which calls CUDA will fail to compile, complaining about `_NV_RSQRT_SPECIFIER` being undefined.

I ran across this while working on VFX Platform 2027 container images. The fix does not appear to have been backported to LLVM 21.1.8, but I was able to compile the CUDA sections with LLVM 22.1.7.

lgritz:
CUDA 13.2+ requires LLVM 22.1.2+ for the _NV_RSQRT_SPECIFIER fix.

Fixes #2129

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -132,6 +132,16 @@ if (OSL_USE_OPTIX)
             message (FATAL_ERROR "NVPTX target is not available in the provided LLVM build")
         endif()
 
+        # CUDA 13.2+ requires LLVM 22.1.2+ for the _NV_RSQRT_SPECIFIER fix.
+        # See: https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/issues/2129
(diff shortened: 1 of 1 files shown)


jfpanisset (approved):
That should hopefully avoid bug reports against that combo.

PR #2134, opened 2026-06-23
docs: docs GHA workflow + overaul testshade, oslquery, osltoy docs

lgritz:
Add a docs.yml workflow to build the docs. Triggered when we push branches or submit PRs that change them to rebuild the docs (and make them available as a downloadable artifact that people can check even if they can't build the docs on their local machine).  How did we not have this before?

Convert OSLQuery docs from docdeep to doxygen+markdown.  We had a cobbled together thing called "docdeep" that would scrape comments from a header file (like oslquery.h) in a "poor man's doxygen" style, producing "Markdeep", and turning it into HTML.  I can barely justify that now, especially after having moved all the rest of the OSL docs to Readthedocs years ago.  So just go ahead and convert it all to the standard approach we are using for the rest of the OSL docs, to appear on RTD with the rest.

Also change the testshade and osltoy docs from markdeep to markdown and just bundle them into our regular OSL documentation as will appear on Readthedocs.

Assisted-by: Claude Code / sonnet-4.6

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,89 @@
+# Copyright Contributors to the Open Shading Language project.
+# SPDX-License-Identifier: BSD-3-Clause
+# https://github.com/AcademySoftwareFoundation/OpenShadingLanguage
+
+name: docs
+
+on:
+  push:
+    # Skip jobs when only non-doc files are changed.
+    paths-ignore:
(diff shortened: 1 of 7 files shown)


lgritz:
Sorry, has sat around for a couple weeks over my vacation. Any objections?

lgritz:
Last chance for objections

PR #2136, opened 2026-07-16
Added AMD GPU backend support.

Kiniaxx:


lgritz:
There is no commit message or explanation at all in this PR. Please read other recent PRs to see what we expect. #2130 is a good recent example, that's even about a back end.

Code comments appear to not be in English.

Lots of files checked in or changed that look accidental?

lgritz:
This PR is unusable, and the author never returned to answer any questions about it. Closing.

PR #2137, opened 2026-07-27
docs: Rearrange doxygen comments within oslquery.h

lgritz:
Follow-on to PR #2134: we switched away from our old homegrown "docdeep" to a doxygen+sphinx+breathe workflow, but I neglected to change the comment style (which had a quirky comments-after-function layout) to the more typical doxygen layout.


lgritz:
No objections for a week + comment-only change = merging

PR #2138, opened 2026-07-27
platform.h: Define OSL_DEPRECATED_EXTERNAL

lgritz:
Similar to OSL_DEPRECATED, but for things where we want it to look deprecated externally, but still need to use it internally for now and don't want the warnings when building OSL itself.

Also added OSL_CONSTEXPR23 which we don't use yet, but I should have added it previously, just happened to notice now.


lgritz:
This is over a week old, no objections, and it only adds a new macro (can't break anything that already exists), so I am merging.

PR #2139, opened 2026-07-29
testing: Additional ref output for render-microfacet for Mac ARM + cpp

lgritz:
I don't know, something must have changed on the GHA runner that once again slightly changed the output for this one test, only on one platform, only when testing the new C++ debugging output. Visually, it's fine, just slightly changed sampling firefly locations in a small number of pixels. It's equally "right", just different. The expedient fix is to check in an additional reference image to let it pass.


lgritz:
I'm planning to merge this immediately after it fully passes CI, since it fixes what's breaking every other PR's CI.

PR #2140, opened 2026-07-29
testing: reduce cpp test cost

lgritz:
The rendering tests that are using the cpp debugging output are even more expensive then those without, and are also more serialized (both due to the fact that they have to shell out to invoke the C++ compiler).

Bump up their "cost" and reduce their core allocation (i.e. let other tests be more inclined to parallelize against them) in the hopes of speeding up testing, especially on the CI runners.

Also I noticed that the "render-*.cpp" tests were running their "unoptimized" variants (in addition to opt), whereas the regular render-* tests only run the optimized version. So fix so that we don't need to run the render-*.cpp unoptimized tests, saving more time.


lgritz:
This is a week old, no objections, and all it does is get rid of some expensive, redundant tests in our CI. Merging.

PR #2141, opened 2026-07-29
ci: Test against PugiXML 1.16 and raise default for CI

lgritz:


lgritz:
No objections for a week -> merging

PR #2143, opened 2026-08-02
ci: turn unity builds off when building OIIO for CI

lgritz:
This fixes a break where OIIO has a declaration in exrinput.cpp that is mismatched if a unity build combines it with exrinput_c.cpp.  I have a fix in progress on the OIIO side but this is the expedient workaround here -- just don't use unity build. When we are sure it's fixed on the OIIO side for any version we will build in CI, then we can go back to unity builds.


lgritz:
Merging without review because this is needed to unbreak CI (and changes no code so can't break any users).

PR #2145, opened 2026-08-15
fix: Repair bad string formatting - missing args, dangerous format spec

lgritz:
RendererServices default implementations of errorfmt(), warningfmt(), etc, passed the formatted error string directly to Context::errorfmt() as the sole argument, like

    context->errorfmt(some_string);

But even though it's a single plain string with no intentional arguments to format, what if that string happens to containing std::format-like formatting directives, like "{hello world}", that will confuse the heck out of the underlying string formatting engine? The correct code all along was:

    context->errorfmt("{}", any_string_ok_here);

Also happened to find a use of fmt in typecheck.cpp where it didn't pass an argument implied by the format spec.

Additionally, for safety in a somewhat related way, make decode_message not crash, and print debug info, if the underlying format_to_n throws an exception.

### DIFF
--- a/src/cmake/testing.cmake
+++ b/src/cmake/testing.cmake
@@ -342,7 +342,7 @@ macro (osl_add_all_tests)
                 debugnan debug-uninit
                 derivs derivs-muldiv-clobber
                 draw_string
-                error-dupes error-serialized
+                error-dupes error-malformed error-serialized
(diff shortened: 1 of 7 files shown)


grdanny (approved):
thanks for the quick fix!

PR #2012, opened 2025-07-28
Serialize builds in testshade

jfpanisset:
## Description

When building with:

- OptiX enabled (OSL_USE_OPTIX)
- Code Coverage disabled (NOT CODECOV)
- a parallel build

various components of the clang toolchain are likely to crash as the testshade and libtestshade builds run in parallel and can overwrite each other's intermediate results.

This chain adds explicit dependencies to serialize these builds, avoiding toolchain crashes at a minimal performance cost.

## Tests

```
cmake -DOSL_USE_OPTIX=ON ..
cmake --build  -j$(nproc) --verbose
```
and looking at the verbose log file, observe the serialization of the build steps in `testshade`:

```
grep gmake /tmp/oslbuild.txt | grep testshade
```

## Checklist:

- [x] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [x] I have updated the documentation, if applicable.
- [x] I have ensured that the change is tested somewhere in the testsuite (adding new test cases if necessary).
- [x] My code follows the prevailing code style of this project. If I haven't
  already run clang-format v17 before submitting, I definitely will look at
  the CI test that runs clang-format and fix anything that it highlights as
  being nonconforming.

PR #2157, opened 2026-08-29
fix(build): Correctly serialize CUDA build in testrender

### LINKED ISSUE #2156: [BUG] testrender CUDA parallel build missing dependencies
In [PR #2012](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/2012)
dependencies were added in `src/testshade/CMakeLists.txt` to avoid clang crashes while compiling
CUDA code when building across larger numbers of cores.

The same problem exists in `src/testrender/CMakeLists.txt` with potentially the same fix, here is a patch against
1.15.6.0 :

```
diff --git src/testrender/CMakeLists.txt src/testrender/CMakeLists.txt
index d99dd79ac..bfda2778a 100644
--- src/testrender/CMakeLists.txt
+++ src/testrender/CMakeLists.txt
@@ -85,6 +85,10 @@
         pugixml::pugixml
         Threads::Threads)

+if (OSL_USE_OPTIX)
+    add_dependencies(testrender testrender_ptx)
+endif ()
+
 osl_optix_target (testrender)

 install ( TARGETS testrender RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} )
```

The patch seems to be working in the context of the aswf-docker project, if it seems appropriate I can submit as a PR.

jfpanisset:
When building with CUDA / OptiX, src/testrender/CMakeLists.txt was not specifying a dependency of testrender on testrender_ptx, allowing GNU Make to run these two sets of instructions as parallel jobs. When building on multiple cores, this could result in llvm crashing when the testrender build steps would try to access files still being build by the testrender_ptx build steps.

Fixes #2156 and is the same fix which was previously applied to testshade in [PR 2012](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/2012).

Assisted-by: Claude/Sonnet 5

### DIFF
--- a/src/testrender/CMakeLists.txt
+++ b/src/testrender/CMakeLists.txt
@@ -85,6 +85,10 @@ target_link_libraries (testrender
         pugixml::pugixml
         Threads::Threads)
 
+if (OSL_USE_OPTIX)
+    add_dependencies(testrender testrender_ptx)
+endif ()
+
 osl_optix_target (testrender)
 
 install ( TARGETS testrender RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} )


lgritz (approved):
LGTM

PR #2161, opened 2026-09-07
build: Fix name and install destination of llvm_macros.cmake

### LINKED ISSUE #1771: Bad file install locations
### Problem

The 1.13.6.1 release installs some new files that are not in acceptable locations.

All `.cmake` files should be installed into one location. Extra support scripts should be installed under `share` not at the top of the filesystem.

**Expected behavior:**

A clean install with files in expected locations

**Actual behavior:**

cmake config files and extra support scripts are installed into two locations.

### Versions

* OSL branch/version: 1.13.6.1

### Solution

`llvm_macros.cmake` should have `DESTINATION ${OSL_CONFIG_INSTALL_DIR}` to put it with the other cmake files.

The `serialize-bc.py` script would be better installed somewhere like `DESTINATION ${CMAKE_INSTALL_DATADIR}/build-scripts`

lgritz:
Fixes #1771

Primarily, this was about llvm_macros.cmake installing in the same location where we put the exported cmake config files.

But I also realized, maybe it's not good to have exported cmake files that aren't prefixed with "OSL_" like the others, so I renamed it, too. I will only have that in 1.16, so it will not break people in release branches.


jfpanisset (approved):
Looks good, building and installing "everything" into `/usr/local/` in the ci-vfxall images is a good way to catch projects which land files in unexpected places.

PR #2162, opened 2026-09-10
Add nanobind as an alternative Python bindings backend for oslquery

lgritz:
The oslquery Python module can now be built with nanobind, pybind11, or both, as selected by the CMake variable OSL_PYTHON_BINDINGS_BACKEND.

When the option is unset, the backend is auto-selected: nanobind when OpenImageIO is >= 3.2 and Python is >= 3.10, pybind11 otherwise. Matching OpenImageIO's own framework keeps OSLQuery.Parameter.type interoperable with the paired OpenImageIO module. nanobind need not be installed; it is built locally when missing.

This is modelled from, and uses basically the same techniques as, the [recent work done to convert OpenImageIO to nanobind](https://github.com/AcademySoftwareFoundation/OpenImageIO/pulls?q=is%3Apr%20nanobind)

Design notes are in: docs/dev/specs/003-nanobind-python-bindings/

Assisted-by: Claude Code / claude-opus-5


lgritz:
No objections -> merging

